PR #4, opened 2025-06-17
chore: add .gitignore to keep our repo neat and sparkling ✨

madflojo:
- Removing the .DS_Store file
- Adding a .gitignore so it doesn't come back


sougou:
Thanks for the contribution, and catching my oversight. For clarification, I've just created an announcement on our contribution policy https://github.com/multigres/multigres/discussions/6.

madflojo:
No problem, glad to see support for Postgres becoming a reality!

PR #29, opened 2025-07-22
replace multi-writer with highly-available

deepthi:
And another small typo fix.


sougou:
Let's instead replace multi-writer with HA. That's what the term implies.

PR #30, opened 2025-07-23
doc: HA and PG full-sync blog

sougou:


site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+1b: Primary crashes after writing a transaction to the WAL: Multigres can automate the repair of the primary when it comes back up.
+
+1c: Client reading phantom records: There is no mitigation, but this is a rare occurrence.
+
+1d: Replica that is not a standby receives errant transaction: Multigres can automate the repair.
+
+2: Replica committing before primary: There is no mitigation, but the application can work around this possibility on a case-by-case basis.
+
+3: Failures during failover: Assuming clocks are reliable, Multigres can use the WAL timestamps to determine which of the failover attempts was the latest and can use it. We will cover why this is the most authoritative timeline when we explain consensus algorithms in upcoming blogs.
```
kiwicopple:
```suggestion
- **1a: CDC picking up a transaction early:** Run logical replication on a replica.
- **1b: Primary crashes after writing a transaction to the WAL:** Multigres can automate the repair of the primary when it comes back up.
- **1c: Client reading phantom records:** There is no mitigation, but this is a rare occurrence.
- **1d: Replica that is not a standby receives errant transaction:** Multigres can automate the repair.
- **2: Replica committing before primary:** There is no mitigation, but the application can work around this possibility on a case-by-case basis.
- **3: Failures during failover:** Assuming clocks are reliable, Multigres can use the WAL timestamps to determine which of the failover attempts was the latest and can use it. We will cover why this is the most authoritative timeline when we explain consensus algorithms in upcoming blogs.
```

suggested for readability

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
+# High Availability and Postgres full-sync
```
NikolayS: maybe to add word "replication" ("full-sync replication")?

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+In order to achieve High Availability (HA) in Postgres, we need to have at least one other standby replica that keeps up with the primary’s changes near-real-time. Using Postgres physical replication is the most common method to achieve this.
+
+High Availability also requires us to solve the problem of distributed durability. After all, we have to make sure that no transactions are lost when we failover to a standby. So, if we can make this work, we can avoid the need for an external system like a mounted cloud drive or other exotic solutions to ensure that we don’t lose data. We could just have all the servers use their local NVME drive for storage. This will serendipitously improve performance due to the drives being an order of magnitude faster than the network, and reduce costs since disk I/O does not incur network cost.
+
+## Multigres features
+
+We plan to support HA using full-sync replication, mainly because this is a widely used feature. However, there are some pitfalls to watch out for. We will cover these in the next section.
```
NikolayS: I would clarify the term "full-sync"... 

Widely used – do you mean outside Postgres ecosystem, or inside?

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+Multigres will bring the flexibility of a pluggable durability policy. There will be a few predefined ones, but if your needs are bespoke, you can just write an extension. Examples of durability policies include:
+
+- cross-availability zone (cross-az): a replica in at least one other az must have my data
```
NikolayS:
```suggestion
- cross-availability zone (cross-AZ): a replica in at least one other AZ must have my data
```

(just for readability)

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
+# High Availability and Postgres full-sync
+
+In order to achieve High Availability (HA) in Postgres, we need to have at least one other standby replica that keeps up with the primary’s changes near-real-time. Using Postgres physical replication is the most common method to achieve this.
+
+High Availability also requires us to solve the problem of distributed durability. After all, we have to make sure that no transactions are lost when we failover to a standby. So, if we can make this work, we can avoid the need for an external system like a mounted cloud drive or other exotic solutions to ensure that we don’t lose data. We could just have all the servers use their local NVME drive for storage. This will serendipitously improve performance due to the drives being an order of magnitude faster than the network, and reduce costs since disk I/O does not incur network cost.
```
NikolayS, line 5: maybe mention why async can be not enough? e.g., `maximum_lag_on_failover: 1048576` in Patroni -- data loss up to 1 MiB is allowed

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
...
+The advantages of policy based durability go beyond ease of use. Essentially, the policy does not need to depend on the number of nodes in the quorum. This flexibility allows you to deploy more nodes or additional zones without affecting the performance of the cluster.
+
+More about how this works will be covered in subsequent blog posts related to generalized consensus.
+
+## Existing replication pitfalls
+
+Postgres has many options on how you can configure replication. They broadly fall into two categories:
+
+1. Asynchronous replication (async)
+    * The primary commits the data and returns success to the caller.
```
NikolayS:
```suggestion
    * the primary commits the data and returns success to the caller.
```

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+The problem with async replication is that there are failure modes where you may lose data. For example, if a primary node crashes after acknowledging a commit but before sending the data to the replica, that commit is lost. In the case of a network partition, this can go on for a long time risking substantial data loss.
+
+Synchronous replication addresses this problem, but it also has its own pitfalls explained below. This [Video from Kukushkin](https://youtu.be/PFn9qRGzTMc?si=neDR5sgti16Hfni0) covers many possible ways full-sync replication can fail on you. Overall, the problems are due to one of the following reasons:
+
+1. The Primary writes to the WAL and crashes before sending the data to the standby:
+    
+    <img src="/img/blog/fsync1.svg" alt="fsync1" />
+    
+    * setups where there are clients subscribed to the WAL on the primary (like a CDC process), they would view and transmit this as a committed transaction. However, a failover will abandon this transaction, while the subscriber might have irreversibly acted on this.
```
NikolayS, line 45: I recently learned about [synchronized_standby_slots](https://postgresqlco.nf/doc/en/param/synchronized_standby_slots/) -- can be helpful in this context

NikolayS: (PG17+)

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
...
+    * the primary is restarted after a failover, it will have a timeline that is incompatible with the rest of the cluster, and will not be able to rejoin the cluster.
```
NikolayS, line 46: hmm, if it has older timeline, it can rejoin as replica, right?
and if it has more commits (that were not acknowledged by new primary), we can pg_rewind it, but it's data loss again..

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+- **1a: CDC picking up a transaction early:** Run logical replication on a replica.
+- **1b: Primary crashes after writing a transaction to the WAL:** Multigres can automate the repair of the primary when it comes back up.
+- **1c: Client reading phantom records:** There is no mitigation, but this is a rare occurrence.
+- **1d: Replica that is not a standby receives errant transaction:** Multigres can automate the repair.
+- **2: Replica committing before primary:** There is no mitigation, but the application can work around this possibility on a case-by-case basis.
+- **3: Failures during failover:** Assuming clocks are reliable, Multigres can use the WAL timestamps to determine which of the failover attempts was the latest and can use it. We will cover why this is the most authoritative timeline when we explain consensus algorithms in upcoming blogs.
+
+Many of these problems can be tolerated by the application, but it will be nice to eventually eliminate them altogether. This is why we intend to implement a two-phase sync replication solution, which can be the foundation for consensus protocols.
```
NikolayS, line 70: I'm not sure I understand the term; "two-phase sync replication" - what are two phases here? Or it's about 2PC?..

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+Furthermore, the two-phase approach will have the same performance overhead as the full-sync approach, so you have nothing to lose and everything to gain by transitioning to it. In any case, until it becomes robust, the full-sync approach can continue be used as an acceptable compromise.
```
NikolayS, line 72: that's an interesting statement! I understand that you feel it, and I'm still curious

I'm still learning and would be great to do some benchmarks / source code research to see how they compare performance-wise

NikolayS:
@sougou enjoyed the reading, thanks for inviting me!

my questions are mostly in the area of terminology - I have a feeling that sometimes we speak in slightly different terms, maybe worth introducing terms in a more formal way

also, very curious if you're planning to implement some tests showing problems that are listed as to be solved by Multigres!

thanks!

site/blog/2025-07-22-ha-postgres-full-sync.mdx
DylanGriffith: Out of curiosity, how does this differ with 2PC? I'm reading https://en.wikipedia.org/wiki/Two-phase_commit_protocol and thinking that we still have a problem where one node gets the "Commit" message before the other node. In that case I assume that first node will write to the WAL and it's still possible that the coordinator or the other node crashes at that point before the 2nd node can commit.

The transaction is meant to be abandoned but there may still be some nodes that have it.

Or maybe this falls under the "Disadvantages" section of that wiki article where it says:

> After a participant has sent an agreement message as a response to the commit-request message from the coordinator, it will block until a commit or rollback is received.

So that 2nd node would just be stuck forever and the client will assume the transaction was abandoned?

But still anyone reading the WAL of any node that has received a Commit will receive this data I assume.

site/static/img/blog/fsync1.svg
za-arthur, line 1: It seems there might be some bug in SVG rendering, but Safari doesn't show arrows
Screenshot 2025-07-24 at 12 45 18

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: Does this make it more clear? "We plan to support HA by configuring Postgres in full-sync replication mode".

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: This is good to know. I'll clarify this feature of Patroni in a new paragraph. And, I think we don't want to support it.

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: Good to hear about this new feature. Added synchronized standby slots to 1a as an option.

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: Yes. Rewind is the correct solution. Technically, it's not a data loss because the commit would not have returned. So, the application cannot assume that it succeeded.

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: Actually, two-phase sync (2PS) is very different from 2PC. The only thing common between them is the fact that it's two phases :). They are actually orthogonal. In Multigres, 2PC will rely on 2PS for durability; It would ensure completion of a prepared transaction even if there are node failures during a distributed commit.

I wrote an old document about how to reliably complete 2PC transactions. That scheme is implemented in Vitess. I'll see if it can be resurrected.

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: I clarified this a bit in a previous answer. It feels like it may be worth writing a separate blog explaining two-phase sync, and also talk about how it's different from 2pc?

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: This will be awesome!

sougou:
> @sougou enjoyed the reading, thanks for inviting me!
> 
> my questions are mostly in the area of terminology - I have a feeling that sometimes we speak in slightly different terms, maybe worth introducing terms in a more formal way

This is a great idea. I found myself using the wrong terminology a few times. Will you be able to send me a list? We could work together on getting them defined.

> 
> also, very curious if you're planning to implement some tests showing problems that are listed as to be solved by Multigres!

Yes. We have to :). We have to make sure Multigres handles those corner cases.

site/static/img/blog/fsync1.svg
sougou: I'll get these redone.

site/blog/2025-07-22-ha-postgres-full-sync.mdx
NikolayS: Yes, would be great -- I didn't know this term (2PS). I know 2PC and 2PL, also very very different, so with 2PS, we have 3 very different two-phase things, explanation would be definitely great to have!

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: @DylanGriffith: I found the document on the Vitess 2PC design here: https://github.com/vitessio/vitess/blob/release-21.0/doc/design-docs/TwoPhaseCommitDesign.md.

The PG implementation will be simpler because we had to emulate PREPARE in MySQL, but it's natively available in PG.

Let me know if you have questions.

site/blog/2025-07-22-ha-postgres-full-sync.mdx
```diff
@@ -0,0 +1,84 @@
...
```
samrose, line 72: Will multigres need to resort to network-level fencing, to guarantee only one primary node during partitions, or do other aspects of the implementation negate the need for this?

sougou: We'll be using principles of consensus (like RAFT) to guarantee revocation of previous leadership during failover. There is a longer answer, and I plan to cover this in detail in an upcoming blog post.

site/blog/2025-07-22-ha-postgres-full-sync.mdx
DylanGriffith: Thanks. A couple of interesting and surprising details in there which relate to my confusion:

> Isolation requires additional work. If a client tries to read data in the middle of a distributed commit, it could see partial commits. In order to prevent this, databases put read locks on rows that are involved in a 2PC. So, anyone that tries to read them will have to wait till the transaction is resolved. This type of locking is so contentious that it often defeats the purpose of distributing the data.
> 
> In reality, this level of Isolation guarantee is overkill for most code paths of an application. So, it’s more practical to relax this for the sake of scalability, and let the application use explicit locks where it thinks better Isolation is required.

This statement seems to relate to the above comments about whether or not we should allow anomalous reads. It sounds like we can just use locks if it's critical or live with the risk if it's not. I guess this corresponds to sacrificing "Availability" for "Consistency" in a distributed database.

> If Prepares are successful, then the state is transitioned to ‘Commit’. In the Commit state, only commits are allowed. By the guarantee given by the Prepare contract, all databases will eventually accept the commits.

I think I was missing this piece of information reading the 2PC wiki and the "watchdog" seems to be necessary to accomplish this. I'm a little skeptical about the claim "all databases will eventually accept the commits" but maybe that's possible assuming we can evict databases that are misbehaving.

> The watchdog will grab this transaction and invoke a ResolveTransaction. The VTGate will then make a decision to rollback, because all it sees is a 2PC in Prepare state. It will attempt to rollback all participants, while some might have already committed. A failure like this will be undetectable.

Also seems like the Vitess design is saying it's possible to have data committed to some nodes while the overall transaction was considered to have failed.

Anyway thanks Sugu for pointing me to these docs. I'm looking forward to seeing some similar Multigres specific docs coming out as these features are built.

site/blog/2025-07-22-ha-postgres-full-sync.mdx
sougou: The 2PS name is my own invention :).

sougou:
Merging this early so that @mildtomato can fix the drawings.

PR #35, opened 2025-08-05
Adds metadata

rafael:
* Takes over https://github.com/multigres/multigres/pull/34
* Fixes another date.


sougou:
Looks like it's the same as https://github.com/multigres/multigres/pull/34? If so, we can close it.

PR #42, opened 2025-08-13
Rename all docs to use the new Multi prefix naming for components

sougou:


doc/mvp_project_plan.md
```diff
@@ -55,22 +55,22 @@ The existing Postgres primitives are insufficient to build a robust consensus pr
 
 ### Multigres
 
-Multigres will build the coordination part of the consensus protocol using a brand new VTOrc. This will not be ported from Vitess, because the VTOrc from Vitess has a large amount of legacy code that was inherited from the MySQL Orchestrator.
+Multigres will build the coordination part of the consensus protocol using a brand new MultiOrch. This will not be ported from Vitess, because the MultiOrch from Vitess has a large amount of legacy code that was inherited from the MySQL Orchestrator.
```
deepthi:
```suggestion
Multigres will build the coordination part of the consensus protocol using a brand new component called MultiOrch. This will not be ported from Vitess, because the VTOrc from Vitess has a large amount of legacy code that was inherited from the MySQL Orchestrator.
```

PR #43, opened 2025-08-14
Initial set of binaries

deepthi:
Adding the initial set of binaries - MultiGateway, MultiPooler, MultiOrch and pgctld.
Basic flags and config have been added using viper. 
Added a basic test and CI for it.

Note to reviewers: I have not added a http port to these binaries yet. That is something we can decide on and add if necessary.


sougou (approved):

cmd/multiorch/main.go
```diff
@@ -0,0 +1,116 @@
...
+	"syscall"
+
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+func setupConfig() {
+	// Define flags
+	pflag.StringP("grpc-port", "p", "15300", "gRPC port to listen on")
+	pflag.StringP("topology-addr", "t", "localhost:2379", "etcd topology server address")
```
rafael, line 35: Wondering if we should be explicit that this refers to the global topo. So something like: `global-topo-addr`

cmd/multigateway/main.go
```diff
@@ -0,0 +1,114 @@
...
+	"os/signal"
+	"syscall"
+
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+func setupConfig() {
+	// Define flags
+	pflag.StringP("port", "p", "5432", "Port to listen on")
```
rafael, line 34: We could add a comment here to be explicit that this is the postgres port and the main entry point to interact with the cluster as if it were a postgres server

rafael (commented):
Nice. This makes sense. Left a few comments

venkat-murty:
🚀

PR #44, opened 2025-08-15
topology: creates initial protobuf definitions for topology metadata

rafael:
# Description

This PR is making two changes:
1. It adds the initial protobuf definitions for the clustermetadata topology information.
2. In order to compile the proto definition, it wires up protoc into the Makefile of the repo.


cmd/multigateway/BUILD.bazel
```diff
@@ -0,0 +1,19 @@
+load("@rules_go//go:def.bzl", "go_binary", "go_library")
```
rafael: we don't have to write this manually. These are generated automatically by bazel

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  PRIMARY = 1;
+
+  // REPLICA replicates from leader. It is used to read only traffic
+  REPLICA = 2;
+}
+
+// PoolerServingStatus represents the serving status of the given MultiPooler.
+enum PoolerServingStatus {
+  // SERVING is the status a server during normal operations when is serving traffic.
+  SERVING = 0;
```
v0idpwn: Probably all enums should have UNKNOWN, as `PoolerType` does. See https://protobuf.dev/best-practices/dos-donts/#unspecified-enum

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+// This is not stored in the topo itself, but it's the entry point to
+// configure the global topo.
+message GlobalTopoConfig {
+  // topo_type indicates the type of topology service in use
+  // (for example: "etcd2",  "memory").
+  string implementation = 1;
+
+  // server is the network address (host:port) of the topology service.
+  // Can be a comma-separated list for HA setups.
+  string server_address = 2;
```
v0idpwn: Perhaps worth making addresses a message type, and expecting `repeated`.

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+// Global Topology
+// =============================================================================
+
+// TopoConfig defines the connection parameters for a topology service.
+// It specifies the type of topology backend, where it's hosted, and the
+// logical root path within that backend.
+// This is not stored in the topo itself, but it's the entry point to
+// configure the global topo.
+message GlobalTopoConfig {
+  // topo_type indicates the type of topology service in use
```
deepthi:
```suggestion
  // implementation indicates the type of topology service in use
```

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  
+  // If range based sharding is used, range for the pooler's shard.
+  KeyRange key_range = 5;
+
+  // PoolerType is the current type of the pooler.
+  PoolerType type = 6;
+  
+  // PoolerServingStatus is the current type of the pooler.
+  PoolerServingStatus serving_status = 7;
+
```
deepthi: Do we need to prefix these with `Pooler`? It seems more consistent to just call them Type and ServingStatus. Is there something we would get from having a prefix? Maybe you think Type is too generic?

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  // Fully qualified domain name of the host.
+  string hostname = 8;
+  
+  // Map of named ports. Normally this should include postgres and grpc.
```
deepthi: Pooler won't have a postgres port open.

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  map<string, int32> port_map = 9;
+  
+  // tags
+  map<string, string> tags = 10;
```
deepthi: It will be good to write down why tags and how they may be used. I know of one use case, which is to support directing queries to a group of poolers which have the same tag. Maybe there are others?

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  // Fully qualified domain name of the host.
+  string hostname = 2;
+  
+  // Map of named ports. Normally this should include postgres and grpc.
+  map<string, int32> port_map = 3;
+}
+
+// MultiOrc represents information about a running instance of multiorc.
+message MultiOrc {
+   // identifier is the unique name of the MultiOrc in the cluster.
```
deepthi:
```suggestion
// MultiOrch represents information about a running instance of multiorch.
message MultiOrch {
   // identifier is the unique name of the MultiOrch in the cluster.
```

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  ID identifier = 1;
+
```
deepthi: It seems to me that if multiorch's are registered in the topo, then their hostname and port_map should also be there.

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+   // identifier is the unique name of the MultiOrc in the cluster.
+  ID identifier = 1;
+  
+}
+
+// =============================================================================
+// Auxiliary Data Types
+// =============================================================================
+
+// PoolerID is a globally unique pooler identifier.
```
deepthi:
```suggestion
// ID is a globally unique identifier.
```

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+message ID {
+  // component that this ID is used for (i.e multigateway / multiorc / multipooler)
```
deepthi:
```suggestion
  // component that this ID is used for (i.e multigateway / multiorch / multipooler)
```

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  MultigresComponent component = 1;
+  
+  // cell is the cell the pooler is located
```
deepthi:
```suggestion
  // cell is the cell the component is located in
```

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  string cell = 2;
+
+  // uid is a unique id for this pooler within the shard
+  // (this is the Postgres server id as well).
```
deepthi: This comment needs to be rewritten.

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+// PoolerType represents the type of a given MultiPooler.
+enum PoolerType {
+  // UNKNOWN this is not a valid type. 
+  // It is useful during upgrades to detect when a value is not initialized
+  UNKNOWN = 0;
+
+  // PRIMARY is the primary server for the shard. Only PRIMARY allows DMLs.
+  PRIMARY = 1;
+
+  // REPLICA replicates from leader. It is used to read only traffic
```
deepthi:
```suggestion
  // REPLICA replicates from leader. It can be used for read only traffic
```

deepthi: I'm saying `can be` because we may or may not automatically direct read traffic to replicas.

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  REPLICA = 2;
+}
+
+// PoolerServingStatus represents the serving status of the given MultiPooler.
+enum PoolerServingStatus {
+  // SERVING is the status a server during normal operations when is serving traffic.
+  SERVING = 0;
+  
+  NOT_SERVING = 1;
```
deepthi: Needs a comment.

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  // PRIMARY is the primary server for the shard. Only PRIMARY allows DMLs.
+  PRIMARY = 1;
+
+  // REPLICA replicates from leader. It is used to read only traffic
+  REPLICA = 2;
+}
+
+// PoolerServingStatus represents the serving status of the given MultiPooler.
+enum PoolerServingStatus {
+  // SERVING is the status a server during normal operations when is serving traffic.
```
deepthi:
```suggestion
  // SERVING is the status a server during normal operations when it is serving traffic.
```

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+  SERVING = 0;
+  
+  NOT_SERVING = 1;
+  
+  // BACKUP is the status of a server when is taking a backup. No queries
```
deepthi:
```suggestion
  // BACKUP is the status of a server when it is taking a backup. No queries
```

proto/clustermetadata.proto
```diff
@@ -0,0 +1,205 @@
...
+enum MultigresComponent {
+  // MULTIGATEWAY handles query routing and traffic distribution across clusters.
+  MULTIGATEWAY = 0;
+
+  // MULTIPOOLER manages connection pooling and load balancing for database clients.
+  MULTIPOOLER = 1;
+
+  // MULTIORC is the orchestrator responsible for cluster topology management,
+  // failover coordination, and automated repairs.
+  MULTIORC = 2;
```
deepthi:
```suggestion
  // MULTIORCH is the orchestrator responsible for cluster topology management,
  // failover coordination, and automated repairs.
  MULTIORCH = 2;
```

proto/clustermetadata.proto
rafael: I brought this over from Vitess blindly. It could be handy in the future. But  now thinking to keep this minimum and only add it when we need it.

proto/clustermetadata.proto
rafael: This is a good idea. Changing.

test/startup_test.go
```diff
@@ -47,9 +48,30 @@ func TestBinaryStartupShutdown(t *testing.T) {
...
+		if _, err := os.Stat(binaryPath); err == nil {
+			runBinaryTest(t, binaryPath, binaryName, port)
+			return
+		}
+		require.Fail(t, "Binary not found: %s", binaryName)
+	}
+
+	binaryPath := filepath.Join(runfilesDir, "_main", "cmd", binaryName, binaryName+"_", binaryName)
+
+	runBinaryTest(t, binaryPath, binaryName, port)
```
egor-romanov:
```suggestion
	if runfilesDir != "" {
		binaryPath := filepath.Join(runfilesDir, "_main", "cmd", binaryName, binaryName+"_", binaryName)
		runBinaryTest(t, binaryPath, binaryName, port)
		return
	}
	
	// Fallback for non-Bazel execution (like make test)
	binaryPath := filepath.Join("..", "bin", binaryName)
	if _, err := os.Stat(binaryPath); err == nil {
		runBinaryTest(t, binaryPath, binaryName, port)
		return
	}
	
	// If ../bin doesn't exist, try current directory
	binaryPath = filepath.Join("bin", binaryName)
	if _, err := os.Stat(binaryPath); err == nil {
		runBinaryTest(t, binaryPath, binaryName, port)
		return
	}
	require.Fail(t, "Binary not found: %s", binaryName)
```

proto/clustermetadata.proto
rafael: Good flag

deepthi (approved):

PR #45, opened 2025-08-15
general: adds hooks, linters and topo renames

rafael:
# Description

* Adds some initial linters and pre commit hooks checks.
* Follow up in some naming from previous PR.

### DIFF
--- a/misc/hooks/golangci-lint
+++ b/misc/hooks/golangci-lint
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # Required version of golangci-lint
-REQUIRED_VERSION="v2.0.2"
+REQUIRED_VERSION="1.62.2"
 
 # Function to compare versions in pure Bash
 version_greater_or_equal() {


proto/clustermetadata.proto
```diff
@@ -94,56 +93,19 @@ message MultiPooler {
   // If range based sharding is used, range for the pooler's shard.
   KeyRange key_range = 5;
 
-  // Type is the kind of pooler: PRIMARY or REPLICA
-  Type type = 6;
+  // PoolerType is the kind of pooler: PRIMARY or REPLICA
```
rafael, line 96: We discussed this rename in previous PR, but missed the commit before it was merged 😅

.golangci.yml
```diff
@@ -0,0 +1,71 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+version: "2"
+run:
+  go: "1.24"
```
deepthi, line 17: Should be 1.25?

rafael: I got into some issues with `golangci-lint`  that need follow up. It did not seem to work with 1.25 even in the latest version. 

Let me do some research to check what's up with that.

PR #47, opened 2025-08-16
Rchacon/update golang

rafael:
# Desc

* Fixes some indentation issues in the golangci config that were preventing some settings to be picked up correctly.
* Make sure that during the automatic install we use v2.
* Fixes existing lint errors in the project.


misc/hooks/golangci-lint
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # Required version of golangci-lint
-REQUIRED_VERSION="1.62.2"
+REQUIRED_VERSION="v2.4.0"
```
rafael, line 17: when I did the first pass, I installed manually first with go get and it was not using v2, so it was a very old version.

rafael:
Closed in favor of https://github.com/multigres/multigres/pull/49, so it's opened this same [github workflow](https://vitess.io/docs/contributing/github-workflow/).

PR #51, opened 2025-08-18
fixes: check license

rafael:
# Desc

* When I first added this hook, I did not have the integration for golang-ci and accidentally broke the expectations on how this is called from the [pre-commit](https://github.com/multigres/multigres/blob/main/misc/git/pre-commit#L0-L1) hook. These fixes that regression.
* It also makes sure addlicense is installed when running the hook.

### DIFF
--- a/misc/git/pre-commit
+++ b/misc/git/pre-commit
@@ -24,6 +24,6 @@ fi
 
 for hook in $GIT_DIR/../misc/hooks/*; do
  if [ -x "$hook" ]; then
-   $hook $@
+   $hook
  fi
 done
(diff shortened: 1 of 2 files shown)


misc/git/pre-commit
```diff
@@ -24,6 +24,6 @@ fi
 
 for hook in $GIT_DIR/../misc/hooks/*; do
  if [ -x "$hook" ]; then
-   $hook $@
+   $hook
```
rafael, line 27: same as vitess, the scripts don't expect any args

PR #56, opened 2025-08-18
clustermetadata: topo foundations

rafael:
# Summary

This PR implements introduces the clustermedata package. It mostly replicates what the topology does in Vitess. We are making some changes such that the topology does not store as much data as it did in vitess. We are going to be moving schema and shard metadata to postgres itself. That will be happening in a future PR. In the new model, the global topo and the cells topo store this data:

1. **Global topology**: Cluster level minimum static metadata. It contains the pieces of information that allows other components to find what databases and where they are located. 
2. **Cell topology**:  Cell level catalog that contains the *dynamic* data about the components of the cluster.

We will publish a design doc for our decisions here soon.

##   Core Topology Infrastructure

- Interfaces (conn.go, store.go): Foundational abstractions for topology connections and store operations
- Error Handling (errors.go, mterrors/): Robust error types and handling mechanisms ported from Vitess.
- Protocol Buffers (proto/): Cluster metadata and RPC definitions

### In-Memory Implementation

This is a port of Vitess. Thank Alain, your code is still alive. I added a few new capabilities to enhance the tests. 

### API Support
  
*Database Management*

  - CRUD Operations (database.go): Create, read, update, delete databases in topology
  - Field Updates with retry logic for version conflicts
  - Path-based organization with proper validation
  
*Cell Location Management*

  - Cell Operations (cell_location.go): Manage geographical cell configurations
  - Server Addressing: Configure topology server addresses per cell
  - Field Updates with the same retry logic as databases
  
  The methods here should give us the core functionality to initialize the global topo. 

## Next Steps

* Add API methods to store cell data.
* There is some functionality that I did not port yet:
   - Servenv
   - Observability.
    
##  Test Coverage

I ported all the tests that we had in Vitess for the topo and added the following:
  - Error Injection Framework:
    - AddOneTimeOperationError() for single-failure testing 
    - ClearOperationErrors() for test cleanup
  - Retry Behavior Tests:
    - Database update retry on BadVersion error 
    - Cell location update retry on BadVersion error 
    - Verification that update functions are called multiple times during retries.


go/clustermetadata/topo/cell_location.go
```diff
@@ -0,0 +1,156 @@
...
+// available cells, if necessary. A CellLocation can only be removed if no
+// Shard record references the corresponding cell in its Cells list.
+
+func pathForCellLocation(cell string) string {
+	return path.Join(CellsPath, cell, CellLocationFile)
+}
+
+// GetCellLocationNames returns the names of the existing cells. They are
+// sorted by name.
+func (ts *store) GetCellLocationNames(ctx context.Context) ([]string, error) {
```
deepthi: If this is just the cell names, we should call it GetCellNames

go/clustermetadata/topo/store.go
```diff
@@ -0,0 +1,364 @@
...
+    locations.
+
+ 2. Cell topology: cell-level catalogs that store dynamic metadata about local
+    components (gateways, poolers, orchestrators, etc). Each cell is logically
+    distinct and accessed through a separate connection. In practice, a
+    deployment may choose to run the global and cell topologies on the same
+    etcd cluster, but they remain separate in terms of naming and client
+    management.
+
+Below a diagram representing the architecture:
```
deepthi, line 37: Nice!!

go/clustermetadata/topo/store.go
```diff
@@ -0,0 +1,364 @@
...
+	factories = make(map[string]Factory)
+
+	FlagBinaries = []string{"multigateway", "multiorch", "multipooler", "pgctld"}
+
+	// DefaultReadConcurrency Default read concurrency to use in order to avoid overwhelming the topo server.
+	DefaultReadConcurrency int64 = 32
+)
+
+func init() {
+	// TODO: Follow up PR perform this hooking
```
deepthi: 👍 
When we do this, the flags will move out of the various main.go files under cmd.

go/clustermetadata/topo/store.go
```diff
@@ -0,0 +1,364 @@
...
+// Filenames for all object types.
+const (
+	CellLocationFile = "CellLocation"
+)
+
+// Path for all object types.
+const (
+	DatabasesPath = "databases"
+	CellsPath     = "cells"
+	MultiGateways = "multi_gateways"
```
deepthi: do we really need the multi prefix in topo paths?

deepthi: I like the decision to use only lowercase for paths, and PascalCase for files.

proto/mtrpc.proto
```diff
@@ -0,0 +1,191 @@
...
+  // Topo server connection is read-only
+  READ_ONLY = 18;
+}
+
+// RPCError is an application-level error structure returned by
+// MultiPooler (and passed along by MultiGateway if appropriate).
+// We use this so the clients don't have to parse the error messages,
+// but instead can depend on the value of the code.
+message RPCError {
+  reserved 1; reserved "legacy_code";
```
GuptaManan100: We don't need reserved numbers right now I think 😄

go/clustermetadata/topo/memorytopo/lock.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	return c.lock(ctx, dirPath, contents, false)
+}
+
+// LockName is part of the topo.Conn interface.
+func (c *conn) LockName(ctx context.Context, dirPath, contents string) (topo.LockDescriptor, error) {
+	// c.factory.callstats.Add([]string{"LockName"}, 1)
+	return c.lock(ctx, dirPath, contents, true)
+}
```
GuptaManan100, line 122: Shouldn't this just return the name of the lock and not actually lock the file? How is this different from Lock?

go/clustermetadata/topo/database.go
```diff
@@ -0,0 +1,149 @@
...
+	_, err = ts.globalTopo.Create(ctx, filePath, contents)
+	return err
+}
+
+// UpdateDatabaseFields is a high level helper method to read a Database
+// object, update its fields, and then write it back. If the write fails due to
+// a version mismatch, it will re-read the record and retry the update.
+// If the update method returns ErrNoUpdateNeeded, nothing is written,
+// and nil is returned.
+func (ts *store) UpdateDatabaseFields(ctx context.Context, database string, update func(*clustermetadatapb.Database) error) error {
```
rafael, line 96: In Vitess, UpdateDatabase takes a lock, but cell location updates use optimistic concurrency (retry on version mismatch). For now I kept the optimistic approach for both. We can revisit later if needed.

go/clustermetadata/topo/store.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// registerTopoFlags registers command-line flags for topology configuration.
+// This function is currently commented out but will be implemented in a follow-up PR.
+// func registerTopoFlags(fs *pflag.FlagSet) {
+// 	fs.StringVar(&topoImplementation, "topo_implementation", topoImplementation, "the topology implementation to use")
+// 	fs.StringSliceVar(&topoGlobalServerAddresses, "topo_global_server_addresses", topoGlobalServerAddresses, "the addresses of the global topology servers")
+// 	fs.StringVar(&topoGlobalRoot, "topo_global_root", topoGlobalRoot, "the path of the global topology data in the global topology server")
+// 	fs.Int64Var(&DefaultReadConcurrency, "topo_read_concurrency", DefaultReadConcurrency, "Maximum concurrency of topo reads per global or local cell.")
+// }
```
deepthi: Can we remove this commented code? Anyhow when we get to it, the flag names may be slightly different. At the very least, we won't be using underscores.

go/clustermetadata/topo/test/directory.go
```diff
@@ -0,0 +1,208 @@
...
+)
+
+// checkDirectory tests the directory part of the topo.Conn API.
+func checkDirectory(t *testing.T, ctx context.Context, ts topo.Store) {
+	// global topo
+	t.Logf("===   checkDirectoryInCell global")
+	conn, err := ts.ConnForCell(ctx, topo.GlobalTopo)
+	if err != nil {
+		t.Fatalf("ConnForCell(global) failed: %v", err)
+	}
```
deepthi: This whole test file should be refactored to use require and assert instead of Fatalf and Errorf.

go/clustermetadata/topo/test/file.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+// checkFile tests the file part of the Conn API.
+func checkFile(t *testing.T, ctx context.Context, ts topo.Store) {
+	// global cell
+	t.Logf("===   checkFileInCell global")
+	conn, err := ts.ConnForCell(ctx, topo.GlobalTopo)
+	if err != nil {
+		t.Fatalf("ConnForCell(global) failed: %v", err)
+	}
```
deepthi: Replace with require/assert.

go/clustermetadata/topo/test/lock.go
```diff
@@ -0,0 +1,371 @@
...
+	t.Log("===      checkTryLockUnblocks")
+	checkTryLockUnblocks(ctx, t, conn)
+}
+
+// checkTryLockTimeout test the fail-fast nature of TryLock
+func checkTryLockTimeout(ctx context.Context, t *testing.T, conn topo.Conn) {
+	databasePath := path.Join(topo.DatabasesPath, "test_database")
+	lockDescriptor, err := conn.TryLock(ctx, databasePath, "")
+	if err != nil {
+		require.Fail(t, "TryLock failed", err.Error())
```
deepthi: Can we rewrite these? 
```
require.NoError(t, err, "TryLock failed with:%v", err.Error())
```

go/clustermetadata/topo/test/lock.go
```diff
@@ -0,0 +1,371 @@
...
+	}()
+
+	waitUntil := time.Now().Add(10 * time.Second)
+	var firstTime = true
+	// after attempting the `TryLock` and getting an error `NodeExists`, we will cancel the context deliberately
+	// and expect `context canceled` error in next iteration of `for` loop.
+	for {
+		if time.Now().After(waitUntil) {
+			t.Fatalf("Unlock(test_database) timed out")
+		}
```
deepthi: One more Fatalf that can be rewritten.

go/clustermetadata/topo/test/topo_test_suite.go
```diff
@@ -0,0 +1,86 @@
...
+
+// LocalCellName is the cell name used by this test suite.
+const LocalCellName = "test"
+
+// TopoServerTestSuite runs the full topo.Server/Conn test suite.
+// The factory method should return a topo.Server that has a single cell
+// called LocalCellName.
+// Not all tests are applicable for each Topo server, therefore we provide ignoreList in order to
+// avoid them for given Topo server tests. For example `TryLock` implementation is same as `Lock` for some Topo servers.
+// Hence, for these Topo servers we ignore executing TryLock Tests.
```
deepthi: This isn't true any more for TryLock, is it?

proto/mtrpc.proto
```diff
@@ -0,0 +1,191 @@
...
```
deepthi: Do we really need to reserve this? Can't we just use 1 and 2?

deepthi (approved):
Some minor changes are needed. Rest LGTM

go/clustermetadata/topo/cell_location.go
```diff
@@ -0,0 +1,156 @@
...
+// CellLocation records are not meant to be changed while the system is
+// running.  In a running system, a CellLocation can be added, and
+// topology server implementations should be able to read them to
+// access the cells upon demand. Topology server implementations can
+// also read the available CellLocation at startup to build a list of
+// available cells, if necessary. A CellLocation can only be removed if no
+// Shard record references the corresponding cell in its Cells list.
+
+func pathForCellLocation(cell string) string {
+	return path.Join(CellsPath, cell, CellLocationFile)
```
GuptaManan100: Instead of CellLocation, I think we can call it just Cell i.e. CellFile instead of CellLocationFile. It can to begin with only store the cell location, but it might store something more too in which case we're better off having a more generic name

GuptaManan100 (approved):
Rest LGTM!

go/clustermetadata/topo/memorytopo/lock.go
GuptaManan100: I got the answer by looking at the interface, never mind

go/clustermetadata/topo/store.go
rafael: oh probably not. Going to remove the prefix in the paths.

go/clustermetadata/topo/test/topo_test_suite.go
rafael: oh yeah. I actually implemented, I did not like that pattern of skipping some tests. Seemed incidental complexity. Will remove this comment.

proto/mtrpc.proto
rafael: oh lol. My bad

go/clustermetadata/topo/test/lock.go
rafael: Makes sense. Updated

GuptaManan100 (approved):

PR #61, opened 2025-08-20
Initial pgctld implementation

deepthi:
pgctld - PostgreSQL Control Daemon for Multigres

# Overview

 pgctld is a comprehensive PostgreSQL lifecycle management daemon designed for the Multigres cluster. It provides both CLI and gRPC interfaces for managing PostgreSQL server instances.

##  Command-Line Interface (CLI)

  - Commands Available:
    - start - Initialize and start PostgreSQL server
    - stop - Stop PostgreSQL server with configurable shutdown modes (smart/fast/immediate)
    - restart - Restart PostgreSQL server
    - status - Check PostgreSQL server status and health
    - reload-config - Reload PostgreSQL configuration without restart
    - server - Start gRPC server for remote management

##  gRPC Service Interface

  - Service Methods:
    - Start() - Start PostgreSQL with configuration parameters
    - Stop() - Stop PostgreSQL with shutdown mode options
    - Restart() - Restart PostgreSQL server
    - Status() - Get detailed server status information
    - ReloadConfig() - Reload server configuration
    - Version() - Get PostgreSQL version information
    - InitDataDir() - Initialize PostgreSQL data directory

### PostgreSQL Lifecycle Management

  - Automatic Initialization: Runs initdb if data directory is not initialized
  - Process Monitoring: Tracks PostgreSQL processes via PID files with real process validation
  - Graceful Shutdown: Supports smart, fast, and immediate shutdown modes
  - Health Checking: Uses pg_isready for connection readiness verification
  - Configuration Management: Supports custom config files and runtime parameter overrides

### Configuration Management

  - Flexible Configuration: Supports Viper-based configuration with YAML files
  - CLI Flag Override: Command-line flags override configuration file settings
  - gRPC Parameter Passing: Request-specific parameters override defaults
  - Environment Variables: Supports PGCTLD_* environment variable prefix

 ### TODO
- generate random port numbers to avoid conflicts ✅ 
- fix CI setup so that e2e tests can run

### Future Enhancements
- Generate postgres config files instead of requiring them to be provided
- config-file for the binary to be passed in as a startup option instead of being hard-coded. We want to do this in one place and have all the binaries use it.
- Move RPCs behind an interface, and move to a shared package outside cmd/pgctld


go/pb/pgctld_grpc.pb.go
```diff
@@ -0,0 +1,382 @@
...
+
+// UnsafePgCtldServer may be embedded to opt out of forward compatibility for this service.
+// Use of this interface is not recommended, as added methods to PgCtldServer will
+// result in compilation errors.
+type UnsafePgCtldServer interface {
+	mustEmbedUnimplementedPgCtldServer()
+}
+
+func RegisterPgCtldServer(s grpc.ServiceRegistrar, srv PgCtldServer) {
+	// If the following call pancis, it indicates UnimplementedPgCtldServer was
```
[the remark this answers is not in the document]
deepthi: This is coming from the protoc compiler via code generation. Not something we can fix.

go/cmd/pgctld/command/stop.go
```diff
@@ -0,0 +1,171 @@
...
+func stopPostgreSQL(dataDir, mode string, timeout int) error {
+	// Legacy function - use defaults for other config
+	config := NewPostgresConfigFromDefaults()
+	config.DataDir = dataDir
+	config.Timeout = timeout
+	return stopPostgreSQLWithConfig(config, mode)
+}
+
+func stopPostgreSQLWithConfig(config *PostgresConfig, mode string) error {
+	// First try using pg_ctl
```
NikolayS, line 139: Before this, it makes sense to issue an explicit CHECKPOINT so shutdown checkpoint becomes fast -- cc @sougou

go/cmd/pgctld/command/stop.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Wait for process to exit
+	for i := 0; i < config.Timeout; i++ {
+		if !isProcessRunning(pid) {
+			return nil
+		}
+		time.Sleep(1 * time.Second)
+	}
+
+	// If still running after timeout, force kill for immediate mode
```
NikolayS: Uh, I would be really unhappy if this happened with my cluster.

GuptaManan100 (commented):
So we are doing two ways of using the pgctld binary? A server mode which accepts RPCs and also the command mode that does the work for the command and exits? Like `pgctld start`. Do we really need both modes?

GuptaManan100 (approved):
otherwise it looks good to me

.github/workflows/test.yml
```diff
@@ -32,6 +32,26 @@ jobs:
...
+        sudo systemctl start postgresql
+        sudo systemctl enable postgresql
+        # Create postgres user and set up basic auth
+        sudo -u postgres createuser --superuser $USER || true
+        sudo -u postgres psql -c "ALTER USER $USER CREATEDB;" || true
+
+    - name: Verify PostgreSQL Installation
+      run: |
+        which initdb
+        which postgres
```
thelovekesh: I think `initdb` and `postgres` binaries are not added to path since it's db version specific and can be found in `/usr/lib/postgresql/${version}/bin/`

.github/workflows/test.yml
```diff
@@ -32,6 +32,26 @@ jobs:
       with:
         go-version: '1.25.0'
 
+    - name: Install PostgreSQL
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y postgresql postgresql-contrib
```
thelovekesh: A better way to install it could be to setup PostgreSQL ppa and download it from there. It could be possible that official apt repo is behind by many patches you want to test or could be outdated.

With PostgreSQL ppa, you can also run the tests with different pg version by using matrix and downloading them by version like `postgresql-${version}`

.github/workflows/test.yml
```diff
@@ -32,6 +32,26 @@ jobs:
...
+        sudo systemctl start postgresql
+        sudo systemctl enable postgresql
```
thelovekesh: I think this is not required given this is `ephemeral` instance.

go/cmd/pgctld/command/e2e_test.go
```diff
@@ -0,0 +1,345 @@
+// Copyright 2025 The Multigres Authors.
```
rafael, line 1: nit - about the location. For topo, I created a topo/test package that has the full topo test suite. In spirit similar to the e2e_test.go. 

Could we do something similar? so we have `commnand/test/e2e_test.go`. 

So we have a consistent pattern

go/cmd/pgctld/command/e2e_test.go
```diff
@@ -0,0 +1,345 @@
...
+		// but we can at least verify the server starts without errors
+
+		// Check if server process is still running (not crashed)
+		if serverCmd.Process != nil {
+			// Check if process is still alive by checking if ProcessState is nil
+			// If the process has exited, ProcessState will be non-nil
+			if serverCmd.ProcessState == nil {
+				t.Log("gRPC server is running successfully")
+			} else {
+				t.Errorf("gRPC server process died: exit code %d", serverCmd.ProcessState.ExitCode())
```
rafael: should this be assert / require ?

go/cmd/pgctld/command/e2e_test.go
```diff
@@ -0,0 +1,345 @@
...
+		require.NoError(t, err)
+		defer func() {
+			if serverCmd.Process != nil {
+				_ = serverCmd.Process.Kill()
+				_ = serverCmd.Wait()
+			}
+		}()
+
+		// Give server time to start
+		time.Sleep(3 * time.Second)
```
rafael: nit - we could refactor this so is not a full sleep, we can do something like:
```
deadline := time.Now().Add(5 * time.Second)

for time.Now().Before(deadline) {
	if serverCmd.Process != nil {
		// Check if still alive
		if serverCmd.ProcessState == nil {
			t.Log("gRPC server is running successfully")
			return
		}
		t.Fatalf("gRPC server process died: exit code %d", serverCmd.ProcessState.ExitCode())
	}
	time.Sleep(100 * time.Millisecond)
}

t.Fatal("timeout: gRPC server process was never started")
```

So if it starts earlier, the test runs faster :D

go/cmd/pgctld/command/e2e_test.go
```diff
@@ -0,0 +1,345 @@
...
+
+		// Test basic gRPC connectivity by checking if the server is listening
+		// We can't easily test full gRPC functionality without importing the pb package
+		// but we can at least verify the server starts without errors
+
+		// Check if server process is still running (not crashed)
+		if serverCmd.Process != nil {
```
rafael: Is this missing the else clause? What happens if serverCmd.Process is nil for some reason?

.github/workflows/test.yml
deepthi: thanks for that pointer!

go/cmd/pgctld/command/stop.go
deepthi: Thanks for the feedback. What is the best practice for immediate shutdown?

proto/pgctld.proto
```diff
@@ -0,0 +1,211 @@
+// Copyright 2025 The Multigres Authors
```
rafael: should we name thiss  file `pgctldservice.proto`? So we keep them seperate from the messages

go/cmd/pgctld/command/stop.go
hunleyd: `pg_ctl -m immediate` @deepthi is how you'd do an 'immediate' from the cli in a PG world

.github/workflows/test.yml
hunleyd: and `/usr/pgsql-${version}/bin` on RPM-based distros

.github/workflows/test.yml
hunleyd: yes, the official repo is always best. See https://wiki.postgresql.org/wiki/Apt

go/cmd/pgctld/command/server.go
```diff
@@ -0,0 +1,355 @@
...
+
+	// Create config from request parameters
+	config := NewPostgresConfigFromStartRequest(req)
+
+	if config.DataDir == "" {
+		return nil, fmt.Errorf("data-dir is required")
+	}
+
+	// Check if PostgreSQL is already running before attempting to start
+	if isPostgreSQLRunning(config.DataDir) {
```
rafael: It looks like this is symmetric to the implementation of the start command, but with some subtle differences. For example, here we call readPostmasterPID, while in the other one we don’t. Is that intentional? If so, could we add comments explaining how they differ?

For the parts that are the same, what do you think about encapsulating more of the Start functionality so that both server and CLI can share as much as possible? (or another idea, what if from the get-go, the cli just makes the RPC call. So they are always a very thin layer).

proto/pgctld.proto
```diff
@@ -0,0 +1,211 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package pgctld;
+
+option go_package = "github.com/multigres/multigres/go/pb";
+
+// PostgreSQL Control Service
+service PgCtld {
```
rafael: These methods were not exposed via CLI (I think):  

rpc Status(StatusRequest) returns (StatusResponse);
rpc Version(VersionRequest) returns (VersionResponse);
rpc InitDataDir(InitDataDirRequest) returns (InitDataDirResponse);

rafael (approved):
Some minor observations on my end.  One more QoL is to make sure we add installing the required tools as part of make tools, so the environment is setup automatically for developers. 

Stamping, I think we can iterate.

.github/workflows/test.yml
thelovekesh: By this, I only mean this command - `sudo systemctl enable postgresql`

.github/workflows/test.yml
samrose: I can help you convert this (doesn't have to be immediately) to use the exactly sources and packages for https://github.com/supabase/postgres + our configurations, migrations, etc probably isn't urgent just to keep moving forward. But I'll help when ready!

.github/workflows/test.yml
```diff
@@ -16,24 +16,63 @@ name: Tests
...
+          go-version: "1.25.0"
 
-    - name: Build
-      run: make build
+      - name: Install PostgreSQL
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y postgresql postgresql-contrib postgresql-client-common postgresql-common
+          sudo systemctl start postgresql
+          sudo systemctl enable postgresql
```
hunleyd: these can be combined into `sudo systemctl enable --now postgresql`

.github/workflows/test.yml
```diff
@@ -16,24 +16,63 @@ name: Tests
...
+          # Create postgres user and set up basic auth
+          sudo -u postgres createuser --superuser $USER || true
+          sudo -u postgres psql -c "ALTER USER $USER CREATEDB;" || true
```
hunleyd: why are these separate instead of `sudo -u postgres createuser --superuser --createdb $USER || true` ?

go/cmd/pgctld/command/stop.go
deepthi: That's what we do already in `stopWithPgCtl`. This signal-based fallback of killing the server is not really required, so I've removed it.

proto/pgctld.proto
deepthi: Done

go/cmd/pgctld/command/e2e_test.go
deepthi: go best practices are that unit tests live with the code. for endtoend tests, it is debatable whether they should go under the top level test directory instead of in each package. In future, when we want to start tracking code coverage, it will be better to keep them all alongside code.

go/cmd/pgctld/command/start.go
```diff
@@ -0,0 +1,371 @@
...
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to start PostgreSQL with pg_ctl: %w", err)
+	}
+
+	// Wait for PostgreSQL to be ready using pg_isready
+	return waitForPostgreSQLWithConfig(config)
+}
+
+// isPostgreSQLReady checks if PostgreSQL is ready to accept connections
+// by reading the postgresql.log file and checking for "ready" status
```
hunleyd: it's better to use the `pg_isready` binary for this. see [here](https://www.postgresql.org/docs/current/app-pg-isready.html)

go/cmd/pgctld/command/e2e_test.go
deepthi: your suggestion above takes care of this case too.

go/cmd/pgctld/command/start.go
thelovekesh: there also a exist a function `waitForPostgreSQLWithConfig()` which uses `pg_ready`. not sure if the function using log file to check status is even being used anywhere.

thelovekesh: Oh but that one only waits to be ready. I think the functionality to check if pg is ready can be abstracted away in a different function to be used at both places.

.github/workflows/test.yml
thelovekesh: why even need to enable it given github action runner only comes online once and destroyed after that, so running `systemctl enable` doesn't makes much sense IMO.

.github/workflows/test.yml
deepthi: I looked at the PostgreSQL ppa. It seems to be available only until Jammy (ubuntu 22.04). This is where I'm looking, so let me know if there's some other location to check
https://www.ubuntuupdates.org/ppa/postgresql
As hunleyd said, the preference is to download from official apt repo as documented in the project docs.

go/cmd/pgctld/command/server.go
deepthi: The CLI command doesn't return anything, it only logs. Do you think it's useful to log the PID of the running postgres?
The point about refactoring is good. I'll take care of that.

proto/pgctld.proto
deepthi: We do have status, but not the other two. Do you think they are useful? it's easy to add them.

go/cmd/pgctld/command/server.go
deepthi: I have implemented the refactor.

.github/workflows/test.yml
hunleyd: thats fair

proto/pgctld.proto
rafael: We can come back if we need to. I just noticed and flag it. After thinking about it, not obvious if they will be useful .

GuptaManan100 (approved):
LGTM

proto/pgctld.proto
deepthi: I've added these two.

deepthi:
> So we are doing two ways of using the pgctld binary? A server mode which accepts RPCs and also the command mode that does the work for the command and exits? Like `pgctld start`. Do we really need both modes?

Yes. server mode is for production deployments and CLI mode is for local / dev deployments.

.github/workflows/test.yml
deepthi: we don't even need $USER, we can do everything using the postgres superuser, so I'll remove these two lines.

.github/workflows/test.yml
deepthi: actually they are being added to the path in line 52 and line 68

go/cmd/pgctld/command/start.go
deepthi: removed the log check and used pg_isready everywhere.

.github/workflows/test.yml
```diff
@@ -16,24 +16,58 @@ name: Tests
...
-    branches: [ main ]
+    branches: [main]
 
 jobs:
   test:
     runs-on: ubuntu-24.04
 
     steps:
-    - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
```
rafael, line 28: one idea here: why don't we start splitting these into separate actions and we leave the main one just for the short tests. 

I started doing that in the topo etcd [PR](https://github.com/multigres/multigres/pull/72)

deepthi: agreed. right now everything runs fast, so I left them all in.

PR #62, opened 2025-08-20
topo: Adds CRUD operations for multipoolers in the cells topo

rafael:
# Desc
* Straight forward CRUD operations for the multipooler.
* I'll follow up with adding the same for gateways and multiorchs. 
# Tests
* Added unit tests.


proto/clustermetadata.proto
```diff
@@ -138,9 +138,6 @@ message MultiOrch {
 
 // ID is a globally unique pooler identifier.
 message ID {
-  // component that this ID is used for (i.e multigateway / multiorch / multipooler)
```
rafael, line 141: I'll rework this on our discussion to have the id be: `component-cell-string`. That can be a follow up PR.

go/mterrors/mterrors.go
```diff
@@ -287,6 +287,10 @@ type wrapping struct {
 func (w *wrapping) Error() string { return w.msg + ": " + w.cause.Error() }
 func (w *wrapping) Cause() error  { return w.cause }
 
+// Unwrap implements Go's standard error unwrapping interface.
+// This allows errors.Is() and errors.As() to work with wrapped errors.
+func (w *wrapping) Unwrap() error { return w.cause }
```
rafael, line 292: Without this, wrapped errors don't work when you do `errors.Is`

deepthi (approved):

PR #72, opened 2025-08-26
topo: Adds etcd2 topo implementation

rafael:
## Desc

Adds etcd topology implementation for cluster metadata management. This is a port of the Vitess implementation. No changes here.

In addition to that, some minor changes to GH actions setup to divide short and integration tests in separate actions

### DIFF
--- a/.github/workflows/test-short.yml
+++ b/.github/workflows/test-short.yml
@@ -33,11 +33,8 @@ jobs:
         with:
           go-version: "1.25.0"
 
-      - name: Install PostgreSQL
-        run: |
-          sudo apt-get update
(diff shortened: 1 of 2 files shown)


.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,80 @@
...
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  test-integration:
+    name: Cluster Metadata Tests
```
deepthi: Now that this also includes pgctld tests, this name isn't right any more.

PR #75, opened 2025-08-27
topo: add CRUD for multiorch

deepthi:
Added CRUD operations for MultiOrch in the topo with unit tests. This follows the same patterns as MultiPooler and MultiGateway.


rafael (approved):
Makes sense. I did not look too closely the tests, assuming we have parity with what implemented in the other ones. 

LGTM

PR #79, opened 2025-09-04
adds more tests and fixes issue with start connectivity errors

rafael:
# Desc
* Small PR following up in some small issues from the start / stop in previous PR
* Adds more tests

### DIFF
--- a/go/provisioner/local/local.go
+++ b/go/provisioner/local/local.go
@@ -738,14 +738,6 @@ func (p *localProvisioner) provisionMultigateway(ctx context.Context, req *provi
 		return nil, fmt.Errorf("multigateway process validation failed: %w", err)
 	}
 
-	// Wait for multigateway to be ready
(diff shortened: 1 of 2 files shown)


go/provisioner/local/local.go
```diff
@@ -763,6 +755,14 @@ func (p *localProvisioner) provisionMultigateway(ctx context.Context, req *provi
 		fmt.Printf("Warning: failed to save service state: %v\n", err)
 	}
 
+	// Wait for multigateway to be ready
```
rafael, line 758: @deepthi  - This makes sure that this connectivity checks happen after the process state file is saved. It should fix the issue you found

This is tricky to add an integration test though. Do you have thoughts on that?

deepthi: Agree that it is tricky. For now it should be fine to simulate the failure locally and make sure it works.

deepthi (approved):

PR #80, opened 2025-09-04
tools: install etcd

deepthi:
`make tools` will now also install etcd so that there's no need to do it manually.
Cluster bootstrap expects etcd in the path. We symlink etcd into multigres/bin, so that it's available from the same place as all other binaries.
It is necessary to add ./bin from the multigres root directory for this version of etcd to be detected and used.

### DIFF
--- a/build.env
+++ b/build.env
@@ -12,3 +12,4 @@ export PROTOC_VER=25.1
 export ADDLICENSE_VER=v1.2.0
 export PROTOC_GEN_GO_VER=v1.36.7
 export PROTOC_GEN_GO_GRPC_VER=v1.3.0
+export ETCD_VER=v3.6.4
--- a/tools/setup_build_tools.sh
+++ b/tools/setup_build_tools.sh
@@ -25,6 +25,7 @@ PROTOC_VERSION="$PROTOC_VER"
 ADDLICENSE_VERSION="$ADDLICENSE_VER"
 PROTOC_GEN_GO_VERSION="$PROTOC_GEN_GO_VER"
 PROTOC_GEN_GO_GRPC_VERSION="$PROTOC_GEN_GO_GRPC_VER"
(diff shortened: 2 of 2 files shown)


tools/setup_build_tools.sh
```diff
@@ -83,27 +87,68 @@ install_dep() {
...
+
+    case $(uname) in
+        Linux)  local platform=linux; local ext=tar.gz;;
+        Darwin) local platform=darwin; local ext=zip;;
+        *)   echo "ERROR: unsupported platform for etcd"; exit 1;;
+    esac
+
+    case $(uname -m) in
+        aarch64)  local arch=arm64;;
+        x86_64)  local arch=amd64;;
```
dweitzman, line 124: Gotta have glasses on to spot the different between amd64 and arm64!

dweitzman (approved):

dweitzman:
Could be nice at some point to run shellcheck on bash scripts as a linter, since bash has so many footguns

deepthi:
> Could be nice at some point to run shellcheck on bash scripts as a linter, since bash has so many footguns

I can add a hook for it to run locally, and add it into CI as well (in a follow up PR).

build.env
```diff
@@ -11,4 +11,5 @@ ln -sf "$PWD/misc/git/commit-msg" .git/hooks/commit-msg
 export PROTOC_VER=25.1
 export ADDLICENSE_VER=v1.2.0
 export PROTOC_GEN_GO_VER=v1.36.7
-export PROTOC_GEN_GO_GRPC_VER=v1.3.0
\ No newline at end of file
+export PROTOC_GEN_GO_GRPC_VER=v1.3.0
+export ETCD_VER=v3.6.4
```
rafael, line 15: For convenience, we could consider adding MTROOT/bin to the path in this script. That way people don't have to remember to add this directory to the path.

PR #83, opened 2025-09-08
pgctld: bootstrap preparation

rafael:
# Desc
We are getting ready to hook pgctld into the cluster bootstrap. As part of this work, making these changes:
- Renaming some flags to make it clearer that they are related to postgres configs
- Making sure that pgtctld uses servenv correctly. 
- Cleans up some files that were  not being used.

### DIFF
--- a/go/cmd/pgctld/command/start.go
+++ b/go/cmd/pgctld/command/start.go
@@ -148,6 +148,14 @@ Examples:
 func runStart(cmd *cobra.Command, args []string) error {
 	config := NewPostgresConfigFromDefaults()
 
+	// Override with command-specific flags if provided
+	if cmd.Flags().Changed("pg-socket-dir") {
+		config.SocketDir, _ = cmd.Flags().GetString("pg-socket-dir")
+	}
+	if cmd.Flags().Changed("pg-config-file") {
+		config.ConfigFile, _ = cmd.Flags().GetString("pg-config-file")
+	}
+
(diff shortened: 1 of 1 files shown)


go/cmd/pgctld/command/restart.go
```diff
@@ -100,15 +100,15 @@ func RestartPostgreSQLWithResult(config *PostgresConfig, mode string) (*RestartR
...
 
 func runRestart(cmd *cobra.Command, args []string) error {
-	config := NewPostgresConfigFromViper()
+	config := NewPostgresConfigFromDefaults()
 	mode, _ := cmd.Flags().GetString("mode")
 
 	// Override with command-specific flags if provided
-	if cmd.Flags().Changed("socket-dir") {
-		config.SocketDir, _ = cmd.Flags().GetString("socket-dir")
+	if cmd.Flags().Changed("pg-socket-dir") {
```
rafael, line 107: I don't think we need this anymore. Because it will be either the default or the value from Viper. But would love extra pair of eyes before removing.

go/viperutil/config.go
```diff
@@ -51,7 +51,7 @@ var (
 	configType = Configure(
 		"config.type",
 		Options[string]{
-			EnvVars:  []string{"VT_CONFIG_TYPE"},
+			EnvVars:  []string{"MT_CONFIG_TYPE"},
```
deepthi, line 54: Oops 😅

deepthi (approved):

PR #88, opened 2025-09-10
Multiadmin server, RPCs and CLI

deepthi:
Initial implementation of multiadmin server.
- Two RPCs GetCell and GetDatabase
- CLI commands for the same
- Cluster init/start/stop all include multiadmin now.
- Cluster start writes database metadata into topo


go/provisioner/local/local.go
```diff
@@ -774,6 +789,138 @@ func (p *localProvisioner) provisionMultigateway(ctx context.Context, req *provi
...
+		"--topo-global-server-addresses", etcdAddress,
+		"--topo-global-root", topoGlobalRoot,
+		"--topo-implementation", topoBackend,
+		"--log-level", logLevel,
+		"--log-output", logFile,
+		"--service-map", "grpc-multiadmin",
+	}
+
+	// Start multiadmin process
+	multiadminCmd := exec.CommandContext(context.Background(), multiadminBinary, args...)
```
deepthi: This should use the passed in context. I'll fix it in the next commit.

go/cmd/multiadmin/internal/server/server.go
```diff
@@ -0,0 +1,117 @@
+/*
```
rafael: nit - I think we landed on using `#` for the license. Maybe we can do one PR where we fix all licenses and also add a pre-commit hook / linter so doesn't sneak in.

go/cmd/multiadmin/internal/server/server.go
```diff
@@ -0,0 +1,117 @@
...
+Copyright 2025 The Multigres Authors.
```
rafael: Another nit - that we can fast follow. We should align on the layout. I wonder if internal, should be a multiadmin component at the top level and be outside cmd. 

That's what I started to do with some things in `pgctld`. 

Slight preference towards having at at the top level, otherwise I think the `cmd` start having too much logic that is not strictly related to the cmd.

go/cmd/multigres/command/topo/getcell.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if adminServerFromConfig == "" {
+		return "", fmt.Errorf("either --admin-server flag or --config-path with multiadmin configuration must be provided")
+	}
+
+	return adminServerFromConfig, nil
+}
+
+// getAdminServerFromConfig extracts the multiadmin server address from config
+func getAdminServerFromConfig(configPaths []string) (string, error) {
```
rafael, line 83: We could have the admin-server address at the top level:
```
cli:
  admin-server: 'init_sets_this_according_to_the_provisioner'
provisioner: local
provisioner-config:
  ...
```

Later we can figure out how this is populated for other provisioners, but for now we can leave it by convention. Thinking that it will be better that we don't couple the provisioner config with configuring other parts of the CLI.

go/test/startup_test.go
```diff
@@ -73,6 +74,13 @@ func testBinaryStartupShutdown(t *testing.T, binaryName, port string) {
 			"--topo-global-root", "/",
 			"--topo-implementation", "memory",
 			"--log-level", "info")
+	case "multiadmin":
```
rafael, line 77: I'm starting to think we can kill this test. The e2e cluster covers this and as the binaries start to become more complex, there is some overlap here.

go/cmd/multiadmin/internal/server/server.go
deepthi: Right. We can do that as a separate PR.

go/cmd/multiadmin/internal/server/server.go
deepthi: The way I'm thinking about it is that if something is in cmd, no one else can depend on it. If some other module needs to depend on this code, then it should move out to a separate package under go/
Do we expect anything else to include the server implementation? or even interface? I don't think so.
The protobufs are available for any callers of the RPCs.

go/cmd/multigres/command/topo/getcell.go
deepthi: That's a good point. Though Multiadmin is not just for the CLI. And this point applies to etcd as well.
It's a bit awkward that provisioner is the one creating Multiadmin though. So if I move it out of provisioner-config, it may not work to start/stop it using the provisioner. Any ideas how we can get around this?

go/cmd/multiadmin/internal/server/server.go
deepthi: I'll create go/admin and move the internal/server package into it. When we get to writing more code for gateway and pooler we can create those packages. Note that I'm not prefixing the package names with `multi` :)

go/test/startup_test.go
deepthi: Maybe we should move the other e2e and integration tests under go/test instead. Otherwise this module will have only utils (which are probably used only in startup_test.go)

PR #93, opened 2025-09-11
feat: add CLI commands and associated RPCs to get various entities from topo

deepthi:
Following up on #88 with more CLI commands and RPCs.
- Fixed a bug with multiorch not being registered properly in the topo
- Write database into MultiPooler object in topo
- Add database as a multipooler config property
- Add zone2 as a cell to local provisioner and bring up all components in zone2 during cluster start
- test all commands with cell filtering and database filtering
- Available commands from help output
```
Available Commands:
  cluster          Manage cluster lifecycle
  completion       Generate the autocompletion script for the specified shell
  getcell          Get information about a specific cell
  getcellnames     Get all cell names in the cluster
  getdatabase      Get information about a specific database
  getdatabasenames Get all database names in the cluster
  getgateways      Get gateways filtered by cells
  getorchs         Get orchestrators filtered by cells
  getpoolers       Get poolers filtered by cells and/or database
  help             Help about any command
```

Note: If you have already generated a cluster config, you should delete and regenerate it to get the new config with two cells.


go/cmd/multigres/command/cluster/e2e_test.go
```diff
@@ -153,7 +153,7 @@ func createTestConfigWithPorts(tempDir string, portConfig *testPortConfig) (stri
 	binPath := filepath.Join(tempDir, "bin")
 	localConfig := &local.LocalProvisionerConfig{
 		RootWorkingDir: tempDir,
-		DefaultDbName:  "default",
+		DefaultDbName:  "postgres",
```
deepthi: @rafael does this make sense?

go/test/startup_test.go
```diff
@@ -67,13 +66,6 @@ func testBinaryStartupShutdown(t *testing.T, binaryName, port string) {
 	switch binaryName {
 	case "pgctld":
 		cmd = exec.CommandContext(ctx, binaryPath, "--grpc-port", port, "--log-level", "info")
-	case "multiorch":
```
deepthi, line 68: Removed multiorch from here because it also needs a `--cell` to be specified. We can delete this test in a follow up PR.

go/cmd/multipooler/main.go
```diff
@@ -146,6 +153,7 @@ func run(cmd *cobra.Command, args []string) error {
 		multipooler := topo.NewMultiPooler(serviceID, cell, hostname)
 		multipooler.PortMap["grpc"] = int32(servenv.GRPCPort())
 		multipooler.PortMap["http"] = int32(servenv.HTTPPort())
+		multipooler.Database = database
```
deepthi, line 154: This actually writes the Database into topo, which we were not doing earlier.

go/admin/server/server.go
```diff
@@ -115,3 +119,171 @@ func (s *MultiAdminServer) GetDatabase(ctx context.Context, req *multiadminpb.Ge
...
+			return nil, status.Errorf(codes.Internal, "failed to retrieve cell names: %v", err)
+		}
+		cellsToQuery = allCells
+	}
+
+	var allPoolers []*clustermetadatapb.MultiPooler
+
+	// Query each cell for poolers
+	for _, cellName := range cellsToQuery {
+		poolerInfos, err := s.ts.GetMultiPoolersByCell(ctx, cellName, nil)
```
rafael: This method already accepts options that will filter by Database. So we can push down the filter there.

go/admin/server/server_test.go
```diff
@@ -152,3 +152,350 @@ func TestMultiAdminServerGetDatabase(t *testing.T) {
...
+	t.Run("get gateways with empty topology", func(t *testing.T) {
+		req := &multiadminpb.GetGatewaysRequest{}
+		resp, err := server.GetGateways(ctx, req)
+
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+		assert.Empty(t, resp.Gateways)
+	})
+
+	t.Run("get gateways filtered by non-existent cell", func(t *testing.T) {
```
rafael, line 241: should this error or return empty? I could go either way.

go/cmd/multigres/command/topo/getgateways.go
```diff
@@ -0,0 +1,94 @@
...
+// runGetGateways handles the getgateways command
+func runGetGateways(cmd *cobra.Command, args []string) error {
+	// Get admin server address
+	adminServer, err := getAdminServerAddress(cmd)
+	if err != nil {
+		return err
+	}
+
+	// Get flag values
+	cellsFlag, _ := cmd.Flags().GetString("cells")
```
rafael: should we validate this?

go/cmd/multipooler/main.go
rafael: Good catch

go/admin/server/server.go
deepthi: Oh nice, I'll fix it up.

go/admin/server/server_test.go
deepthi: Empty is fine, I think. Could be user error.

go/cmd/multigres/command/topo/getgateways.go
deepthi: good point. I'll add that.

go/admin/server/server_test.go
deepthi: On second thought, I'm going to collect and return any errors with partial results. That is better also for the scenario where one of the cells is unreachable from the admin server.

PR #94, opened 2025-09-11
pgctld: explicit init, unified config + validation

rafael:
This refactor removes auto-initialization and introduces a required `pgctld init` step, making behavior more predictable and safer.

Key changes:
- `pgctld start` no longer auto-inits; all commands validate init state
- Centralized postgres config system (150+ settings, template-backed)
- Shared validation across CLI + gRPC, clear NOT_INITIALIZED errors
- Server refactor to single config object; added `--pg-port` flag

Breaking: all commands now require explicit `pgctld init` before use.

### DIFF
--- a/go/cmd/pgctld/command/e2e_test.go
+++ b/go/cmd/pgctld/command/e2e_test.go
@@ -412,7 +412,7 @@ func TestPostgreSQLAuthentication(t *testing.T) {
 
 	// Check if PostgreSQL binaries are available
 	if !hasPostgreSQLBinaries() {
-		t.Skip("PostgreSQL binaries not found, skipping authentication test")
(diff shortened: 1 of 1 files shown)


config/postgres/template.cnf
```diff
@@ -0,0 +1,307 @@
...
+max_parallel_maintenance_workers = {{.MaxParallelMaintenanceWorkers}}	# taken from max_parallel_workers
+
+
+#------------------------------------------------------------------------------
+# WRITE-AHEAD LOG
+#------------------------------------------------------------------------------
+
+# - Settings -
+
+wal_level = logical			# minimal, replica, or logical
```
deepthi, line 89: We should make this `replica`.

PR #112, opened 2025-09-17
Pgctld bootstrap local

rafael:
# Description
* This integrates into the local bootstrap pctctld. With this, we technically have all the core components starting and stopping locally. They don't anything, but hey, it boots. 
* This pgctld setup, integrates the postgres bootstrap. It starts two postgres instances in local (one for each pooler). They still don't have replication set up. Just two vanilla postgres. 
# Desc
* Integration test just works. Added few more assertions to make sure postgres is looking operational. Tested this in my local as well.

### DIFF
--- a/go/provisioner/local/healthcheck.go
+++ b/go/provisioner/local/healthcheck.go
@@ -1,0 +1,126 @@
+// Copyright 2025 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


go/provisioner/local/healthcheck.go
```diff
@@ -0,0 +1,132 @@
+// Copyright 2025 Supabase, Inc.
```
rafael, line 1: This mostly moves some functions from `local.go` to this file + it adds pgctld health check integration for local

go/test/endtoend/cluster_test.go
```diff
@@ -569,7 +603,7 @@ func TestInitCommand(t *testing.T) {
 		{
 			name: "successful init with multiple valid paths",
 			setupDirs: func(t *testing.T) ([]string, func()) {
-				tempDir1, err := os.MkdirTemp("", "multigres_init_test1")
+				tempDir1, err := os.MkdirTemp("/tmp/", "mlt")
```
deepthi: nit: the trailing / is not required. `/tmp` works fine.

PR #114, opened 2025-09-19
Fix typo and use slices.Contains for cell check

cxljs:
- Used slices.Contains for cell check in cellsInTopo.
- Fixed comment typos.

Multigres is a great project. I read the code tonight (except for the parser) and made minor changes. I checked the Discussion #6 , so if this PR isn't helpful, feel free to close it.

### DIFF
--- a/go/clustermetadata/topo/store.go
+++ b/go/clustermetadata/topo/store.go
@@ -461,7 +461,7 @@ func (ts *store) Close() error {
 	}
 
 	// Clear the map to release references
-	ts.cellConns = nil
+	ts.cellConns = make(map[string]cellConn)
 
 	// Return combined error if any occurred during cleanup
 	if len(errs) > 0 {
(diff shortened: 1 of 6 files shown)


go/clustermetadata/topo/store.go
```diff
@@ -461,7 +461,7 @@ func (ts *store) Close() error {
 	}
 
 	// Clear the map to release references
-	ts.cellConns = make(map[string]cellConn)
+	ts.cellConns = nil
```
sougou:
```suggestion
	ts.cellConns = make(map[string]cellConn)
```

This looks unsafe to me. It can cause future panics if functions get called after Close.

cxljs: Calling other functions after Store Close (before reopening) seems like a bug. This panic during the testing phase can help us identify the bug. What do you think?

sougou: We prefer that bugs cause functions to fail by returning an error. If a test doesn't catch the bug, a panic will result in a production outage.
The only time we "crash' a program is when we detect a misconfiguration that makes it chronically impossible to recover from. If so, we crash with an explicit fatal log message.

cxljs: @sougou ,You are right, and this is a great practice for building systems. Thank you.

PR #119, opened 2025-09-22
Multipooler: Implement ExecuteQuery RPC

deepthi:
- ExecuteQuery RPC with necessary proto message definitions
- Query will be executed on postgres and results returned to caller.

Note that we are using lib/pq to connect to postgres. This is a simple/naive implementation which will be replaced with a pooled connection in a future PR.


go/multipooler/server/plugin_pq.go
```diff
@@ -0,0 +1,19 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	_ "github.com/lib/pq" // PostgreSQL driver
```
deepthi, line 18: @sougou I spent some time trying to replace this with jackc/pgx and reverted that for a couple of reasons
- pgx has it's own pooling which we may not want given that we are going to implement our own
- there was a certain amount of translation code needed to convert from pgx results to proto messages. If this is anyway throw-away, it didn't seem worth writing that code.

go/test/endtoend/multipooler_client.go
```diff
@@ -0,0 +1,287 @@
...
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
+	multipoolerpb "github.com/multigres/multigres/go/pb/multipoolerservice"
+	querypb "github.com/multigres/multigres/go/pb/query"
+)
+
+// MultiPoolerTestClient wraps the gRPC client for testing
```
sougou, line 33: Why is this a test client? Don't you need a real client for MultiGateway?

go/multipooler/server/server.go
```diff
@@ -0,0 +1,294 @@
...
+
+// RegisterService registers the MultiPooler gRPC service with the given configuration
+func RegisterService(config *Config) {
+	servenv.OnRun(func() {
+		if servenv.GRPCServer == nil {
+			return
+		}
+
+		// Check if the pooler service should be registered
+		if servenv.GRPCCheckServiceMap("pooler") {
```
sougou, line 46: Optional nit: return on false

go/multipooler/server/server.go
```diff
@@ -0,0 +1,294 @@
...
+// executeQuery executes a SQL query and returns the result
+func (s *MultiPoolerServer) executeQuery(ctx context.Context, query string, maxRows uint64) (*querypb.QueryResult, error) {
+	// Determine if this is a SELECT query or a modification query
+	trimmedQuery := strings.TrimSpace(strings.ToUpper(query))
+	isSelect := strings.HasPrefix(trimmedQuery, "SELECT") ||
+		strings.HasPrefix(trimmedQuery, "WITH") ||
+		strings.HasPrefix(trimmedQuery, "SHOW") ||
+		strings.HasPrefix(trimmedQuery, "EXPLAIN")
+
+	if isSelect {
```
sougou, line 209: I assume this is temporary.

sougou (approved):

go/test/endtoend/multipooler_client.go
deepthi: We can move it when we get there.

go/multipooler/server/server.go
deepthi: yes

go/multipooler/server/server.go
deepthi: Actually, it seems like we are doing things slightly differently for different binaries. I'm going to make them uniform.

go/test/endtoend/cluster_test.go
```diff
@@ -1062,13 +1072,13 @@ func TestClusterLifecycle(t *testing.T) {
...
 		tempDir, err := os.MkdirTemp("/tmp", "mlt")
 		require.NoError(t, err)
-		defer os.RemoveAll(tempDir)
 
 		// Always cleanup processes, even if test fails
 		defer func() {
 			if cleanupErr := cleanupTestProcesses(tempDir); cleanupErr != nil {
 				t.Logf("Warning: cleanup failed: %v", cleanupErr)
 			}
+			os.RemoveAll(tempDir)
```
deepthi, line 1081: @Rafael I moved this here because sometimes the tempDir would get deleted before cleanup, which caused test processes to not get cleaned up properly.

go/cmd/multipooler/main.go
```diff
@@ -138,10 +144,25 @@ func run(cmd *cobra.Command, args []string) error {
 			"cell", cell,
 			"database", database,
 			"table_group", tableGroup,
+			"socket_file_path", socketFilePath,
```
rafael, line 147: nit: Should we be consistent in the logging with the actual args params (i.e `socket-file-path` vs `socket_file_path`) ?

go/test/endtoend/cluster_test.go
```diff
@@ -1123,3 +1133,39 @@ func assertDirectoryTreeEmpty(rootPath string) error {
...
+func testMultipoolerGRPC(t *testing.T, grpcPort int) {
+	t.Helper()
+
+	// Connect to multipooler gRPC service
+	multipoolerAddr := fmt.Sprintf("localhost:%d", grpcPort)
+	client, err := NewMultiPoolerTestClient(multipoolerAddr)
+	require.NoError(t, err, "Failed to connect to multipooler gRPC at %s", multipoolerAddr)
+	defer client.Close()
+
+	// Test basic SELECT query
```
rafael, line 1147: From an end-to-end  tests organization, we could leave here just the basic select, and create a separate one focused on multipooler. 

Otherwise we risk making the main cluster lifecycle too big as we start adding more and more functionality. 

We could keep this test just focust on making sure we have general cluster lifecycle working, and more detail tests we can have them by component.

PR #133, opened 2025-09-23
bootstrap: some code re-org for more clarity

rafael:
# Desc
* This doesn't change any functionality. It re-organizes code for better readability. The `local.go` was becoming too large so this splits out some logical parts into other files.


rafael:
@deepthi  let me know if you think we should do this. I had this kind of ready before going, but now there is a merge hell. I think it will be easier to start from scratch. 

I can do that, but let's double check that we want to do this.

deepthi:
Good idea. Let's do it.

PR #135, opened 2025-09-24
feat: a ticker with exponential backoff

sougou:
We can use this ticker for all our retries. It implements a starting interval, max interval, progressing with an exponential backoff. It also addes a 10% jitter to prevent systemic spikes.

### DIFF
--- a/go/tools/ticker/backoff.go
+++ b/go/tools/ticker/backoff.go
@@ -1,0 +1,142 @@
+// Copyright 2025 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package ticker provides a ticker implementation with exponential backoff and jitter.
+package ticker
+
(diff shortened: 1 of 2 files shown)


go/tools/ticker/backoff.go
```diff
@@ -0,0 +1,142 @@
...
+	if initialInterval <= 0 {
+		panic("ticker: non-positive interval for NewBackoffTicker")
+	}
+	if maxInterval < initialInterval {
+		panic("ticker: maxInterval must be >= initialInterval")
+	}
+
+	bt := &BackoffTicker{
+		C:            make(chan time.Time, 1),
+		initialDelay: initialInterval,
```
rafael, line 54: I would name this initialDelay, so it's consistent.

go/tools/ticker/backoff.go
```diff
@@ -0,0 +1,142 @@
...
+	mu           sync.Mutex
+	timer        *time.Timer
+	initialDelay time.Duration
+	maxDelay     time.Duration
+	currentDelay time.Duration
+	stopped      bool
+	rand         *rand.Rand
+}
+
+// NewBackoffTicker creates a new BackoffTicker with the given initial and maximum intervals.
```
rafael, line 41: One flag: I think there should be an option to fire immediately. With exponential backoff something that I think is a common use case is to want the very first attempt to happen right away, then apply the backoff for subsequent retries.

rafael (approved):
Makes sense. I left two comments!

go/tools/ticker/backoff.go
sougou: Responding late to this. I thought about adding this option, but realized that it's not necessary. The channel can also receive. So, you can just:
```
  ticker := NewBackoffTicker(1*time.Second, 30*time.Second)
  ticker.C <- time.Now()
```

PR #136, opened 2025-09-25
misc: small improvements to hooks

rafael:
# Desc
* Small tweaks to misc scripts.

### DIFF
--- a/misc/hooks/protolint
+++ b/misc/hooks/protolint
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -euo pipefail
+set -uo pipefail
 
 # Get list of proto files to lint
 protofiles=$(git diff --cached --name-only --diff-filter=ACM -- '*.proto')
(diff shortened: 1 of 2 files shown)


misc/hooks/protolint
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -euo pipefail
+set -uo pipefail
```
rafael, line 16: The `e` makes the script to fail immediately, so the output from `go tool` was not printed and it was not obvious why it was not failing. Something llke this:
```
git ci -m "proto: adds multipooler admin gRPC api. Exposes methods related to replication and HA"
Linting go/pb/multipoolermanager go/pb/multipoolermanagerdata go/pb/pgctldata go/pb/pgctldservice
0 issues.
Prettier not found. Skipping...
```

dweitzman: The example is for prettier -- was something similar happening with protolint? I would expect protolint to have better odds of success since it's not depending on any tooling other than `go tool`

rafael: Hmm the example was for protolint, it was failing, but it doesn't tell you :D. After my change I do see the output. The prettier not found, doesn't fail right now, it just tell you.

dweitzman (approved):

Issue #111, opened 2025-09-17
High availability (MultiOrch) v1


deepthi:
### Definitions
- Postgres Cluster: A set of n postgres servers that are connected via replication. Only one of these can accept writes (aka primary). All others are replica / standby servers
- [Multigres Cluster](https://multigres.com/docs/architecture): See Fig.3
- MultiGateway: The multigres component that provides a postgres endpoint that clients can connect to.
- MultiPooler: The multigres component that controls access to an underlying postgres server. There will be as many MultiPoolers in a multigres cluster as postgres servers.
- MultiOrch: The multigres component that monitors and fixes the postgres servers. This component will always communicate with postgres servers via MultiPoolers using gRPC.

### Scope
- MultiOrch
    - Ensures that a primary is elected on cluster bootstrap
    - Monitors all postgres servers through multipoolers
        - If it detects that there is no active primary (DeadPrimary), it will elect a new one
    - Configures replication on replica/standby servers according to the durability policy
    - Respects durability policy
- Planned leadership change can be done either through MultiOrch or from CLI

### Pre-requisites
- MultiPooler 
    - [x] postgres connection and query execution https://github.com/multigres/multigres/pull/119
    - [ ] gRPC management interface and implementation

### Tasks
- [ ] Read cluster state through poolers
- [ ] Store cluster state (in-memory structs)
- [ ] Implement initial primary election
    - [ ] Setup replication on replica/standby servers
- [ ] Implement planned primary change
    - [ ] CLI to invoke
- [ ] Run failure detection on a ticker (5 second default?)
- [ ] Emergency primary change aka failure recovery
- [ ] Replication failure detection and fixing
- [ ] Multigateway Buffering

PR #137, opened 2025-09-25
proto: adds multipooler admin gRPC api. HA. 

rafael:
This PR extends the protobuf definitions to support leadership changes in Multigres consensus, adding the key methods needed for coordinator-driven failover operations.

Part of: https://github.com/multigres/multigres/issues/111

## New Methods Added

**multipoolermanagerservice.proto**
- **WaitForLSN** – Wait for PostgreSQL to reach specific LSN positions  
- **SetReadOnly** – Make a PostgreSQL instance read-only  
- **PromoteStandby** – Promote a standby to primary (calls `pg_promote`)  
- **GetPrimaryLSN** – Get current primary LSN position  
- **IsReadOnly** – Check if instance is in read-only mode  
- **SetStandbyPrimaryConnInfo** – Configure standby-to-primary connection  
- **StartStandbyReplication / StopStandbyReplication** – Control WAL replay  
- **ResetStandbyReplication** – Reset standby connection (prevents commit acknowledgment)  
- **ConfigureSynchronousReplication** – Configure sync replication settings with support for `ANY` / `FIRST` / quorum methods  

**pgctldservice.proto**  
- Only basic PostgreSQL server lifecycle methods (no new replication methods added)

## Supporting Consensus Leadership Changes

These methods directly support the coordinator’s goals during leadership transitions:

- **Revocation** – `SetReadOnly`, `ResetStandbyReplication` (disable old leaders)  
- **Discovery** – `GetPrimaryLSN`, `IsReadOnly`, `WaitForLSN` (inspect current state)  
- **Establishment** – `PromoteStandby`, `ConfigureSynchronousReplication` (establish new leader)  
- **Propagation** – `WaitForLSN`, replication controls (ensure durability)


proto/multipoolermanagerservice.proto
```diff
@@ -0,0 +1,113 @@
...
+
+  // StandbyReplicationStatus gets the current replication status of the standby
+  rpc StandbyReplicationStatus(multipoolermanagerdata.StandbyReplicationStatusRequest)
+    returns (multipoolermanagerdata.StandbyReplicationStatusResponse);
+
+  // ResetStandbyReplication resets the standby's connection to its primary by clearing primary_conninfo,
+  // reloading PostgreSQL configuration, and terminating the WAL receiver process. This effectively
+  // disconnects the replica from the primary and prevents it from acknowledging commits, making it
+  // unavailable for synchronous replication until reconfigured.
+  rpc ResetStandbyReplication(multipoolermanagerdata.ResetStandbyReplicationRequest)
```
rafael, line 68: In my mind this is what we will be calling when recruiting. The response contains LSN and latest transaction timestamp.

proto/multipoolermanagerdata.proto
```diff
@@ -0,0 +1,276 @@
...
+  string lsn = 1;
+
+  // Result of pg_is_wal_replay_paused()
+  bool is_wal_replay_paused = 2;
+
+  // Result of pg_get_wal_replay_pause_state()
+  string wal_replay_pause_state = 3;
+
+  // Replication lag in seconds (optional, may not always be available)
+  int64 lag_seconds = 4;
```
dweitzman, line 35: What's the thinking here for having replication lag in seconds instead of something more granular like milliseconds?

Wondering if the folks who've worked on Vitess more recently have any feelings on whether tracking replication lag with higher granularity is worthwhile

proto/multipoolermanagerdata.proto
```diff
@@ -0,0 +1,276 @@
...
+  string last_xact_replay_timestamp = 5;
+}
+
+// Wait for PostgreSQL server to reach a specific LSN position
+message WaitForLSNRequest {
+  // Target LSN position to wait for (PostgreSQL LSN format: X/XXXXXXXX)
+  string target_lsn = 1;
+
+  // Timeout in seconds (0 means no timeout)
+  int32 timeout_seconds = 2;
```
dweitzman, line 47: This is maybe a higher-level team discussion: do we have a stance of whether or not to use unsigned integer types in proto3 for this sort of thing?

proto/multipoolermanagerdata.proto
```diff
@@ -0,0 +1,276 @@
...
+// SetStandbyPrimaryConnInfo sets the primary connection info for a standby server
+message SetStandbyPrimaryConnInfoRequest {
+  // Primary server hostname
+  string host = 1;
+
+  // Primary server port
+  int32 port = 2;
+
+  // Heartbeat interval in seconds
+  double heartbeat_interval = 3;
```
dweitzman, line 106: This should probably have the unit in the name (secs or milliseconds)

Also, wonder if there's any reason not to make this a uint64 with milliseconds instead of a double with seconds?

proto/multipoolermanagerdata.proto
```diff
@@ -0,0 +1,276 @@
...
+// PostgreSQL leader server status
+message PrimaryStatus {
+  // Current WAL LSN position (PostgreSQL format: X/XXXXXXXX)
+  string lsn = 1;
+
+  // Whether server is accepting connections
+  bool ready = 2;
+
+  // Connected follower servers
+  repeated string followers = 3;
```
dweitzman, line 145: What's the format of followers?

sougou (approved):
It's too early to tell how how many of these functions we'll need. I do think that we'll eventually need something like `RequestVote`. But this is good for a first pass, and we can iterate forward as needed.

proto/multipoolermanagerdata.proto
rafael: @deepthi  / @GuptaManan100  what do you think? I think this is a good suggestion.

rafael:
> It's too early to tell how how many of these functions we'll need. I do think that we'll eventually need something like `RequestVote`. But this is good for a first pass, and we can iterate forward as needed.

Agreed, there are some missing here. But was thinking to iterate forward as well

proto/multipoolermanagerdata.proto
deepthi: Yes. I agree with the direction in #144. Given that the underlying timestamp type has microsecond precision, we should be able to provide lag values at least in milliseconds.

proto/multipoolermanagerservice.proto
```diff
@@ -0,0 +1,113 @@
...
+
+  // WaitForLSN waits for PostgreSQL server to reach a specific LSN position
+  rpc WaitForLSN(multipoolermanagerdata.WaitForLSNRequest) returns (multipoolermanagerdata.WaitForLSNResponse);
+
+  // SetReadOnly makes the PostgreSQL instance read-only
+  rpc SetReadOnly(multipoolermanagerdata.SetReadOnlyRequest) returns (multipoolermanagerdata.SetReadOnlyResponse);
+
+  // PromoteStandby PostgreSQL standby server to primary
+  // this will call pg_promote() in the underlying instance.
+  rpc PromoteStandby(multipoolermanagerdata.PromoteStandbyRequest)
```
deepthi, line 37: We should drop `Standby` from all the RPC names.

proto/multipoolermanagerservice.proto
```diff
@@ -0,0 +1,113 @@
...
+  // ChangeType changes the pooler type (LEADER/FOLLOWER)
+  rpc ChangeType(multipoolermanagerdata.ChangeTypeRequest)
+    returns (multipoolermanagerdata.ChangeTypeResponse);
+
+  // GetFollowers gets the list of follower servers
+  rpc GetFollowers(multipoolermanagerdata.GetFollowersRequest)
+    returns (multipoolermanagerdata.GetFollowersResponse);
+
+  // DemoteLeader demotes the current leader server
+  rpc DemoteLeader(multipoolermanagerdata.DemoteLeaderRequest)
```
deepthi, line 102: Drop Leader and Follower from RPCs. The rationale is that you are invoking an RPC on a pooler which is already a leader/follower. There's no need to have them in the function names.

PR #142, opened 2025-09-26
feat(prototype): add retries during server initialization

### LINKED ISSUE #104: Components should not exit if they cannot connect to topo

sougou:
This includes a range of improvements in many places. This also fixes #104.

* Tuned various timeouts
* Use BackoffTicker in various places
* Changed topo to use WrapperConn
* Fix some bugs in WrapperConn
* Create a generic topopublish package for all components
* Create separate packages for each component with component-specific init.go and status.go behaviors
* Simplify all main.go files to call into the component packages
* Tuned provisioner: start finishes in 3-6 seconds, stop finishes near instantly
* Add component specific index pages that auto-refresh every 30s
* Add `/ready` endpoints for all components for readiness checks

As a side effect, tests also finish faster.

I tried parallelizing the launch of components in local provisioner, but that had diminishing returns. Also, it introduced flakiness in the tests that were expecting things to happen in a certain order.
This also means that we have to add tests to bring components in different order, and make sure they self-heal.

Screenshot of a green multigateway index page:

image

multigateway that could not connect to etcd:

image


sougou:
Here are some screenshots from the last commit

Error case
image

Success case
image

go/clustermetadata/topo/wrapperconn.go
```diff
@@ -82,12 +81,14 @@ func (c *Conn) retryConnection(conn topo.Conn) {
 		return
 	}
 
-	// TODO(sougou): Implement exponential backoff.
-	ticker := time.NewTicker(1 * time.Millisecond)
+	ticker := timertools.NewBackoffTicker(1*time.Millisecond, 30*time.Second)
 	defer ticker.Stop()
 
 	for range ticker.C {
```
dweitzman, line 126: When `c.wrapped` is `nil`, if there are multiple racing calls + errors it seems like `retryConnection` could be running in parallel with itself here trying to create an initial connection instead of uniting around a single attempt with exponential backoff?

Would it be worth adding counter to the mutex-protected state and make it an argument to `retryConnection` as an alternative to checking that `c.wrapped != conn` for situations where `c.wrapped = nil`?

sougou: Nice catch. This code is too complex anyways. I was actually planning to enhance this wrapper to call-back an alarm function, so that failed connections can be reported on the status page. You can see how it works in topopublish.
So, a separate variable to set the alarm on entry and reset it on exit should work well. Then I can rid of all the conn checks.

go/clustermetadata/topo/wrapperconn.go
```diff
@@ -116,7 +145,7 @@ func (c *Conn) retryConnection(conn topo.Conn) {
 	}
```
dweitzman, line 153: This is probably not a big deal in practice, but I think the most "ideal" behavior would involve setting `retrying = false` while still holding the lock in `mustContinue` instead of briefly releasing and re-acquiring it

Otherwise it's possible (but unlikely) to have a race where:
- a new connection `C1` was created and is available for use
- `retrying` is still `true` and the `retryConnection` goroutine A that created `C1` is still alive, just waiting to set `retrying` to false -- but already released the lock to access it.
- `C1` raises an error when someone tries to use it
- A new `retryConnection` goroutine B starts, observes that `retrying` is still true, and aborts
- goroutine `A` finishes and sets `retrying` to false

Final state:
- Connection `C1` is new but returned an error and isn't actually retrying

Expected state:
- `C1` should be closed due to the error and we should be trying to create a replacement connection `C2`

Seems like the race could be avoided by setting `c.wrapped` to `C1` and `retrying` to `false` in the same locked / critical section?

sougou: Yeah. I did consider that case, and felt that it wasn't worth handling. But your suggestion is very nice and simple. I'll make the change.

sougou: Ohhh. Actually, this is a bigger problem:
* goroutine A: Inner function sets retrying to false, releases lock.
* Another call starts goroutine B: it sets `retrying` to true, and gets past `mustReturn`
* goroutine A invokes `defer`, which sets `retrying` to false and returns.
* We now have goroutine B still running, but `retrying` is false.

go/clustermetadata/topo/wrapperconn.go
sougou: We could use sync.Once, which would be a readability load.

The other option is to leave it as is: it's a near impossible race to repro. Also, it's not harmful. The next attempt to do something will start the retry anyways.

go/clustermetadata/topo/wrapperconn.go
sougou: I found another bug about that defer. I've added a test to reproduce it, and confirm the fix doesn't cause it. For now, I'm going to punt on this race condition. I've added a detailed explanation in the code for it, just in case.

This PR has grown. I'm going to finalize it for review. I'll summarize everything on the main comment.

go/clustermetadata/topopublish/topopublish.go
```diff
@@ -0,0 +1,106 @@
...
+	logger *slog.Logger
+
+	unpublish func(ctx context.Context) error
+}
+
+// Publish publishes the component using the publish function. If the publish function
+// returns an error, it will be retried with exponential backoff until successful.
+// The alarm will be invoked with the latest error message during retries. If the
+// publishing succeeds, the alarm will be invoked with an empty string.
+func Publish(publish func(ctx context.Context) error, unpublish func(ctx context.Context) error, alarm func(string)) *TopoPublisher {
```
rafael: Idea on naming: Something like Register seems clearer in this context than publish.

go/cmd/multiadmin/main.go
```diff
@@ -53,98 +44,15 @@ func main() {
...
-		servenv.HTTPHandleFunc("/admin/status", handleStatusEndpoint)
-		servenv.HTTPHandleFunc("/admin/clusters", handleClustersEndpoint)
-		logger.Info("Admin HTTP endpoints available at /admin/status and /admin/clusters")
-	})
-
-	servenv.OnClose(func() {
-		logger.Info("multiadmin shutting down")
-	})
-
+	servenv.OnRun(multiadmin.Init)
```
rafael, line 47: This is cleaner 👯

go/multigateway/init.go
```diff
@@ -0,0 +1,94 @@
...
+		"http_port", servenv.HTTPPort(),
+		"grpc_port", servenv.GRPCPort(),
+	)
+
+	// Create MultiGateway instance for topo registration
+	multigateway := topo.NewMultiGateway(serviceID, cell, servenv.Hostname)
+	multigateway.PortMap["grpc"] = int32(servenv.GRPCPort())
+	multigateway.PortMap["http"] = int32(servenv.HTTPPort())
+
+	tp = topopublish.Publish(
```
rafael: Where do we wait in the startup process to check that this eventually succeeds?

sougou: The startup doesn't wait. As we spin up services, they're expected to get stuck in their own retry loops waiting for topo to be up. If they don't depend on it, then they can come up, but won't be discoverable.

go/clustermetadata/topopublish/topopublish.go
sougou: I debated about this. I thought of `Publish`, because the component is letting the world know of its existence. But I'm ok with `Register` also.
Right now, the topo call is `InitComponent`. Shall we rename that accordingly?

PR #143, opened 2025-09-26
refactor(proto): update proto definitions to address peer feedback

rafael:
# Desc
* Follow up from: https://github.com/multigres/multigres/pull/137

### DIFF
--- a/proto/multipoolermanagerdata.proto
+++ b/proto/multipoolermanagerdata.proto
@@ -17,6 +17,7 @@ syntax = "proto3";
 package multipoolermanagerdata;
 
 import "clustermetadata.proto";
+import "google/protobuf/duration.proto";
 
 option go_package = "github.com/multigres/multigres/go/pb/multipoolermanagerdata";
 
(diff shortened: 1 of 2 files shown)


proto/multipoolermanagerdata.proto
```diff
@@ -32,7 +33,7 @@ message ReplicationStatus {
   string wal_replay_pause_state = 3;
 
   // Replication lag in seconds (optional, may not always be available)
-  int64 lag_seconds = 4;
+  uint32 lag_ms = 4;
```
sougou, line 36: I'd recommend against unsigned numbers. They are future nightmares. Also, why not just use duration? This will make us future-proof: For example, if we introduce heartbeats.

dweitzman: If we want to go with Duration, here's a PR to update all the protos: https://github.com/multigres/multigres/pull/144

rafael: Great, closing this one in favor of yours. Thanks @dweitzman !

dweitzman: My PR didn't rename the follower field to standby_names or clarify it's format, FYI

PR #149, opened 2025-10-01
Removing global variables in favour of structs

GuptaManan100:
### Description

TBD

### DIFF
--- a/proto/pgctldservice.proto
+++ b/proto/pgctldservice.proto
@@ -185,6 +185,9 @@ message InitDataDirRequest {
 
   // Additional initdb arguments
   repeated string extra_args = 3;
+
+  // Path to password file for PostgreSQL user authentication
+  string pg_pwfile = 4;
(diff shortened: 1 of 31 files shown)


sougou:
I think this looks cleaner. But you have an older snapshot. I've refactored substantial parts of this. Hopefully, the refactor should make it easier. Can you update the branch and see where things land?

proto/pgctldservice.proto
```diff
@@ -185,6 +185,9 @@ message InitDataDirRequest {
 
   // Additional initdb arguments
   repeated string extra_args = 3;
+
+  // Path to password file for PostgreSQL user authentication
+  string pg_pwfile = 4;
```
rafael, line 190: This is small detail. In my original thinking it was by design that this is not a parameter for Init. Thinking that should only be set when the server starts, but not something that is provided by requests.

PR #151, opened 2025-10-01
feat: topo conn status export

sougou:
This exports the status for all topo connections to the index page of every component.

Sample screenshot
image


go/clustermetadata/topo/store.go
```diff
@@ -416,45 +429,69 @@ func (ts *store) ConnForCell(ctx context.Context, cell string) (Conn, error) {
 	// Connect to the cell topology server while holding the lock.
 	// This ensures only one connection is established at any given time.
 	// Create the connection and cache it for future use.
-	conn := NewWrapperConn(func() (Conn, error) {
-		return ts.factory.Create(cell, ci.Root, ci.ServerAddresses)
-	})
+
+	ts.statusMu.Lock()
+	defer ts.statusMu.Unlock()
```
dweitzman: Does the `ts.statusMu.Lock()` need to be held until the end of the function? Can it be released immediately after setting the status?

go/clustermetadata/topo/wrapperconn.go
```diff
@@ -118,6 +123,7 @@ func (c *WrapperConn) retryConnection() {
 		c.mu.Lock()
 		defer c.mu.Unlock()
 		c.retrying = false
+		c.alarm("")
```
dweitzman: Makes me a little nervous that alarm() is client-provided but gets called while locks are held.

Maybe a harder-to-break alternative could be to keep the current error message on the WrappedConn itself and send a no-argument alarm()/notify() in a separate gorountine?

go/multiadmin/status.go
```diff
@@ -54,6 +55,7 @@ func init() {
 
 // handleIndex serves the index page
 func handleIndex(w http.ResponseWriter, r *http.Request) {
+	serverStatus.TopoStatus = ts.Status()
```
dweitzman: Looks like an in-place modification of a global without a lock? I might be missing something, since it looks like this same pattern is in other status handlers, but should this be creating a copy of the default status locally and then modifying it instead of changing the global?

go/clustermetadata/topo/mock_test.go
```diff
@@ -0,0 +1,311 @@
+// Copyright 2025 Supabase, Inc.
```
rafael, line 1: Thinking we should remove `_test` from this file. There are no tests here.

go/clustermetadata/topo/store.go
```diff
@@ -416,45 +429,69 @@ func (ts *store) ConnForCell(ctx context.Context, cell string) (Conn, error) {
...
 	// Close all cell connections
-	ts.mu.Lock()
-	defer ts.mu.Unlock()
-
-	for cell, cc := range ts.cellConns {
-		if cc.conn != nil {
-			if err := cc.conn.Close(); err != nil {
-				errs = append(errs, fmt.Errorf("failed to close cell connection %s: %w", cell, err))
+	ts.cellMu.Lock()
+	defer ts.cellMu.Unlock()
```
rafael: It seems we need this lock inside the go routine. 

I'm thinking that the errgroup adds unnecessary complexity here. I would keep it simple and do it sequentially.

go/clustermetadata/topo/store.go
```diff
@@ -416,45 +429,69 @@ func (ts *store) ConnForCell(ctx context.Context, cell string) (Conn, error) {
...
 			}
 		}
-	}
+		return nil
+	})
+	err := g.Wait()
 
-	// Clear the map to release references
 	ts.cellConns = make(map[string]cellConn)
+	ts.status = make(map[string]string)
```
rafael: Do we need the `statusMu` lock here?

go/clustermetadata/topo/store.go
```diff
@@ -300,19 +312,25 @@ func GetAvailableImplementations() []string {
...
+		cellConns: make(map[string]cellConn),
+		status:    make(map[string]string),
+	}
+	ts.status[GlobalCell] = ""
+	conn := NewWrapperConn(
+		func() (Conn, error) {
+			return factory.Create(GlobalCell, root, serverAddrs)
+		},
+		func(s string) {
+			ts.statusMu.Lock()
```
rafael: nit - maybe better to have a small method: setStatus and the we call that method consistently wherever we want to set the status. You can then use this same method when setting the status for a cell: `ts.setStatus(cell, "status...")`

go/clustermetadata/topo/store.go
```diff
@@ -393,21 +411,16 @@ func (ts *store) ConnForCell(ctx context.Context, cell string) (Conn, error) {
...
+	defer ts.cellMu.Unlock()
 	cc, ok := ts.cellConns[cell]
 	if ok {
-		// Client exists in cache. Verify that it's for the same cell configuration.
-		// The cell name can be reused with different ServerAddresses and/or Root,
-		// in which case we should get a new connection and update the cache.
-		cellAddrs := strings.Join(cc.Cell.ServerAddresses, ",")
-		if serverAddrsStr == cellAddrs && ci.Root == cc.Cell.Root {
+		// Verify that the connection parameters match.
+		if proto.Equal(ci, cc.Cell) {
```
rafael: The change here looks good. But I'm not sure if we have a test for this. Did you verify this works?

go/clustermetadata/topo/mock_test.go
sougou: The `_test` ensures that it's not included in the main build. Better to keep the suffix.

go/clustermetadata/topo/store.go
sougou: There are two tests: `TestConnForCell_CachedConnection` and `TestConnForCell_UpdatedCellConfig`

go/clustermetadata/topo/store.go
sougou: This looks correct to me. If I move it inside, I'll have to lock it again for resetting the map. I can do it for improved readability. But I found a real bug: I need to lock statusMu before updating ts.status. I'll make both changes.

go/clustermetadata/topo/store.go
sougou: Ohh. Yes, we do :).

go/clustermetadata/topo/wrapperconn.go
sougou: This is good hygiene. I've refactored the code to call alarm outside of locks.

go/multiadmin/status.go
sougou: Yeah. This is a bug. Needs a more invasive fix. I'm wondering if I should make a separate PR for this.

go/clustermetadata/topo/store.go
```diff
@@ -223,13 +227,18 @@ type store struct {
 	// It is set at construction time and used to create cell-specific connections.
 	factory Factory
 
-	// mu protects the following fields from concurrent access.
-	mu sync.Mutex
 	// cellConns contains cached connections to cell-specific topology services.
 	// These connections should be accessed through the ConnForCell() method, which
 	// will read the cell configuration from the global cluster and create clients
 	// as needed.
+	cellMu    sync.Mutex
```
dweitzman: Worth adding a comment not to lock `cellMu` when you already have `statusMu`?

go/clustermetadata/topo/wrapperconn.go
```diff
@@ -142,6 +148,7 @@ func (c *WrapperConn) retryConnection() {
 				return false
 			}
 			if err != nil {
+				c.alarm(err.Error())
```
dweitzman: This alarm() call still seems to have an internal lock held?

go/clustermetadata/topo/store_test.go
```diff
@@ -0,0 +1,665 @@
...
+	// All should get the same connection (or very few if there was a race)
+	firstConn := conns[0]
+	sameCount := 0
+	for i := 1; i < numGoroutines; i++ {
+		if conns[i] == firstConn {
+			sameCount++
+		}
+	}
+	// Most should be the same (allowing for some race conditions during initialization)
+	assert.Greater(t, sameCount, numGoroutines-3, "Most connections should be the same cached instance")
```
dweitzman: Would this ever create more than 1 connection? ConnForCell() holds a lock while it creates the WrapperConn, so it seems like even under concurrent access there should only be one connection created?

go/clustermetadata/topo/store_test.go
```diff
@@ -0,0 +1,665 @@
...
+	require.NotNil(t, ts, "Store should not be nil")
+
+	// Verify factory was called
+	assert.GreaterOrEqual(t, factory.getCreateCount(), int32(1), "Factory should be called")
+
+	// Clean up
+	ts.Close()
+}
+
+// Note: We cannot easily test the error cases of Open() because they call os.Exit(1)
```
dweitzman, line 457: Could handle testing by factoring out testable openInternal() that returns an err and having Open() just call it and os.Exit(1) if the err is non-nil?

go/clustermetadata/topo/store_test.go
```diff
@@ -0,0 +1,665 @@
...
+func TestStatus_MultipleCells(t *testing.T) {
+	factory := newMockFactory()
+	ts := NewWithFactory(factory, "/test", []string{"localhost:2181"})
+	require.NotNil(t, ts)
+	defer ts.Close()
+
+	ctx := context.Background()
+
+	// Create multiple cells
+	for i := 1; i <= 3; i++ {
```
dweitzman: Maybe change the `<= 3` to `< 4` here and lower in the function?

go/clustermetadata/topo/store.go
sougou: Renamed to `cellConnsMu`.

go/clustermetadata/topo/wrapperconn.go
sougou: Nice catch. I forgot to remove this.

go/clustermetadata/topo/store_test.go
sougou: I'll defer this for later.

PR #153, opened 2025-10-02
fix(servenv): enable unix socket for gRPC services

rafael:
# Desc
* Easy PR, this closes a gap on the initial pass of servenv. We were not wiring the unix socket configuration for gRPC. This fixes that issue. For the cases where pgtctld is running in the same node as multipooler, we will want to use the socket. 
* This is mostly plumbing and adding tests as the functioanlity was already there, it was just not connected.

### DIFF
--- a/go/cmd/multigres/command/cluster/start.go
+++ b/go/cmd/multigres/command/cluster/start.go
@@ -109,7 +109,8 @@ func (s *ServiceSummary) PrintSummary() {
 		}
 	}
 	fmt.Println("- 🐘 Connect to PostgreSQL via Multigateway: TODO")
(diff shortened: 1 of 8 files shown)


go/cmd/multigres/command/cluster/start.go
```diff
@@ -109,7 +109,8 @@ func (s *ServiceSummary) PrintSummary() {
 		}
 	}
 	fmt.Println("- 🐘 Connect to PostgreSQL via Multigateway: TODO")
-	fmt.Println("- 🛑 Run \"multigres cluster stop\" to stop the cluster")
+	fmt.Println("- 🟢 Cluster started successfully. Enjoy!")
```
rafael, line 112: @sougou it is fixed! :D

sougou (approved):

PR #154, opened 2025-10-02
feat(multiadmin): add path-based proxy to node debug pages

dweitzman:
## Summary

Adds HTTP proxy functionality to multiadmin for a consistent, more bookmarkable way to discover and browse cluster nodes during development and in production.

Note: since gate and orch names are randomly-assigned by local provisioner, they're not super-duper bookmarkable. In the future we can either add the ability to name them consistently in local provisioner or, maybe more usefully, add a path format that routes to a random gate within a replica set instead of a specific one by name.

### Service Discovery (`/services` endpoint)
- New web UI displaying all cluster services discovered from topology
- Shows both proxied URLs (through multiadmin) and, for local develop, direct URLs for each service with their real ports
- Organizes services by global scope and per-cell

### HTTP Reverse Proxy (`/proxy/*` routes)
- Path-based routing to backend services:
  - `/proxy/admin/multiadmin` - multiadmin itself
  - `/proxy/gate/{cell}/{name}` - multigateway instances  
  - `/proxy/pool/{cell}/{name}` - multipooler instances
  - `/proxy/orch/{cell}/{name}` - multiorch instances
- Automatic HTML rewriting for correct asset paths and links behind the proxy

Routing currently involves an on-demand query to the topology server to resolve a name into a host & port. In the future we may want to pre-fetch and cache backend information in multiadmin

### Port Defaults Consolidation
- New `go/provisioner/local/ports` package with centralized default constants

### Local Provisioner Updates
- Local provisioner sets the new `--hostname localhost` flag on services. This avoids getting URLs like `http://mac:15000/` which don't work because web browsers are less forgiving about not using https on non-`localhost` domains
- Fixed default behavior when port numbers are omitted from `multigres.yaml`. Previously it was parsed as a 0 port and you'd get random port assignment instead of the default. Now 0 is assumed to mean you want to use the default value

### HTML Rewriting (`proxy.go`)
- Injects `<base href="/proxy/.../"/>` tag for relative URL resolution
- Rewrites absolute paths (e.g., `/css/style.css` → `/proxy/.../css/style.css`)
- Prevents double-prefixing of URLs already starting with `/proxy/`
- Preserves external URLs and query parameters

### Service Discovery (`discovery.go`)
- Queries topology store on demand for registered services in each cell
- Sorted alphabetically for consistent display

In the future we may want to have all this backend information pre-fetched and cached in memory instead of querying the topology service on demand.

## Testing

Unit tests for HTML rewriting functionality


go/provisioner/local/local.go
```diff
@@ -1031,6 +1036,8 @@ func (p *localProvisioner) stopService(ctx context.Context, req *provisioner.Dep
 	case "multipooler":
 		fallthrough
 	case "multiorch":
+		fallthrough
+	case "multiadmin":
```
dweitzman, line 1054: Also fixed deprovisioning of multiadmin

dweitzman:
I'll update this in a bit so that global/local doesn't look different in the URL path, since `GlobalCell` (`"global"`) can just be the cell name for a global cell in the path

go/multiadmin/proxy.go
```diff
@@ -0,0 +1,245 @@
...
+		http.Error(w, "Service hostname not found", http.StatusNotFound)
+		return
+	}
+	if targetPort == 0 {
+		http.Error(w, "Service port not found", http.StatusNotFound)
+		return
+	}
+
+	// Create reverse proxy to the target service
+	targetURL, _ := url.Parse(fmt.Sprintf("http://%s:%d", targetHost, targetPort))
```
rafael: nit - maybe all the code above here can be extracted in another function that just handled the mapping.

go/provisioner/local/local.go
rafael: oh we were not stopping multiadmin? I can fast follow adding a test for this once this lands

go/multiadmin/proxy.go
```diff
@@ -0,0 +1,245 @@
...
+
+			// Update response body
+			resp.Body = io.NopCloser(bytes.NewReader(rewrittenHTML))
+			resp.Header.Set("Content-Length", fmt.Sprintf("%d", len(rewrittenHTML)))
+		}
+
+		return nil
+	}
+
+	proxy.ServeHTTP(w, r)
```
rafael, line 208: Do you think is worth doing something to make sure this only starts when we are running in local? So we make sure we enforce that

rafael (approved):
Left some minor comments. Stamping, looks good to me based on our discussion

go/multiadmin/proxy.go
dweitzman: I was thinking we'd also want the ability to proxy on k8s to help with things like getting pprof data off of instances that don't have a public IP address

Maybe the easiest thing for now would be to support this proxying on k8s and if/when the react code makes it obsolete we could turn it off?

go/multiadmin/proxy.go
dweitzman: Done

go/provisioner/local/local.go
dweitzman: Seems like we weren't stopping it, but I might be missing something

go/multiadmin/proxy.go
rafael: Sounds good to me

PR #156, opened 2025-10-06
feat(bootstrap): QoL validations when initializing and starting cluster

rafael:
# Desc

This part adds a few quality of life improvements when initializing and starting the cluster:

1) If the config path for multigres is too long, it will not allow to initialize as unix socket path can be an issue in postgres:
```
Initializing Multigres cluster configuration...
Error: configuration validation failed: provisioner validation failed: unix socket path would exceed system limit (108 bytes): /Users/rafael/sandboxes/multigres/multigres_local/long/path/test/rafael/data/pooler_xxxxxxxx/pg_sockets/.s.PGSQL.5432

To fix this issue:
1. Initialize multigres from a directory with a shorter path
2. Provide config-path to multigres (--config-path target_dir) that has a shorter length

Example:
  Current: multigres cluster init --config-path ./multigres_local/long/path/test/rafael/
  Better:  multigres cluster init --config-path /tmp/mt/

This will generate socket paths like:
  /tmp/mt/data/pooler_xxxxxxxx/pg_sockets/.s.PGSQL.5432 (53 bytes)

Current path length: 117 bytes (limit: 108 bytes)
```
2. Make sure that required binaries are available when starting the cluster, otherwise it fails to start. 

## Tests

* Adds relevant tests to cover the use case

### DIFF
--- a/go/provisioner/local/local.go
+++ b/go/provisioner/local/local.go
@@ -1165,6 +1165,16 @@ func (p *localProvisioner) waitForProcessExit(process *os.Process, timeout time.
 
 // Bootstrap sets up etcd and creates the default database
 func (p *localProvisioner) Bootstrap(ctx context.Context) ([]*provisioner.ProvisionResult, error) {
+	// Validate binary paths before starting
+	if err := p.validateBinaryPaths(p.config); err != nil {
+		return nil, err
+	}
+
+	// Validate required system binaries before starting
+	if err := p.validateSystemBinaries(); err != nil {
+		return nil, err
+	}
+
 	fmt.Println("=== Bootstrapping Multigres cluster ===")
 	fmt.Println("")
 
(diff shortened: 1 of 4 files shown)


go/provisioner/local/local.go
```diff
@@ -1765,6 +1775,159 @@ func (p *localProvisioner) ValidateConfig(config map[string]any) error {
...
+		return err
+	}
+
+	return nil
+}
+
+// validateUnixSocketPathLength validates that Unix socket paths won't exceed system limits
+func (p *localProvisioner) validateUnixSocketPathLength(config *LocalProvisionerConfig) error {
+	// Unix domain socket path length limit is typically 108 bytes on most systems
+	const maxSocketPathLength = 108
```
dweitzman: Maybe make a function to return the max socket path length and have it return 108 by default and 104 on a mac?

dweitzman (approved):

go/provisioner/local/local.go
rafael: Good idea

PR #157, opened 2025-10-06
Adding Dockerfile and docker-compose

fernando-villalba:
Adding a Dockerfile to create an image for all multigres components and also a docker-compose for development and testing.

Run with:

```bash
docker-compose up -d
```

Tear down with:

```bash
docker-compose down -v
```

Docker compose will watch for changes in the `go` directory and rebuild images on code changes.

Please note that currently none of the components will be in a read state when running docker-compose as there is a topology initialisation operation missing.


fernando-villalba:
Removed docker-compose, other solutions will be explored for local development with containers

rafael (approved):
This makes sense to me.

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN go mod download
+
+# Copy source and build static binaries
+COPY . .
+RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o /bin/multigres ./go/cmd/multigres
+RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o /bin/multiadmin ./go/cmd/multiadmin
+RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o /bin/multigateway ./go/cmd/multigateway
+RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o /bin/multiorch ./go/cmd/multiorch
+RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o /bin/multipooler ./go/cmd/multipooler
+RUN CGO_ENABLED=0 go build -ldflags="-w -s" -o /bin/pgctld ./go/cmd/pgctld
```
Neurostep: [q]: Curious, why don't we use the [make build](https://github.com/multigres/multigres/blob/5531b332fd0556e5bc975b849780081161cedf0d/Makefile#L65) here? If `build` target misses the `ldflags` and `CGO_ENABLED=0`, shouldn't we add them to the Makefile as well?

.dockerignore
```diff
@@ -0,0 +1,18 @@
...
+# Local build artifacts
+bin/
+build/
+
+# Documentation and miscellaneous files
+*.md
+Makefile
+
+# Local development directory
+multigres_local/
```
Neurostep: [q]: Shouldn't we add an extra empty line here to make some [POSIX tools](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) happy?

Dockerfile
fernando-villalba: Excellent point! I have moved this build logic to the Makefile and the makefile into docker. Please review now

.dockerignore
fernando-villalba: Good catch, have addressed this in the latest commit.

Makefile
```diff
@@ -65,12 +65,12 @@ parser:
 build:
 	mkdir -p bin/
 	cp external/pico/pico.* go/web/templates/css/
-	go build -o bin/multigateway ./go/cmd/multigateway
-	go build -o bin/multipooler ./go/cmd/multipooler
-	go build -o bin/pgctld ./go/cmd/pgctld
-	go build -o bin/multiorch ./go/cmd/multiorch
-	go build -o bin/multigres ./go/cmd/multigres
-	go build -o bin/multiadmin ./go/cmd/multiadmin
+	CGO_ENABLED=0 go build -ldflags="-w -s" -o bin/multigateway ./go/cmd/multigateway
```
rafael: For local development we still want the debug info and symbol names. 

Thinking we should allow these flags to be parametrized. Something like:
```
build:
	mkdir -p $(OUTPUT_DIR)
	cp external/pico/pico.* go/web/templates/css/
	@for cmd in $(CMDS); do \
		echo "Building $$cmd ($(MODE))"; \
		if [ "$(MODE)" = "release" ]; then \
		     build with flags
		else \
			normal build
		fi \
	done
```

fernando-villalba: @rafael I added a new target called `build-release` and done the above right now. Thanks Rafa!

Makefile
```diff
@@ -7,7 +7,7 @@
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
+# distributed under the License is a distributed on an "AS IS" BASIS,
```
rcambrj:
```suggestion
# distributed under the License is distributed on an "AS IS" BASIS,
```

PR #159, opened 2025-10-06
feat(multipooler): Implements pooler manager rdonly API

rafael:
# Description
## Pooler API
- Implements the rdonly API for the manager. There are some follow-ups we’ll need to handle later, as this should likely be paired with the multipooler query service. If we set rdonly in the Postgres instance, we should also shut down the query service in the multipooler.
- Unlike MySQL, Postgres doesn’t have a way to enforce read-only mode at the database level. It’s just a session setting that can be overridden. In Multigres, this makes it even more important to pair this method with guardrails in the pooler code to prevent writes to the database.

## Misc
- Ports the mterrors gRPC wrapper. This follows a pattern that also exists in Vitess and seems like a sensible approach. I had a TODO to port it.

### DIFF
--- a/go/test/endtoend/multipooler_manager_test.go
+++ b/go/test/endtoend/multipooler_manager_test.go
@@ -629,7 +629,6 @@ func TestMultipoolerReplicationApi(t *testing.T) {
 		require.NotNil(t, isReadOnlyResp)
 		assert.False(t, isReadOnlyResp.ReadOnly, "Instance should not be in read-only mode after SetReadOnly(false)")
 
-		t.Logf("Verify writes work again after disabling read-only mode")
 		_, err = poolerClient.ExecuteQuery("INSERT INTO test_readonly VALUES (2)", 0)
 		require.NoError(t, err, "Should be able to insert after disabling read-only mode")
 		t.Logf("✓ Successfully inserted data after disabling read-only mode")


proto/multipoolermanagerdata.proto
```diff
@@ -58,8 +58,10 @@ message StartReplicationRequest {
 message StartReplicationResponse {
 }
 
-// SetReadOnly makes the PostgreSQL instance read-only
+// SetReadOnly toggles the PostgreSQL instance read-only mode
 message SetReadOnlyRequest {
```
dweitzman, line 62: I think in vitess / mysql the behavior of SetReadOnly is to allow in-progress write transactions to finish but block starting new transactions?

Not sure where we'll end up with this RPC in a world where the transaction timeout is not a given, but whatever we go with here we'll probably want to at least document the expected behavior here in the .proto

go/multipooler/manager/manager.go
```diff
@@ -210,13 +210,75 @@ func (pm *MultiPoolerManager) WaitForLSN(ctx context.Context, targetLsn string)
...
+
+	// Set default_transaction_read_only using ALTER SYSTEM
+	query := "ALTER SYSTEM SET default_transaction_read_only = "
+	if readOnly {
+		query = query + "'on'"
+	} else {
+		query = query + "'off'"
+	}
+
+	_, err := pm.db.ExecContext(ctx, query)
```
dweitzman, line 238: I think this might also work without converting to to an on/off string value?

`pm.db.ExecContext(ctx, "ALTER SYSTEM SET default_transaction_read_only = $1", readOnly)`

go/multipooler/manager/manager.go
```diff
@@ -210,13 +210,75 @@ func (pm *MultiPoolerManager) WaitForLSN(ctx context.Context, targetLsn string)
...
+		pm.logger.Error("Failed to reload configuration", "error", err)
+		return mterrors.Wrap(err, "failed to reload configuration")
+	}
+
+	// Terminate all backend connections except our own to ensure they reconnect with the new setting
+	// Note: This is disruptive and will abort in-progress transactions
+	terminateCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	rows, err := pm.db.QueryContext(terminateCtx, `
+		SELECT pg_terminate_backend(pid)
```
dweitzman, line 256: I think this is intended to kill in-progress write transactions, but at the cost of also killing in-progress reads as a side-effect?

Long term it feels like we'll probably want to kill in-progress writes more gracefully using custom logic instead of dropping a hammer so strong is also kills reads

go/multipooler/manager/manager.go
```diff
@@ -225,7 +287,24 @@ func (pm *MultiPoolerManager) IsReadOnly(ctx context.Context) (bool, error) {
...
+
+	// Ensure database connection
+	if err := pm.connectDB(); err != nil {
+		pm.logger.Error("Failed to connect to database", "error", err)
+		return false, mterrors.Wrap(err, "database connection failed")
+	}
+
+	// Query the current setting of default_transaction_read_only
+	var setting string
+	err := pm.db.QueryRowContext(ctx, "SELECT current_setting('default_transaction_read_only')").Scan(&setting)
```
dweitzman, line 299: I guess we'll need to assume here that there isn't some per-database or per-user override that would make the value here inconsistent with what query traffic would see. Worth mentioning for completeness in the gRPC spec or somewhere what user + database are used for checking the read-only setting, maybe?

go/test/endtoend/multipooler_manager_test.go
```diff
@@ -565,4 +565,75 @@ func TestMultipoolerReplicationApi(t *testing.T) {
 		// TODO: This should error because this is not yet a real standby.
 		require.NoError(t, err)
 	})
+
+	t.Run("SetReadOnly_IsReadOnly", func(t *testing.T) {
```
dweitzman, line 569: Might be useful to have test(s) showing what happens to an in-progress transaction or query (like `pg_sleep()` maybe) when switching into read-only mode?

go/multipooler/manager/manager.go
dweitzman: Another thought: maybe it'd eventually be sufficient to terminate only backends that bypassed multigres and connected to postgres directly?

proto/multipoolermanagerdata.proto
rafael: That's correct. From my recollection, this happens before it even reaches this point. This `SetReadOnly` was only applied at the database level.

Looking a bit deeper, it seems we used this RPC in Vitess to enable writes on a replica that was promoted to primary. In Vitess / MySQL, replicas start with `super_read_only` enabled. Refreshing my memory here, in MySQL, unless you explicitly set that flag, you could write to a replica. So I think the intent was mainly to prevent accidental writes to replicas.

In Postgres, however, it's not possible to write to a standby.

So I’m thinking we don’t actually need this RPC.

@deepthi what do you think?

After your comment, I’m leaning toward punting on implementing this and only adding it later if we find a concrete use case.

dweitzman: Looks like it's not possible to toggle into `hot_standby` mode without restarting a server, so maybe what multigres needs is just:
- A way to make a best-effort attempt to drain/buffer traffic before a switchover
- A way to retry read-committed queries after a switchover is complete if they were interrupted

rafael: Just to check: I think that's consistent with I'm saying. Hence why we don't really need `SetReadOnly` as a method in the RPC interface for the manager.  

The way to dran/buffer before switchover is something we need to implement, but is outside this context (same as vitess)

dweitzman: Yes, for switchover within a shard your assessment makes sense to me

Would we want `SetReadOnly` for a resharding, though? If so, could be a future problem

rafael:
Per discussion, we are gonna close this for now and revisit at a future time

PR #163, opened 2025-10-08
feat(pooler): Implements SetPrimaryConnInfo and consensus Term

rafael:
# Description

This PR implements continues the implementation of the API manager RPC calls. It adds `SetPrimaryConnInfo` to set replication. In addition, it adds the term management for consensus in the pooler.

### Core Changes

**Consensus Term Management**
- Introduces `ConsensusTerm` protobuf message tracking current term, voted-for pooler, and leader ID
- Implements disk-based term persistence at `$PGDATA/consensus/consensus_term.json` with atomic writes
- Adds `SetTerm` RPC for explicit term initialization by orchestrator
- Enforces Raft-like consensus rules: reject stale terms, accept equal terms, update and accept newer terms
- Manager startup now waits for both topology record AND consensus term to be loaded before entering ready state

**SetPrimaryConnInfo Implementation**
- Fully implements replication configuration for standby instances with consensus-aware validation
- Pre-flight guardrails: validates pooler type is `REPLICA` and PostgreSQL is in recovery mode
- Configures `primary_conninfo` via `ALTER SYSTEM` and `pg_reload_conf()` with keepalive settings
- Supports optional replication control via `stopReplicationBefore` and `startReplicationAfter` flags
- Term validation prevents, operations with stale terms are rejected unless `force=true`

**Testing**
- End-to-end tests covering full replication setup flow: pg_basebackup, configuration, data validation
- Term validation test suite verifying all consensus rules (equal/stale/newer terms, force bypass)
- Replication control flag tests for stop/start behavior


go/multipooler/grpcmanagerservice/service.go
```diff
@@ -74,7 +75,20 @@ func (s *managerService) IsReadOnly(ctx context.Context, req *multipoolermanager
...
+	// Convert protobuf Duration to time.Duration
+	var heartbeatInterval time.Duration
+	if req.HeartbeatInterval != nil {
+		heartbeatInterval = req.HeartbeatInterval.AsDuration()
+	}
+
+	err := s.manager.SetPrimaryConnInfo(ctx,
+		req.Host,
+		req.Port,
+		heartbeatInterval,
```
sougou: Does standby need to know heatbeatInterval?

go/multipooler/manager/manager.go
```diff
@@ -229,13 +247,276 @@ func (pm *MultiPoolerManager) IsReadOnly(ctx context.Context) (bool, error) {
...
+			LeaderId:     nil,
+		}
+
+		// Update term in pgctld
+		conn, err := pm.createPgctldClient()
+		if err != nil {
+			pm.logger.Error("Failed to create pgctld client for term update", "error", err)
+			return mterrors.Wrap(err, "failed to create pgctld client for term update")
+		}
+		client := pgctldpb.NewPgCtldClient(conn)
```
sougou: I'm wondering if multipooler can track its own term. May be simpler.

go/multipooler/grpcmanagerservice/service.go
rafael: Oh this is to control connection keepalive for the replication stream. It is the postgres setting.  Maybe we don't expose this yet? We can set a sensible default and evaluate later if we want to allow people to tweak it. It might be an early optimization allowing this to be set. 

What do you think?

go/multipooler/manager/manager.go
rafael: Yeah, I thought about that.  But on the fence. It felt odd to keep this state outside where postgres is.  My thinking is that it could work, but storing next to postgres seemed like the safest approach. In practice, if they are running next to each other and it will be a call via the unix socket, it's effectively the same.

rafael: We discussed this offline and aligned on multipooler writing directly

PR #167, opened 2025-10-10
feat(pooler): Multipooler manager API

rafael:
# Description

This PR continues to implement more RPC endpoints in the manager API. In this case, it implements the following RPCs:
* StopReplication
* StartReplication
* ResetReplication  

## Misc changes
* Unifies timeouts in the integration test for the manager API.
* Removes the SetReadOnly RPC. After https://github.com/multigres/multigres/pull/159, we aligned on not needing it.

### DIFF
--- a/go/multipooler/grpcmanagerservice/service_test.go
+++ b/go/multipooler/grpcmanagerservice/service_test.go
@@ -230,6 +230,22 @@ func TestManagerServiceMethods_ManagerNotReady(t *testing.T) {
 			},
 		},
 		{
+			name: "StartReplication",
+			method: func() error {
+				req := &multipoolermanagerdata.StartReplicationRequest{}
+				_, err := svc.StartReplication(ctx, req)
+				return err
+			},
+		},
+		{
+			name: "StopReplication",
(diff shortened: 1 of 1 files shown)


go/multipooler/grpcmanagerservice/service_test.go
```diff
@@ -89,42 +89,6 @@ func TestManagerServiceMethods_NotImplemented(t *testing.T) {
...
-		},
-		{
-			name: "StopReplication",
-			method: func() error {
-				req := &multipoolermanagerdata.StopReplicationRequest{}
-				_, err := svc.StopReplication(ctx, req)
-				return err
-			},
-			expectedMethod: "StopReplication",
-		},
```
deepthi, line 127: Were these deleted by mistake?

rafael: Oh good catch. 

I meant to delete one group. There is a test suite here that asserts that the method is not implemented.

But there is another suite that tests that all RPC return error if the manager is not initialized. Added those back.

PR #170, opened 2025-10-21
refactor(pgtctld): Postgres config doesn't contain host specific data

rafael:
# Description

Previously, we were setting the data directory, port, and Unix socket in the `postgresql.conf`. This caused issues since these values are host-specific and end up being included in the backup. When restoring to a standby, we had to re-initialize the configuration manually.

With this change, the configuration no longer contains any host-specific data. This makes the `postgresql.conf` simpler, cleaner, and more portable across environments.

# Tests

All tests continue to pass. The changes were minimal, mostly updating test cases to pass the port flag to pgctld and adjusting a few expectations around the generated configuration (which no longer includes data_directory, port, or related settings).

### DIFF
--- a/go/test/endtoend/multipooler_manager_test.go
+++ b/go/test/endtoend/multipooler_manager_test.go
@@ -646,20 +646,6 @@ func setupStandbyReplication(t *testing.T, primaryPgctld *ProcessInstance, stand
 	err := os.MkdirAll(configBackupDir, 0o755)
 	require.NoError(t, err)
 
-	// Backup important config files
-	configFiles := []string{"postgresql.auto.conf", "pg_hba.conf", "pg_ident.conf"}
-	for _, configFile := range configFiles {
-		srcPath := filepath.Join(standbyPgDataDir, configFile)
-		dstPath := filepath.Join(configBackupDir, configFile)
-
-		// Copy file if it exists
-		if data, err := os.ReadFile(srcPath); err == nil {
-			err = os.WriteFile(dstPath, data, 0o644)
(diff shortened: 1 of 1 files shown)


go/test/endtoend/multipooler_manager_test.go
```diff
@@ -646,20 +646,6 @@ func setupStandbyReplication(t *testing.T, primaryPgctld *ProcessInstance, stand
 	err := os.MkdirAll(configBackupDir, 0o755)
 	require.NoError(t, err)
 
-	// Backup important config files
```
rafael, line 649: This is the proof that this PR works :D

go/pgctld/postgresconfig.go
```diff
@@ -199,29 +199,19 @@ func ReadPostgresServerConfig(pgConfig *PostgresServerConfig, waitTime time.Dura
...
+	// This ensures backups remain portable across different environments
 	if pgConfig.MaxConnections, parseErr = pgConfig.lookupInt("max_connections"); parseErr != nil {
 		pgConfig.MaxConnections = 100 // default
 	}
 	if val, err := pgConfig.lookupWithDefault("listen_addresses", pgConfig.ListenAddresses); err == nil {
 		pgConfig.ListenAddresses = val
 	}
-	if val, err := pgConfig.lookupWithDefault("unix_socket_directories", pgConfig.UnixSocketDirectories); err == nil {
-		pgConfig.UnixSocketDirectories = val
-	}
```
deepthi, line 211: This was subtle. We still set this in the config, we just don't _get_ it from the config template. 😅

cuongdo (approved):

PR #174, opened 2025-10-23
feat(multipooler): Adds rpc locks pooler manager

rafael:
# Description

We want to make sure that actions that are performing mutations are not run concurrently. This is something we also did in the tablet manager in Vitess. This ports this functionality into the pooler manager. 

# Tests

* Added a relevant unit test.

### DIFF
--- a/go/multipooler/manager/manager.go
+++ b/go/multipooler/manager/manager.go
@@ -102,7 +102,10 @@ func NewMultiPoolerManagerWithTimeout(logger *slog.Logger, config *Config, loadT
 // lock is used at the beginning of an RPC call, to acquire the
 // action semaphore. It returns ctx.Err() if the context expires.
 func (pm *MultiPoolerManager) lock(ctx context.Context) error {
-	return pm.actionSema.Acquire(ctx, 1)
+	if err := pm.actionSema.Acquire(ctx, 1); err != nil {
+		return mterrors.Wrap(err, "failed to acquire action lock")
+	}
+	return nil
 }
 
 // unlock is the symmetrical action to lock.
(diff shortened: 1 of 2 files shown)


go/multipooler/manager/manager.go
```diff
@@ -627,6 +654,13 @@ func (pm *MultiPoolerManager) SetPrimaryConnInfo(ctx context.Context, host strin
 	if err := pm.checkReady(); err != nil {
 		return err
 	}
+
+	// Acquire the action lock to ensure only one mutation runs at a time
+	if err := pm.lock(ctx); err != nil {
+		return mterrors.Wrap(err, "failed to acquire action lock")
```
dweitzman: Would it make sense to put the `mterrors.Wrap()` with the message into the `lock()` function itself?

go/multipooler/manager/manager_test.go
```diff
@@ -965,3 +966,188 @@ func TestParseAndRedactPrimaryConnInfo(t *testing.T) {
...
+	holdLock := func(duration time.Duration) context.CancelFunc {
+		lockCtx, cancel := context.WithCancel(context.Background())
+		lockAcquired := make(chan struct{})
+		go func() {
+			if err := manager.lock(lockCtx); err == nil {
+				// Signal that the lock was acquired
+				close(lockAcquired)
+				// Hold the lock for the duration or until cancelled
+				select {
+				case <-time.After(duration):
```
dweitzman: Could change to `WithDeadline()` above and avoid the need for a separate `time.After()` case here?

rafael:
For future reference, we are using a semaphore here because of this change: https://github.com/vitessio/vitess/pull/6625

PR #175, opened 2025-10-23
feat(multipooler): Adds update synchronous standby list

rafael:
## Summary

Implements UpdateSynchronousStandbyList RPC for dynamically managing the PostgreSQL synchronous standby list, and completes the PrimaryStatus implementation.

## Main RPC Methods

**UpdateSynchronousStandbyList (new)**: Dynamically updates the synchronous standby list on primary servers:
  - ADD operation - idempotently adds standbys to existing list
  - REMOVE operation - idempotently removes standbys from list
  - REPLACE operation - replaces entire standby list
  - Preserves existing quorum method (FIRST/ANY) and num_sync count from current configuration
  - Parses and validates current synchronous_standby_names before applying changes

**PrimaryStatus**:  Returns structured status information from primary servers including:
  - Synchronous replication configuration
  - Connected standby replication state and LSN positions
  - Sync status for each standby

## Key Implementation Details
  - Idempotent operations: ADD and REMOVE skip changes when configuration already matches desired state
  - Primary guardrails: Validates PRIMARY pooler type and non-recovery mode before modifications
  - Consistent naming: Uses shared generateApplicationName() helper for standby identification

## Testing
  - Standby list operations with edge cases (duplicates, empty results, ordering preservation)
  - PostgreSQL config format parsing including wildcards and various method formats
  - Primary status retrieval with various replication states
  - Idempotency validation for ADD/REMOVE operations

### DIFF
--- a/go/test/endtoend/multipooler_manager_test.go
+++ b/go/test/endtoend/multipooler_manager_test.go
@@ -3004,10 +3004,19 @@ func TestPrimaryStatus(t *testing.T) {
 				!statusResp.Status.IsWalReplayPaused
 		}, 5*time.Second, 200*time.Millisecond, "Replication should be established")
 
-		// Get primary status
-		statusResp, err := primaryManagerClient.PrimaryStatus(utils.WithShortDeadline(t), &multipoolermanagerdata.PrimaryStatusRequest{})
-		require.NoError(t, err, "PrimaryStatus should succeed")
-		require.NotNil(t, statusResp.Status, "Status should not be nil")
+		// Wait for primary to register the follower in pg_stat_replication
+		// There can be a delay between standby connection and primary's view update
+		t.Log("Waiting for primary to register the follower...")
+		var statusResp *multipoolermanagerdata.PrimaryStatusResponse
+		require.Eventually(t, func() bool {
+			var err error
+			statusResp, err = primaryManagerClient.PrimaryStatus(utils.WithShortDeadline(t), &multipoolermanagerdata.PrimaryStatusRequest{})
+			if err != nil {
+				t.Logf("PrimaryStatus error: %v", err)
+				return false
+			}
+			return statusResp.Status != nil && len(statusResp.Status.Followers) > 0
+		}, 10*time.Second, 200*time.Millisecond, "Primary should register the follower")
 
 		// Verify followers list contains the standby
 		require.NotEmpty(t, statusResp.Status.Followers, "Should have at least one follower")


go/test/endtoend/multipooler_manager_test.go
```diff
@@ -1571,27 +1602,23 @@ func TestConfigureSynchronousReplication(t *testing.T) {
...
-
-		t.Log("Verifying synchronous_standby_names is configured...")
-		queryResp, err = primaryPoolerClient.ExecuteQuery(context.Background(), "SHOW synchronous_standby_names", 1)
-		require.NoError(t, err)
-		require.Len(t, queryResp.Rows, 1)
-		syncStandbyNames := string(queryResp.Rows[0].Values[0])
-		assert.Contains(t, syncStandbyNames, "FIRST 1", "synchronous_standby_names should contain 'FIRST 1'")
-		assert.Contains(t, syncStandbyNames, standbyAppName, "synchronous_standby_names should contain standby name")
+		// Wait for configuration to converge and verify using PrimaryStatus API
+		t.Log("Waiting for configuration to converge...")
```
rafael, line 1861: Now that we have PrimaryStatus, this simplifies this test.

go/test/endtoend/multipooler_manager_test.go
```diff
@@ -1631,26 +1658,23 @@ func TestConfigureSynchronousReplication(t *testing.T) {
...
-		syncCommit := string(queryResp.Rows[0].Values[0])
-		assert.Equal(t, "remote_apply", syncCommit, "synchronous_commit should be 'remote_apply'")
-
-		t.Log("Verifying synchronous_standby_names is configured with ANY method...")
-		queryResp, err = primaryPoolerClient.ExecuteQuery(context.Background(), "SHOW synchronous_standby_names", 1)
-		require.NoError(t, err)
-		require.Len(t, queryResp.Rows, 1)
-		syncStandbyNames := string(queryResp.Rows[0].Values[0])
-		assert.Contains(t, syncStandbyNames, "ANY 1", "synchronous_standby_names should contain 'ANY 1'")
+		// Wait for configuration to converge and verify using PrimaryStatus API
```
rafael, line 1899: This is the same as above, do not get the status manually, use the RPC

go/test/endtoend/multipooler_manager_test.go
```diff
@@ -2076,6 +2082,70 @@ func TestConfigureSynchronousReplication(t *testing.T) {
 		t.Log("End-to-end synchronous replication test completed successfully")
 	})
 
+	t.Run("ConfigureSynchronousReplication_ClearConfig", func(t *testing.T) {
```
rafael, line 2270: This is a new test, I made changes to allow this via the RPC

go/multipooler/manager/manager.go
```diff
@@ -1044,13 +1006,290 @@ func (pm *MultiPoolerManager) ConfigureSynchronousReplication(ctx context.Contex
...
+	// Get current LSN and recovery status in a single query
+	var lsn string
+	var isInRecovery bool
+	err := pm.db.QueryRowContext(ctx, "SELECT pg_current_wal_lsn()::text, pg_is_in_recovery()").Scan(&lsn, &isInRecovery)
+	if err != nil {
+		pm.logger.Error("Failed to query LSN and recovery status", "error", err)
+		return nil, mterrors.Wrap(err, "failed to query LSN and recovery status")
+	}
+	status.Lsn = lsn
+	status.Ready = !isInRecovery
```
rafael: @deepthi this is what seemed sensible to me in this context. What do you think?

go/multipooler/manager/manager.go
```diff
@@ -1108,13 +1070,302 @@ func (pm *MultiPoolerManager) ConfigureSynchronousReplication(ctx context.Contex
...
+
+	pm.logger.Info("UpdateSynchronousStandbyList called",
+		"operation", operation,
+		"standby_ids", standbyIDs,
+		"reload_config", reloadConfig,
+		"consensus_term", consensusTerm,
+		"force", force)
+
+	// Validate and update consensus term following consensus rules
+	if err := pm.validateAndUpdateTerm(ctx, consensusTerm, force); err != nil {
```
rafael, line 1142: @deepthi - I think we need this here as well. Similar to SetPrimaryConnInfo. Tagging you to double check my reasoning here.

go/multipooler/manager/manager.go
```diff
@@ -970,7 +911,15 @@ func (pm *MultiPoolerManager) setSynchronousCommit(ctx context.Context, synchron
 // Note: Use '*' to match all connected standbys, or specify explicit standby application_name values
 // Application names are generated from multipooler IDs using the shared generateApplicationName helper
 func (pm *MultiPoolerManager) setSynchronousStandbyNames(ctx context.Context, synchronousMethod multipoolermanagerdata.SynchronousMethod, numSync int32, standbyIDs []*clustermetadatapb.ID) error {
```
dweitzman, line 955: Higher level style question: do we want do something like a "Locked" suffix for method that should be called when a lock is already held, and if so would this method be appropriate for that sort of suffix?

go/multipooler/manager/manager.go
```diff
@@ -1108,13 +1070,302 @@ func (pm *MultiPoolerManager) ConfigureSynchronousReplication(ctx context.Contex
...
+	for i, id := range updatedStandbys {
+		appName := generateApplicationName(id)
+		quotedNames[i] = fmt.Sprintf(`"%s"`, appName)
+	}
+	standbyList := strings.Join(quotedNames, ", ")
+
+	// Convert method enum back to string
+	var methodStr string
+	switch cfg.Method {
+	case multipoolermanagerdata.SynchronousMethod_SYNCHRONOUS_METHOD_FIRST:
```
dweitzman: Worth factoring out some shared code with setSynchronousStandbyNames(), like formatting the method + NumSync + standbyList into a string and/or some of the SQL statements?

go/multipooler/manager/pg_config.go
```diff
@@ -0,0 +1,204 @@
...
+		return nil, mterrors.New(mtrpcpb.Code_INVALID_ARGUMENT,
+			"wildcard (*) is not supported in Multigres - standby list must be explicit")
+	}
+
+	// Parse format: METHOD NUM (member1, member2, ...)
+	// Regex: ^(FIRST|ANY)\s+(\d+)\s*\(([^)]*)\)$
+	// Note: this regex assumes standby_names are being controlled by multigres
+	// and will have the format we expect (i.e cell_name). We are not validating
+	// for this format here.
+	re := regexp.MustCompile(`^(FIRST|ANY)\s+(\d+)\s*\(([^)]*)\)$`)
```
dweitzman: Should the MustCompile result be saved as a top-level constant instead of recompiling for every call?

Any chance FIRST or ANY could be lowercase?

go/test/endtoend/multipooler_manager_test.go
```diff
@@ -1571,27 +1602,23 @@ func TestConfigureSynchronousReplication(t *testing.T) {
...
+				config.SynchronousCommit == multipoolermanagerdata.SynchronousCommitLevel_SYNCHRONOUS_COMMIT_ON &&
+				config.SynchronousMethod == multipoolermanagerdata.SynchronousMethod_SYNCHRONOUS_METHOD_FIRST &&
+				config.NumSync == 1 &&
+				containsStandbyIDInConfig(config, "test-cell", standbyAppName)
+		}, "Configuration should converge to expected values")
+
+		// Verify the configuration using PrimaryStatus
+		t.Log("Verifying configuration using PrimaryStatus...")
+		status := getPrimaryStatusFromClient(t, primaryManagerClient)
+		assert.Equal(t, multipoolermanagerdata.SynchronousCommitLevel_SYNCHRONOUS_COMMIT_ON, status.SyncReplicationConfig.SynchronousCommit, "synchronous_commit should be ON")
```
dweitzman: Doesn't the waitForSyncConfigConvergenceWithClient() above already verify everything in these assert.Equal() checks?

Same comment for a bunch of other tests in this file

go/test/endtoend/multipooler_manager_test.go
```diff
@@ -2098,3 +2168,574 @@ func TestConfigureSynchronousReplication(t *testing.T) {
...
+			ReloadConfig:  true,
+			ConsensusTerm: 0,
+			Force:         true,
+		}
+		_, err = primaryManagerClient.UpdateSynchronousStandbyList(utils.WithShortDeadline(t), updateReq)
+		require.NoError(t, err, "ADD should be idempotent")
+
+		// Wait for config to settle - should remain unchanged (idempotent)
+		// We wait for the same configuration to ensure pg_reload_conf() completes
+		waitForSyncConfigConvergenceWithClient(t, primaryManagerClient, func(config *multipoolermanagerdata.SynchronousReplicationConfiguration) bool {
```
dweitzman: Since there's no change in the value, does this wait do anything?

Would it also make sense somewhere in this test case to try to add both an existing a new a new standby in a single request?

go/test/endtoend/multipooler_manager_test.go
```diff
@@ -2098,3 +2168,574 @@ func TestConfigureSynchronousReplication(t *testing.T) {
...
+
+		status := getPrimaryStatusFromClient(t, primaryManagerClient)
+		assert.Equal(t, multipoolermanagerdata.SynchronousMethod_SYNCHRONOUS_METHOD_ANY, status.SyncReplicationConfig.SynchronousMethod)
+		assert.True(t, containsStandbyIDInConfig(status.SyncReplicationConfig, "test-cell", "standby2"))
+		t.Log("Initial configuration verified")
+
+		// REMOVE standby2
+		updateReq := &multipoolermanagerdata.UpdateSynchronousStandbyListRequest{
+			Operation:     multipoolermanagerdata.StandbyUpdateOperation_STANDBY_UPDATE_OPERATION_REMOVE,
+			StandbyIds:    []*clustermetadatapb.ID{makeMultipoolerID("test-cell", "standby2")},
```
dweitzman, line 2653: Worth testing remove for a standby that isn't in the list? Could add a random bogus value into this request too

go/test/endtoend/multipooler_manager_test.go
```diff
@@ -2098,3 +2168,574 @@ func TestConfigureSynchronousReplication(t *testing.T) {
...
+		assert.Len(t, status.SyncReplicationConfig.StandbyIds, 2)
+		assert.False(t, containsStandbyIDInConfig(status.SyncReplicationConfig, "test-cell", "standby1"))
+		assert.False(t, containsStandbyIDInConfig(status.SyncReplicationConfig, "test-cell", "standby2"))
+		assert.True(t, containsStandbyIDInConfig(status.SyncReplicationConfig, "test-cell", "standby3"))
+		assert.True(t, containsStandbyIDInConfig(status.SyncReplicationConfig, "test-cell", "standby4"))
+
+		t.Log("REPLACE operation verified successfully")
+
+		// Cleanup
+		resetReq := &multipoolermanagerdata.ConfigureSynchronousReplicationRequest{
```
dweitzman: Wondering if it'd make sense to have a `setupReplicationTest(t)` helper or something like that to register a `t.Cleanup()` that'll reset replication when the test is over so it doesn't become part of the test case code

go/multipooler/manager/manager.go
deepthi: if `isInRecovery` is true, we won't even get this far, right? we'll return on line 1315. So if this is really the primary, Ready can never be false.

go/multipooler/manager/manager.go
deepthi: We'll need to refactor this when we implement consensus properly, but when we get there, the term should already have been updated before this RPC is called. This RPC should not be the one that is updating the term.

go/multipooler/manager/manager.go
rafael: Yeah, I think that's right. I'll set it to true and add a comment.

go/test/endtoend/multipooler_manager_test.go
rafael: Good flag. This was from a previous version when I was not waiting. Removed these

go/multipooler/manager/manager.go
deepthi: It's valid to say that a REPLICA is notified of a new term via SetPrimaryConnInfo and updates its local term. It cannot be the case that a PRIMARY is notified of a new term via an RPC.

deepthi: Actually, this makes sense but we need to expand it. If a PRIMARY receives this request with a newer term than it knows about, it may or may not be the real PRIMARY any more. So it needs to validate that and demote itself if necessary. We can do all that as follow up work.

rafael: Agreed. Thanks for helping me reason through this. Per our discussion in Slack, I will follow up removing this for now and add note to come back to this when we make more progress in consensus.

PR #176, opened 2025-10-23
refactor(tools): New retry package with iterator API for retry with exponential backoff

dweitzman:
This offers a little better protection against immediate retries than the ticker-based approach, and overall seems like it should be a simple, extensible, and relatively safe, generic approach to retries with backoff.

I think it's a little nicer and less error-prone than the ticker approach. Thoughts?

Example usage:

```go
r := retry.New(10*time.Millisecond, 30*time.Second)
for attempt, err := range r.Attempts(ctx) {
    if err != nil {
        // Context cancelled or timed out
        return err
    }
    
    // Perform operation with automatic backoff between iterations
    if tryOperation() {
        return nil // Success
    }

    // In a forever-loop like pooler discovery, you can optionally reset
    // the backoff to the initial amount once you've confirmed that
    // the system is in a good state
    r.Reset();
}
```

A downside of the iterator-based API is that `errcheck` won't complain if you don't look for an error response. An alternative API that doesn't have that issue would be this:

```go
for {
  err := r.StartAttempt(ctx)
  // ...
}
```

That API exists internally as `r.startAttempt()` but isn't exported since the iterator API feels a little more elegant, and maybe there's a way in the future to make a linter that would encourage checking the error value there. Open to other opinions.

If you want to delay before the first attempt instead of making the first attempt immediately:

```go
r := retry.New(10*time.Millisecond, 30*time.Second, retry.WithInitialDelay())
```

This code changes the jitter algorithm from 10% fractional jitter (non-jitter value plus or minus 10%) to full jitter (anywhere from 0% to 100% of the non-jitter value). [AWS argues](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/) that fractional 50% ("equal") jitter does more work and takes more time than full or decorrelated jitter in a load shedding scenario, so full jitter may be a safer default than 10% fractional jitter

### DIFF
--- a/go/tools/retry/retry_test.go
+++ b/go/tools/retry/retry_test.go
@@ -235,7 +235,7 @@ func TestRetry_startAttempt_ContextCancelledDuringWait(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, 1, r.attempt)
 
-	// Cancel context then try second attempt (which would wait)
+	// Cancel context then try second attempt
 	cancel()
 	err = r.startAttempt(ctx)
 	assert.Error(t, err)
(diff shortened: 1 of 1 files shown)


go/multipooler/manager/manager.go
```diff
@@ -345,43 +346,38 @@ func (pm *MultiPoolerManager) loadMultiPoolerFromTopo() {
 		return
 	}
 
-	ticker := timertools.NewBackoffTicker(100*time.Millisecond, 30*time.Second)
-	<-ticker.C
```
dweitzman, line 372: Not sure why previous code did a delay before its first attempt, so I removed the delay. Open to adding initial delay back if there's a reason I'm overlooking

go/tools/retry/retry.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// startAttempt prepares for the next retry attempt by waiting for the backoff delay.
+// On the first call (attempt 0), it returns immediately unless WithInitialDelay was configured.
+// On subsequent calls, it waits for the exponentially increasing backoff delay.
+//
+// Returns:
+//   - nil if the caller should proceed with the next attempt
+//   - ctx.Err() if the context was cancelled or timed out during the wait
+func (r *Retry) startAttempt(ctx context.Context) error {
```
sougou, line 122: I think this can be renamed to `nextAttempt` or even just `wait`?

dweitzman: The first attempt won't do any waiting unless initial delay is toggled on, so `wait` doesn't feel 100% accurate to me

`startAttempt` feels more clear to me than `nextAttempt` to indicate that it's doing something (potentially waiting), but I'm not strongly attached to the name. It's a non-exported API, so the name is only for unit tests

dweitzman: `shouldAttempt` could be another option?

sougou: Let's leave it as it is.

PR #177, opened 2025-10-24
feat(multipooler): implement manager RPC for stop and get replication status

rafael:
# Description

Implements `StopReplicationAndGetStatus` RPC method for the multipooler manager API. Similar to `StopReplication`, this method stops WAL replay on a standby, but with a key difference: **it waits for replication to actually be paused before capturing and returning the status**. This ensures the LSN returned represents the exact point at which replication stopped, not a potentially still-advancing LSN.

  ## Implementation Details

  - Calls `pg_wal_replay_pause()` to request the pause
  - Polls `pg_is_wal_replay_paused()` until it returns true (handles PostgreSQL's async pause behavior)
  - Captures all replication status (LSN, pause state, primary_conninfo) atomically once paused
  - Uses a context with 10-second timeout for the polling loop
  - Implemented via helper function `waitForReplicationPause()` for clean separation of concerns

  ## Why This Matters

PostgreSQL's `pg_wal_replay_pause()` is **asynchronous**. Without waiting for the pause to complete, we risk returning an LSN that continues to advance. This implementation guarantees the returned LSN is stable and represents the true stopping point.

## Tests
  - **Unit tests**: Removed from "not implemented" test suite
  - **E2E tests**: Added `TestStopReplicationAndGetStatus` with 3 scenarios:
    1. Correctly rejects on PRIMARY pooler (FAILED_PRECONDITION)
    2. Stops replication on REPLICA and **verifies LSN doesn't advance** after subsequent writes to primary
    3. Handles already-paused replication gracefully (idempotent)

## Question for Review

  **Should our existing `StopReplication` method also wait for pause completion?** Currently it doesn't, which means callers can't be sure when replication has actually stopped. This  could lead to race conditions in orchestration code.

### DIFF
--- a/go/test/endtoend/multipooler_manager_test.go
+++ b/go/test/endtoend/multipooler_manager_test.go
@@ -1604,7 +1604,7 @@ func TestStopReplicationAndGetStatus(t *testing.T) {
 
 		// Verify LSN is populated
 		assert.NotEmpty(t, stopResp.Status.Lsn, "LSN should not be empty")
-		assert.Contains(t, stopResp.Status.Lsn, "/", "LSN should be in PostgreSQL format (e.g., 0/1234ABCD)")
+		assert.Contains(t, stopResp.Status.Lsn, "/", "LSN should be in PostgreSQL format")
 
 		// Verify PrimaryConnInfo is populated (should still be set even though replication is paused)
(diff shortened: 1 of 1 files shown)


sougou:
> Should our existing StopReplication method also wait for pause completion? Currently it doesn't, which means callers can't be sure when replication has actually stopped. This could lead to race conditions in orchestration code.

Agree. StopReplication should wait.

PR #178, opened 2025-10-25
feat(pooler): add columns to heartbeat table to use for consensus

deepthi:
Prepare heartbeat table changes to use for consensus / leader elections.

### DIFF
--- a/go/multipooler/heartbeat/writer.go
+++ b/go/multipooler/heartbeat/writer.go
@@ -197,15 +197,18 @@ func (w *Writer) write() error {
 	// Get current leader term
 	leaderTerm := w.leaderTerm.Load()
 
+	// Get current timestamp in nanoseconds
+	tsNano := w.now().UnixNano()
+
 	_, err = conn.ExecContext(ctx, `
 		INSERT INTO multigres.heartbeat (shard_id, pooler_id, ts, leader_term, leader_wal_position)
-		VALUES ($1, $2, NOW(), $3, $4)
+		VALUES ($1, $2, $3, $4, $5)
 		ON CONFLICT (shard_id) DO UPDATE
 		SET pooler_id = EXCLUDED.pooler_id,
-		    ts = NOW(),
+		    ts = EXCLUDED.ts,
 		    leader_term = EXCLUDED.leader_term,
 		    leader_wal_position = EXCLUDED.leader_wal_position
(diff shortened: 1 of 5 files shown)


go/multipooler/manager/db.go
```diff
@@ -101,9 +101,11 @@ func CreateSidecarSchema(db *sql.DB) error {
 	// Create the heartbeat table
 	_, err = db.Exec(`
 		CREATE TABLE IF NOT EXISTS multigres.heartbeat (
-			shard_id BYTEA PRIMARY KEY,
+			shard_id TEXT PRIMARY KEY,
 			pooler_id TEXT NOT NULL,
-			ts BIGINT NOT NULL
+			ts TIMESTAMP NOT NULL,
```
deepthi: Is there a reason this needs to be BIGINT? It seems to me that timestamp works as well.
cc @cuongdo

go/multipooler/heartbeat/writer.go
```diff
@@ -189,20 +194,40 @@ func (w *Writer) write() error {
...
+	var lsn string
+	err := w.db.QueryRowContext(ctx, "SELECT pg_current_wal_lsn()").Scan(&lsn)
+	if err != nil {
+		return "", mterrors.Wrap(err, "failed to get WAL position")
+	}
+	return lsn, nil
+}
+
+// SetLeaderTerm updates the leader term for consensus tracking
+func (w *Writer) SetLeaderTerm(term int64) {
```
rafael: are we using this?

go/multipooler/manager/db.go
cuongdo: Timestamp should work. BIGINT was used in the Vitess schema, but I can't see why we would need more than the microsecond resolution of TIMESTAMP.

go/multipooler/heartbeat/writer.go
deepthi: good point. we are not yet. We can leave it in or delete it in this PR and add it later.

rafael: I would say let's remove it and we come back to it

go/multipooler/manager/db.go
sougou: Is there a downside to using higher precision? If not, we might as well use it.

PR #179, opened 2025-10-26
feat: Add PostgreSQL Wire Protocol Server

GuptaManan100:
## Summary
Implements the PostgreSQL wire protocol server to enable PostgreSQL clients (psql, pgx, psycopg2, etc.) to connect to multigres and execute queries.

  ## What's Included

  ### Core Infrastructure
  - **Buffer pooling system** - Vitess-inspired bucket pool with exponential sizing strategy
  - **Low-level packet I/O** - Message reading/writing with buffering and type-safe packet handling
  - **Reusable package structure** - Standalone `go/pgprotocol/` package organized into `protocol/`, `bufpool/`, and `server/` subdirectories

  ### Connection Lifecycle
  - **Startup handshake** - StartupMessage parsing, parameter extraction
  - **Authentication flow** - Trust mode authentication with BackendKeyData and ParameterStatus messages
  - **Session initialization** - ReadyForQuery state management

  ### Query Execution (Simple Protocol)
  - **Query message handling** - 'Q' message parsing and routing to handler
  - **Response encoding** - RowDescription, DataRow, CommandComplete, EmptyQueryResponse
  - **Error handling** - ErrorResponse and NoticeResponse with field encoding
  - **Command loop** - Main message dispatcher with connection state management

  ## Future Work

  - Extended query protocol (prepared statements)
  - Advanced authentication methods
  - Additional PostgreSQL features (COPY, transactions, etc.)
  - Wiring up the multi gateway handler to actually send the queries to the multi poolers.


go/pgprotocol/server/listener.go
```diff
@@ -0,0 +1,196 @@
...
+	listener net.Listener
+
+	// handler processes queries for connections.
+	handler Handler
+
+	// logger for logging.
+	logger *slog.Logger
+
+	// connBufferPooling enables buffer pooling for connections.
+	connBufferPooling bool
```
sougou: Is there a reason we need this option? I feel like we should just choose an optimal one and live with that. I also think the default option would be to pool because it has the best of both worlds.

GuptaManan100: I initially added it to be able to test without needing buffering, but after you pointed it out, I've reconsidered. It's probably not required. Tests can also use buffering, its fine.

GuptaManan100:
> I see that client is not implemented yet. Do you plan to implement one eventually, or can we continue living with third party libraries?

I don't see the need to implement this right away. We'll only be using the client side of the postgres protocol for connection from the multi pooler to the postgres instance. We can use the existing third party libraries for now, but I wanted to leave the door open for us to be able to write one of our own incase we want to. We already know the third party libraries have issues with dropped connections etc, and we might want to write our own implementation for that reason. So I left it as a possibility for the future.

PR #182, opened 2025-10-27
feat(multipooler): small fixes for Manager API

rafael:
# Desc

These are two small follow up from previous PR's:
1) Remove update consensus term call when updating standby list: https://github.com/multigres/multigres/pull/175#discussion_r2457234859
2) Stop replication should make sure to wait: https://github.com/multigres/multigres/pull/177#issuecomment-3448857415

### DIFF
--- a/go/multipooler/manager/manager.go
+++ b/go/multipooler/manager/manager.go
@@ -1147,7 +1147,7 @@ func (pm *MultiPoolerManager) UpdateSynchronousStandbyList(ctx context.Context, 
 		"force", force)
 
 	// === Validation ===
-	// TODO: We need to validate consesnsus term here.
+	// TODO: We need to validate consensus term here.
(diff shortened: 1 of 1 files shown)


go/multipooler/manager/manager.go
```diff
@@ -1138,12 +1146,11 @@ func (pm *MultiPoolerManager) UpdateSynchronousStandbyList(ctx context.Context,
 		"consensus_term", consensusTerm,
 		"force", force)
 
-	// Validate and update consensus term following consensus rules
-	if err := pm.validateAndUpdateTerm(ctx, consensusTerm, force); err != nil {
-		return err
-	}
-
 	// === Validation ===
+	// TODO: We need to validate consesnsus term here.
```
deepthi:
```suggestion
	// TODO: We need to validate consensus term here.
```

PR #183, opened 2025-10-27
feat(multipooler): implements GetFollowers from admin API

rafael:
# Description

Implements the GetFollowers RPC endpoint to provide comprehensive replication monitoring for primary servers. This endpoint returns the complete configured cohort of standby servers from synchronous_standby_names along with their connection status and detailed replication metrics from pg_stat_replication.

Follows the same patterns we've been using for all the other methods

# Tests
- Added e2e TestGetFollowers with 5 subtests:
    - No sync replication configured
    - Single connected follower
    - Follower disconnection detection
    - Mixed connected/disconnected followers
    - Standby-only server (should fail with PRIMARY guardrail)

### DIFF
--- a/go/multipooler/manager/manager.go
+++ b/go/multipooler/manager/manager.go
@@ -755,6 +755,14 @@ func (pm *MultiPoolerManager) StopReplication(ctx context.Context) error {
 		return mterrors.Wrap(err, "failed to pause WAL replay")
 	}
 
+	// Wait for WAL replay to actually be paused
+	// pg_wal_replay_pause() is asynchronous, so we need to wait for it to complete
+	pm.logger.Info("Waiting for WAL replay to complete pausing")
+	_, err = pm.waitForReplicationPause(ctx)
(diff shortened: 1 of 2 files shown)


go/multipooler/manager/manager.go
```diff
@@ -1348,7 +1349,7 @@ func (pm *MultiPoolerManager) PrimaryStatus(ctx context.Context) (*multipoolerma
 		pm.logger.Error("Error iterating pg_stat_replication rows", "error", err)
 		return nil, mterrors.Wrap(err, "failed to read connected followers")
 	}
-	status.Followers = followers
+	status.ConnectedFollowers = followers
```
rafael, line 1359: This returns only followers that are currently connected. Open for discussion if we want to make it match what I did in GetFollowers, which returns the full list of configured standbys (whether they are currently connected or not).

Leaning towards revisiting at a later point when start using these API's in the consensus work.

PR #187, opened 2025-10-29
feat: Implement query serving pipeline from multi gateway to multipooler

GuptaManan100:
## Query Serving Path: Complete Pipeline Implementation

  This PR implements a complete query execution pipeline for MultiGateway, enabling PostgreSQL clients to execute queries that are routed to multipooler instances via gRPC.

  - Integrated all layers from PostgreSQL wire protocol → multigateway handler → executor → planner → engine primitives → pooler execution. They are all simple implementations right now, which can be expanded on later.

### DIFF
--- a/go/multigateway/planner/planner.go
+++ b/go/multigateway/planner/planner.go
@@ -42,12 +42,6 @@ func NewPlanner(defaultTableGroup string, logger *slog.Logger) *Planner {
 
 // Plan creates an execution plan for the given SQL query.
 //
-// Phase 1 Implementation:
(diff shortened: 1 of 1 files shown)


GuptaManan100:
> I see that you've decided to hide context inside Conn. This approach may bite you back. I think go standards state that context should always be the first and explicit parameter to all functions. I'll leave this judgment call to you.
> 
> However, I see that `callback` does not have a context. We may need to add that, unless you think it won't be necessary.

Yes, I fixed the code to address both the issues.

PR #188, opened 2025-10-29
feat(consensus): implement pooler RPCs

deepthi:
# Consensus RPCs

  ConsensusStatus
  - Returns node's consensus state for health monitoring
  - Reports: local term (from disk), leader term (from Postgres heartbeat table), WAL position, role (primary/replica), health status, eligibility
  - Lazy-loads consensus term from disk if not in memory

  GetLeadershipView
  - Returns current leader information from the heartbeat table
  - Delegates to replication tracker's heartbeat reader
  - Provides: leader ID, leader term, WAL position, last heartbeat timestamp, replication lag

  GetWALPosition
  - Simple wrapper around GetCurrentWALPosition
  - Returns current WAL position for this node (different fields for primary vs standby)

  CanReachPrimary
  - Checks if this standby can reach the specified primary
  - Queries pg_stat_wal_receiver to verify WAL receiver is active ("streaming" status)
  - Parses conninfo to validate host/port match the expected primary
  - Returns reachable=false if: no WAL receiver, WAL receiver stopping, host/port mismatch, or DB unavailable

  BeginTerm
  - Validates term, checks if already voted, verifies replication lag (30s threshold)
  - Accepts term if: newer/same term + not accepted different candidate + caught up on replication
  - Persists decision to disk for crash recovery
  - Note: the 30s threshold is a placeholder, whether this node can vote or not will be refined as we go along.

---

Along the way
- renamed pooler_id to leader_id in heartbeat table
- factored out endtoend pooler manager test setup into a separate file

### DIFF
--- a/go/clustermetadata/topo/memorytopo/store.go
+++ b/go/clustermetadata/topo/memorytopo/store.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"log/slog"
 	"math/rand/v2"
-	"os"
 	"regexp"
 	"strings"
 	"sync"
@@ -324,12 +323,7 @@ func NewServerAndFactory(ctx context.Context, cells ...string) (topo.Store, *Fac
 	}
 	f.cells[topo.GlobalCell] = f.newDirectory(topo.GlobalCell, nil)
 
-	ts, err := topo.NewWithFactory(f, "" /*root*/, []string{""} /*serverAddrs*/)
-	if err != nil {
-		// CHECK THIS BEFORE MERGING
-		slog.Error("topo.NewWithFactory() failed", "error", err)
-		os.Exit(1)
-	}
+	ts := topo.NewWithFactory(f, "" /*root*/, []string{""} /*serverAddrs*/)
 	for _, cell := range cells {
 		f.cells[cell] = f.newDirectory(cell, nil)
 		// Create cell with mock server addresses for testing
--- a/go/clustermetadata/topo/mock_test.go
+++ b/go/clustermetadata/topo/mock_test.go
@@ -1,0 +1,311 @@
+// Copyright 2025 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package topo
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/multigres/multigres/go/mterrors"
+	"github.com/multigres/multigres/go/pb/mtrpc"
+)
+
+// mockConn is a mock implementation of Conn for testing
+type mockConn struct {
+	id              int
+	closed          bool
+	shouldFailCalls bool
+	mu              sync.Mutex
+	data            map[string][]byte
+}
+
+func newMockConn(id int) *mockConn {
+	return &mockConn{
+		id:   id,
+		data: make(map[string][]byte),
+	}
+}
+
+func (m *mockConn) setShouldFailCalls(fail bool) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.shouldFailCalls = fail
+}
+
+func (m *mockConn) checkError() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if m.shouldFailCalls {
+		return mterrors.Errorf(mtrpc.Code_UNAVAILABLE, "connection error")
+	}
+	if m.closed {
+		return mterrors.Errorf(mtrpc.Code_UNAVAILABLE, "connection closed")
+	}
+	return nil
+}
+
+func (m *mockConn) ListDir(ctx context.Context, dirPath string, full bool) ([]DirEntry, error) {
+	if err := m.checkError(); err != nil {
+		return nil, err
+	}
+	return []DirEntry{{Name: "test"}}, nil
+}
+
+func (m *mockConn) Create(ctx context.Context, filePath string, contents []byte) (Version, error) {
+	if err := m.checkError(); err != nil {
+		return nil, err
+	}
+	m.mu.Lock()
+	m.data[filePath] = contents
+	m.mu.Unlock()
+	return &mockVersion{version: "1"}, nil
+}
+
+func (m *mockConn) Update(ctx context.Context, filePath string, contents []byte, version Version) (Version, error) {
+	if err := m.checkError(); err != nil {
+		return nil, err
+	}
+	m.mu.Lock()
+	m.data[filePath] = contents
+	m.mu.Unlock()
+	return &mockVersion{version: "2"}, nil
+}
+
+func (m *mockConn) Get(ctx context.Context, filePath string) ([]byte, Version, error) {
+	if err := m.checkError(); err != nil {
+		return nil, nil, err
+	}
+	m.mu.Lock()
+	data, ok := m.data[filePath]
+	m.mu.Unlock()
+	if !ok {
+		return nil, nil, &TopoError{Code: NoNode}
+	}
+	return data, &mockVersion{version: "1"}, nil
+}
+
+func (m *mockConn) GetVersion(ctx context.Context, filePath string, version int64) ([]byte, error) {
+	if err := m.checkError(); err != nil {
+		return nil, err
+	}
+	return []byte("test"), nil
+}
+
+func (m *mockConn) List(ctx context.Context, filePathPrefix string) ([]KVInfo, error) {
+	if err := m.checkError(); err != nil {
+		return nil, err
+	}
+	return []KVInfo{{Key: []byte("key"), Value: []byte("value")}}, nil
+}
+
+func (m *mockConn) Delete(ctx context.Context, filePath string, version Version) error {
+	return m.checkError()
+}
+
+func (m *mockConn) Lock(ctx context.Context, dirPath, contents string) (LockDescriptor, error) {
+	if err := m.checkError(); err != nil {
+		return nil, err
+	}
+	return &mockLockDescriptor{}, nil
+}
+
+func (m *mockConn) LockWithTTL(ctx context.Context, dirPath, contents string, ttl time.Duration) (LockDescriptor, error) {
+	if err := m.checkError(); err != nil {
+		return nil, err
+	}
+	return &mockLockDescriptor{}, nil
+}
+
+func (m *mockConn) LockName(ctx context.Context, dirPath, contents string) (LockDescriptor, error) {
+	if err := m.checkError(); err != nil {
+		return nil, err
+	}
+	return &mockLockDescriptor{}, nil
+}
+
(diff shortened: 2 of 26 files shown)


go/multipooler/grpcconsensusservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+			srv := &consensusService{
+				manager: pm,
+			}
+			consensuspb.RegisterMultiPoolerConsensusServer(grpc.Server, srv)
+		}
+	})
+}
+
+// RequestVote handles vote requests during leader election
+func (s *consensusService) RequestVote(ctx context.Context, req *consensusdata.RequestVoteRequest) (*consensusdata.RequestVoteResponse, error) {
```
sougou: Rename to `StartNewTerm`

deepthi: renamed to BeginTerm

go/multipooler/manager/rpc_consensus.go
```diff
@@ -0,0 +1,284 @@
...
+		term, err := GetTerm(pm.config.PoolerDir)
+		if err != nil {
+			pm.mu.Unlock()
+			return nil, fmt.Errorf("failed to load consensus state: %w", err)
+		}
+		pm.consensusTerm = term
+	}
+
+	currentTerm := pm.consensusTerm.GetCurrentTerm()
+	votedFor := pm.consensusTerm.GetVotedFor()
```
deepthi, line 53: I'm going to change how we read and store consensusTerm in a follow up PR. At that time I will also rename this field.

go/multipooler/grpcconsensusservice/service.go
```diff
@@ -0,0 +1,89 @@
...
+func (s *consensusService) GetLeadershipView(ctx context.Context, req *consensusdata.LeadershipViewRequest) (*consensusdata.LeadershipViewResponse, error) {
+	resp, err := s.manager.GetLeadershipView(ctx, req)
+	if err != nil {
+		return nil, mterrors.ToGRPC(err)
+	}
+	return resp, nil
+}
+
+// GetWALPosition returns the current WAL position
+func (s *consensusService) GetWALPosition(ctx context.Context, req *consensusdata.GetWALPositionRequest) (*consensusdata.GetWALPositionResponse, error) {
```
rafael, line 74: Do we need this? We already have `PrimaryStatus` and `ReplicationStatus` to get WAL information.

go/multipooler/manager/manager.go
```diff
@@ -535,6 +535,12 @@ func (pm *MultiPoolerManager) validateAndUpdateTerm(ctx context.Context, request
 		pm.consensusTerm = newTerm
 		pm.mu.Unlock()
 
+		// Synchronize term to heartbeat writer if it exists
+		if pm.replTracker != nil {
```
rafael, line 539: I wonder if we need this extra synchronization. From what I see, the existing writer logic that updates leader_term on write via the heartbeat should be sufficient.

Doing it here feels brittle by design, there will be times when we need to write the term but can’t write to the DB. It might be better to avoid that extra complexity altogether.

go/multipooler/manager/manager.go
```diff
@@ -571,6 +577,12 @@ func (pm *MultiPoolerManager) loadConsensusTermFromDisk() {
 		pm.consensusLoaded = true
 		pm.mu.Unlock()
 
+		// Synchronize term to heartbeat writer if it exists
```
rafael, line 580: same for these other synchronizations

go/multipooler/manager/rpc_consensus.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	// Test database connectivity
+	if err := pm.db.PingContext(ctx); err != nil {
+		return nil, fmt.Errorf("postgres unhealthy, cannot accept new term: %w", err)
+	}
+
+	// Load consensus term from disk if not already loaded
+	pm.mu.Lock()
+	if pm.consensusTerm == nil {
```
rafael, line 43: Related to your comment below, for follow up. Probably we want to encapsulate all this is a convenience function

go/multipooler/manager/rpc_consensus.go
```diff
@@ -0,0 +1,284 @@
...
+	pm.mu.Unlock()
+
+	response := &consensusdatapb.BeginTermResponse{
+		Term:     currentTerm,
+		Accepted: false,
+		PoolerId: pm.serviceID.GetName(),
+	}
+
+	// Reject if term is outdated
+	if req.Term < currentTerm {
```
rafael, line 63: validateAndUpdateTerm was doing this logic already. Probably for your follow up PR, but we should have just one place where we do this logic.

go/multipooler/manager/rpc_consensus.go
```diff
@@ -0,0 +1,284 @@
...
+	localTerm := pm.consensusTerm.GetCurrentTerm()
+	pm.mu.Unlock()
+
+	// Get last successful leader term from Postgres (replicated data)
+	var leaderTerm int64
+	isHealthy := false
+
+	if pm.db != nil {
+		err := pm.db.QueryRowContext(ctx, `
+			SELECT COALESCE(MAX(leader_term), 0)
```
rafael, line 146: Would it be safe to select the term from the last row? This query could get expensive as more rows get added. Also thinking that it seems that another invariant would be broken if the max leader_term is not the newest row

go/multipooler/manager/rpc_consensus.go
```diff
@@ -0,0 +1,284 @@
...
+	if isHealthy {
+		var err error
+		walPosition, err = pm.GetCurrentWALPosition(ctx)
+		if err != nil {
+			pm.logger.Warn("Failed to get WAL position", "error", err)
+		}
+	}
+
+	// Determine role (primary/replica)
+	role := "replica"
```
rafael, line 170: Let' talk about this. It seems that we should be able to consolidate here with logic already implemented for PrimaryStatus / ReplicaStatus.

go/multipooler/manager/rpc_consensus.go
```diff
@@ -0,0 +1,284 @@
...
+	reader := pm.replTracker.HeartbeatReader()
+	view, err := reader.GetLeadershipView()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get leadership view: %w", err)
+	}
+
+	return &consensusdatapb.LeadershipViewResponse{
+		LeaderId:          view.LeaderID,
+		LeaderTerm:        view.LeaderTerm,
+		LeaderWalPosition: view.LeaderWALPosition,
```
rafael, line 207: Thinking we should not expose this WAL here. This will be lagged and it might not be the actual LeaderWalPosition.

go/multipooler/manager/rpc_consensus_test.go
```diff
@@ -0,0 +1,780 @@
...
+	consensusdatapb "github.com/multigres/multigres/go/pb/consensusdata"
+	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
+)
+
+// TestBeginTerm_AlreadyVotedInOlderTerm tests the scenario where:
+// 1. Node has voted for candidate A in term 5
+// 2. Candidate B requests vote in term 10 (newer term)
+// Expected: Vote should be accepted because the newer term should reset the vote
+// Bug: Currently the code checks votedFor before updating the term, which would reject the vote
+func TestBeginTerm_AlreadyVotedInOlderTerm(t *testing.T) {
```
rafael, line 43: Going to punt on reviewing the tests for the follow up PR

go/multipooler/manager/rpc_consensus_test.go
```diff
@@ -0,0 +1,780 @@
...
+	loadedTerm, err := GetTerm(tmpDir)
+	require.NoError(t, err)
+	assert.Equal(t, int64(10), loadedTerm.CurrentTerm)
+	assert.Equal(t, "candidate-B", loadedTerm.VotedFor.GetName())
+
+	// Verify all expectations were met
+	assert.NoError(t, mock.ExpectationsWereMet())
+}
+
+// TestBeginTerm_AlreadyVotedInSameTerm tests the scenario where:
```
rafael, line 106: Idea: for these suite, we can rely more in unit tests of the core consensus logic around term validation. I was taking that approach in here: https://github.com/multigres/multigres/blob/main/go/multipooler/manager/manager_test.go#L288 

I think for this kind assertion a matrix style makes it easier to parse all the tests

go/test/endtoend/multipooler_manager_test.go
```diff
@@ -33,758 +26,13 @@ import (
...
 	"github.com/multigres/multigres/go/test/utils"
-	"github.com/multigres/multigres/go/tools/pathutil"
 
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	multipoolermanagerpb "github.com/multigres/multigres/go/pb/multipoolermanager"
 	multipoolermanagerdata "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
-	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
 )
 
-var (
```
rafael, line 48: Ah nice, I wanted to do this

go/multipooler/manager/rpc_consensus.go
deepthi: There is expected to be only one row, because we always do an upsert. Technically we can change it to select without any aggregation and add LIMIT 1

rafael: ah - I think that would be more intuitive

go/multipooler/manager/rpc_consensus.go
```diff
@@ -0,0 +1,284 @@
...
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+	consensusdatapb "github.com/multigres/multigres/go/pb/consensusdata"
+	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
+)
+
+// BeginTerm handles coordinator requests during leader appointments
+func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatapb.BeginTermRequest) (*consensusdatapb.BeginTermResponse, error) {
```
rafael, line 30: Per our discussion this should also take into account:
- Revocation
- Discovery

rafael (approved):
For those following at home, we discussed this PR and comments will be addressed in the next PR's.

go/multipooler/manager/rpc_consensus.go
deepthi: PR https://github.com/multigres/multigres/pull/202

go/multipooler/manager/manager.go
deepthi: Actually there are two parts to this
- when there is a leader transition, we need to initialize the leaderTerm with the database value, that is missing right now and is a bug.
- when we get a new leader appointed, we need to write the new term into the heartbeat table. this is being achieved by setting the in memory value on the writer and it gets written on the next write. 

This still feels wrong. We should wait for the appointment to be complete. i.e. a new leader has been accepted by the required quorum, and only then the new leader should write the new term value into the heartbeat table. So for now, it would make sense to delete this, and add it back in the right place when we implement leader appointment fully.
The point is that there was no "existing writer logic that updates leader_term". It was forever 0.

go/multipooler/manager/rpc_consensus.go
deepthi: PR #202

go/multipooler/manager/rpc_consensus_test.go
deepthi: Implemented in #202

Issue #189, opened 2025-10-30
Flaky test TestMultiGateway_PostgreSQLConnection


sougou:
Pasting log here:

```
Cleaning up multigateway process (PID: 58123)...
Cleaning up etcd process (PID: 58119)...
Cleaning up multiadmin process (PID: 58121)...
{"level":"warn","ts":"2025-10-29T21:52:22.019981-0700","logger":"etcd-client","caller":"v3@v3.6.5/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0x140004da5a0/localhost:14829","method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = context deadline exceeded"}
{"level":"warn","ts":"2025-10-29T21:52:22.052171-0700","logger":"etcd-client","caller":"v3@v3.6.5/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0x140004da5a0/localhost:14829","method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = Canceled desc = context canceled"}
{"level":"warn","ts":"2025-10-29T21:52:22.083460-0700","logger":"etcd-client","caller":"v3@v3.6.5/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0x140004da5a0/localhost:14829","method":"/etcdserverpb.Lease/LeaseRevoke","attempt":0,"error":"rpc error: code = NotFound desc = etcdserver: requested lease not found"}
{"level":"warn","ts":"2025-10-29T21:52:23.238942-0700","logger":"etcd-client","caller":"v3@v3.6.5/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0x140003601e0/localhost:14829","method":"/etcdserverpb.Lease/LeaseRevoke","attempt":0,"error":"rpc error: code = NotFound desc = etcdserver: requested lease not found"}
{"level":"warn","ts":"2025-10-29T21:52:24.321609-0700","logger":"etcd-client","caller":"v3@v3.6.5/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0x140003603c0/localhost:14829","method":"/etcdserverpb.KV/Txn","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = context deadline exceeded"}
[...]

PR #190, opened 2025-10-30
fix: flaky test TestMultiGateway_PostgreSQLConnection

sulthonmb:
# Fix Port Conflicts in multipooler_manager_test.go

## Problem

The `multipooler_manager_test.go` test was experiencing flaky failures due to port conflicts, as reported in issue [#189](https://github.com/multigres/multigres/issues/189). The root cause was inconsistent port allocation mechanisms across the test suite:

- **`testutil.GenerateRandomPort()`**: Used in `multipooler_manager_test.go`, generates random ports between 10000-65535 without any coordination or conflict detection
- **`utils.GetNextPort()`**: Used in other tests like `cluster_test.go`, employs a systematic approach with process-based offsets to avoid port conflicts

This inconsistency led to race conditions where multiple tests could attempt to use the same randomly generated port, causing test failures.

### DIFF
--- a/go/test/endtoend/multipooler_manager_test.go
+++ b/go/test/endtoend/multipooler_manager_test.go
@@ -552,13 +552,13 @@ func getSharedTestSetup(t *testing.T) *MultipoolerTestSetup {
 
 		t.Logf("Created topology cell '%s' at etcd %s", cellName, etcdClientAddr)
 
-		// Generate ports for shared instances
-		primaryGrpcPort := testutil.GenerateRandomPort()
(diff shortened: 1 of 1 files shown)


rafael (approved):
LGTM. I ran into this issue as well in https://github.com/multigres/multigres/pull/191. I started to update the ports there. Will merge this and rebase.

Good catch!

deepthi:
@rafael we should delete GenerateRandomPort altogether. We can do it in one of the cleanup PRs that are planned.

PR #192, opened 2025-10-30
feat(consensus): maintain consensus state in memory

deepthi:
- Consensus state: save to disk and load from disk as needed

The rest of the changes in this PR are based on reviews of #188, and code inspection.
- Terminology changes to remove voting and elections
- remove `GetWALPosition` RPC, we already have `PrimaryStatus` and `ReplicaStatus`
- reduce code duplication and reuse `validateAndUpdateTerm`
- remove `leader_wal_position` from `heartbeat` table, it is guaranteed to be stale
- initialize `leaderTerm` in heartbeat writer from database if it already exists
- refactor tests for readability
- use `sync.Mutex` instead of `sync.RWMutex`

Enhancing BeginTerm with Revocation and Discovery will be done in a follow up PR.

### DIFF
--- a/go/event/event.go
+++ b/go/event/event.go
@@ -81,7 +81,7 @@ import (
 )
 
 var (
-	listenersMutex sync.RWMutex // protects listeners and interfaces
+	listenersMutex sync.Mutex // protects listeners and interfaces
 	listeners      = make(map[reflect.Type][]any)
 	interfaces     = make([]reflect.Type, 0)
 )
@@ -128,26 +128,28 @@ func AddListener(fn any) {
 // Dispatch sends an event to all registered listeners that were declared
 // to accept values of the event's type, or interfaces that the value implements.
 func Dispatch(ev any) {
-	listenersMutex.RLock()
-	defer listenersMutex.RUnlock()
-
 	evType := reflect.TypeOf(ev)
-	vals := []reflect.Value{reflect.ValueOf(ev)}
+	listenersMap := make(map[reflect.Type][]any)
 
-	// call listeners for the actual static type
-	callListeners(evType, vals)
+	// Build a map of listeners to be called while holding the lock
+	func() {
+		listenersMutex.Lock()
+		defer listenersMutex.Unlock()
 
-	// also check if the type implements any of the registered interfaces
-	for _, in := range interfaces {
-		if evType.Implements(in) {
-			callListeners(in, vals)
+		listenersMap[evType] = append(listenersMap[evType], listeners[evType]...)
+
+		// also check if the type implements any of the registered interfaces
+		for _, in := range interfaces {
+			if evType.Implements(in) {
+				listenersMap[in] = append(listenersMap[in], listeners[in]...)
+			}
 		}
-	}
-}
+	}()
 
-func callListeners(t reflect.Type, vals []reflect.Value) {
-	for _, fn := range listeners[t] {
-		reflect.ValueOf(fn).Call(vals)
+	for _, list := range listenersMap {
+		for _, fn := range list {
+			reflect.ValueOf(fn).Call([]reflect.Value{reflect.ValueOf(ev)})
+		}
 	}
 }
 
--- a/go/multigateway/discovery.go
+++ b/go/multigateway/discovery.go
@@ -44,7 +44,7 @@ type PoolerDiscovery struct {
 	wg         sync.WaitGroup
 
 	// State
-	mu          sync.RWMutex
+	mu          sync.Mutex
 	poolers     map[string]*topo.MultiPoolerInfo // pooler ID -> pooler info
 	lastRefresh time.Time
 }
@@ -232,8 +232,8 @@ func (pd *PoolerDiscovery) processPoolerChange(watchData *topo.WatchDataRecursiv
 
 // GetPoolers returns a list of all discovered poolers.
 func (pd *PoolerDiscovery) GetPoolers() []*clustermetadatapb.MultiPooler {
-	pd.mu.RLock()
-	defer pd.mu.RUnlock()
+	pd.mu.Lock()
+	defer pd.mu.Unlock()
 
 	poolers := make([]*clustermetadatapb.MultiPooler, 0, len(pd.poolers))
 	for _, pooler := range pd.poolers {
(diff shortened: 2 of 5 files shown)


go/multipooler/grpcconsensusservice/service_test.go
```diff
@@ -33,15 +33,16 @@ import (
...
 )
 
-func TestConsensusService_BeginTerm(t *testing.T) {
+// setupConsensusService creates a test manager with consensus service initialized
+func setupConsensusService(t *testing.T) (*consensusService, func()) {
+	t.Helper()
+
 	ctx := context.Background()
 	logger := slog.New(slog.NewTextHandler(os.Stderr, nil))
 	ts, _ := memorytopo.NewServerAndFactory(ctx, "zone1")
```
dweitzman, line 42: Would it make sense to use t.Cleanup() here to schedule a cleanup at the end of the test instead of returning a cleanup function?

Returning a cleanup() function seems more error-prone if anything goes wrong during setupConsensusService(), so unless there are tests that need to do multiple setups + multiple cleanups it may be cleaner to just defer to t.Cleanup()

go/multipooler/heartbeat/writer.go
```diff
@@ -92,6 +92,18 @@ func (w *Writer) Open() {
 
 	// TODO: open connection pools here
 
+	// Initialize leaderTerm from database if a heartbeat row already exists
+	var existingTerm sql.NullInt64
+	err := w.db.QueryRow("SELECT leader_term FROM multigres.heartbeat WHERE shard_id = $1", w.shardID).Scan(&existingTerm)
+	if err != nil && err != sql.ErrNoRows {
+		w.logger.Warn("Failed to read existing leader_term from database", "error", err)
+	} else if existingTerm.Valid {
+		w.leaderTerm.Store(existingTerm.Int64)
```
dweitzman: Should this have any defense or assertion to make sure it's not storing a lower term number than what was already in leaderTerm in memory?

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,242 @@
...
+	cs.mu.RLock()
+	defer cs.mu.RUnlock()
+
+	if cs.term == nil {
+		return 0
+	}
+	return cs.term.GetCurrentTerm()
+}
+
+// GetAcceptedLeader returns the candidate ID this node voted for in the current term.
```
dweitzman: voted for -> accepted, and some similar change on the next line of the comment?

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,242 @@
...
+// Changes are marked dirty but not persisted - call Save() to persist.
+func (cs *ConsensusState) SetTerm(term *multipoolermanagerdatapb.ConsensusTerm) {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+
+	cs.term = cloneTerm(term)
+	cs.dirty = true
+}
+
+// AcceptAppointment records a vote for the specified candidate in the current term.
```
dweitzman: Another "vote"

go/multipooler/manager/manager.go
```diff
@@ -159,6 +160,27 @@ func (pm *MultiPoolerManager) unlock() {
...
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+
+	if pm.consensusState == nil {
+		pm.consensusState = NewConsensusState(pm.config.PoolerDir, pm.serviceID)
+
+		// Load existing consensus term from disk if it exists
+		if err := pm.consensusState.Load(); err != nil {
+			pm.logger.Error("Failed to load consensus state from disk", "error", err)
+			// Don't return error here - we'll initialize with default values (term=0)
```
dweitzman: Seems like there are two reasons for failing to read the file:
- A newly created shard has no consensus history. In this case we want to create a new file
- Somehow access to the consensus file was lost or corrupted (disk issue, permissions issue, operator error, etc.). In this case it seems less obvious how to safely recover. If we acknowledged joining a term and that state is corrupted on disk, maybe the safest option is fail pooler startup, similar to what we would do if postgres data on disk was unreadable?...

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,242 @@
...
+
+	if cs.term == nil || cs.term.AcceptedLeader == nil {
+		return ""
+	}
+	return cs.term.AcceptedLeader.GetName()
+}
+
+// GetTerm returns a copy of the current consensus term.
+// Returns nil if state has not been loaded.
+func (cs *ConsensusState) GetTerm() *multipoolermanagerdatapb.ConsensusTerm {
```
dweitzman, line 117: Reflecting on potential risks of updating the term in memory while exposing/exporting that through GetTerm() or GetCurrentTerm() when it's possible Save() failed and didn't persist the data to disk. I'm nervous about people reading, sharing, or acting on term data that didn't make it to disk.

Would it be any safer for ConsensusState to track both a `proposedTerm` and `durableTerm`, where the proposed term would be promoted into the durable term only when `Save()` succeeds?

go/multipooler/heartbeat/writer.go
deepthi: Agreed. For now I'm going to log an error, and we don't enable writes if we detect this condition. we need to pass it all the way up the chain and prevent a primary promotion if this happens. I'll leave those as TODOs for later.

go/multipooler/manager/consensus_state.go
deepthi: That might get complicated. But this reminded me that we should retry saving to disk if it fails, so I am adding that.

go/multipooler/manager/manager.go
deepthi: See my comments in code along with the change I made to return the error and fail startup.

go/multipooler/manager/consensus_state.go
deepthi: Ok, I think what we should do is only update the in-memory value after we are able to save to disk. Any failure to save should percolate up the stack and return from the RPC as an error. Callers will then need to retry.

deepthi:
I've moved the cleanup/review parts of this PR to #202. That will allow me to do a clean PR for the consensus state file instead of mixing multiple things into one PR.

deepthi:
We can close this now. I'll create a new PR for the consensus state feature.

PR #202, opened 2025-11-02
refactor(pooler): remove code duplication, change names to be consistent

deepthi:
The changes in this PR are based on reviews of https://github.com/multigres/multigres/pull/188, and code inspection.

- Terminology changes to remove voting and elections
- remove `GetWALPosition` RPC, we already have `PrimaryStatus` and `ReplicaStatus`
- reduce code duplication and reuse `validateAndUpdateTerm`
- remove `leader_wal_position` from `heartbeat` table, it is guaranteed to be stale
- refactor tests for readability
- use `sync.Mutex` instead of `sync.RWMutex`
- use `t.Cleanup` instead of `defer`


go/multipooler/manager/rpc_consensus.go
```diff
@@ -156,23 +140,40 @@ func (pm *MultiPoolerManager) ConsensusStatus(ctx context.Context, req *consensu
...
 		var inRecovery bool
 		err := pm.db.QueryRowContext(ctx, "SELECT pg_is_in_recovery()").Scan(&inRecovery)
-		if err == nil && !inRecovery {
-			role = "primary"
+		if err == nil {
+			if inRecovery {
+				// On standby: get receive and replay positions
+				var receiveLsn, replayLsn sql.NullString
+				err = pm.db.QueryRowContext(ctx,
+					"SELECT pg_last_wal_receive_lsn(), pg_last_wal_replay_lsn()").
```
deepthi, line 158: The reason we want to return both of these is that when we get to candidate selection, we will start using `last_receive_lsn`.

go/multipooler/grpcconsensusservice/service.go
```diff
@@ -70,15 +70,6 @@ func (s *consensusService) GetLeadershipView(ctx context.Context, req *consensus
 	return resp, nil
 }
 
-// GetWALPosition returns the current WAL position
```
rafael, line 73: nice

go/multipooler/grpcconsensusservice/service_test.go
```diff
@@ -37,11 +37,11 @@ func TestConsensusService_BeginTerm(t *testing.T) {
 	ctx := context.Background()
 	logger := slog.New(slog.NewTextHandler(os.Stderr, nil))
 	ts, _ := memorytopo.NewServerAndFactory(ctx, "zone1")
-	defer ts.Close()
+	t.Cleanup(func() { ts.Close() })
```
dweitzman: I think the `defer` here and elsewhere in this file should be fine since t.Run() is blocking

In the other PR my suggestion for t.Cleanup() was in a helper function where the helper was going to return while the test case was still running

Changing to t.Cleanup() here doesn't seem harmful, but might feel less idiomatic

deepthi: I wasn't completely sure if was needed either. I can change it back and then merge.

PR #204, opened 2025-11-03
refactor(multipooler): move rpc functions into it's own file

rafael:
# Description
- This is another small PR, refactoring `manager.go` to move RPC functions into it's own file. This follows same pattern we have for this in Vitess and what @deepthi  is already doing for rpc_consensus.go


go/multipooler/manager/manager.go
```diff
@@ -460,76 +459,6 @@ func (pm *MultiPoolerManager) loadMultiPoolerFromTopo() {
...
-		return err
-	}
-	defer pm.unlock()
-
-	pm.logger.Info("SetReadOnly called")
-	return mterrors.New(mtrpcpb.Code_UNIMPLEMENTED, "method SetReadOnly not implemented")
-}
-
-// IsReadOnly checks if PostgreSQL instance is in read-only mode
-func (pm *MultiPoolerManager) IsReadOnly(ctx context.Context) (bool, error) {
```
rafael, line 525: This was deleted altogether as we removed this from the service

PR #205, opened 2025-11-03
build(endtoend): kill orphan subprocesses after test runs

dweitzman:
Currently it's easy for integration tests run to leave behind orphan processes (etcd, postgres, multipooler, etc.)

It's not just an annoyance or a minor waste of resources: it can use up all the interprocess communication slots on OS X and block you from being able to run more tests

The goal of this PR is to have an approach to terminating orphan processes that's relatively simple but also reliable: we ideally don't want an approach that only fixes orphans the next time you run tests, since in local development you might not run tests often

The approach taken in this PR is:
- add a hidden flag to servenv processes to enable ophan detection, where they kill themselves if their parent process dies
- add a helper bash wrapper script for starting non-multigres processes (etcd & postgres) with the same behavior
- add a helper bash script for daemons like postgres that can detect if a parent process terminates and run a custom command, in this case `pg_ctl stop <data-dir>` after the parent test run dies

This only works on unix-like operating systems, but should work on both linux and mac


go/cmd/pgctld/command/restart.go
```diff
@@ -135,7 +135,7 @@ func RestartPostgreSQLWithResult(logger *slog.Logger, config *pgctld.PostgresCtl
 	} else {
 		logger.Info("Starting PostgreSQL server")
 	}
-	startResult, err := StartPostgreSQLWithResult(logger, config)
+	startResult, err := StartPostgreSQLWithResult(logger, config, false)
```
rafael: super nit - we had this convention in Vitess (which I think is good), where we add inline comments for booleans flags for clarity:
```
StartPostgreSQLWithResult(logger, config, false /* testOrphanDetection */)
```

I'm not sure where that came from, but always founded useful.

go/servenv/servenv_unix.go
```diff
@@ -118,3 +119,54 @@ func (sv *ServEnv) populateHostname() {
...
+				currentParentPID := os.Getppid()
+				if currentParentPID != parentPID {
+					slog.Warn("Parent process died, initiating graceful shutdown",
+						"original_parent_pid", parentPID,
+						"current_parent_pid", currentParentPID)
+
+					if !gracefulShutdownStarted {
+						gracefulShutdownStarted = true
+						// Trigger graceful shutdown by sending signal to exitChan
+						if sv.exitChan != nil {
```
rafael: do we need to hold the mutex here?

go/cmd/pgctld/command/start.go
```diff
@@ -239,6 +239,23 @@ func startPostgreSQLWithConfig(logger *slog.Logger, config *pgctld.PostgresCtlCo
...
+	// If test orphan detection is enabled, spawn a watchdog process that will stop postgres if parent dies
+	if testOrphanDetection {
+		logger.Info("Spawning watchdog process for orphan detection")
+		watchdogCmd := exec.Command(
+			"run_command_if_parent_dies.sh",
+			"pg_ctl", "stop",
+			"-D", config.PostgresDataDir,
+			"-m", "fast",
+		)
+		if err := watchdogCmd.Start(); err != nil {
```
rafael, line 251: do we need a `Wait()` for the watchdogCmd?  It seems like we could have a who watches the watcher problem

go/cmd/pgctld/command/start.go
```diff
@@ -145,7 +145,7 @@ func (s *PgCtlStartCmd) runStart(cmd *cobra.Command, args []string) error {
 }
 
 // StartPostgreSQLWithResult starts PostgreSQL with the given configuration and returns detailed result information
-func StartPostgreSQLWithResult(logger *slog.Logger, config *pgctld.PostgresCtlConfig) (*StartResult, error) {
+func StartPostgreSQLWithResult(logger *slog.Logger, config *pgctld.PostgresCtlConfig, testOrphanDetection bool) (*StartResult, error) {
```
rafael, line 148: Maybe we can add a test to double check this is working. Something like:
```
  func TestOrphanDetection(t *testing.T) {
      // Start a pgctld server with --test-orphan-detection
      cmd := exec.Command("pgctld", "server", "--test-orphan-detection", ...)
      require.NoError(t, cmd.Start())

      // Kill the parent process
      cmd.Process.Kill()

      // Wait and verify the child terminated itself
      time.Sleep(2 * time.Second)

      // Check that pgctld is no longer running
      err := cmd.Process.Signal(syscall.Signal(0))
      assert.Error(t, err, "Process should have terminated")
  }
```

go/servenv/servenv_unix.go
```diff
@@ -118,3 +119,54 @@ func (sv *ServEnv) populateHostname() {
...
+
+	parentPID := os.Getppid()
+	slog.Info("Starting orphan detection", "parent_pid", parentPID)
+
+	go func() {
+		ticker := time.NewTicker(1 * time.Second)
+		defer ticker.Stop()
+
+		forceKillTimer := time.NewTimer(0)
+		forceKillTimer.Stop()
```
rafael: this felt a bit confusing. Could we create this timer when we are gonna use it? So we don't have to stop and then re-start it

go/cmd/pgctld/command/start.go
dweitzman: We don't need to wait for the watchdog to finish as long as it starts correctly since it'll kill itself when we die

There's a potential risk here of a bug like the watchdog script isn't in the `PATH` and this call does nothing

go/cmd/pgctld/command/start.go
dweitzman: Added a test to endtoend/pgctld_test.go

go/servenv/servenv_unix.go
dweitzman: Simplified

go/cmd/pgctld/command/restart.go
dweitzman: Done

go/servenv/servenv_unix.go
dweitzman: Moved the `exitChan` creation into `NewServEnvWithConfig` to ensure it'll aways exist prior to `Init()` being called on the servenv

With that, exitChan should never be `nil` so I removed the `if` statement

go/test/endtoend/multipooler/setup_test.go
```diff
@@ -55,9 +55,9 @@ var (
 
 // TestMain sets the path and cleans up after all tests
 func TestMain(m *testing.M) {
-	// Set the PATH so etcd can be found
+	// Set the PATH so dependencies like etcd and run_in_test.sh can be found
 	// Use automatic module root detection instead of hard-coded relative paths
-	if err := pathutil.PrependBinToPath(); err != nil {
+	if err := pathutil.PrependModuleSubdirsToPath("bin", "go/test/endtoend"); err != nil {
```
dweitzman, line 60: FYI: Updated this for a merge conflict

makinje16:
Does this also fix #189? I know you mentioned this is for integration tests but could having left over processes also be the reason for that flaky end to end test as well since it says port 14832 is already in use?

I've kinda started to look at that issue but came across this PR

makinje16:
> Does this also fix #189? I know you mentioned this is for integration tests but could having left over processes also be the reason for that flaky end to end test as well since it says port 14832 is already in use?
> 
> I've kinda started to look at that issue but came across this PR

update: pretty sure yes

dweitzman:
> > Does this also fix #189? I know you mentioned this is for integration tests but could having left over processes also be the reason for that flaky end to end test as well since it says port 14832 is already in use?
> > I've kinda started to look at that issue but came across this PR
> 
> update: pretty sure yes

Wasn't specifically trying to fix that, but if it got fixed as a result all the better!

PR #208, opened 2025-11-04
refactor(multipooler): move replication methods into it's own file.

rafael:
# Description

This is another cleanup PR, refactoring some functions of manager.go into it's own file and improving unit test coverage.


go/multipooler/manager/rpc_manager.go
```diff
@@ -919,9 +919,9 @@ func (pm *MultiPoolerManager) Demote(ctx context.Context, consensusTerm int64, d
 	}
 
 	// Capture State & Make PostgreSQL Read-Only
-
-	finalLSN, err := pm.captureFinalLSN(ctx, state)
+	finalLSN, err := pm.getPrimaryLSN(ctx)
```
rafael, line 922: CaptureFinal LSN was doing effectively the same as this. So that I removed

go/multipooler/manager/pg_replication.go
```diff
@@ -0,0 +1,575 @@
+// Copyright 2025 Supabase, Inc.
```
deepthi, line 1: Should we call this something like replication_utils.go? @dweitzman wdyt?

go/multipooler/manager/pg_replication_test.go
```diff
@@ -0,0 +1,1510 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/multigres/multigres/go/mterrors"
+
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+)
+
```
deepthi, line 37: Lots of new tests. Nice work!

go/multipooler/manager/pg_replication.go
dweitzman: Since it has part of the manager implementation and not just helpers, my inclination is that `_utils.go` could be a little misleading

I don't have strong conviction either way

deepthi: What about the prefix? Does `pg_` make sense? I don't feel strongly about these, so if there is no agreement, we can go ahead with no changes.

dweitzman:
The pg doesn't seem super necessary to me

PR #209, opened 2025-11-04
build: enable sloglint to encourage using context versions of logging where possible

dweitzman:
Using methods like `ErrorContext()` instead of just `Error()` to log creates a path in the future to do things like annotate all the logs associated with a particular trace ID, or automatically tag logs with the RPC they relate to (`Demote`, `ExecuteQuery`, etc.)

There's a hypothetical future where we could see an error spike in metrics, use exemplars to navigate to a relevant example distributed trace where the error happened, and also see all log messages from that trace

Or we could also do things like look at an aggregated view of all the log statements related to a switchover operation across orchestrator and all the impacted poolers

We'd need future work to think about what context to automatically add into log statements and when/why to do it, but it feels like a safe bet that we'll eventually want trace ID at minimum


rafael (approved):
LGTM - Sorry for the merge conflict

PR #211, opened 2025-11-04
test: switch to env-based orphan detection for test cleanup

dweitzman:
Replace command-line flag approach with environment variables for more robust postgres process cleanup in integration tests.

This helps with supporting tests that do `multigres cluster start` because the environment variables propagate across multiple levels of subprocesses. The process ID monitored for subprocesses to kill themselves is now passed explicitly instead of assuming it's the direct parent process.

Changes:
- Remove `--test-orphan-detection` flag from servenv
- Add `MULTIGRES_TEST_PARENT_PID` baseline protection in TestMain functions
- Use `MULTIGRES_TESTDATA_DIR` for directory-deletion triggered cleanup. Ideally if tests use temporary directories, as soon as they delete those it'll trigger cleanup soon without waiting for the whole test run to finish
- Update shell scripts (run_in_test.sh, run_command_if_parent_dies.sh) to monitor both environment variables
- Environment variables automatically inherit to subprocesses
- Also cleanup after non-multipooler endtoend tests
- Fix two test flakes

Benefits:
- Simpler: No flag propagation needed, environment variables inherit naturally
- Defense in depth: Baseline PID monitoring + optional directory monitoring

### DIFF
--- a/go/test/endtoend/multipooler/setup_test.go
+++ b/go/test/endtoend/multipooler/setup_test.go
@@ -1140,17 +1140,6 @@ func setupPoolerTest(t *testing.T, setup *MultipoolerTestSetup, opts ...cleanupO
 
 	// Register cleanup handler
 	t.Cleanup(func() {
-		// if t.Failed() {
-		// 	if setup.PrimaryPgctld != nil {
-
-		// 		logFile := filepath.Join(setup.PrimaryPgctld.DataDir, "primary-multipooler", "multipooler.log")
-		// 		contentBytes, err := os.ReadFile(setup.PrimaryPgctld.DataDir)
-		// 		if err != nil {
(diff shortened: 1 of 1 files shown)


go/test/endtoend/multipooler/setup_test.go
```diff
@@ -1113,6 +1140,17 @@ func setupPoolerTest(t *testing.T, setup *MultipoolerTestSetup, opts ...cleanupO
 
 	// Register cleanup handler
 	t.Cleanup(func() {
+		// if t.Failed() {
```
dweitzman: Need to remove this before merging

dweitzman: Done

PR #212, opened 2025-11-05
feat(consensus): in-memory state management backed by filesystem

deepthi:
This PR replaces #192.

### Consensus State Management
- We maintain ConsensusState in memory, but read/write to disk as needed instead of going to disk for every read.
- Saving to disk is "pessimistic", i.e. we write to disk before we update in-memory state. In-memory state does not change if the disk write fails for any reason. In @sougou's words: "if it didn't get written to disk it didn't happen".
- Given that the consensus service implemented in #188 is optional, state initialization is conditional on the service being enabled.
- If state initialization fails, the poolerManager never goes into ready state. It goes into ERROR state instead.
- The exported functions GetTerm/SetTerm have been made private so that they are used only by the state/file management code.

Some questions:
- Is it worth retrying disk writes (once) for transient errors vs permanent errors?

### Term vs Term Number
- Some of the fields and functions have been renamed to make this distinction clearer. Term refers to the whole ConsensusTerm struct, whereas term number is just one field in that struct.

### Bugs fixed
- BeginTerm should take the action lock.
- BeginTermRequest should pass a full ID for the candidate, not just the name.

### DIFF
--- a/go/multipooler/manager/rpc_consensus.go
+++ b/go/multipooler/manager/rpc_consensus.go
@@ -134,7 +134,7 @@ func (pm *MultiPoolerManager) ConsensusStatus(ctx context.Context, req *consensu
 		`, []byte(req.ShardId)).Scan(&leaderTerm)
 
 		if err != nil {
-			pm.logger.Warn("Failed to query postgres leader term", "error", err)
+			pm.logger.WarnContext(ctx, "Failed to query postgres leader term", "error", err)
 			leaderTerm = 0
 		} else {
 			isHealthy = true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,6 +23,7 @@ linters:
     - errcheck
     - govet
     - ineffassign
+    - sloglint
     - staticcheck
   settings:
     depguard:
@@ -56,6 +57,8 @@ linters:
           deny:
             - pkg: "math/rand$"
               desc: "Please use math/rand/v2"
+    sloglint:
+      context: scope
     govet:
     disable:
       - nilness
--- a/go/admin/server/server.go
+++ b/go/admin/server/server.go
@@ -59,7 +59,7 @@ func (s *MultiAdminServer) RegisterWithGRPCServer(grpcServer *grpc.Server) {
 
 // GetCell retrieves information about a specific cell
 func (s *MultiAdminServer) GetCell(ctx context.Context, req *multiadminpb.GetCellRequest) (*multiadminpb.GetCellResponse, error) {
-	s.logger.Debug("GetCell request received", "cell_name", req.Name)
+	s.logger.DebugContext(ctx, "GetCell request received", "cell_name", req.Name)
 
 	// Validate request
 	if req.Name == "" {
@@ -69,7 +69,7 @@ func (s *MultiAdminServer) GetCell(ctx context.Context, req *multiadminpb.GetCel
 	// Get cell from topology
 	cell, err := s.ts.GetCell(ctx, req.Name)
 	if err != nil {
-		s.logger.Error("Failed to get cell from topology", "cell_name", req.Name, "error", err)
+		s.logger.ErrorContext(ctx, "Failed to get cell from topology", "cell_name", req.Name, "error", err)
 
 		// Check if it's a not found error
 		if errors.Is(err, &topo.TopoError{Code: topo.NoNode}) {
@@ -84,13 +84,13 @@ func (s *MultiAdminServer) GetCell(ctx context.Context, req *multiadminpb.GetCel
 		Cell: cell,
 	}
 
-	s.logger.Debug("GetCell request completed successfully", "cell_name", req.Name)
+	s.logger.DebugContext(ctx, "GetCell request completed successfully", "cell_name", req.Name)
 	return response, nil
 }
 
 // GetDatabase retrieves information about a specific database
 func (s *MultiAdminServer) GetDatabase(ctx context.Context, req *multiadminpb.GetDatabaseRequest) (*multiadminpb.GetDatabaseResponse, error) {
-	s.logger.Debug("GetDatabase request received", "database_name", req.Name)
+	s.logger.DebugContext(ctx, "GetDatabase request received", "database_name", req.Name)
 
 	// Validate request
 	if req.Name == "" {
@@ -100,7 +100,7 @@ func (s *MultiAdminServer) GetDatabase(ctx context.Context, req *multiadminpb.Ge
 	// Get database from topology
 	database, err := s.ts.GetDatabase(ctx, req.Name)
 	if err != nil {
-		s.logger.Error("Failed to get database from topology", "database_name", req.Name, "error", err)
+		s.logger.ErrorContext(ctx, "Failed to get database from topology", "database_name", req.Name, "error", err)
 
 		// Check if it's a not found error
 		if errors.Is(err, &topo.TopoError{Code: topo.NoNode}) {
@@ -115,17 +115,17 @@ func (s *MultiAdminServer) GetDatabase(ctx context.Context, req *multiadminpb.Ge
 		Database: database,
 	}
 
-	s.logger.Debug("GetDatabase request completed successfully", "database_name", req.Name)
+	s.logger.DebugContext(ctx, "GetDatabase request completed successfully", "database_name", req.Name)
 	return response, nil
 }
 
 // GetCellNames retrieves all cell names in the cluster
 func (s *MultiAdminServer) GetCellNames(ctx context.Context, req *multiadminpb.GetCellNamesRequest) (*multiadminpb.GetCellNamesResponse, error) {
-	s.logger.Debug("GetCellNames request received")
+	s.logger.DebugContext(ctx, "GetCellNames request received")
 
 	names, err := s.ts.GetCellNames(ctx)
 	if err != nil {
-		s.logger.Error("Failed to get cell names from topology", "error", err)
+		s.logger.ErrorContext(ctx, "Failed to get cell names from topology", "error", err)
 		return nil, status.Errorf(codes.Internal, "failed to retrieve cell names: %v", err)
 	}
 
@@ -133,17 +133,17 @@ func (s *MultiAdminServer) GetCellNames(ctx context.Context, req *multiadminpb.G
 		Names: names,
 	}
 
-	s.logger.Debug("GetCellNames request completed successfully", "count", len(names))
+	s.logger.DebugContext(ctx, "GetCellNames request completed successfully", "count", len(names))
 	return response, nil
 }
 
 // GetDatabaseNames retrieves all database names in the cluster
 func (s *MultiAdminServer) GetDatabaseNames(ctx context.Context, req *multiadminpb.GetDatabaseNamesRequest) (*multiadminpb.GetDatabaseNamesResponse, error) {
-	s.logger.Debug("GetDatabaseNames request received")
+	s.logger.DebugContext(ctx, "GetDatabaseNames request received")
 
 	names, err := s.ts.GetDatabaseNames(ctx)
 	if err != nil {
-		s.logger.Error("Failed to get database names from topology", "error", err)
+		s.logger.ErrorContext(ctx, "Failed to get database names from topology", "error", err)
 		return nil, status.Errorf(codes.Internal, "failed to retrieve database names: %v", err)
 	}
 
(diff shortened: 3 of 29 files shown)


go/multipooler/manager/term_storage.go
```diff
@@ -68,8 +68,8 @@ func GetTerm(poolerDir string) (*multipoolermanagerdatapb.ConsensusTerm, error)
 	return term, nil
 }
 
-// SetTerm sets the consensus term information to disk
-func SetTerm(poolerDir string, term *multipoolermanagerdatapb.ConsensusTerm) error {
+// setConsensusTerm saves the consensus term information to disk
+func setConsensusTerm(poolerDir string, term *multipoolermanagerdatapb.ConsensusTerm) error {
```
Neurostep, line 72: [q]: Just a side question, not related to this PR: does `setConsensusTerm` have to be durable?

I was curious about the content of this function and noticed that the [os.WriteFile](https://github.com/multigres/multigres/blob/8a144b19a9ff4c7332c822f6aae85da8cded0519/go/multipooler/manager/term_storage.go#L97) is used. For durablity, shouldn't we call `fsync` to make sure the filesystem flushes buffers to disc?

Also, shouldn't we remove temp file when the `WriteFile` fails? On a next call the file will be truncated, but could be a good idea to cleanup potentially corrupted file.

Not sure if the `os.Rename` is atomic enough and does not need a `consensusDir` fsync as well 🤔

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	// Return a copy to prevent external modifications
+	return cloneTerm(cs.term)
+}
+
+// PrepareAcceptance creates a new term with the acceptance recorded.
+// This does NOT modify in-memory state - the caller must call SaveAndUpdate.
+// Returns an error if already accepted a different candidate in this term.
+func (cs *ConsensusState) PrepareAcceptance(candidateID string) (*multipoolermanagerdatapb.ConsensusTerm, error) {
```
rafael: Could we add some comments on when we will use `PrepareAcceptance` vs `PrepareTermUpdate`. In this context it's not obvious why we need this nuance.

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,208 @@
...
+func (cs *ConsensusState) PrepareTermUpdate(newTerm int64) (*multipoolermanagerdatapb.ConsensusTerm, error) {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+
+	currentTerm := int64(0)
+	if cs.term != nil {
+		currentTerm = cs.term.GetTermNumber()
+	}
+
+	if newTerm < currentTerm {
```
rafael, line 158: I think we also need to check if we've already accepted a leader here.

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,208 @@
...
+	return nil
+}
+
+// cloneTerm creates a deep copy of a ConsensusTerm
+func cloneTerm(term *multipoolermanagerdatapb.ConsensusTerm) *multipoolermanagerdatapb.ConsensusTerm {
+	if term == nil {
+		return nil
+	}
+
+	clone := &multipoolermanagerdatapb.ConsensusTerm{
```
rafael: Could we use `proto.Clone` here?  I think that will be more future proof

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,208 @@
...
+// Returns nil if state has not been loaded.
+func (cs *ConsensusState) GetTerm() *multipoolermanagerdatapb.ConsensusTerm {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+
+	if cs.term == nil {
+		return nil
+	}
+
+	// Return a copy to prevent external modifications
```
rafael, line 100: nice

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	if newTerm < currentTerm {
+		return nil, fmt.Errorf("cannot update to older term: current=%d, new=%d", currentTerm, newTerm)
+	}
+
+	// Create new term with reset acceptance
+	term := &multipoolermanagerdatapb.ConsensusTerm{
+		TermNumber:         newTerm,
+		AcceptedLeader:     nil,
```
rafael, line 170: This is something I did not catch in the previous review. But after re-reading this PR in more depth. I think `AcceptedLeader` is a bit confusing here. This field is only used during the process of starting a new term. It is not really a leader yet. 

The reason why then is confusing, is that it is ok for this to be nil (like in this case) when we are outside the context of starting a new term. But having AcceptedLeader being nil, when there is an actual leader seems odd. 

What do you think?

go/multipooler/manager/consensus_state.go
deepthi: In the context of BeginTerm, there are two distinct phases. The first one is revocation. Updating the term with nil for AcceptedLeader and the new term_number persists the revocation decision. The next step is accepting the candidate as the new leader. That gets persisted separately.

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,208 @@
...
+
+	return term, nil
+}
+
+// SaveAndUpdate atomically saves the term to disk, then updates memory ONLY on success.
+// This is the key method that ensures memory never diverges from disk.
+// If the save fails, memory remains unchanged and the error is returned.
+func (cs *ConsensusState) SaveAndUpdate(newTerm *multipoolermanagerdatapb.ConsensusTerm) error {
+	// Save to disk FIRST (outside of lock to allow concurrent reads)
+	if err := setConsensusTerm(cs.poolerDir, newTerm); err != nil {
```
dweitzman, line 195: Does this method also assume that the caller is responsible for holding a lock during both the prepare and the save? If so, that feels worth mentioning explicitly but also feels potentially subtle or easy to get wrong in the caller.

Having consensus_state.go hold an exclusive lock between verifying that update is valid and saving it to disk feels like a simple way to stay safe that's easy to feel confident about. The downside is that we'd hold a lock during a disk write, but short of implementing some kind of on-disk representation of prepared commits with crash recovery maybe that's unavoidable?

If the prepare and save methods are separate, one option for an exclusive lock could be something like this:

```go
// Once we prepare a term update, all other attempts to update a term will be blocked until we either
// succeed or fail at the term update
preparedTerm, cancel, err := cs.PrepareTermUpdate(...)
defer cancel() // Releases the lock if we decide not to commit

if err != nil {
    return ...
}

err = cs.CommitTermUpdate(preparedTerm) // Releases the lock
// ...
```

go/multipooler/manager/term_storage.go
deepthi: This is intended to be durable but you are correct that it is possible that an fsync has not occurred before a crash. 
The eventual consensus implementation will converge correctly even if the file ~gets corrupted~ did not get persisted on a server, but it is probably still worthwhile hardening this.
Thank you for raising it. We'll do a follow up PR to fix it up.

go/multipooler/manager/consensus_state.go
rafael: Right, I meant add comment to the code so future readers can have that context

go/multipooler/manager/rpc_consensus.go
```diff
@@ -55,20 +60,19 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
 	}
 
-	// Validate and update term if needed (handles req.Term >= currentTerm)
-	// This will update the term to req.Term if req.Term > currentTerm, resetting AcceptedLeader to nil
-	// If req.Term == currentTerm, it does nothing (which is correct for idempotent acceptance)
-	if err := pm.validateAndUpdateTerm(ctx, req.Term, false); err != nil {
-		return nil, fmt.Errorf("failed to validate term: %w", err)
+	// Update term if needed (only if req.Term > currentTerm)
+	// This will reset AcceptedLeader to nil
+	if req.Term > currentTerm {
```
rafael, line 80: We can remove all this block right? Because now we are doing the Prepare and Save below

go/multipooler/manager/consensus_state.go
deepthi: `AcceptedLeader` is not just used while starting a new term. It is our persistent view of who the leader is for a given term. If we are given a new term without a leader specified (like what happens right now in Demote and SetPrimaryConnInfo), we cannot keep the old value for `AcceptedLeader`, we have to nil it out.
I think the real problem is that we have not implemented the full leader appointment flow yet. The answer might be that whenever we send a Demote or SetPrimaryConnInfo RPC to a replica/standby, we can't just send a term number, we also need to send the accepted_leader value for that term.
I'm not ruling out the possibility that we may need to introduce another field into this - AcceptedLeader vs ConfirmedLeader or something of that sort. But for now, it is correct to say that if we are informed of a term_number change without being told who the new leader is, we have to nil out the `AcceptedLeader` field.

go/multipooler/manager/consensus_state.go
deepthi: I debated whether to hold the lock while writing to disk, and ended up with
```
	// Save to disk FIRST (outside of lock to allow concurrent reads)
```
But this may be one instance where it is actually worth using an RWMutex to allow concurrent reads.

dweitzman: The scenario I'm concerned about is interleaving between two writes to disk:

- Term is 3
- Orchestrator A does prepare at term 4. 4 > 3 so prepare succeeds
- Orchestrator B does prepare at term 5. 5 > 3 so prepare succeeds
- Orch B commits (5 saved to disk) and returns ACK at 5 to an RPC
- Orch A commits (4 saved to disk) <- !!! term number decreases 5 -> 4 after commit !!!

That scenario isn't competing with in-memory reads, so it feels possible to guard against it with a disk-writing lock that doesn't impact reads?

go/multipooler/manager/consensus_state.go
deepthi: I think there's a bug here for the == condition. We should not clear out accepted_leader if PrepareTermUpdate is called with newTerm == currentTerm.

go/multipooler/manager/rpc_consensus.go
deepthi: We could combine them as an optimization. I tried to keep them separate so that Revocation and Acceptance (aka voting) are kept logically separate.
We discussed this offline, but there's still a gap here, because we need to also revoke at the postgres level by clearing out the primaryConnInfo on a standby (but only if this pooler is a standby). That is still TODO and planned for a future PR.

go/multipooler/manager/consensus_state.go
deepthi: In the current version of code, this bug never gets triggered because of higher level checks in callers. But still good to add safety.

go/multipooler/manager/consensus_state.go
deepthi: This is also correct in the sense that if this node accepted a new term and never voted for anyone in that term, then AcceptedLeader has to be nil.

go/multipooler/manager/consensus_state.go
deepthi: Hopefully the latest commit addresses this properly.

go/multipooler/manager/consensus_state.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// Return a copy to prevent external modifications
+	return cloneTerm(cs.term)
+}
+
+// AcceptCandidateAndSave atomically records acceptance of a candidate in the current term.
+// This is called when a node accepts a candidate during BeginTerm.
+// Returns error if already accepted a different candidate in this term.
+// Idempotent: succeeds if already accepted the same candidate.
+func (cs *ConsensusState) AcceptCandidateAndSave(candidateID *clustermetadatapb.ID) error {
```
dweitzman, line 108: Should this also take the expected term number for defense in depth? In theory whoever called this should be holding a lock and should know the term number, but if the expected current term was a parameter to this function you wouldn't have to look at the calling code to know that this isn't going to save the candidate to the wrong term

go/multipooler/manager/consensus_state.go
dweitzman: Just to record some conversation about this in writing: consensus state is not a sufficient place to defend against interleaving because we need to lock at a higher level

Suppose we're on term 3:
- Orchestrator A requests an update to term 4 with leader X
- Orchestrator B requests an update to term 5 with leader Y

Even if we get the ordering we want of term 3 -> 4 -> 5, there's another race condition with applying the new standby settings which we can't do atomically when we update the terms

We could get out-of-order updates to the follower settings and end up at term 5 but following the leader from term 4

We'll need to solve that with locking at a higher level, like the action lock, so I'm focusing on making sure we can enforce that the action lock is held when making consensus-related changes

go/multipooler/manager/consensus_state.go
rafael: Yeah,  I think this is a place where this kind of guardrails makes sense,

rafael (approved):
This is looking good from my perspective. I think there are some new comments from @dweitzman . I think we can address them in a follow up PR

dweitzman (commented):
Agreed, I don't have any blocking comments

PR #213, opened 2025-11-05
feat: Basic implementation of Extended query protocol

GuptaManan100:
### Description
This PR implements basic support for PostgreSQL's Extended Query Protocol in Multigres. The Extended Query Protocol is an alternative to the Simple Query Protocol that provides support for parameterized prepared statements and offers improved performance and flexibility.

**What is Extended Query Protocol?**
The Extended Query Protocol breaks down query execution into multiple steps:
- Parse - Parses the query string and creates a prepared statement
- Bind - Binds parameter values to the prepared statement, creating a portal
- Execute - Executes the portal (the bound prepared statement)
- Sync - Provides a synchronization point for error recovery

### Implementation Notes
This is a basic implementation that provides core Extended Query Protocol functionality. `Describe` doesn't work correctly because we aren't sending them to Postgres yet, and that means we have no way to spoof the data without extensive query planning support.

This implementation will need to be reworked after adding connection pooling. At that point we will make the changes proposed in https://github.com/multigres/multigres/blob/main/docs/query_serving/prepared_statements_design.md


go/multigateway/handler/connection_state.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// StorePreparedStatement stores a prepared statement.
+// If a statement with the same name already exists, it is replaced.
+func (cs *ConnectionState) StorePreparedStatement(stmt *PreparedStatement) {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+
+	cs.preparedStatements[stmt.Name] = stmt
+}
```
v0idpwn, line 126: fwiw, in Postgres it errors in parse if name != "" and it already exists. Only the empty statement is overriden

GuptaManan100: I will address this, when I rework the implementation as described in https://github.com/multigres/multigres/blob/main/docs/query_serving/prepared_statements_design.md

sougou (approved):
As discussed, I assume this is incomplete, and will be completed after the pooler part is done.

PR #214, opened 2025-11-05
fix(multipooler): fix replTracker writer after demotion & promotion

dweitzman:
The failures in `TestConsensus_GetLeadershipView` look like this: `rpc error: code = Unknown desc = failed to get leadership view: sql: database is closed` if you run it multiple times

The manager Close()s and recreates the DB during demotion to standby, but the replTracker still has a reference to that old (now-closed) DB pool

My hacky workaround is to avoid closing the DB pool and just gently encourage it to recycle all its connections instead. Once we implement our own custom pooling we can add a `Reconnect()` method, if we want


deepthi:
Can we close this now that this bug has been fixed in #227?

dweitzman:
The test issue should be fixed by https://github.com/multigres/multigres/pull/227

PR #216, opened 2025-11-05
feat(multipooler): enforce holding the action lock to get the current term

dweitzman:
This caught a bug: `BeginTerm` wasn't holding the lock

The approach here is to make the action lock a standalone struct that you can only acquire with a `Context`, and it returns a new context. We can verify that the lock is held (defense in depth) by checking the `Context` to protect against calls into critical code (consensus, etc.) without holding the action lock


go/multipooler/manager/manager.go
```diff
@@ -301,15 +285,21 @@ func (pm *MultiPoolerManager) checkPoolerType(expectedType clustermetadatapb.Poo
 	return nil
 }
 
-// getCurrentTerm returns the current consensus term in a thread-safe manner
-func (pm *MultiPoolerManager) getCurrentTerm() int64 {
+// getCurrentTerm returns the current consensus term in a thread-safe manner.
+// This method must only be called while holding the action lock.
+func (pm *MultiPoolerManager) getCurrentTerm(ctx context.Context) (int64, error) {
```
dweitzman: This is the new defense in depth

go/multipooler/manager/rpc_consensus.go
```diff
@@ -27,18 +27,29 @@ import (
 
 // BeginTerm handles coordinator requests during leader appointments
 func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatapb.BeginTermRequest) (*consensusdatapb.BeginTermResponse, error) {
+	// Acquire the action lock to ensure only one mutation runs at a time
```
dweitzman: This is the bug it caught

go/multipooler/manager/rpc_consensus.go
```diff
@@ -110,6 +119,13 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
 
 // ConsensusStatus returns the current status of this node for consensus
 func (pm *MultiPoolerManager) ConsensusStatus(ctx context.Context, req *consensusdatapb.StatusRequest) (*consensusdatapb.StatusResponse, error) {
+	// Acquire the action lock to read consensus state
+	ctx, err := pm.actionLock.Acquire(ctx, "ConsensusStatus")
```
dweitzman: Another potential bug caught. Might not matter in this case because an RPC to read consensus state can probably get away with reading stale data, but maybe we want to avoid reading inconsistent data where the consensus term and replication status are out of sync

deepthi: All of the read-only RPCs in this file currently do not take the action lock, the reasoning being that they are not "actions". It is debatable whether that can lead to subtle errors. On balance, if one multiorch is asking a pooler to BeginTerm and another one is querying for Status, we should probably serialize them, so I agree with adding this here.

deepthi (approved):

go/multipooler/manager/rpc_consensus.go
dweitzman: Maybe it'd make sense to add something like `GetInconsistentCurrentTermNumber()` or `GetCurrentTermNumberForMonitoring()` that doesn't require holding the action lock?

dweitzman: I introduced `GetInconsistentCurrentTermNumber()` as a way to retrieve the information without a lock for status monitoring. Any objections?

deepthi (approved):

PR #222, opened 2025-11-07
feat: Add automated AST Rewrite and Clone code generation

GuptaManan100:
### Description 

This PR adds automated code generation for AST tree traversal and deep cloning, eliminating manual maintenance for 200+ AST node types.

- **New tool**: `go/tools/asthelpergen/` - Code generator based on Vitess's approach
  - `asthelpergen.go` (368 lines) - Type discovery and generator orchestration
  - `clone_gen.go` (249 lines) - Deep clone method generator
  - `rewrite_gen.go` (410 lines) - Tree rewrite method generator

- **Generated code**: `go/parser/ast/`
  - `ast_clone.go` (4,787 lines) - Deep cloning for all AST types
  - `ast_rewrite.go` (12,190 lines) - Visitor pattern traversal with node replacement
  - `rewriter_api.go` (126 lines) - Public API: `Rewrite(node, pre, post ApplyFunc)`

- **Tests**: 14 comprehensive tests (all passing)
  - 5 rewrite tests: traversal, replacement, skipping, prepared statements
  - 9 clone tests: nil handling, deep copy, structure preservation

### Usage
  ```go
  // Clone an AST
  cloned := ast.CloneNode(stmt)

  // Rewrite an AST (replace all integers)
  modified := ast.Rewrite(stmt, func(c *ast.Cursor) bool {
      if intNode, ok := c.Node().(*ast.Integer); ok {
          c.Replace(ast.NewInteger(999))
      }
      return true
  }, nil)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,6 @@ tools:
 	ln -sf "$(MTROOT)/misc/git/commit-msg" .git/hooks/commit-msg
 	./tools/setup_build_tools.sh
 	go install golang.org/x/tools/cmd/goyacc@latest
-	go install golang.org/x/tools/cmd/goimports@latest
 
 # Generate protobuf files
 proto: tools $(PROTO_GO_OUTS)
--- a/go/tools/asthelpergen/internal/utils.go
+++ b/go/tools/asthelpergen/internal/utils.go
@@ -74,7 +74,7 @@ func GoImports(fullPath string) error {
 	}
 
 	// Run goimports
-	cmd = exec.Command("goimports", "-local", "github.com/supabase/multigres", "-w", fullPath)
+	cmd = exec.Command("go", "tool", "goimports", "-local", "github.com/supabase/multigres", "-w", fullPath)
 	cmd.Stderr = os.Stderr
 	if err := cmd.Run(); err != nil {
 		return err


sougou (approved):
Nice work! I see that the rewriter generates code. I'm wondering if there's something that can trigger a requirement to write new tests, or better: review a generated test. Otherwise, in the future, people may make changes and forget that tests need to be written.

Makefile
```diff
@@ -41,6 +41,7 @@ tools:
 	ln -sf "$(MTROOT)/misc/git/commit-msg" .git/hooks/commit-msg
 	./tools/setup_build_tools.sh
 	go install golang.org/x/tools/cmd/goyacc@latest
+	go install golang.org/x/tools/cmd/goimports@latest
```
dweitzman: FYI, elsewhere we're invoking as `go tool goimports` instead of installing `goimports` globally in the path

GuptaManan100: That's a good idea, I'll fix it

GuptaManan100:
@sougou Thank you! To address your concerns, we have protections in place to ensure that when new AST types are added, the code generator is rerun. However, I'm not convinced of the value of generated tests in this context.

If we generate both the code and the tests, what additional confidence do the tests actually provide? Generated tests tend to become boilerplate that simply verify the generated code does what the generator intended, without validating whether the functionality actually meets our real-world requirements.

I believe tests should verify the actual use cases and behaviors we care about, not just mechanically confirm the generator's output which is bound to be correct. That's why in this PR, I specifically wrote `TestRewritePreparedStatement` - it tests the actual functionality I intend to use the rewriter for, which is much more valuable than a generated test would be.

Makefile
GuptaManan100: Fixed! I removed this install 👍

PR #223, opened 2025-11-07
test(endtoend): wait for replication reset to finish in t.Cleanup()

dweitzman:
We've seen some flakes where setupPoolerTest() claims a previous test failed to reset replication state. I think it's actually because GUC changes can take a moment to propagate, so it's safest if the Cleanup() function confirms they've taken effect before returning

### DIFF
--- a/go/test/endtoend/multipooler/setup_test.go
+++ b/go/test/endtoend/multipooler/setup_test.go
@@ -890,15 +890,16 @@ func queryStringValue(ctx context.Context, client *endtoend.MultiPoolerTestClien
 	return string(resp.Rows[0].Values[0]), nil
 }
 
-// validateGUCValue queries a GUC and fails the test if it doesn't match the expected value.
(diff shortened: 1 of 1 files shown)


go/test/endtoend/multipooler/setup_test.go
```diff
@@ -1185,5 +1203,11 @@ func setupPoolerTest(t *testing.T, setup *MultipoolerTestSetup, opts ...cleanupO
 				}
 			}
 		}
+
+		// Step 4: Validate that cleanup fully applied and state is clean
+		// Use Eventually to give the system time to reach clean state
+		require.Eventually(t, func() bool {
```
dweitzman, line 1209: This is the essence of the change. Everything else is just making `validateCleanState()` reusable at both the beginning of setup and the end of cleanup

rafael (approved):
LGTM

PR #224, opened 2025-11-07
refactor(viperutil): eliminate global config state in favor of dependency injection

dweitzman:
Pass Registry instances through constructors instead of using global Static/Dynamic registries.

This prevents leaking state across tests, so several tests which previously failed when you run them multiple times (`go test -count=5 ./...`) now succeed after this PR. Running tests with count > 1 is helpful for investigating flakes

### DIFF
--- a/go/viperutil/debug/debug.go
+++ b/go/viperutil/debug/debug.go
@@ -23,15 +23,15 @@ import (
 // Debug provides the Debug functionality normally accessible to a given viper
 // instance, but for a combination of the static and dynamic registries.
 func Debug(reg *viperutil.Registry) {
-	viperutil.Combined(reg).Debug()
+	reg.Combined().Debug()
 }
 
 // WriteConfigAs writes the config into the given filename.
 func WriteConfigAs(reg *viperutil.Registry, filename string) error {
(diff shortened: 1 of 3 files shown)


go/viperutil/internal/registry/registry.go
```diff
@@ -24,16 +24,6 @@ import (
 )
 
 var (
-	// Static is the registry for static config variables. These variables will
-	// never be affected by a Watch-ed config, and maintain their original
-	// values for the lifetime of the process.
-	Static = viper.New()
```
dweitzman, line 30: These are the deleted global variables

go/viperutil/internal/registry/registry.go
```diff
@@ -47,13 +37,3 @@ type Bindable interface {
 	RegisterAlias(alias string, key string)
 	SetDefault(key string, value any)
 }
-
-// Combined returns a viper combining the Static and Dynamic registries.
```
dweitzman, line 51: This moved into the non-internal registry.go because it now depends on the Registry type

GuptaManan100 (approved):
This is great! Thank you! I wanted to fix this too, but I hadn't spent the time to see how!

PR #225, opened 2025-11-10
feat(heartbeat): sync leader_term from database when starting writer

deepthi:
While reviewing #188, we realized that the heartbeat writer needs to notified of a new `leader_term` (term number). However, that should only happen after a new leader appointment is complete.
This PR removes the code that implements that notification in the wrong place. We will reimplement it when we write the full leader appointment code.
In addition, we also need to initialize the `leader_term` from the underlying database when a multipooler restarts in place. That feature has been implemented in this PR with some guardrails.


deepthi:
@Rafael brought up a good point. Why do we even need a `leader_term` column in the database when we are already storing the term number on disk? I'm going to close this one and open another PR that removes the `leader_term` column entirely.

PR #227, opened 2025-11-10
refactor(multipooler): change init to improve multipooler lifecycle management

rafael:
## Description
Refactors multipooler initialization and lifecycle management such that the manager controls the multipooler. Before this change, both the Multipooler and the Manager had a reference to the manager config which was confusing. 

The way this was implemented was similar to how it is done in Vitess. This is a step in hardening this initialization, but there is more hardening required. We can do this in other PR's as we continue to make progress. Calling some notable next steps in the TODO

  ### Key Changes

  #### 1. **Introduced PoolerController interface**
  - Created `controller.go` defining the `PoolerController` interface
  - Separates concerns: manager handles orchestration, controller handles query serving
  - Follows Vitess `TabletManager` → `TabletServer` pattern
  
  #### 2. **Database connection ownership**
  - Moved DB connection from MultiPooler to Executor
  - Executor creates and owns its connection using provided config
  
  ## TODO in Follow up PR's
- We need a more robust state manager. Something like tm_state.go in Vitess. That way we can handle transition between different states in a proper way. 
- The executor owns it's db, but talks to the DB directly. This should be replaced with the query pooler.
- The manager also has a db object for the administrative operations (e.g the ones we call from rpc_manager). We should decide how are we going to handle this kind of connections.

### DIFF
--- a/go/multipooler/manager/manager.go
+++ b/go/multipooler/manager/manager.go
@@ -298,6 +298,7 @@ func (pm *MultiPoolerManager) Close() error {
 	// Close resources (safe to call even if nil/never opened)
 	if pm.replTracker != nil {
 		pm.replTracker.Close()
+		pm.replTracker = nil
 	}
 
 	if pm.db != nil {


go/test/endtoend/multipooler/rpc_manager_promotion_demotion_test.go
```diff
@@ -275,6 +275,8 @@ func TestDemoteAndPromote(t *testing.T) {
 	})
 
 	t.Run("Idempotency_Promote", func(t *testing.T) {
+		setupPoolerTest(t, setup)
```
rafael, line 278: These seemed like a few places where we missed this.

deepthi:
> The manager also has a db object for the administrative operations (e.g the ones we call from rpc_manager). We should decide how are we going to handle this kind of connections.

A separate DBA pool for the manager.

go/multipooler/poolerserver/pooler.go
```diff
@@ -17,119 +17,180 @@ package poolerserver
...
 
 	"github.com/multigres/multigres/go/multipooler/executor"
-	"github.com/multigres/multigres/go/multipooler/manager"
 	"github.com/multigres/multigres/go/multipooler/queryservice"
-	"github.com/multigres/multigres/go/servenv"
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 )
 
 // MultiPooler is the core pooler implementation
+// It implements the PoolerController interface
```
GuptaManan100: We should probably expland on this comment, to write out more explicitly what the purpose of this is. And also we should probably rename the struct. MultiPooler doesn't seem like a very descriptive name

go/multipooler/init.go
```diff
@@ -227,19 +225,6 @@ func (mp *MultiPooler) Init() {
 	mp.senv.HTTPHandleFunc("/", mp.handleIndex)
 	mp.senv.HTTPHandleFunc("/ready", mp.handleReady)
 
-	// Initialize and start the MultiPooler
```
rafael, line 230: In vitess, we create the tablet server and pass it to the manager. I couldn't think of a good reason why not just create it  inside the manager. So I decided to fully push it down

go/multipooler/manager/manager.go
```diff
@@ -233,15 +277,41 @@ func (pm *MultiPoolerManager) startHeartbeat(ctx context.Context, shardID []byte
...
+
+	// Always cancel context to stop async loaders
 	pm.cancel()
+
+	// Always set isOpen to false
+	pm.isOpen = false
+
+	// Close resources (safe to call even if nil/never opened)
 	if pm.replTracker != nil {
 		pm.replTracker.Close()
```
dweitzman: Non-blocking comment: should replTrack be set to `nil` after Close()ing?

go/multipooler/poolerserver/controller.go
```diff
@@ -0,0 +1,107 @@
...
+// This follows the Vitess Controller pattern (see vitess/go/vt/vttablet/tabletserver/controller.go)
+//
+// The controller is responsible for:
+// - Managing query serving state (SERVING, NOT_SERVING, SERVING_RDONLY, DRAINED)
+// - Handling query execution through the executor
+// - Providing health status
+//
+// The MultiPoolerManager creates and controls the lifecycle of the PoolerController,
+// similar to how TabletManager controls TabletServer in Vitess.
+type PoolerController interface {
```
deepthi, line 44: Controller doesn't seem obvious as a name. @GuptaManan100 any suggestions?

deepthi:
Merging this now, but we will need to revisit the naming of all the components of MultiPooler.

PR #228, opened 2025-11-11
feat(observability): initial support for opentelemetry

dweitzman:
This PR is setting up a minimal amount of boilerplate to make exporting metrics and traces possible, and annotating structured logs with trace IDs possible

`autoexport` is used so metric and tracing export can be enabled with standard otel env variables (https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/). However... by default autoexport tries to push data even if no environment variables are configured, so I've changed the default tracing and metrics exporters to "none" for now

There's also a simple integration-y test that starts up a servenv and verifies that HTTP and gRPC requests can be traced across the network request with client-side and server-side spans and metrics. The test also verifies that exemplars can be recorded into metrics with trace IDs that match the corresponding span

There's more future work to be done to make using opentelemetry easy, flexible, or useful, such as:
- Improving instrumentation (propagating context to preserve trace IDs, adding new metrics, adding new spans etc.)
- Setting up prometheus export at /metrics on a servenv HTTP mux so there's an easy place to look at metrics during development
- Adding scripts or docker compose to make it easier to bring up jaeger, prometheus, and grafana for local testing
- Provide some mechanism so that "multigres cluster start" can configure trace sampling differently for itself than for the subprocesses it starts -- normally they'd inherit the same env variables

### DIFF
--- a/go/servenv/telemetry.go
+++ b/go/servenv/telemetry.go
@@ -61,18 +61,8 @@ func (t *Telemetry) WithTestExporters(spanExporter sdktrace.SpanExporter, metric
 }
 
 // InitTelemetry initializes OpenTelemetry providers and exporters
-// This should be called early in the service lifecycle, typically in OnInit or OnRun hooks
-// Configuration is done via standard OpenTelemetry environment variables:
-// - OTEL_SERVICE_NAME: Service name (defaults to defaultServiceName)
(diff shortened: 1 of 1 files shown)


go/servenv/telemetry_test.go
dweitzman, line 1: I haven't added tests for pgctld or the multigres command, but I'm open to it

PR #229, opened 2025-11-11
fix(test): resolve TestConsensus_GetLeadershipView flake due to shared state

dweitzman:
The test was failing intermittently because it expected consensus term to be exactly 1, but TestConsensus_BeginTerm (which runs first alphabetically) increments the term to 2.

Since tests use getSharedTestSetup() which creates a singleton cluster via sync.Once, the term mutation persists across tests.

Changed assertions from Equal(1) to GreaterOrEqual(1) since the actual term value doesn't matter for validating GetLeadershipView functionality.

Example failure: https://github.com/multigres/multigres/actions/runs/19256401238?pr=228


dweitzman:
This is obsolete after https://github.com/multigres/multigres/pull/231

PR #233, opened 2025-11-12
feat(telemetry): support tracing local provisioning

dweitzman:
### Key changes

- Move `Telemetry` under `go/tools` instead of `go/servenv` so the local provisioner can depend on it. This feels reasonable anyway, but it's necessary because we set up a dependency test to prevent the local provisioner from depending on multigres code except for shared libraries
- Add `SetCmdEnvTraceContext()` and `WithEnvTraceparent()` to the `telemetry` package for propagating context across process boundaries by reading `TRACEPARENT` from an environment variable or writing it into the env variable parameters for a command to a subprocess
- Update root CLI commands to run telemetry collection and call `WithEnvTraceparent()` as appropriate to record the parent span into the command's context. The subtlety here is that some commands are one-offs like `multigres cluster start` where the command finishes quickly and everything it does should be considered part of the parent span. For long-running servers like `multipooler`, we should probably only inherit the traceparent into the "initialization" part of the command so that `multigres cluster start` gives visibility into the startup process for multipooler and pgctld but not the entire lifetime of the server

To keep this PR minimal and easier to review, it doesn't fix code that isn't propagating `Context` and doesn't add any spans. I'll make a follow-up PR that'll make traces of local provisioning with `multigres cluster start` much more complete and useful

### How to test

Start jaeger to collect trace and provide a UI for browsing them. It'll look something like this:
```bash
$ docker run -d --name jaeger-all-in-one \
  -e COLLECTOR_OTLP_ENABLED=true \
  -e COLLECTOR_OTLP_HTTP_PORT=4138 \
  -p 16686:16686 \
  -p 4138:4138 \
  jaegertracing/all-in-one:latest
```

Start a cluster with tracing enabled:
```bash
$ OTEL_EXPORTER_OTLP_ENDPOINT="http://localhost:4318" \
  OTEL_TRACES_SAMPLER=always_on \
  OTEL_TRACES_EXPORTER=otlp \
     ./bin/multigres cluster start --config-path multigres_local
```

Visit http://localhost:16686/ to view traces. The cluster start will show up as the `multigres-cli` service

The trace will look much better after my follow-up PR to preserve parent context and add some spans for slow actions that aren't blocking on an HTTP request

### DIFF
--- a/go/cmd/multigres/command/root.go
+++ b/go/cmd/multigres/command/root.go
@@ -82,7 +82,7 @@ Configuration:
 				return err
 			}
 
-			if span, err = mc.telemetry.InitForCommand(cmd, "multigres-cli", true); err != nil {
+			if span, err = mc.telemetry.InitForCommand(cmd, "multigres-cli", true /* startSpan */); err != nil {
 				return fmt.Errorf("failed to initialize OpenTelemetry: %w", err)
 			}
 
--- a/go/cmd/multipooler/main.go
+++ b/go/cmd/multipooler/main.go
@@ -46,7 +46,7 @@ func CreateMultiPoolerCommand() (*cobra.Command, *multipooler.MultiPooler) {
 			return run(cmd, args, mp)
 		},
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			if _, err := telemetry.InitForCommand(cmd, "multipooler", false); err != nil {
+			if _, err := telemetry.InitForCommand(cmd, "multipooler", false /* startSpan */); err != nil {
 				return fmt.Errorf("failed to initialize OpenTelemetry: %w", err)
 			}
 
--- a/go/cmd/pgctld/command/root.go
+++ b/go/cmd/pgctld/command/root.go
@@ -98,7 +98,7 @@ management for PostgreSQL servers.`,
 			pc.lg.SetupLogging()
 			// Initialize telemetry for CLI commands (server command will re-initialize via ServEnv.Init)
 			var err error
-			if span, err = pc.telemetry.InitForCommand(cmd, "pgctld", cmd.Use != "server"); err != nil {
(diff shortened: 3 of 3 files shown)


go/tools/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+func TestNewTelemetry(t *testing.T) {
+	tel := NewTelemetry()
+	require.NotNil(t, tel)
+	assert.False(t, tel.initialized)
+	assert.Nil(t, tel.tracerProvider)
+	assert.Nil(t, tel.meterProvider)
+}
+
+func TestInitTelemetry_DefaultServiceName(t *testing.T) {
```
dweitzman, line 36: These tests are moved from go/servenv/telemetry_test.go. The test cases themselves are unchanged

go/tools/telemetry/telemetry_test_helpers.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// testTelemetrySetup holds test telemetry infrastructure
+type testTelemetrySetup struct {
+	Telemetry    *Telemetry
+	SpanExporter *tracetest.InMemoryExporter
+	MetricReader *metric.ManualReader
+}
+
+func (t *testTelemetrySetup) ForceFlush(ctx context.Context) error {
```
dweitzman, line 35: Pulled these helpers out and exported them for reuse by the integration-y test at go/servenv/telemetry_test.go

deepthi:
Should the "test instructions" themselves become part of the source in GH? Maybe it's time to create examples/. The other option is to put them into one of the source files - maybe local.go.

dweitzman:
> Should the "test instructions" themselves become part of the source in GH? Maybe it's time to create examples/. The other option is to put them into one of the source files - maybe local.go.

Yeah, I think we'll want some testing instructions and/or helper scripts but I'm going to say it's out of scope for this particular PR

PR #234, opened 2025-11-12
feat(multipooler): add shard initialization RPCs

deepthi:
Add three pooler RPCs for bootstrap initialization:
  - InitializationStatus: returns pooler initialization state (data dir, role, WAL position, etc.)
  - InitializeEmptyPrimary: initializes pooler as empty primary, creates multigres schema
  - InitializeAsStandby: initializes pooler as standby from primary backup. Note: the actual backup hasn't been implemented in this PR, we will get it from #226 

These RPCs will be called by MultiOrch coordinator during shard bootstrap, repair, and leader appointment workflows.

### DIFF
--- a/go/multipooler/manager/rpc_initialization.go
+++ b/go/multipooler/manager/rpc_initialization.go
@@ -20,6 +20,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/multigres/multigres/go/mterrors"
+	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
 	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
 )
@@ -33,56 +35,61 @@ func (pm *MultiPoolerManager) InitializeEmptyPrimary(ctx context.Context, req *m
 	var err error
 	ctx, err = pm.actionLock.Acquire(ctx, "InitializeEmptyPrimary")
 	if err != nil {
-		return nil, fmt.Errorf("failed to acquire action lock: %w", err)
+		return nil, mterrors.Wrap(err, "failed to acquire action lock")
 	}
 	defer pm.actionLock.Release(ctx)
 
-	// 1. Check if already initialized
+	// 1. Validate consensus term must be 1 for new primary
+	if req.ConsensusTerm != 1 {
+		return nil, mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT, "consensus term must be 1 for new primary initialization, got %d", req.ConsensusTerm)
+	}
+
+	// 2. Check if already initialized
 	if pm.isInitialized() {
 		pm.logger.InfoContext(ctx, "Pooler already initialized", "shard", pm.getShardID())
 		return &multipoolermanagerdatapb.InitializeEmptyPrimaryResponse{Success: true}, nil
 	}
 
-	// 2. Initialize data directory via pgctld if needed
+	// 3. Initialize data directory via pgctld if needed
 	if !pm.hasDataDirectory() {
 		pm.logger.InfoContext(ctx, "Initializing data directory", "shard", pm.getShardID())
 		if pm.pgctldClient == nil {
-			return nil, fmt.Errorf("pgctld client not available")
+			return nil, mterrors.New(mtrpcpb.Code_UNAVAILABLE, "pgctld client not available")
 		}
 
 		initReq := &pgctldpb.InitDataDirRequest{}
 		if _, err := pm.pgctldClient.InitDataDir(ctx, initReq); err != nil {
-			return nil, fmt.Errorf("failed to initialize data directory: %w", err)
+			return nil, mterrors.Wrap(err, "failed to initialize data directory")
 		}
 	}
 
-	// 3. Start PostgreSQL if not running
+	// 4. Start PostgreSQL if not running
 	if !pm.isPostgresRunning(ctx) {
 		pm.logger.InfoContext(ctx, "Starting PostgreSQL", "shard", pm.getShardID())
 		if pm.pgctldClient == nil {
-			return nil, fmt.Errorf("pgctld client not available")
+			return nil, mterrors.New(mtrpcpb.Code_UNAVAILABLE, "pgctld client not available")
 		}
 
 		startReq := &pgctldpb.StartRequest{}
 		if _, err := pm.pgctldClient.Start(ctx, startReq); err != nil {
-			return nil, fmt.Errorf("failed to start PostgreSQL: %w", err)
+			return nil, mterrors.Wrap(err, "failed to start PostgreSQL")
 		}
 	}
 
-	// 4. Wait for database connection
+	// 5. Wait for database connection
 	if err := pm.waitForDatabaseConnection(ctx); err != nil {
-		return nil, fmt.Errorf("failed to connect to database: %w", err)
+		return nil, mterrors.Wrap(err, "failed to connect to database")
 	}
 
-	// 5. Create multigres schema and heartbeat table
+	// 6. Create multigres schema and heartbeat table
 	if err := pm.initializeMultigresSchema(ctx); err != nil {
-		return nil, fmt.Errorf("failed to initialize multigres schema: %w", err)
+		return nil, mterrors.Wrap(err, "failed to initialize multigres schema")
 	}
 
-	// 6. Set consensus term
+	// 7. Set consensus term
 	if pm.consensusState != nil {
 		if err := pm.consensusState.UpdateTermAndSave(ctx, req.ConsensusTerm); err != nil {
-			return nil, fmt.Errorf("failed to set consensus term: %w", err)
+			return nil, mterrors.Wrap(err, "failed to set consensus term")
 		}
 	}
 
@@ -97,21 +104,25 @@ func (pm *MultiPoolerManager) InitializeAsStandby(ctx context.Context, req *mult
 		"shard", pm.getShardID(),
 		"primary", fmt.Sprintf("%s:%d", req.PrimaryHost, req.PrimaryPort),
 		"term", req.ConsensusTerm,
-		"force_reinit", req.ForceReinit)
+		"force", req.Force)
 
 	// Acquire action lock
 	var err error
 	ctx, err = pm.actionLock.Acquire(ctx, "InitializeAsStandby")
 	if err != nil {
-		return nil, fmt.Errorf("failed to acquire action lock: %w", err)
+		return nil, mterrors.Wrap(err, "failed to acquire action lock")
 	}
 	defer pm.actionLock.Release(ctx)
 
-	// 1. Remove data directory if force_reinit
-	if req.ForceReinit && pm.hasDataDirectory() {
+	// 1. Check for existing data directory
+	if pm.hasDataDirectory() {
+		if !req.Force {
+			return nil, mterrors.New(mtrpcpb.Code_FAILED_PRECONDITION, "data directory already exists, use force=true to reinitialize")
+		}
+		// Remove data directory if force
 		pm.logger.InfoContext(ctx, "Force reinit: removing data directory", "shard", pm.getShardID())
 		if err := pm.removeDataDirectory(); err != nil {
-			return nil, fmt.Errorf("failed to remove data directory: %w", err)
+			return nil, mterrors.Wrap(err, "failed to remove data directory")
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,383 @@
...
+
+	pm.logger.InfoContext(ctx, "Creating multigres schema and heartbeat table")
+
+	// Create schema
+	if _, err := pm.db.ExecContext(ctx, "CREATE SCHEMA IF NOT EXISTS multigres"); err != nil {
+		return fmt.Errorf("failed to create multigres schema: %w", err)
+	}
+
+	// Create heartbeat table
+	createTableSQL := `
```
deepthi, line 348: @cuongdo I think it's fine that we create this table here if needed, and also in MultiPoolerManager.Open. WDYT?

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,383 @@
...
+	"path/filepath"
+
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
+)
+
+// InitializeEmptyPrimary initializes this pooler as an empty primary
+// Used during bootstrap initialization of a new shard
+func (pm *MultiPoolerManager) InitializeEmptyPrimary(ctx context.Context, req *multipoolermanagerdatapb.InitializeEmptyPrimaryRequest) (*multipoolermanagerdatapb.InitializeEmptyPrimaryResponse, error) {
+	pm.logger.InfoContext(ctx, "InitializeEmptyPrimary called", "shard", pm.getShardID(), "term", req.ConsensusTerm)
```
deepthi, line 32: We are calling this Initialize_Empty_Primary but getting a term_number passed in. Is there any reasonable situation where we call this, but term_number is not exactly 1? cc @rafael

go/multipooler/manager/rpc_initialization.go
cuongdo: That sounds fine. This is idempotent, and it seems like the only requirement is that the heartbeat table exists before standbys start writing heartbeats.

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,390 @@
...
+	// 3. TODO: Restore from primary using pgBackRest (PR #226)
+	// For now, we'll use pg_basebackup as a placeholder
+	pm.logger.InfoContext(ctx, "Performing backup from primary", "primary", fmt.Sprintf("%s:%d", req.PrimaryHost, req.PrimaryPort))
+
+	// Placeholder: In production, this would call pm.Restore() from PR #226
+	// For now, we skip the actual backup to avoid dependencies
+	finalLSN := ""
+
+	// 4. Create standby.signal
+	if err := pm.createStandbySignal(); err != nil {
```
rafael: I added an option to Restart that allows to pass Standby parameter to pgctld:
```
	// Call pgctld to restart as standby
	// This will create standby.signal and restart the server
	req := &pgctldpb.RestartRequest{
		Mode:      "fast",
		Timeout:   nil, // Use default timeout
		Port:      0,   // Use default port
		ExtraArgs: nil,
		AsStandby: true, // Create standby.signal before restart
	}

	// Close the query service controller to release its stale database connection
	if err := pm.Close(); err != nil {
		pm.logger.WarnContext(ctx, "Failed to close query service controller after restart", "error", err)
		// Continue - we'll try to reconnect anyway
	}

	resp, err := pm.pgctldClient.Restart(ctx, req)
	if err != nil {
		pm.logger.ErrorContext(ctx, "Failed to restart PostgreSQL as standby", "error", err)
		return mterrors.Wrap(err, "failed to restart as standby")
	}
```
That handles creating the signal. I think we can do the same here, instead of setting this manually.

go/multipooler/manager/rpc_initialization.go
rafael: Nothing that comes to mind. I think it should always be 1

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,390 @@
...
+			return nil, fmt.Errorf("pgctld client not available")
+		}
+
+		stopReq := &pgctldpb.StopRequest{}
+		if _, err := pm.pgctldClient.Stop(ctx, stopReq); err != nil {
+			return nil, fmt.Errorf("failed to stop PostgreSQL: %w", err)
+		}
+	}
+
+	// 2. Remove data directory if force_reinit
```
rafael: If force is not passed and there is data,  it seems like this should error

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	// 7. Wait for database connection
+	if err := pm.waitForDatabaseConnection(ctx); err != nil {
+		return nil, fmt.Errorf("failed to connect to database: %w", err)
+	}
+
+	// 8. Set consensus term
+	if pm.consensusState != nil {
+		if err := pm.consensusState.UpdateTermAndSave(ctx, req.ConsensusTerm); err != nil {
```
rafael, line 163: I think for the primary, it should always be `1`. For replicas, we use what is provided in the request.

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,390 @@
...
+		return false
+	}
+
+	if pm.db == nil {
+		return false
+	}
+
+	// Check if multigres schema exists
+	var exists bool
+	query := "SELECT EXISTS(SELECT 1 FROM information_schema.schemata WHERE schema_name = 'multigres')"
```
rafael: I've been adding all queries to `pg_replication.go` , so we don't have those inline in files. This one doesn't belong there, but maybe that all queries should have their function and we don't have raw queries in this layer.  I don't feel strongly about this, but sharing the idea

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,390 @@
...
+	statusResp, err := pm.pgctldClient.Status(ctx, statusReq)
+	if err != nil {
+		return false
+	}
+
+	return statusResp.Status == pgctldpb.ServerStatus_RUNNING
+}
+
+// getRole returns the current role of this pooler ("primary", "standby", or "unknown")
+func (pm *MultiPoolerManager) getRole(ctx context.Context) string {
```
rafael, line 251: This already exists as a function `isPrimary` in `pg_replication` . I think we can use that function and then derive the role from there.

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	if inRecovery {
+		return "standby"
+	}
+	return "primary"
+}
+
+// getWALPosition returns the current WAL position
+func (pm *MultiPoolerManager) getWALPosition(ctx context.Context) string {
```
rafael: Similarly, we already have `getPrimaryLSN` and `getStandbyReplayLSN`. Given that, I don’t think we need a separate role function.

Instead, I’d add a new helper `getWalPosition` in `pg_replication` that does something like:

```go
isPrimary := pm.isPrimary()

if isPrimary {
    return pm.getPrimaryLSN(ctx)
}

return pm.getStandbyReplayLSN(ctx)
```

We already have similar logic in `manager.go` in `checkDemotionState`. We could reuse this helper both there and here 🥳

go/multipooler/manager/rpc_initialization.go
rafael: Once we hook everything, I'm thinking we should remove this from the `Open`. So it's only in one place. All poolers need to be initialized, so if it's not there by the time we Open seems like a broken invariant.

In the future, we can have ways to introduce data migrations, but I think doing it when we call `Open` is a big hammer

proto/multipoolermanagerdata.proto
```diff
@@ -489,3 +489,73 @@ message ConsensusTerm {
...
+  string primary_host = 1;
+
+  // Primary server port
+  int32 primary_port = 2;
+
+  // Consensus term to set for this standby
+  int64 consensus_term = 3;
+
+  // Whether to force reinitialization (removes existing data directory)
+  bool force_reinit = 4;
```
rafael: I think we can call this just `force`

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,390 @@
...
+	// 4. Wait for database connection
+	if err := pm.waitForDatabaseConnection(ctx); err != nil {
+		return nil, fmt.Errorf("failed to connect to database: %w", err)
+	}
+
+	// 5. Create multigres schema and heartbeat table
+	if err := pm.initializeMultigresSchema(ctx); err != nil {
+		return nil, fmt.Errorf("failed to initialize multigres schema: %w", err)
+	}
+
```
rafael, line 88: something here should Update the multipooler record in the topo (I think )

go/multipooler/manager/rpc_initialization.go
deepthi: There's always the question of receive_lsn vs replay_lsn. So far it is unclear which one we should be returning where.

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,390 @@
...
+	// 4. Create standby.signal
+	if err := pm.createStandbySignal(); err != nil {
+		return nil, fmt.Errorf("failed to create standby.signal: %w", err)
+	}
+
+	// 5. Configure primary_conninfo
+	if err := pm.SetPrimaryConnInfo(ctx, req.PrimaryHost, req.PrimaryPort, false, false, req.ConsensusTerm, false); err != nil {
+		return nil, fmt.Errorf("failed to set primary_conninfo: %w", err)
+	}
+
```
cuongdo, line 141: It would be good to set `archive_mode = on` somewhere before starting PostgreSQL. `archive_command` can be set to `/bin/true`. This streamlines promotion later on, because `archive_mode` cannot be changed without a Postgres restart.

This isn't needed for this PR, but it'd be good to add a TODO.

go/multipooler/manager/rpc_initialization.go
deepthi: What would we update? The only thing that can possibly change is ServingStatus. Now I'm wondering whether we should even keep that in the topo at all?
```
  // PoolerType is the kind of pooler: PRIMARY or REPLICA
  PoolerType type = 6;

  // PoolerServingStatus is the current type of the pooler.
  PoolerServingStatus serving_status = 7;
```
also "current type" -> "current status"

go/multipooler/manager/rpc_initialization.go
deepthi: Good point! That should go into postgresconfig.go and postgresconfig_gen.go

go/multipooler/manager/rpc_initialization.go
deepthi: I've replaced the explicit query with a call to `isPrimary`, but it seems less convenient to de-duplicate the rest of the code here with what is in checkDemotionState.
However, I remembered that we have similar code in `rpc_consensus.go:ConsensusStatus`. I'm able to compact that code by calling `isPrimary` and `getPrimaryLSN`.

go/multipooler/manager/rpc_initialization.go
rafael: I was thinking `PoolerType`. Wondering if this is all part of what will be the status state management transition we were talking

go/multipooler/manager/rpc_initialization.go
deepthi: After discussion, we decided that we can in fact refactor and extract the common code. I have implemented that.

go/multipooler/manager/rpc_initialization.go
deepthi: Yeah, I would defer this for now.

go/multipooler/manager/rpc_initialization.go
deepthi: Added a check to fail the RPC if term number is not 1.

go/multipooler/manager/rpc_initialization.go
```diff
@@ -0,0 +1,360 @@
...
+	if err != nil {
+		return false
+	}
+
+	return info.IsDir()
+}
+
+// isPostgresRunning checks if PostgreSQL is currently running
+func (pm *MultiPoolerManager) isPostgresRunning(ctx context.Context) bool {
+	if pm.pgctldClient == nil {
```
rafael, line 237: small nit - I wonder, if we should remove this check. Seems like calling Status should be the canonical way of checking this

rafael (approved):

go/multipooler/manager/rpc_initialization.go
deepthi: It's a standard nil check to avoid a possible nil dereference on line 242. It doesn't hurt to leave it in.
If you mean the whole function is redundant, and should be replaced by a call to some existing function, we can do that as a follow up.

deepthi:
@Rafael thank you for the detailed review. We ended up with a much cleaner PR because of the time you took to do that.

PR #236, opened 2025-11-13
feat(multiorch): Discovery plumbing

rafael:
# Desc

Implements the foundational recovery engine for MultiOrch, inspired by Vitess VTOrc's architecture. This PR establishes the core maintenance loop infrastructure that will support future health checking and automated recovery capabilities.

### Core Components:
  - RecoveryEngine: Main orchestrator with configurable maintenance loop
  - ShardWatchTarget: Hierarchical filtering for databases/tablegroups/shards
  - Cluster Metadata Refresh: Periodic polling of topology service filtered by watch targets
  - Dynamic Config Reloading: Hot-reload of shard watch targets without restart
  - Goroutine Management: Pile-up prevention for slow operations using runIfNotRunning() helper.

###  Configuration Flags:
  - --shard_watch_targets: List of db/tablegroup/shard targets to monitor (dynamic, required)
  - --bookkeeping_interval: Cleanup task interval (default: 1m)
  - --cluster_metadata_refresh_interval: Topology polling interval (default: 15s)
  - --cluster_metadata_refresh_timeout: Timeout for metadata refresh operation (default: 30s)
  
## Tests
- Added relevant unit tests for the added components.

## Next Steps
1)  Starting to have some state management for all the data that is being tracked.
2) Plumbing to perform health checks on the poolers that are being watched.

### DIFF
--- a/go/multiorch/init.go
+++ b/go/multiorch/init.go
@@ -60,7 +60,7 @@ func (mo *MultiOrch) RunDefault() {
 // Register flags that are specific to multiorch.
 func (mo *MultiOrch) RegisterFlags(fs *pflag.FlagSet) {
 	fs.String("cell", mo.cell.Default(), "cell to use")
-	fs.StringSlice("shard-watch-targets", mo.shardWatchTargets.Default(), "list of db/tablegroup/shard targets to watch")
+	fs.StringSlice("watch-targets", mo.shardWatchTargets.Default(), "list of db/tablegroup/shard targets to watch")
 	fs.Duration("bookkeeping-interval", mo.bookkeepingInterval.Default(), "interval for bookkeeping tasks")
 	fs.Duration("cluster-metadata-refresh-interval", mo.clusterMetadataRefreshInterval.Default(), "interval for refreshing cluster metadata from topology")
 	fs.Duration("cluster-metadata-refresh-timeout", mo.clusterMetadataRefreshTimeout.Default(), "timeout for cluster metadata refresh operation")
@@ -79,8 +79,8 @@ func NewMultiOrch() *MultiOrch {
 			Dynamic:  false,
 			EnvVars:  []string{"MT_CELL"},
 		}),
-		shardWatchTargets: viperutil.Configure(reg, "shard-watch-targets", viperutil.Options[[]string]{
-			FlagName: "shard-watch-targets",
+		shardWatchTargets: viperutil.Configure(reg, "watch-targets", viperutil.Options[[]string]{
+			FlagName: "watch-targets",
 			Dynamic:  true,
 			EnvVars:  []string{"MT_SHARD_WATCH_TARGETS"},
 		}),
@@ -128,13 +128,13 @@ func (mo *MultiOrch) Init() {
 	// Validate and parse shard watch targets
 	targetsRaw := mo.shardWatchTargets.Get()
 	if len(targetsRaw) == 0 {
-		logger.Error("shard-watch-targets is required")
+		logger.Error("watch-targets is required")
 		os.Exit(1)
 	}
 
 	targets, err := ParseShardWatchTargets(targetsRaw)
 	if err != nil {
-		logger.Error("failed to parse shard-watch-targets", "error", err)
+		logger.Error("failed to parse watch-targets", "error", err)
 		os.Exit(1)
 	}
 
--- a/go/multiorch/recovery_engine.go
+++ b/go/multiorch/recovery_engine.go
@@ -98,7 +98,7 @@ func runIfNotRunning(logger *slog.Logger, inProgress *atomic.Bool, taskName stri
 // # Configuration
 //
 // The RecoveryEngine requires:
-//   - shard-watch-targets: List of database/tablegroup/shard targets to monitor
+//   - watch-targets: List of database/tablegroup/shard targets to monitor
 //   - bookkeeping-interval: How often to run cleanup tasks (default: 1m)
 //   - cluster-metadata-refresh-interval: How often to refresh from topology (default: 15s)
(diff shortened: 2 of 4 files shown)


go/provisioner/local/local.go
```diff
@@ -949,6 +949,7 @@ func (p *localProvisioner) provisionMultiOrch(ctx context.Context, req *provisio
 		"--topo-global-root", topoGlobalRoot,
 		"--topo-implementation", topoBackend,
 		"--cell", cell,
+		"--shard_watch_targets", req.DatabaseName,
```
rafael: This will effectively watch all tablegroups / shards in this db 💃

go/multiorch/recovery_engine_test.go
```diff
@@ -0,0 +1,425 @@
...
+	// Verify initial targets
+	re.mu.Lock()
+	currentTargets := re.shardWatchTargets
+	re.mu.Unlock()
+
+	expectedInit := []ShardWatchTarget{{Database: "db1"}}
+	require.Equal(t, expectedInit, currentTargets, "initial targets mismatch")
+
+	// Update viper config using Set()
+	shardWatchTargets.Set([]string{"db1", "db2", "db3"})
```
rafael, line 409: Excited to see this in action

go/multiorch/init.go
```diff
@@ -51,7 +60,11 @@ func (mo *MultiOrch) RunDefault() {
 // Register flags that are specific to multiorch.
 func (mo *MultiOrch) RegisterFlags(fs *pflag.FlagSet) {
 	fs.String("cell", mo.cell.Default(), "cell to use")
-	viperutil.BindFlags(fs, mo.cell)
+	fs.StringSlice("shard_watch_targets", mo.shardWatchTargets.Default(), "list of db/tablegroup/shard targets to watch")
+	fs.Duration("bookkeeping_interval", mo.bookkeepingInterval.Default(), "interval for bookkeeping tasks")
+	fs.Duration("cluster_metadata_refresh_interval", mo.clusterMetadataRefreshInterval.Default(), "interval for refreshing cluster metadata from topology")
+	fs.Duration("cluster_metadata_refresh_timeout", mo.clusterMetadataRefreshTimeout.Default(), "timeout for cluster metadata refresh operation")
+	viperutil.BindFlags(fs, mo.cell, mo.shardWatchTargets, mo.bookkeepingInterval, mo.clusterMetadataRefreshInterval, mo.clusterMetadataRefreshTimeout)
```
deepthi: We might need a custom linter to prevent underscores in flags. All flags need to use only dashes.

go/multiorch/recovery_engine.go
```diff
@@ -0,0 +1,350 @@
...
+		for _, target := range targets {
+			opt := &topo.GetMultiPoolersByCellOptions{
+				DatabaseShard: &topo.DatabaseShard{
+					Database:   target.Database,
+					TableGroup: target.TableGroup, // empty = all tablegroups
+					Shard:      target.Shard,      // empty = all shards
+				},
+			}
+
+			poolers, err := re.ts.GetMultiPoolersByCell(ctx, cell, opt)
```
deepthi, line 318: Do we have to fetch these every time? Is it possible to fetch them once and then set a watch on the path/prefix?
Doesn't have to be in this PR, but something to think about.

go/multiorch/config.go
```diff
@@ -0,0 +1,138 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+// ShardWatchTarget represents a target to watch in the format:
+// - "database" - watch entire database
+// - "database/tablegroup" - watch specific tablegroup
+// - "database/tablegroup/shard" - watch specific shard
+type ShardWatchTarget struct {
```
deepthi: Should we have Shard in this name at all? WatchTarget?

go/multiorch/recovery_engine.go
rafael: I'm down to explore watches. Here I was following the pattern we already used in Vitess. I recall @sougou  mentioned there were some issues with watches in etcd and we decided to use poll instead in vitess. But maybe those things are more stable now. 

@sougou  do you have thoughts on this?

go/multiorch/init.go
rafael: Oh that's a good idea. I added it

go/tools/linter/flagname/flagname.go
```diff
@@ -0,0 +1,123 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package flagname defines an analyzer that checks Go flag names.
+// It enforces that flag names use dashes (-) and never underscores (_),
+// and match the pattern ^[a-z][a-z0-9-]*.
+package flagname
```
rafael: I got help from code to write this checker. My manual checking seems 👍 . We can add some unit tests for this, but is a bit convoluted. That could be a follow up PR

go/multiorch/config.go
rafael: Good idea. Renamed

go/multiorch/recovery_engine.go
rafael: We spoke about this. We will explore using watches and it will be part of production readiness. Creating linear task to keep track of this internally

go/tools/linter/flagname/flagname.go
deepthi: This is being fixed in #238

deepthi (approved):

PR #238, opened 2025-11-13
build: linter to check for flag names with underscores in them

dweitzman:
Checks FlagSet and viper flags

All the name fixes in this PR were reported by the linter

Context: https://github.com/multigres/multigres/pull/236#discussion_r2521040123

This PR also separate golangci-lint into a separate github action from super-linter because without that I'm not sure how to make ruleguard work within super-linter. It's probably better to use the golangci-lint official github action anyway -- even just setting it up flagged bugs in our configuration

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -14,6 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+        with:
+          persist-credentials: false
       - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6
         with:
           go-version: stable


.github/linters/.golangci.yml
dweitzman, line 1: This is just a symlink that isn't needed after moving this off super-linter

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,25 @@
...
+  contents: read
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+        with:
+          persist-credentials: false
```
dweitzman, line 18: Hat tip to zizmor for catching the need to put `persist-credentials: false` here, which I would have missed otherwise

.golangci.yml
```diff
@@ -60,23 +63,31 @@ linters:
     sloglint:
       context: scope
     govet:
```
dweitzman, line 65: Switching the the official golangci-lint github action caused `golangci-lint config verify` to run, which cause some issues here I had to fix

go/clustermetadata/topo/conn.go
```diff
@@ -134,7 +134,7 @@ type ConnLock interface {
 
 	// LockWithTTL is similar to `Lock` but the difference is that it allows
 	// you to override the global default TTL that is configured for the
-	// implementation (--topo_etcd_lease_ttl).
+	// implementation (--topo-etcd-lease-ttl).
```
deepthi, line 137: Clearly none of these flags is actually being used in tests or local provisioner 🫠

PR #240, opened 2025-11-13
feat(multiorch): implement coordinator package with primitives needed for consensus

deepthi:
## Summary

  Implements the consensus coordinator foundation for orchestrating leader election and shard initialization in Multigres. 

  ## Changes

  ### New Package: `go/multiorch/coordinator/`

  #### Core Files

  **`coordinator.go`** - Main coordinator service
  - `Coordinator` struct with topology store integration
  - `AppointLeader()` - Orchestrates full consensus protocol for a shard
  - `GetShardNodes()` - Retrieves multipooler nodes from topology
  - `updateTopology()` - Updates topology to reflect new primary/standbys

  **`leader_appointment.go`** - Consensus protocol implementation
  - `BeginTerm()` - Stages 1-4 of consensus protocol:
    - Discovers max term from all nodes in parallel
    - Increments term for new election
    - Selects candidate based on WAL position and health
    - Recruits nodes via `BeginTerm` RPC in parallel
    - Validates quorum (majority required for now)
  - `Propagate()` - Stage 5: Promotes candidate and configures standby replication
  - `EstablishLeader()` - Stage 6: Verifies leader is ready and serving

  **`node.go`** - Node abstraction layer
  - `Node` struct wrapping gRPC clients for both MultiPoolerManager and MultiPoolerConsensus services
  - Helper methods for all initialization and consensus RPCs:
    - `InitializationStatus()`, `ConsensusStatus()`, `ManagerStatus()`
    - `BeginTerm()`, `Promote()`, `Demote()`
    - `InitializeEmptyPrimary()`, `InitializeAsStandby()`
    - `SetPrimaryConnInfo()`, `StopReplicationAndGetStatus()`

  **`client_factory.go`** - gRPC client management
  - `CreateNode()` - Creates a Node with connected gRPC clients
  - `CreateNodes()` - Batch creates multiple nodes from topology

  #### Tests

  **`client_factory_test.go`**
  - Node creation with valid/invalid pooler info
  - Batch node creation with error handling

  **`leader_appointment_test.go`**
  - Mock gRPC clients for all consensus and manager RPCs
  - Comprehensive coverage of all protocol stages:
    - `TestDiscoverMaxTerm` - Parallel term discovery
    - `TestSelectCandidate` - WAL-based candidate selection
    - `TestRecruitNodes` - Parallel node recruitment
    - `TestBeginTerm` - Full stages 1-4 with quorum validation
    - `TestPropagate` - Promotion and standby configuration
    - `TestEstablishLeader` - Leader readiness verification

  ## Design Details

  ### Shard-Scoped Operations
  All operations work on a single shard (replication group). Cohort = all multipooler nodes in one shard.

  ### Parallel Execution
  Term discovery and node recruitment execute in parallel using goroutines for performance.

  ### Quorum-Based Consensus
  Requires majority acceptance (`n/2 + 1`) before proceeding. Currently uses simple majority; TODO: future durability policy integration.

  ### WAL Position Selection
  Candidate selection prioritizes healthy nodes with most advanced WAL position. Uses lexicographic comparison as placeholder for proper PostgreSQL LSN comparison.

  ### Error Handling
  Proper use of `mterrors` with gRPC error codes:
  - `INVALID_ARGUMENT` - Validation errors
  - `FAILED_PRECONDITION` - State errors (insufficient quorum)
  - `UNAVAILABLE` - Service unavailable
  - `INTERNAL` - Unexpected states

  ### Topology Integration
  - Uses `GetMultiPoolersByCell()` with `DatabaseShard` filter
  - Updates node types via `UpdateMultiPoolerFields()`
  - Proper handling of `PoolerType_PRIMARY` and `PoolerType_REPLICA`

  ### Graceful Degradation
  Propagate stage continues even if some standbys fail to configure (logs warnings but doesn't fail the entire operation).

  ## What's NOT Included

  This PR focuses on the coordinator foundation. Future work:
  - Bootstrap initialization support (checks and actions for "shard has no primary")
  - Integration with multiorch discovery/recovery loops
  - Async `RepairExcluded()` for pg_rewind of diverged nodes
  - Proper PostgreSQL LSN comparison (currently lexicographic)
  - Durability policy framework integration (currently hardcoded majority)

 ## Related

  - Uses initialization RPCs from PR #234


go/multiorch/coordinator/leader_appointment.go
```diff
@@ -0,0 +1,382 @@
...
+// 2. Increment to get new term
+// 3. Select candidate based on WAL position
+// 4. Send BeginTerm RPC to all nodes in parallel
+// 5. Validate quorum (majority acceptance)
+//
+// Returns the candidate node, standbys, the new term, and any error.
+func (c *Coordinator) BeginTerm(ctx context.Context, shardID string, cohort []*Node) (*Node, []*Node, int64, error) {
+	// Stage 1: Obtain Term Number - Query all nodes for max term
+	c.logger.InfoContext(ctx, "Discovering max term", "shard", shardID, "cohort_size", len(cohort))
+	maxTerm, err := c.discoverMaxTerm(ctx, cohort)
```
rafael, line 38: Idea here:
- Coordinator gets initialized with a MaxTerm on process start. That should be sufficient.
- As part of maintenance tasks in operators, we are periodically checking the current max term. Only reason why that would diverge is that an election happen that this coordinator hasn't discovered yet. I think that could be part of maintance tick instead of trying to do it at the moment of the election.

rafael (commented):
The template is looking good to me on a first pass. We can go in depth once we get to details of each method implementation

go/multiorch/coordinator/client_factory.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Create gRPC connection
+	// TODO: Add proper TLS configuration for production
+	conn, err := grpc.NewClient(
+		addr,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to multipooler at %s: %w", addr, err)
+	}
```
deepthi, line 47: @rafael one thing that strikes me is that instead of multiorch creating a client here for each pooler, it should multiplex everything over the connection that is being used for health checks. Something we should refactor once you implement that part of multiorch discovery.

go/multiorch/coordinator/coordinator.go
```diff
@@ -0,0 +1,163 @@
...
+	c.logger.InfoContext(ctx, "EstablishLeader succeeded", "shard", shardID)
+
+	// Update topology to reflect new primary
+	if err := c.updateTopology(ctx, candidate, standbys); err != nil {
+		// Log but don't fail - the leader is established, topology is just metadata
+		c.logger.WarnContext(ctx, "Failed to update topology",
+			"shard", shardID,
+			"error", err)
+	}
+
```
deepthi, line 99: Should we move this into EstablishLeader?

go/multiorch/coordinator/leader_appointment.go
deepthi: The way I'm understanding this is that the "health" message from poolers will always include the current term, and multiorch can compute the max of those in-memory without any further network calls and initialize the coordinator with that value. Did I get it right?

go/multiorch/coordinator/coordinator.go
deepthi: Upon reflection, I think not. This is updating the topology for both leader and standbys. However, we need to implement retries if the topology update fails.

go/multiorch/coordinator/client_factory.go
rafael: oh yes! This should be something akin to tmcclient right?

PR #242, opened 2025-11-13
feat(multiorch): adds discovery in-memory store 

rafael:
## Summary

This PR adds in-memory state management for the discovery portions of multiorch. 

### Architecture

- Restructured the `multiorch/recovery` package into focused modules:
  - `engine.go` – core engine lifecycle and maintenance loop
  - `discovery.go` – cluster metadata discovery from topology
  - `bookkeeping.go` – cleanup tasks and audit logging
  - `config.go` – watch target configuration
- Introduced a generic `Store[K, V]` with thread-safe (mutex-based) operations.
- Added a `PoolerHealthCheckStatus` struct for tracking pooler state and timestamps.

### Discovery

- Updated the discovery loop to store the poolers in the in-memory store.

### Bookkeeping

- Implemented `forgetLongUnseenInstances()` to remove stale poolers (4+ hour threshold).
- Handles three classes of entries:
  - Broken entries
  - Never health-checked
  - Long-unseen poolers
- Adds audit logging for all cleanup operations with structured fields.

### Observability

- OpenTelemetry metrics:
  - `multiorch.recovery.pooler_store_size` (gauge) – tracks poolers in the store via an observable callback.
  - `multiorch.recovery.refresh_latency` (histogram) – measures metadata refresh duration.
- Emits an audit warning when the store exceeds 1000 poolers (suggesting splitting watch targets).

## Tests
- Added (what I believe) is comprehensive tests for discovery.

### DIFF
--- a/go/multiorch/recovery/bookkeeping.go
+++ b/go/multiorch/recovery/bookkeeping.go
@@ -44,7 +44,7 @@ func (re *Engine) forgetLongUnseenInstances() {
 	now := time.Now()
 	cutoff := now.Add(-threshold)
 
-	allPoolers := re.poolerStore.GetAllWithKeys()
+	allPoolers := re.poolerStore.GetMap()
 	storeSize := len(allPoolers)
 
 	// Warn if store gets too large - operator should consider splitting watchers
--- a/go/multiorch/recovery/discovery.go
+++ b/go/multiorch/recovery/discovery.go
@@ -98,7 +98,7 @@ func (re *Engine) refreshClusterMetadata() {
 					// New pooler - we've discovered it in the topology, but we haven't
 					// performed a health check yet. The health check loop will update
 					// LastSeen, LastCheckAttempted, LastCheckSuccessful, and IsUpToDate.
-					poolerInfo := &store.PoolerHealthCheckStatus{
+					poolerInfo := &store.PoolerHealth{
 						MultiPooler: pooler.MultiPooler,
 						IsUpToDate:  false, // Not yet health checked
 					}
--- a/go/multiorch/store/pooler_info.go
+++ b/go/multiorch/store/pooler_info.go
@@ -20,12 +20,12 @@ import (
 	"github.com/multigres/multigres/go/pb/clustermetadata"
 )
 
-// PoolerHealthCheckStatus represents runtime state of a MultiPooler instance.
-// This is the minimal version that stores:
+// PoolerHealth represents runtime state of a MultiPooler instance.
+// This stores:
 // - The MultiPooler record from topology
 // - Timestamps for staleness detection
 // - Computed fields for quick access
-type PoolerHealthCheckStatus struct {
+type PoolerHealth struct {
 	// MultiPooler record from topology service
 	MultiPooler *clustermetadata.MultiPooler
 
--- a/go/multiorch/store/store.go
+++ b/go/multiorch/store/store.go
@@ -20,6 +20,18 @@ import (
 
 // Store is a generic thread-safe in-memory key-value store.
 // It can be used to store any type of value indexed by any comparable key.
+//
+// Design rationale:
+// While sync.Map could be used here, this custom data structure provides a more
+// flexible API tailored to our access patterns. sync.Map is optimized for
+// "write-once, read-many" scenarios with minimal write contention. In contrast,
+// the recovery engine experiences frequent writes (health check updates, topology
+// changes, bookkeeping operations), making sync.Map's optimizations less effective.
+//
+// This abstraction provides flexibility for future optimizations if contention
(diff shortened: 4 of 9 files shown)


go/multiorch/recovery/discovery.go
```diff
@@ -0,0 +1,120 @@
...
+					"error", err,
+				)
+				continue
+			}
+
+			// Store each pooler in the state store
+			for _, pooler := range poolers {
+				key := topo.MultiPoolerIDString(pooler.Id)
+
+				// Check if we already know about this pooler
```
rafael, line 91: This is the gist of the changes here.

go/multiorch/recovery/bookkeeping.go
```diff
@@ -0,0 +1,121 @@
...
+	// - Expire old recovery history
+	// - Clean up stale data
+}
+
+// forgetLongUnseenInstances removes pooler instances that haven't been successfully
+// health checked in over 4 hours. This handles three cases:
+// 1. Broken entries (nil pointers - should never happen)
+// 2. Instances discovered in topology but never successfully health checked
+// 3. Instances that were previously healthy but haven't been seen in 4+ hours
+func (re *Engine) forgetLongUnseenInstances() {
```
rafael, line 42: This is new

go/multiorch/store/store.go
```diff
@@ -0,0 +1,96 @@
...
+
+package store
+
+import (
+	"sync"
+)
+
+// Store is a generic thread-safe in-memory key-value store.
+// It can be used to store any type of value indexed by any comparable key.
+type Store[K comparable, V any] struct {
```
deepthi, line 35: It is worth writing down why this data structure is needed, and why we are not using sync.Map
Briefly, because sync.Map is optimized for "write once, read many" scenarios and has a more complex API.
The other option of littering code with lock and unlock would be ugly and error-prone.

go/multiorch/store/store.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Clear removes all items from the store.
+func (s *Store[K, V]) Clear() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.items = make(map[K]V)
+}
+
+// GetAllWithKeys returns a snapshot of all key-value pairs in the store as a map.
+func (s *Store[K, V]) GetAllWithKeys() map[K]V {
```
deepthi:
```suggestion
func (s *Store[K, V]) GetAllValuesWithKeys() map[K]V {
```

go/multiorch/store/store.go
```diff
@@ -0,0 +1,96 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	_, existed := s.items[key]
+	delete(s.items, key)
+	return existed
+}
+
+// GetAll returns a snapshot of all values in the store.
+// The order is not guaranteed.
+func (s *Store[K, V]) GetAll() []V {
```
deepthi:
```suggestion
func (s *Store[K, V]) GetAllValues() []V {
```

go/multiorch/store/store.go
deepthi: Or GetMap would be more concise.

go/multiorch/store/pooler_info.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/multigres/multigres/go/pb/clustermetadata"
+)
+
+// PoolerHealthCheckStatus represents runtime state of a MultiPooler instance.
+// This is the minimal version that stores:
+// - The MultiPooler record from topology
+// - Timestamps for staleness detection
+// - Computed fields for quick access
+type PoolerHealthCheckStatus struct {
```
deepthi: I think we will eventually need to store the grpcConn here as well. So it is more than just status, but that is for the future. We can use TabletHealth from vitess for inspiration.

deepthi:
```suggestion
type PoolerHealthCheck struct {
```

rafael: In Vitess for VTOrc, we don't store the connection this way. We use `tmcclient` that internally manages the connections. I was thinking on following that approach here as well. Let me know if you think we should revisit this

deepthi: My inclination is to use a streaming health check similar to what vtgate does with vttablets. The counter argument is that other than periodic health checks, vtorc (and now multiorch) is not constantly sending and receiving grpc messages, so why keep the connection live (with its associated keep alive overhead)? Let's discuss this in more detail offline.

deepthi: This would be part of the same discussion https://github.com/multigres/multigres/pull/240#discussion_r2525036269

PR #246, opened 2025-11-17
feat(multipooler): rpc clients implementation for pooler

rafael:
# Description
* Port grpc client implementation for Multipooler from Vitess to Multigres.
# Desc
* Tests were also mostly ported. 
* Added a few new tests that seemed sensible

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -13,7 +13,7 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


deepthi:
Looking at `go/multiorch/coordinator/client_factory.go`, I'm thinking that `CreateNode` should now be passed in the connection held by multiorch for a specific pooler. Then it should use that to create `consensusClient` and `managerClient`. WDYT?
The only change right now would be to the signature of `CreateNode` and its internal logic. The wiring up of this with multiorch will be done later.

PR #248, opened 2025-11-17
feat(consensus): implement durability policy model, schema and checking in coordinator

deepthi:
### 1. Proto Definitions (proto/clustermetadata.proto)

**Policy Structure:**
- `DurabilityPolicy` - Policy metadata with versioning and activation tracking
- `QuorumRule` - Defines durability requirements with `required_count` and `async_fallback` control
- `QuorumType` enum - `ANY_N` and `MULTI_CELL_ANY_N` types for different availability guarantees
- `AsyncReplicationFallbackMode` enum - Controls behavior when insufficient standbys are available:
  - `ALLOW` (default) - Permits async replication fallback
  - `REJECT` - Fails leader appointment if sync replication cannot be established
  - `UNKNOWN` - Defaults to ALLOW behavior

**RPC Interface:**
- `GetDurabilityPolicy` RPC in MultiPoolerManager service
- Request/response messages in `multipoolermanagerdata.proto`

### 2. Database Schema (go/multipooler/manager/db.go)

- `multigres.durability_policy` table with JSONB quorum_rule storage
- Unique constraint on (policy_name, policy_version)
- Indexed `is_active` flag for fast active policy lookup

### 3. Quorum Validation (go/multiorch/coordinator/quorum.go)

**Validation Logic:**
- `validateAnyNQuorum()` - Requires N nodes from cohort
- `validateMultiCellQuorum()` - Requires nodes from N distinct cells, at least one node from each
- Cell-based logic embedded in implementation (not proto)

**Synchronous Replication Configuration:**
- `buildSyncReplicationConfig()` builds sync replication settings from quorum policy
- Respects `async_fallback` mode:
  - Returns error when mode is `REJECT` and sync replication cannot be established
  - Allows async fallback when mode is `ALLOW`
- Filters standbys by cell for `MULTI_CELL_ANY_N` policies
- Calculates `num_sync` as `required_count - 1` to meet quorum requirements

### 4. Policy Loading (go/multiorch/coordinator/policy.go)

**gRPC-Based Loading:**
- `LoadQuorumRuleFromNode()` retrieves policies via `GetDurabilityPolicy` RPC
- Uses cached gRPC client for efficient communication
- No direct database connections required
- Fallback to default ANY_N majority rule if no policy exists
- Bootstrap function to create default policies

**Load Strategy:**
- `LoadQuorumRule()` tries PRIMARY pooler first, then falls back to REPLICAs
- Queries all nodes until finding latest policy version
- Returns first successful response

### 5. RPC Client Architecture (go/multipooler/rpcclient/)

- `GetDurabilityPolicy` integrated into unified interface

**Implementations:**
- `Client` - Production implementation with connection caching
- `FakeClient` - Test implementation with configurable response maps

### 6. Coordinator Integration (go/multiorch/coordinator/)

**Node Architecture:**
- Node struct uses single `rpcClient` field instead of separate clients
- All 14 Node methods use cached RPC client for efficiency
- Added `GetDurabilityPolicy()` wrapper method

**Leader Appointment Flow:**
- **Stage 0**: Load durability policy via gRPC (AppointLeader)
- **Stage 1**: Discover max term
- **Stage 2**: Increment term
- **Stage 3**: Select candidate
- **Stage 4**: Recruit nodes (BeginTerm RPC)
- **Stage 5**: Validate quorum using loaded policy
- **Stage 6**: Build and validate sync replication config with async fallback control
- **Stage 7**: Propagate (promote candidate, configure standbys)
- **Stage 8**: Establish leader (verify ready state)

### 7. Tests

**Quorum Tests** (go/multiorch/coordinator/quorum_test.go):
- 11 comprehensive test cases covering success/failure scenarios
- Tests for both ANY_N and MULTI_CELL_ANY_N quorum types

**Sync Replication Tests** (go/multiorch/coordinator/sync_replication_test.go):
- 19 test cases covering various quorum configurations
- Cell filtering tests for MULTI_CELL_ANY_N
- Async fallback mode tests (ALLOW, REJECT, UNKNOWN)
- All scenarios validated

**Client Factory Tests** (go/multiorch/coordinator/client_factory_test.go):
- Tests for node creation with shared RPC client
- Validation of cached client usage

**Leader Appointment Tests** (go/multiorch/coordinator/leader_appointment_test.go):
- Full consensus protocol tests using FakeClient
- BeginTerm, Propagate, and EstablishLeader scenarios
- Error handling and quorum validation

---

## Design Notes

- **`required_count` dual purpose**: Node count for ANY_N, cell count for MULTI_CELL_ANY_N
- **Async fallback control**: Prevents silent durability degradation when standbys unavailable
- **gRPC-based policy loading**: Eliminates direct database dependencies, uses coordinator's cached client
- **Unified client interface**: Simplifies coordinator code, all RPCs through single interface

### DIFF
--- a/go/multiorch/coordinator/leader_appointment.go
+++ b/go/multiorch/coordinator/leader_appointment.go
@@ -275,18 +275,32 @@ func (c *Coordinator) recruitNodes(ctx context.Context, cohort []*Node, term int
 func (c *Coordinator) buildSyncReplicationConfig(quorumRule *clustermetadatapb.QuorumRule, standbys []*Node, candidate *Node) (*multipoolermanagerdatapb.ConfigureSynchronousReplicationRequest, error) {
 	requiredCount := int(quorumRule.RequiredCount)
 
-	// Determine async fallback mode (default to ALLOW if unset)
+	// Determine async fallback mode (default to REJECT if unset)
 	asyncFallback := quorumRule.AsyncFallback
 	if asyncFallback == clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_UNKNOWN {
-		asyncFallback = clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_ALLOW
+		asyncFallback = clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_REJECT
 	}
 
-	// If required_count is 1 or there are no standbys, don't configure synchronous replication
+	// If required_count is 1, don't configure synchronous replication
 	// With required_count=1, only the primary itself is needed for quorum, so async replication is sufficient
-	if requiredCount == 1 || len(standbys) == 0 {
+	if requiredCount == 1 {
 		c.logger.Info("Skipping synchronous replication configuration",
 			"required_count", requiredCount,
 			"standbys_count", len(standbys),
+			"reason", "async replication sufficient for quorum")
+		return nil, nil
+	}
+
+	// If there are no standbys, check async fallback mode
+	if len(standbys) == 0 {
+		if asyncFallback == clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_REJECT {
+			return nil, mterrors.New(mtrpcpb.Code_FAILED_PRECONDITION,
+				fmt.Sprintf("cannot establish synchronous replication: no standbys available (required %d standbys, async_fallback=REJECT)",
+					requiredCount-1))
+		}
+		c.logger.Info("Skipping synchronous replication configuration",
+			"required_count", requiredCount,
+			"standbys_count", 0,
 			"reason", "async replication sufficient for quorum")
 		return nil, nil
 	}
@@ -319,13 +333,23 @@ func (c *Coordinator) buildSyncReplicationConfig(quorumRule *clustermetadatapb.Q
 
 	// Calculate num_sync: required_count - 1 (since primary counts as 1)
 	// This ensures that primary + num_sync standbys = required_count total nodes/cells
-	numSync := int32(requiredCount - 1)
+	requiredNumSync := requiredCount - 1
+
+	// Check if we have enough standbys to meet the requirement
+	if requiredNumSync > len(eligibleStandbys) {
+		if asyncFallback == clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_REJECT {
+			return nil, mterrors.New(mtrpcpb.Code_FAILED_PRECONDITION,
+				fmt.Sprintf("cannot establish synchronous replication: insufficient standbys (required %d standbys, available %d, async_fallback=REJECT)",
+					requiredNumSync, len(eligibleStandbys)))
+		}
+		c.logger.Warn("Not enough standbys for required sync count, using all available",
+			"required_num_sync", requiredNumSync,
+			"available_standbys", len(eligibleStandbys))
+	}
 
 	// Cap num_sync at the number of available eligible standbys
+	numSync := int32(requiredNumSync)
 	if int(numSync) > len(eligibleStandbys) {
-		c.logger.Warn("Not enough standbys for required sync count, using all available",
-			"required_num_sync", numSync,
-			"available_standbys", len(eligibleStandbys))
 		numSync = int32(len(eligibleStandbys))
 	}
 
--- a/go/multiorch/coordinator/sync_replication_test.go
+++ b/go/multiorch/coordinator/sync_replication_test.go
@@ -54,6 +54,7 @@ func TestBuildSyncReplicationConfig(t *testing.T) {
 			QuorumType:    clustermetadatapb.QuorumType_QUORUM_TYPE_ANY_N,
 			RequiredCount: 2,
 			Description:   "Two node quorum",
+			AsyncFallback: clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_ALLOW,
 		}
 
 		standbys := []*Node{}
@@ -131,6 +132,7 @@ func TestBuildSyncReplicationConfig(t *testing.T) {
 			QuorumType:    clustermetadatapb.QuorumType_QUORUM_TYPE_ANY_N,
 			RequiredCount: 5,
 			Description:   "Five node quorum",
+			AsyncFallback: clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_ALLOW,
 		}
 
 		standbys := []*Node{
@@ -275,12 +277,13 @@ func TestBuildSyncReplicationConfig(t *testing.T) {
 		require.True(t, names["mp3"], "mp3 should be included (different cell)")
 	})
 
-	t.Run("MULTI_CELL_ANY_N with all standbys in same cell returns nil", func(t *testing.T) {
+	t.Run("MULTI_CELL_ANY_N with all standbys in same cell returns nil with ALLOW mode", func(t *testing.T) {
 		candidate := createTestNode("primary", "us-west-1a")
 		rule := &clustermetadatapb.QuorumRule{
 			QuorumType:    clustermetadatapb.QuorumType_QUORUM_TYPE_MULTI_CELL_ANY_N,
 			RequiredCount: 2,
 			Description:   "Two cell quorum",
+			AsyncFallback: clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_ALLOW,
 		}
 
 		standbys := []*Node{
@@ -291,7 +294,7 @@ func TestBuildSyncReplicationConfig(t *testing.T) {
 
 		config, err := c.buildSyncReplicationConfig(rule, standbys, candidate)
 		require.NoError(t, err)
-		require.Nil(t, config, "Should return nil when all standbys are in same cell as primary")
+		require.Nil(t, config, "Should return nil when all standbys are in same cell as primary with ALLOW mode")
 	})
 
 	t.Run("MULTI_CELL_ANY_N with mixed cells only includes different cells", func(t *testing.T) {
@@ -328,6 +331,7 @@ func TestBuildSyncReplicationConfig(t *testing.T) {
 			QuorumType:    clustermetadatapb.QuorumType_QUORUM_TYPE_MULTI_CELL_ANY_N,
 			RequiredCount: 4, // Would need 3 standbys
 			Description:   "Four cell quorum",
+			AsyncFallback: clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_ALLOW,
 		}
 
 		standbys := []*Node{
@@ -414,7 +418,7 @@ func TestBuildSyncReplicationConfig(t *testing.T) {
 		require.Nil(t, config, "Should return nil and allow async replication")
 	})
 
-	t.Run("default (UNKNOWN) mode behaves like ALLOW", func(t *testing.T) {
+	t.Run("default (UNKNOWN) mode behaves like REJECT", func(t *testing.T) {
 		candidate := createTestNode("primary", "us-west-1a")
 		rule := &clustermetadatapb.QuorumRule{
 			QuorumType:    clustermetadatapb.QuorumType_QUORUM_TYPE_MULTI_CELL_ANY_N,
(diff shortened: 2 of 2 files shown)


go/multiorch/coordinator/leader_appointment.go
```diff
@@ -261,26 +261,99 @@ func (c *Coordinator) recruitNodes(ctx context.Context, cohort []*Node, term int
...
+			"total_standbys", len(standbys),
+			"eligible_standbys", len(eligibleStandbys),
+			"excluded_same_cell", len(standbys)-len(eligibleStandbys))
+
+		// If no eligible standbys remain after filtering, use async replication
+		if len(eligibleStandbys) == 0 {
+			c.logger.Warn("No eligible standbys in different cells, using async replication",
+				"candidate_cell", candidate.ID.Cell,
+				"total_standbys", len(standbys))
+			return nil
```
deepthi: I'm skeptical about this. Should we fail the leader appointment process if there aren't enough nodes to respect the durability policy?
cc @GuptaManan100 and @rafael

go/multiorch/coordinator/coordinator.go
```diff
@@ -45,24 +45,39 @@ func NewCoordinator(coordinatorID *clustermetadatapb.ID, topoStore topo.Store, l
...
 	if len(cohort) == 0 {
 		return mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT, "cohort is empty for shard %s", shardID)
 	}
 
-	// Stage 1-4: BeginTerm (recruit nodes, get consensus)
-	c.logger.InfoContext(ctx, "Stage 1-4: Beginning term", "shard", shardID)
-	candidate, standbys, term, err := c.BeginTerm(ctx, shardID, cohort)
+	// Stage 0: Load durability policy from any available node
+	c.logger.InfoContext(ctx, "Loading durability policy", "shard", shardID)
+	quorumRule, err := c.LoadQuorumRule(ctx, cohort, database)
```
deepthi, line 70: Instead of reading the durability policy with a direct database connection, we should use the cached client that was introduced in #246.
What is the preference from reviewers? Should I integrate it into this PR, or shall I defer it to the next one?

go/multiorch/coordinator/leader_appointment.go
deepthi: We should make this part of the durability policy itself as an option (ASYNC_FALLBACK boolean or enum).

go/multiorch/coordinator/coordinator.go
deepthi: I'm going to go ahead and make this change now.

proto/multipoolermanagerservice.proto
```diff
@@ -79,6 +79,11 @@ service MultiPoolerManager {
   // The following methods are used by MultiOrch to
   // perform HA operations (reparents and failovers)
 
+  // GetDurabilityPolicy retrieves the active durability policy from the local database
+  // Used by MultiOrch to query quorum rules via gRPC instead of direct database connection
+  rpc GetDurabilityPolicy(multipoolermanagerdata.GetDurabilityPolicyRequest)
```
deepthi, line 84: Does this belong here, or should it be part of the consensus service? I couldn't make up my mind about this.

proto/clustermetadata.proto
```diff
@@ -216,3 +218,75 @@ enum PoolerServingStatus {
...
+}
+
+// AsyncReplicationFallbackMode defines the behavior when insufficient standbys are available
+// for synchronous replication to meet the quorum requirement
+enum AsyncReplicationFallbackMode {
+  // ASYNC_REPLICATION_FALLBACK_MODE_UNKNOWN represents an unknown or uninitialized mode
+  ASYNC_REPLICATION_FALLBACK_MODE_UNKNOWN = 0;
+
+  // ASYNC_REPLICATION_FALLBACK_MODE_ALLOW permits falling back to async replication
+  // when no standby servers are available. This is the default behavior.
```
deepthi, line 286: What do people think of this default? Should we prefer REJECT instead? That is the Vitess behavior - preferring durability over availability.

go/multiorch/coordinator/client_factory.go
```diff
@@ -18,57 +18,35 @@ import (
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to connect to multipooler at %s: %w", addr, err)
-	}
-
-	// Create clients for both services on the same connection
-	managerClient := multipoolermanagerpb.NewMultiPoolerManagerClient(conn)
-	consensusClient := consensuspb.NewMultiPoolerConsensusClient(conn)
+// CreateNode creates a Node with a reference to the shared cached RPC client.
+// The node stores the pooler metadata needed for making RPC calls.
+func CreateNode(ctx context.Context, rpcClient rpcclient.MultiPoolerClient, poolerInfo *topo.MultiPoolerInfo) (*Node, error) {
```
rafael: We should be able to delete this file entirely with the client library that we just added

go/multiorch/coordinator/coordinator.go
```diff
@@ -45,24 +48,39 @@ func NewCoordinator(coordinatorID *clustermetadatapb.ID, topoStore topo.Store, l
...
 
 	if len(cohort) == 0 {
 		return mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT, "cohort is empty for shard %s", shardID)
 	}
 
-	// Stage 1-4: BeginTerm (recruit nodes, get consensus)
-	c.logger.InfoContext(ctx, "Stage 1-4: Beginning term", "shard", shardID)
-	candidate, standbys, term, err := c.BeginTerm(ctx, shardID, cohort)
+	// Stage 0: Load durability policy from any available node
+	c.logger.InfoContext(ctx, "Loading durability policy", "shard", shardID)
```
rafael, line 69: I think there is an issue here. But it will be easier to discuss synchronously.

go/multiorch/coordinator/policy.go
```diff
@@ -0,0 +1,170 @@
...
+// LoadQuorumRule attempts to load the quorum rule from any available node in the cohort.
+// It tries each node until it finds one that responds successfully.
+func (c *Coordinator) LoadQuorumRule(ctx context.Context, cohort []*Node, database string) (*clustermetadatapb.QuorumRule, error) {
+	if len(cohort) == 0 {
+		return nil, mterrors.New(mtrpcpb.Code_INVALID_ARGUMENT, "cohort is empty")
+	}
+
+	var lastErr error
+	for _, node := range cohort {
+		rule, err := c.LoadQuorumRuleFromNode(ctx, node, database)
```
rafael: we should this in parallel.

go/multiorch/coordinator/policy.go
```diff
@@ -0,0 +1,296 @@
...
+		"description", quorumRule.Description)
+
+	return quorumRule, nil
+}
+
+// LoadQuorumRule loads the quorum rule using the following strategy:
+// 1. If a PRIMARY node exists, load from it (most up-to-date)
+// 2. Otherwise, load from all REPLICA nodes in parallel
+// 3. Wait for n-1 responses
+// 4. Return the rule with the highest version number
```
deepthi, line 67: This needs review. The thinking behind only waiting for n-1 is that we want to tolerate failures. But if the primary is already unreachable, then we should probably read from _all_ replicas and not n-1 replicas.

deepthi: I'm going to change this to "all replicas" instead of "n-1".

go/multiorch/coordinator/leader_appointment.go
```diff
@@ -24,15 +25,15 @@ import (
...
 // 4. Send BeginTerm RPC to all nodes in parallel
-// 5. Validate quorum (majority acceptance)
+// 5. Validate quorum using durability policy
 //
 // Returns the candidate node, standbys, the new term, and any error.
-func (c *Coordinator) BeginTerm(ctx context.Context, shardID string, cohort []*Node) (*Node, []*Node, int64, error) {
+func (c *Coordinator) BeginTerm(ctx context.Context, shardID string, cohort []*Node, quorumRule *clustermetadatapb.QuorumRule) (*Node, []*Node, int64, error) {
 	// Stage 1: Obtain Term Number - Query all nodes for max term
 	c.logger.InfoContext(ctx, "Discovering max term", "shard", shardID, "cohort_size", len(cohort))
 	maxTerm, err := c.discoverMaxTerm(ctx, cohort)
```
rafael, line 40: Just for my memory. My recollection is that we aligned on not discoveringMaxTerm in this way. 

Is that still the plan?

go/multiorch/coordinator/leader_appointment.go
```diff
@@ -261,24 +262,111 @@ func (c *Coordinator) recruitNodes(ctx context.Context, cohort []*Node, term int
...
-		StandbyIds:        make([]*clustermetadatapb.ID, len(standbys)),
+// buildSyncReplicationConfig creates synchronous replication configuration based on the quorum policy.
+// Returns nil if synchronous replication should not be configured (required_count=1 or no standbys).
+// For MULTI_CELL_ANY_N policies, excludes standbys in the same cell as the candidate (primary).
+func (c *Coordinator) buildSyncReplicationConfig(quorumRule *clustermetadatapb.QuorumRule, standbys []*Node, candidate *Node) (*multipoolermanagerdatapb.ConfigureSynchronousReplicationRequest, error) {
+	requiredCount := int(quorumRule.RequiredCount)
+
+	// Determine async fallback mode (default to ALLOW if unset)
+	asyncFallback := quorumRule.AsyncFallback
+	if asyncFallback == clustermetadatapb.AsyncReplicationFallbackMode_ASYNC_REPLICATION_FALLBACK_MODE_UNKNOWN {
```
rafael, line 280: I think if the mode is unknown, we should error here. Seems risky to default to async if it was not an explicit decision.

go/multiorch/coordinator/node.go
```diff
@@ -20,14 +20,13 @@ import (
 
 	"google.golang.org/protobuf/types/known/durationpb"
 
+	"github.com/multigres/multigres/go/multipooler/rpcclient"
```
rafael, line 22: I wonder if everything that is being stored here, can be part of the PoolerInfo store that we also maintain for recovery. There seems to be some duplication between this and that code

go/multiorch/coordinator/node.go
```diff
@@ -43,25 +42,25 @@ type Node struct {
...
 
 // InitializationStatus gets the initialization status of this node
 func (n *Node) InitializationStatus(ctx context.Context) (*multipoolermanagerdatapb.InitializationStatusResponse, error) {
 	req := &multipoolermanagerdatapb.InitializationStatusRequest{}
-	return n.ManagerClient.InitializationStatus(ctx, req)
+	return n.rpcClient.InitializationStatus(ctx, n.pooler, req)
 }
 
 // WaitForLSN waits for this node to replay WAL up to the target LSN
 func (n *Node) WaitForLSN(ctx context.Context, targetLsn string) error {
```
rafael: These methods are all wrappers around the interface that now the rpc client is already providing.

Proposal: let's remove them and just call the methods for the rpc.Client directly. 

Rationale: Not obvious what's the value of abstracting them under the node.

go/multiorch/coordinator/policy.go
```diff
@@ -0,0 +1,295 @@
...
+		"node", node.ID.Name,
+		"database", database,
+		"quorum_type", quorumRule.QuorumType,
+		"required_count", quorumRule.RequiredCount,
+		"description", quorumRule.Description)
+
+	return quorumRule, nil
+}
+
+// LoadQuorumRule loads the quorum rule using the following strategy:
```
rafael, line 63: Early on I was pushing for a more structured, persisted representation of durability policies, my thinking has shifted. The more I think about this, the more I feel we should follow the Vitess approach: do not persist the full durability policy proto.

I think this model is the right level of simplicity for us right now. Persisting the full proto introduces complexity.

So for V0, my suggestion is that we only persist the name of the durability policy, and let the actual behavioral details be defined in code.

We can always introduce more flexibility later once we have a strong case for it.

rafael (approved):
We reviewed this PR in a call. Pre-stamping. Deepthi will look at the comments and then merge

go/multiorch/coordinator/policy.go
deepthi: Yes, I noticed that and fixed it.

go/multiorch/coordinator/coordinator.go
deepthi: We discussed this. For now the coordinator is being written as a module and it needs to fetch durability policies this way. When we integrate this with multiorch recovery, we will assume that multiorch fetches and caches durability policy during discovery, and uses the cached policy instead of reading it via RPC.
At some point we will also build a way to safely change durability policies on a running cluster.

go/multiorch/coordinator/leader_appointment.go
deepthi: That is correct. Another piece of data that multiorch will use its memory for. This is temporary until then.

go/multiorch/coordinator/leader_appointment.go
deepthi: My thought was to default it to REJECT and fail the leader appointment. I think we are in agreement, so I will make this change.
Ref: https://github.com/multigres/multigres/pull/248#discussion_r2539316279

go/multiorch/coordinator/node.go
deepthi: When we integrate recovery with leader appointment we can de-duplicate.

go/multiorch/coordinator/policy.go
deepthi: We discussed this and decided to meet in the middle
- We probably don't need protos for durability policy, quorum rule etc. Instead they can be structs/enums
- Putting all the behavior into code will lead to an explosion of the number of implementations which are mostly the same. For example, we already have async_fallback_mode which makes the number of policies 2x if it has to be inferred from the policy name. For that reason, we should still have config captured in structs, and used to drive conditional code.

go/multiorch/coordinator/client_factory.go
deepthi: I had already deleted most of it when I integrated the new client library. I will move these remaining two functions (CreateNode and CreateNodes) into node.go and delete this file.

go/multiorch/coordinator/policy.go
deepthi: On second thought, re protobuf definitions for all the durability related structs:
- multiorch will need to fetch durability policy from pooler over gRPC, 
- we need to store a representation of durability policy in the database
For these reasons, I'm retaining them in this PR, but I'm open to changing it in future.

deepthi:
I made most of the agreed upon changes, and left a comment with the reason why I haven't removed protobufs. Merging.

go/multiorch/coordinator/leader_appointment.go
GuptaManan100: Yes! I think we should fail the leader election. That is the only safe thing to do. Reverting to async can cause data loss.

PR #250, opened 2025-11-18
feat(multipooler): adds replication status method to manager

rafael:
# Description
This PR does two things:
1) Renames ReplicationStatus to be StandbyReplicationStatus
2) Adds a new ReplicationStatus method that works regardless of the pooler type. This is the method that MultiOrc will use to check status. 

It is mostly wiring as nothing in the implementation change. Just calling the same underlying implementations depending on the type

### DIFF
--- a/go/multiorch/coordinator/leader_appointment_test.go
+++ b/go/multiorch/coordinator/leader_appointment_test.go
@@ -58,7 +58,7 @@ func createMockNode(fakeClient *rpcclient.FakeClient, name string, term int64, w
 		Accepted: true,
 	}
 
-	fakeClient.StatusResponses[name] = &multipoolermanagerdatapb.StateResponse{
+	fakeClient.StateResponses[name] = &multipoolermanagerdatapb.StateResponse{
 		State: "ready",
 	}
 
@@ -429,7 +429,7 @@ func TestEstablishLeader(t *testing.T) {
 		candidate := createMockNode(fakeClient, "mp1", 5, "0/3000000", true, "primary")
 
 		// Override the status response to indicate not ready
-		fakeClient.StatusResponses["mp1"] = &multipoolermanagerdatapb.StateResponse{
+		fakeClient.StateResponses["mp1"] = &multipoolermanagerdatapb.StateResponse{
 			State: "initializing",
 		}
 


go/mterrors/code.go
```diff
@@ -29,10 +29,14 @@ var (
 	// MT13001 General Error
 	MT13001 = errorWithoutState("MT13001", mtrpcpb.Code_INTERNAL, "[BUG] %s", "This error should not happen and is a bug. Please file an issue on GitHub: https://github.com/multigres/multigres/issues/new/choose.")
 
+	// MT13002 Pooler Type Mismatch
```
rafael, line 32: I thought it would be sensible to have an error code for this as I would like this to be a particular kind of failures that multiorc needs to detect.

go/multipooler/manager/rpc_manager.go
```diff
@@ -250,6 +250,71 @@ func (pm *MultiPoolerManager) ReplicationStatus(ctx context.Context) (*multipool
 	return status, nil
 }
 
+// ReplicationStatus gets unified status that works for both PRIMARY and REPLICA poolers
+// The multipooler returns information based on what type it believes itself to be
```
deepthi: It seems fine to call this `ReplicationStatus` right now because `PrimaryStatus` pretty much returns only replication related info. Are we sure that in future we won't be adding more things to it that are unrelated to replication?
One alternative is to call this `Status` and rename the existing `Status` RPC to `State` - because that is what it is returning.

proto/multipoolermanagerservice.proto
```diff
@@ -41,8 +41,14 @@ service MultiPoolerManager {
...
-  // ReplicationStatus gets the current replication status of the standby
-  rpc ReplicationStatus(multipoolermanagerdata.ReplicationStatusRequest)
+  // StandbyReplicationStatus gets the current replication status of the standby
+  rpc StandbyReplicationStatus(multipoolermanagerdata.StandbyReplicationStatusRequest)
+    returns (multipoolermanagerdata.StandbyReplicationStatusResponse);
+
+  // Status gets unified status that works for both PRIMARY and REPLICA poolers
+  // The multipooler returns information based on what type it believes itself to be,
+  // avoiding disparity between what MultiOrch thinks versus actual state
+  rpc Status(multipoolermanagerdata.ReplicationStatusRequest)
```
deepthi: This seems off to me. The Request and Response names should match the RPC name exactly, so they should be StatusRequest and StatusResponse.

rafael: oh yes. Good catch

PR #252, opened 2025-11-19
feat(consensus): implement shard initialization

deepthi:
# Shard Initialization

## Overview

The shard initialization system provides a library of actions that MultiOrch uses to initialize PostgreSQL shards in different scenarios. The system also provides utilities to detect the state of nodes in a shard in order to call the appropriate initialization action.

### Scenario Detection

The system provides utilities to determine which initialization scenario applies:

**`scenario_detection.go`** contains:
- `InitializationScenario` enum with four states:
  - `ScenarioBootstrap` - All nodes are empty (no initialized schemas)
  - `ScenarioRepair` - Mixed state: some nodes have data, some are empty
  - `ScenarioReelect` - All nodes initialized but no current primary
  - `ScenarioUnknown` - Unable to determine state (e.g., all nodes unavailable)

- `GatherInitializationStatus(ctx, cohort, logger)` - Queries all nodes via the `InitializationStatus` RPC to collect their current state

- `DetermineScenario(statuses, logger)` - Analyzes node statuses and returns the appropriate scenario

### Initialization Actions

The system provides two distinct actions for shard initialization:

#### 1. BootstrapShardAction (`bootstrap_shard.go`)

Used when all nodes in the shard are empty (no data, uninitialized schemas).

**Workflow:**
1. Retrieves the durability policy name from the Database proto in topology (e.g., "ANY_2", "MULTI_CELL_ANY_2")
2. Selects the first reachable node as the bootstrap candidate
3. Initializes the candidate as an empty primary with consensus term = 1 via `InitializeEmptyPrimary` RPC
4. Creates the durability policy record in the primary's `multigres.durability_policy` table
5. Initializes remaining nodes as standbys in parallel via `InitializeAsStandby` RPC

**Key Features:**
- Uses RPC-based initialization (no direct database connections for initialization)
- Parallel standby initialization for efficiency
- Continues even if some standbys fail (at least primary is established)
- Parses policy names into `QuorumRule` proto structures stored as JSONB

#### 2. AppointLeaderAction (`appoint_leader.go`)

Used for both repair and reelect scenarios. This action delegates to the coordinator's consensus protocol.

**Workflow:**
- Calls `coordinator.AppointLeader(ctx, shardID, cohort, database)`
- The coordinator runs the full consensus protocol:
  1. Term discovery (finds maximum consensus term)
  2. Candidate selection (chooses most advanced node by WAL position)
  3. Node recruitment (proposes new term to all nodes)
  4. Quorum validation (ensures durability requirements can be met)
  5. Primary promotion (via `PromoteToPrimary` RPC)
  6. Standby configuration (via `ConfigureAsStandby` RPC with synchronous replication settings)

**Use Cases:**
- **Repair scenario**: Some nodes have data, need to elect the most advanced node
- **Reelect scenario**: All nodes initialized but no primary, need to re-elect a leader

## RPC-Based Initialization

All initialization operations use RPCs instead of direct database connections:

- **InitializationStatus RPC** - Checks if a node is initialized
- **InitializeEmptyPrimary RPC** - Initializes a node as a new primary with term=1
- **InitializeAsStandby RPC** - Initializes a node as a standby from a primary backup
- **PromoteToPrimary RPC** - Promotes an existing node to primary (used in AppointLeader)
- **ConfigureAsStandby RPC** - Configures replication and synchronous commit settings
- **CreateDurabilityPolicy RPC** - Creates durability policy records (used by bootstrap)

This RPC-based approach maintains clean separation between MultiOrch (orchestrator) and MultiPoolerManager (node-level operations).

## Durability Policy Management

During bootstrap initialization, the system:
1. Reads the policy name from the Database proto in topology (e.g., "ANY_2")
2. Parses it into a `QuorumRule` with `QuorumType` and `RequiredCount`
3. Inserts it into `multigres.durability_policy` table as JSONB
4. Sets `policy_version=1` and `is_active=true` for the initial policy

The policy is automatically replicated to standbys through PostgreSQL's built-in replication.

## Schema Initialization Refactoring

The implementation includes a refactoring that eliminated the redundant `initializeMultigresSchema` method and consolidated all multigres schema creation to use `CreateSidecarSchema`. This function creates both the `multigres.heartbeat` table and the `multigres.durability_policy` table with its index, ensuring consistent schema initialization across all code paths (bootstrap initialization via RPC and normal multipooler startup).

## Pending
Integration with MultiOrch discovery/recovery engine to automatically trigger initialization when "shard has no primary" is detected.

## Suggested Usage Pattern

MultiOrch can use the library as follows, but this is optional. We may choose to delete this in favor of multiorch doing this in a different way.

```go
// Step 1: Gather initialization status from all nodes
statuses, err := actions.GatherInitializationStatus(ctx, cohort, logger)

// Step 2: Determine which scenario applies
scenario := actions.DetermineScenario(statuses, logger)

// Step 3: Execute appropriate action
switch scenario {
case actions.ScenarioBootstrap:
    action := actions.NewBootstrapShardAction(topoStore, logger)
    err = action.Execute(ctx, shardID, database, cohort)
    
case actions.ScenarioRepair, actions.ScenarioReelect:
    action := actions.NewAppointLeaderAction(coordinator, logger)
    err = action.Execute(ctx, shardID, database, cohort)
    
case actions.ScenarioUnknown:
    // Handle error - unable to determine state
}
```


go/multiorch/coordinator/node.go
```diff
@@ -38,11 +38,11 @@ type Node struct {
...
 
-	// rpcClient is the shared cached RPC client for calling multipooler services
-	rpcClient rpcclient.MultiPoolerClient
+	// RpcClient is the shared cached RPC client for calling multipooler services
+	RpcClient rpcclient.MultiPoolerClient
 
-	// pooler is the full multipooler metadata needed for RPC calls
-	pooler *clustermetadatapb.MultiPooler
+	// Pooler is the full multipooler metadata needed for RPC calls
+	Pooler *clustermetadatapb.MultiPooler
```
deepthi, line 45: @rafael this is now kind of ugly. Because we removed the RPC wrappers in #248, we have to export these fields so that the actions package can do things like `node.RpcClient.InitializationStatus`.
Maybe we should resurrect the wrappers so that it is more encapsulated.

deepthi:
The main things to review are
- initialize_shard.go
- new RPC to create durability policy

Everything else is because some node fields are now exported.

go/multiorch/coordinator/node.go
rafael: What do you think if we come back to this later? I feel like we don't need the node.go at all once we have all the recovery and coordinator pieces in place. I think we can refactor this part once all the pieces are in place

go/multiorch/actions/initialize_shard.go
```diff
@@ -0,0 +1,512 @@
...
+	// ScenarioUnknown indicates we couldn't determine the scenario
+	ScenarioUnknown InitializationScenario = iota
+
+	// ScenarioBootstrap: All nodes are empty (no data directories or no initialized schemas)
+	// Action: Initialize one node as primary, then initialize others as standbys
+	ScenarioBootstrap
+
+	// ScenarioRepair: Some nodes have data but no current primary
+	// Action: Use coordinator.AppointLeader to elect the most advanced node
+	ScenarioRepair
```
rafael: Why do we have repair scenarios in initialize? Thinking that this should handle a single action. Anything outside this is part of the Analyze, trigger repairs.

go/multiorch/actions/initialize_shard.go
```diff
@@ -0,0 +1,512 @@
...
+		"empty", emptyCount,
+		"unavailable", unavailableCount)
+
+	// All nodes are empty (or unavailable) → Bootstrap
+	if initializedCount == 0 && emptyCount > 0 {
+		return ScenarioBootstrap
+	}
+
+	// All reachable nodes are initialized → Reelect
+	if initializedCount > 0 && emptyCount == 0 {
```
rafael: Similar as the comment above, I think in this context, we shouldn't be doing any other scenario outside Initialize. 

Thinking that the guardrail, where we can't initialize if it's already initialized should be performed, but outside that, this should just perform the Initialization

go/multiorch/actions/initialize_shard.go
```diff
@@ -0,0 +1,512 @@
...
+	if initializedCount > 0 && emptyCount > 0 {
+		return ScenarioRepair
+	}
+
+	// All nodes unavailable or unknown state
+	return ScenarioUnknown
+}
+
+// executeBootstrap handles the bootstrap scenario: initialize a new shard from scratch
+func (a *InitializeShardAction) executeBootstrap(ctx context.Context, shardID string, database string, cohort []*coordinator.Node) error {
```
rafael: not for this PR, but I've been thinking that this action is one where we need a topo Lock. Otherwise, I feel it will be hard to protect from multiple coordinators racing to bootstrap a shard.

go/multiorch/actions/bootstrap_shard.go
```diff
@@ -0,0 +1,512 @@
...
+	if err != nil {
+		return mterrors.Wrap(err, "failed to get durability policy name from topology")
+	}
+
+	a.logger.InfoContext(ctx, "Using durability policy",
+		"shard", shardID,
+		"database", database,
+		"policy_name", policyName)
+
+	// Step 2: Select the first reachable node as the initial primary
```
rafael, line 77: Related to my comment above, without a lock, I would be nervous that different coordinators choose a different node and they could potentially choose a different initial primary

go/multiorch/actions/bootstrap_shard.go
```diff
@@ -0,0 +1,512 @@
...
+	if err := a.createDurabilityPolicy(ctx, candidate, database, policyName); err != nil {
+		return mterrors.Wrap(err, "failed to create durability policy in database")
+	}
+
+	a.logger.InfoContext(ctx, "Created durability policy in database",
+		"shard", shardID,
+		"database", database,
+		"policy_name", policyName)
+
+	// Step 5: Initialize remaining nodes as standbys
```
rafael, line 118: I think there could a different approach here that could be more long term proof. Instead of initializing the renaming nodes as standbys, we do the following:
- Take a backup of the primary.
- Bootstrap is done.
- Start the rest of the standbys. But there is no initialization method, they would come as any normal standby in the cluster.
- They will restore from backup.

What should happen next organically, is that orchestrator will detect that cluster is broken and it will perform a normal repair.

go/multiorch/actions/initialize_shard.go
```diff
@@ -0,0 +1,512 @@
...
+	a.logger.InfoContext(ctx, "Bootstrap initialization complete",
+		"shard", shardID,
+		"database", database,
+		"primary", candidate.ID.Name,
+		"standbys", len(standbys))
+
+	return nil
+}
+
+// executeRepair handles the repair scenario: some nodes have data, elect the most advanced
```
rafael: Same line as the other comment, I think we should wait a bit before adding these actions. I think this will conflict with some of the work I'm doing

go/multiorch/actions/bootstrap_shard.go
```diff
@@ -0,0 +1,512 @@
...
+
+	for _, standby := range standbys {
+		go func(node *coordinator.Node) {
+			req := &multipoolermanagerdatapb.InitializeAsStandbyRequest{
+				PrimaryHost:   primary.Hostname,
+				PrimaryPort:   primary.Port,
+				ConsensusTerm: 1,
+				Force:         false,
+			}
+			resp, err := node.RpcClient.InitializeAsStandby(ctx, node.Pooler, req)
```
rafael, line 200: Looking at the implementation of `InitializeAsStandby`, we left a placeholder to take the backup inside this action. 

In this model, each standby will take a backup of the primary. Which is not a big deal, but just flagging. 

I think the model where bootstrap ends at taking backup from seems better.

What do you think ?

go/multiorch/actions/bootstrap_shard.go
```diff
@@ -0,0 +1,512 @@
...
+
+	// Test the connection
+	if err := db.PingContext(ctx); err != nil {
+		return mterrors.Wrapf(err, "failed to ping primary %s", primary.ID.Name)
+	}
+
+	// Insert the policy into the durability_policy table
+	now := time.Now()
+	query := `
+		INSERT INTO multigres.durability_policy
```
rafael, line 299: Remind me, initialize primary created the table right?

go/multiorch/actions/initialize_shard.go
deepthi: good point.

go/multiorch/actions/initialize_shard.go
deepthi: The idea is that multiorch detects ShardHasNoPrimary, and it may not know whether everything has already been initialized. ShardHasNoPrimary repair can do the right thing depending on initialization status.
If this is confusing, I'm open to reorganizing things in a way that is more intuitive.
In this sense, initialize_shard may be misleading. We can rename things as needed.

go/multiorch/actions/bootstrap_shard.go
deepthi: That is definitely an option. It will probably mean that the initialization process takes longer because multiorch has to detect and repair replication on the new standbys instead of it all happening during bootstrap.
Vitess took the other route - when a replica starts up, replication is initialized without waiting for vtorc to fix it. At all other times, vtorc fixes replication.

go/multiorch/actions/bootstrap_shard.go
deepthi: The code as written today is correct, because that is the way to bootstrap a standby from a primary. However, this will change when we fix the TODO we have to move to pgBackRest.
The way pg_basebackup works is that the backup gets written to the PGDATA directory of the target. In order to reuse that for multiple standbys, you would have to make it write it somewhere else, and copy from there to each standby. It is not worth doing that with pg_basebackup when we are going to replace it with pgBackRest. So we leave this as-is for now, and the PR that replaces it with pgBackRest will do it properly.

go/multiorch/actions/initialize_shard.go
deepthi: I do think that we will need to extract the Repair code out because it is also what we will need to run when multiorch detects a DeadPrimary.

deepthi:
Given the feedback and the dependencies, the best way forward is probably to wait until we have the basic discovery, analysis and recovery loops completed with a placeholder for the recovery for ShardHasNoPrimary. At that point, we can rework this PR to integrate properly with multiorch.
If we were to limit this PR to shard initialization right now, we'd end up deleting approximately two-thirds of the code, when in fact we can use that code for Planned switchovers and Unplanned failovers.

go/multiorch/actions/bootstrap_shard.go
deepthi: This is a great question, I'm glad you asked.
As it happens, `InitializeEmptyPrimary` only creates the `heartbeat` table. And both `heartbeat` and `durability_policy` are created in `CreateSidecarSchema`.
We should actually create all sidecar tables in `InitializeEmptyPrimary` and once we hook up multiorch to be able to do shard_init, we should completely remove that call to `CreateSidecarSchema` from multipooler's `Open`.
We will also need to rework the tests to not create any primary database at all, instead, all poolers come up as standbys and multiorch selects a primary.
tl;dr we should call `CreateSidecarSchema` in `InitializeAsPrimary`.

go/multiorch/actions/initialize_shard.go
rafael: >The idea is that multiorch detects ClusterHasNoPrimary, and it may not know whether everything has already been initialized. ClusterHasNoPrimary repair can do the right thing depending on initialization status.

I'm missing something. In my mental model, MultiOrch will have a repair that is very specific for cluster not being initialized. From the perspective of this code, determining which kind of repair needs to happen is outside this context. 

Otherwise, it seems that this code is also competing with the responsibilities of the analyze loop.

deepthi: s/ClusterHasNoPrimary/ShardHasNoPrimary

deepthi: I think this is all the more reason to wait until we have the analyze loop completed before we continue on this PR. That gives me time to think about which way we want to go. At least in Vitess, you could call `PlannedReparentShard` on a brand new cluster and it would do the right thing BUT `ClusterHasNoPrimary` is a distinct analysis type which is detected only when a cluster never had a primary.

go/multiorch/actions/bootstrap_shard.go
deepthi: I pushed a commit to remove the duplication and only have `CreateSidecarSchema`. The rest we will do when we do the integration / refactor of this PR.

go/multiorch/actions/bootstrap_shard.go
deepthi: Going deeper into this, here's a part of the reason why
  - VTOrc needs to discover the tablet first through its polling mechanism
  - A brand new tablet that just started isn't yet in VTOrc's view
  - The tablet needs to be in a valid initial state before VTOrc can even analyze it
  
In fact, initializeReplication can fail, and then we fail tablet startup, so it seems to make sense to treat bootstrap/init as a special case and not wait for multiorch.

go/multiorch/actions/initialize_shard.go
deepthi: Based on discussion and thinking, I have split this up into two distinct parts, one that is intended to be used for bootstrap, and another that can be used in non-bootstrap situations.

go/multipooler/manager/rpc_initialization.go
```diff
@@ -81,8 +81,8 @@ func (pm *MultiPoolerManager) InitializeEmptyPrimary(ctx context.Context, req *m
 		return nil, mterrors.Wrap(err, "failed to connect to database")
 	}
 
-	// 6. Create multigres schema and heartbeat table
-	if err := pm.initializeMultigresSchema(ctx); err != nil {
+	// 6. Create multigres schema and tables (heartbeat, durability_policy)
```
rafael, line 84: do we have test coverage making sure both tables are created as part of this RPC?

deepthi: I spent an hour trying to write an e2e test for this. We need the initialization of standbys from a base backup before we can get that to work. So it will be a follow up PR after we have that work done.

PR #253, opened 2025-11-19
refactor(otel): pull metric definitions out of instrumented code + lint check

dweitzman:
This PR does a few things:
- Use a ruleguard rule to try to discourage creating metrics outside of files name `*metrics.go`
- Use another ruleguard rule to try to discourage `NewMetrics()` from taking any arguments
- Try to make existing metric definitions comply better with https://opentelemetry.io/docs/specs/semconv/general/naming/ conventions
- Resolve the newly-added linter rules by separating metric definitions from usage

### DIFF
--- a/go/multipooler/rpcclient/metrics.go
+++ b/go/multipooler/rpcclient/metrics.go
@@ -61,8 +61,6 @@ type DialDuration struct {
 }
 
 // Record records a dial operation duration with proper OTel attributes.
-// The path parameter is automatically converted to the correct attribute key
-// internally.
 //
 // Parameters:
 //   - ctx: Context for the metric recording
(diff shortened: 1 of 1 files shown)


go/multiorch/recovery/engine.go
```diff
@@ -217,7 +217,19 @@ func NewEngine(
...
 	// Initialize metrics
-	engine.metrics = NewMetrics(logger, poolerStore)
+	var err error
+	engine.metrics, err = NewMetrics()
+	if err != nil {
+		logger.Error("failed to initialize recovery metrics", "error", err)
+	}
+
+	// Register callback for pooler store size observable gauge
+	err = engine.metrics.RegisterPoolerStoreSizeCallback(func() int {
```
rafael, line 227: Like this. Way cleaner

PR #254, opened 2025-11-19
fix: undo a metric name change

dweitzman:
"pooler poll" and "health check cycle" are nouns, so they can be a single component between dots


rafael (approved):
LGTM

deepthi (approved):

PR #255, opened 2025-11-19
feat(multiorch): Perform actual RPC calls during healthcheck

rafael:
## Description

Implements real health-check RPC calls to multipooler instances, replacing the placeholder stub methods with actual Status API integration. This PR also includes a few small refactors for consistent pooler-key handling.

## Health-Check Implementation 

We have real health checking via a unified `Status` RPC:

* Calls `re.rpcClient.Status()` to retrieve actual health information from multipooler instances.
* Uses **one** Status RPC endpoint for both **PRIMARY** and **REPLICA** poolers.
* Extracts metrics:
  * **Primary:** LSN, readiness, connected followers, synchronous-replication config.
  * **Replica:** replay LSN, receive LSN, lag metrics, WAL replay pause, primary connection details.
* Stores the reported pooler type so we can detect mismatches between topology and reality.
* Handles RPC timeouts (5s) and all error cases cleanly.

#### 1. RPC client integration

* The recovery engine now receives `rpcclient.MultiPoolerClient` at construction time.
* Connection pool limit set to **1000** concurrent connections.
* Wired up in production (`init.go`) and all relevant tests.

#### 2. Standardized pooler-key format

* All test mocks now use `topo.MultiPoolerIDString()` for consistent key generation.
* Ensures `FakeClient` responses match the expected ID format.
* Updates affect `leader_appointment_test.go` and `policy_test.go`.

#### 3. Improved context handling

* `pollPooler` now accepts a context so cancellation propagates correctly (cc: @dweitzman )
* Integrates better with engine lifecycle and graceful shutdown mechanics.

#### 4. Queue API improvement 
*  Refactored `Consume()` to return a closure-based release function, eliminating the need for a separate` Release()` method.

## Test Coverage

* All existing tests pass with the new RPC plumbing.
* Added new tests to validate the RPC call updates the store.
* No behavioral changes to test assertions, only mock setup updated.

### DIFF
--- a/go/multiorch/store/pooler_info.go
+++ b/go/multiorch/store/pooler_info.go
@@ -43,7 +43,7 @@ type PoolerHealth struct {
 	// Health status from Status RPC (populated after successful health check)
 	// This is the type the pooler reports itself as, which may differ from
 	// the topology type if there's a failover in progress or type mismatch.
-	ReportedType clustermetadata.PoolerType
+	PoolerType clustermetadata.PoolerType
 
 	// Primary-specific fields (populated when ReportedType == PRIMARY)
 	PrimaryLSN                string                                                        // Current WAL LSN position (PostgreSQL format: X/XXXXXXXX)
--- a/go/multiorch/recovery/healthcheck.go
+++ b/go/multiorch/recovery/healthcheck.go
@@ -175,7 +175,7 @@ func (re *Engine) pollPooler(ctx context.Context, poolerID *clustermetadata.ID, 
 		LastSeen:            successTime,
 		IsUpToDate:          true,
 		IsLastCheckValid:    true,
-		ReportedType:        statusResp.Status.PoolerType,
+		PoolerType:          statusResp.Status.PoolerType,
 	}
 
 	// Populate type-specific fields based on what the pooler reports
--- a/go/multiorch/recovery/healthcheck_rpc_test.go
+++ b/go/multiorch/recovery/healthcheck_rpc_test.go
@@ -113,7 +113,7 @@ func TestPollPooler_UpdatesStore_Primary(t *testing.T) {
 	require.False(t, updated.LastCheckSuccessful.IsZero(), "LastCheckSuccessful should be set")
 
 	// Check that PRIMARY-specific fields were populated
-	require.Equal(t, clustermetadata.PoolerType_PRIMARY, updated.ReportedType, "should report PRIMARY type")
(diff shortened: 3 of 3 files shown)


go/multiorch/store/pooler_info.go
```diff
@@ -37,4 +39,24 @@ type PoolerHealth struct {
 	// Computed fields (cached)
 	IsUpToDate       bool
 	IsLastCheckValid bool
+
+	// Health status from Status RPC (populated after successful health check)
+	// This is the type the pooler reports itself as, which may differ from
+	// the topology type if there's a failover in progress or type mismatch.
+	ReportedType clustermetadata.PoolerType
```
deepthi: We should just call this PoolerType.

deepthi (approved):
LGTM except for that one name thing.
You probably realized this as well but when we get to real integration tests, we need to either write pooler_type into topo as REPLICA (not UNKNOWN), or pass an `init-pooler-type` flag to pooler startup. Otherwise the poolers never get out of UNKNOWN type.

go/multiorch/store/pooler_info.go
```diff
@@ -37,4 +39,24 @@ type PoolerHealth struct {
 	// Computed fields (cached)
 	IsUpToDate       bool
 	IsLastCheckValid bool
+
+	// Health status from Status RPC (populated after successful health check)
+	// This is the type the pooler reports itself as, which may differ from
+	// the topology type if there's a failover in progress or type mismatch.
+	PoolerType clustermetadata.PoolerType
+
+	// Primary-specific fields (populated when ReportedType == PRIMARY)
```
deepthi, line 48: Not a big deal but the rename did not change comments.

go/multiorch/store/pooler_info.go
```diff
@@ -37,4 +39,24 @@ type PoolerHealth struct {
...
+	PrimaryLSN                string                                                        // Current WAL LSN position (PostgreSQL format: X/XXXXXXXX)
+	PrimaryReady              bool                                                          // Whether server is accepting connections
+	PrimaryConnectedFollowers []*clustermetadata.ID                                         // Follower servers currently connected via replication
+	PrimarySyncConfig         *multipoolermanagerdatapb.SynchronousReplicationConfiguration // Sync replication config
+
+	// Replica-specific fields (populated when ReportedType == REPLICA)
```
deepthi, line 54: This is the other place.

PR #256, opened 2025-11-20
fix(shard init): various bugs in initialization

deepthi:
Added an e2e test for bootstrap and fixed various bugs
- createDurabilityPolicy function has been replaced with CreateDurabilityPolicy RPC call. coordinator cannot directly connect to the postgres database of poolers
- CreateDurabilityPolicy RPC needed to be implemented
- Needed to add new RPCs to the service
- IntiailizeAsStandby was calling SetPrimaryConnInfo which blocks, so refactor it into setPrimaryConnInfoLocked and call that
- InitializationStatus needs to hold the actionLock
- InitializeAsStandby should set poolerType to REPLICA

### DIFF
--- a/go/test/endtoend/bootstrap_test.go
+++ b/go/test/endtoend/bootstrap_test.go
@@ -30,8 +30,6 @@ import (
 	_ "github.com/lib/pq"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/multigres/multigres/go/clustermetadata/topo"
 	"github.com/multigres/multigres/go/clustermetadata/topo/etcdtopo"
@@ -41,7 +39,6 @@ import (
 	"github.com/multigres/multigres/go/test/utils"
 
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
-	multipoolermanagerpb "github.com/multigres/multigres/go/pb/multipoolermanager"
 	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 
 	// Register topo plugins
@@ -399,26 +396,31 @@ func createEmptyNode(t *testing.T, baseDir, cell, shard, database string, index 
 	}
 }
 
+// createMultiPoolerProto creates a MultiPooler proto for the given gRPC port
+func createMultiPoolerProto(grpcPort int) *clustermetadatapb.MultiPooler {
+	return &clustermetadatapb.MultiPooler{
+		Hostname: "localhost",
+		PortMap: map[string]int32{
+			"grpc": int32(grpcPort),
+		},
+	}
+}
+
 // waitForMultipoolerReady polls the multipooler gRPC endpoint until it's ready
 func waitForMultipoolerReady(t *testing.T, grpcPort int, timeout time.Duration) {
 	t.Helper()
 
-	// Create a single gRPC client connection instead of recreating in the loop
-	conn, err := grpc.NewClient(
-		fmt.Sprintf("localhost:%d", grpcPort),
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
-	)
-	require.NoError(t, err, "Failed to create gRPC client")
-	defer conn.Close()
+	client := rpcclient.NewMultiPoolerClient(10) // Small cache for test connections
+	defer client.Close()
 
-	client := multipoolermanagerpb.NewMultiPoolerManagerClient(conn)
+	pooler := createMultiPoolerProto(grpcPort)
 
-	// Use require.Eventually to poll the RPC call rather than recreating connections
+	// Use require.Eventually to poll the RPC call with connection caching
 	require.Eventually(t, func() bool {
 		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
 		defer cancel()
 
-		_, err := client.InitializationStatus(ctx, &multipoolermanagerdatapb.InitializationStatusRequest{})
+		_, err := client.InitializationStatus(ctx, pooler, &multipoolermanagerdatapb.InitializationStatusRequest{})
 		return err == nil
 	}, timeout, 200*time.Millisecond, "Multipooler at port %d did not become ready", grpcPort)
 }
@@ -474,18 +476,15 @@ func cleanupNode(t *testing.T, node *nodeInstance) {
 func checkInitializationStatus(t *testing.T, node *nodeInstance) *multipoolermanagerdatapb.InitializationStatusResponse {
 	t.Helper()
 
-	conn, err := grpc.NewClient(
-		fmt.Sprintf("localhost:%d", node.grpcPort),
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
-	)
-	require.NoError(t, err)
-	defer conn.Close()
+	client := rpcclient.NewMultiPoolerClient(10) // Small cache for test connections
+	defer client.Close()
 
-	client := multipoolermanagerpb.NewMultiPoolerManagerClient(conn)
+	pooler := createMultiPoolerProto(node.grpcPort)
+
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	resp, err := client.InitializationStatus(ctx, &multipoolermanagerdatapb.InitializationStatusRequest{})
+	resp, err := client.InitializationStatus(ctx, pooler, &multipoolermanagerdatapb.InitializationStatusRequest{})
 	require.NoError(t, err)
 
 	return resp
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
         with:
           persist-credentials: false
-      - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+      - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
         with:
           go-version: stable
       - name: golangci-lint
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -33,7 +33,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Go
-        uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+        uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
         with:
           go-version: "1.25.0"
 
--- a/.github/workflows/test-short.yml
+++ b/.github/workflows/test-short.yml
@@ -33,7 +33,7 @@ jobs:
           persist-credentials: false
 
       - name: Set up Go
-        uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+        uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
         with:
           go-version: "1.25.0"
 


go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+			)
+		`).Scan(&tableExists)
+		require.NoError(t, err, "Should query heartbeat table existence")
+		assert.True(t, tableExists, "Heartbeat table should exist")
+
+		t.Logf("Verified heartbeat table exists in multigres schema")
+	})
+
+	t.Run("verify standbys initialized", func(t *testing.T) {
+		t.Skip("Skipping standby initialization verification - backup not yet implemented in bootstrap")
```
deepthi, line 300: @cuongdo this is the test that should start passing when we integrate backups into standby initialization.

go/multiorch/actions/bootstrap_shard.go
```diff
@@ -191,6 +203,16 @@ func (a *BootstrapShardAction) initializeStandbys(ctx context.Context, shardID s
...
+			// Set pooler type to REPLICA before initializing as standby
+			changeTypeReq := &multipoolermanagerdatapb.ChangeTypeRequest{
+				PoolerType: clustermetadatapb.PoolerType_REPLICA,
+			}
+			_, err := node.RpcClient.ChangeType(ctx, node.Pooler, changeTypeReq)
+			if err != nil {
+				results <- result{node: node, err: fmt.Errorf("failed to set pooler type: %w", err)}
+				return
+			}
+
```
deepthi, line 215: @rafael this is another way to solve the poolerType problem. We can debate whether we really need a `--init-pooler-type` flag to multipooler if the bootstrap is handling it already.
We will definitely need it when we introduce more pooler types.

go/multipooler/manager/db.go
```diff
@@ -40,7 +40,7 @@ func CreateDBConnection(logger *slog.Logger, config *Config) (*sql.DB, error) {
 		socketDir := filepath.Join(config.PoolerDir, "pg_sockets")
 		port := fmt.Sprintf("%d", config.PgPort)
 
-		dsn = fmt.Sprintf("user=postgres dbname=%s host=%s port=%s sslmode=disable",
+		dsn = fmt.Sprintf("user=postgres dbname=%s host=%s port=%s sslmode=disable connect_timeout=2",
```
deepthi, line 51: This was added during debugging. What do people feel about it? It is reasonable when postgres is local, but not when we get to `unmanaged` mode in the future.

go/multipooler/manager/rpc_manager.go
```diff
@@ -95,8 +96,19 @@ func (pm *MultiPoolerManager) SetPrimaryConnInfo(ctx context.Context, host strin
...
+	return pm.setPrimaryConnInfoLocked(ctx, host, port, stopReplicationBefore, startReplicationAfter)
+}
+
+// setPrimaryConnInfoLocked sets the primary connection info for a standby server.
+// This function assumes the action lock is already held by the caller.
+func (pm *MultiPoolerManager) setPrimaryConnInfoLocked(ctx context.Context, host string, port int32, stopReplicationBefore, startReplicationAfter bool) error {
+	if err := pm.checkReady(); err != nil {
+		return err
+	}
+
```
deepthi, line 113: We need to do checkReady in both versions (locked and otherwise), otherwise tests fail.

go/multipooler/manager/rpc_initialization.go
```diff
@@ -322,11 +331,39 @@ func (pm *MultiPoolerManager) waitForDatabaseConnection(ctx context.Context) err
 		if err := pm.db.PingContext(ctx); err == nil {
 			return nil
 		}
+		// Close stale connection
+		pm.db.Close()
```
dweitzman, line 336: Hypothetically couldn't this be a context.DeadlineExceeded or context.Canceled that doesn't necessarily prove the db connection is bad?

go/multipooler/manager/rpc_initialization.go
```diff
@@ -322,11 +331,39 @@ func (pm *MultiPoolerManager) waitForDatabaseConnection(ctx context.Context) err
...
-	// TODO: Implement retry logic
-	// For now, assume the database will be available soon
-	return nil
+
+	// Retry for up to 30 seconds
+	maxRetries := 60
+	retryDelay := 500 * time.Millisecond
+
+	var lastErr error
+	for i := 0; i < maxRetries; i++ {
```
dweitzman: If you wanted to use the retry with backoff library here, it might look like this:
```go
r := retry.New(1 * time.Millisecond, 500 * time.Millisecond)
for _, err := range r.Attempts(timeoutCtx) {
    if err != nil {
        return mterrors.Wrap(err, "context ended while waiting for database connection")
    }
    // Try to open the connection
}
```

I'm not sure the logic to stop after 60 attempts is worth it -- seems fine to either do a WithTimeout() to reduce the context length or trust that the caller knows how long they're willing to wait?

go/multipooler/manager/rpc_manager.go
```diff
@@ -95,8 +96,19 @@ func (pm *MultiPoolerManager) SetPrimaryConnInfo(ctx context.Context, host strin
...
 	}
 
+	// Call the locked version that assumes action lock is already held
+	return pm.setPrimaryConnInfoLocked(ctx, host, port, stopReplicationBefore, startReplicationAfter)
+}
+
+// setPrimaryConnInfoLocked sets the primary connection info for a standby server.
+// This function assumes the action lock is already held by the caller.
+func (pm *MultiPoolerManager) setPrimaryConnInfoLocked(ctx context.Context, host string, port int32, stopReplicationBefore, startReplicationAfter bool) error {
+	if err := pm.checkReady(); err != nil {
```
dweitzman, line 110: Could make this method call `AssertActionLockHeld()` for defense in depth?

go/multipooler/manager/rpc_manager.go
```diff
@@ -1027,3 +1039,71 @@ func (pm *MultiPoolerManager) SetTerm(ctx context.Context, term *multipoolermana
...
+	quorumRuleJSON, err := marshaler.Marshal(req.QuorumRule)
+	if err != nil {
+		return &multipoolermanagerdatapb.CreateDurabilityPolicyResponse{
+			Success:      false,
+			ErrorMessage: fmt.Sprintf("failed to marshal quorum rule: %v", err),
+		}, nil
+	}
+
+	// Insert the policy into the durability_policy table
+	_, err = pm.db.ExecContext(ctx, `
```
dweitzman: Not a blocker for the PR, but I think it might be nice to pull all the SQL queries we make into separate helpers of some sort so it's easy to find/audit/change all the SQL queries related to multigres access or changing its own metadata

go/multipooler/manager/db.go
dweitzman: What inspired adding connect_timeout here?

Looking a bit into `sql.Open()` and `lib/pq`, it feels like the Go sql API doesn't leave a lot of good options here. Looks like it opens connections in serial with a `context.Background()` so if a connect attempt hangs it both blocks other attempts from starting and might not ever get cleaned up even when you call `db.Close()`?...

If I'm reading this correctly... all the options seem bad, but setting a connect_timeout to something plausible like 2 seconds might be less bad than omitting it or using a much bigger value

So I guess, no objections from me to `connect_timeout=2` but I feel like a comment with some explanation would be worthwhile.

go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+	pgctldProcess  *exec.Cmd
+	multipoolerCmd *exec.Cmd
+}
+
+func TestBootstrapInitialization(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping end-to-end bootstrap test in short mode")
+	}
+
+	if utils.ShouldSkipRealPostgres() {
```
dweitzman, line 61: `ShouldSkipRealPostgres()` checks of postgres binaries present and `testing.Short()` being true, so if we passed the earlier check for `Short()` above then `ShouldSkipRealPostgres()` will always be false here

go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+		t.Skip("Skipping PostgreSQL test - no postgres binaries")
+	}
+
+	// Require etcd binary
+	_, err := exec.LookPath("etcd")
+	require.NoError(t, err, "etcd binary must be available in PATH")
+
+	// Set PATH for binaries
+	err = pathutil.PrependBinToPath()
```
dweitzman: I think TestMain() should already be configuring the `PATH`?...

go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+	require.NoError(t, err, "Failed to add bin to PATH")
+
+	ctx := context.Background()
```
dweitzman: Can this be `t.Context()`? The only caveat would be that `t.Context()` is canelled before `Cleanup()` runs

More thoughts on context that aren't critical to this PR:
I have an aspiration to remove the vast majority of `context.Background()` calls from the codebase, with most of them either being replaced by parent context (potentially inherited from a Cobra command or `testing.T`), `context.TODO()` if we haven't finished the code, and only `context.Background()` in particularly special cases like test cleanup or code that's genuinely unable to inherit parent context

My reasons for wanting to minimize `context.Background()` are:
- inherited context will make telemetry better
- inherited context will make handling cancellation and deadlines better

go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+	t.Logf("Bootstrap test directory: %s", tempDir)
+
+	// Start etcd
+	etcdDataDir := filepath.Join(tempDir, "etcd_data")
+	require.NoError(t, os.MkdirAll(etcdDataDir, 0o755))
+
+	etcdClientAddr, etcdCmd := startEtcd(t, etcdDataDir)
+	defer func() {
+		if etcdCmd != nil && etcdCmd.Process != nil {
+			_ = etcdCmd.Process.Kill()
```
dweitzman: Not a blocking comment: I think we should have a helper for killing processes for places like that that gives them a chance to respond to a sigterm or something like that before a kill

My worry is that some of the helpers like `run_in_test.sh` for trying to prevent orphan processes after endtoend tests may not be able to detect and handle a kill signal since it can't be intercepted

Seems worthwhile as a general rule to give TERM a chance before escalating to KILL, and we can have a helper for that

go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+	logger := slog.Default()
+	bootstrapAction := actions.NewBootstrapShardAction(ts, logger)
+
+	err = bootstrapAction.Execute(ctx, shardID, database, coordNodes)
+	require.NoError(t, err, "Bootstrap action should succeed")
+
+	t.Logf("Bootstrap action completed successfully")
+
+	// Wait for initialization to complete
+	time.Sleep(2 * time.Second)
```
dweitzman, line 178: FWIW, my personal preference is to favor require.Eventually() over time.Sleep() where possible so there's a possibility it doesn't need to take the entire 2 seconds and tests could potentially run faster

go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+
+	// Start pgctld server
+	logFile := filepath.Join(dataDir, "pgctld.log")
+	pgctldCmd := exec.Command("pgctld", "server",
+		"--pooler-dir", dataDir,
+		"--grpc-port", fmt.Sprintf("%d", pgctldGrpcPort),
+		"--pg-port", fmt.Sprintf("%d", pgPort),
+		"--log-output", logFile)
+
+	// Set MULTIGRES_TESTDATA_DIR for directory-deletion triggered cleanup
```
dweitzman, line 346: No action required, just making an observation: with the logic above to avoid deleting this directory if a test fails, there's a greater chance of orphan processes outlasting any given test case finishing

go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+}
+
+// waitForMultipoolerReady polls the multipooler gRPC endpoint until it's ready
+func waitForMultipoolerReady(t *testing.T, grpcPort int, timeout time.Duration) {
+	t.Helper()
+
+	deadline := time.Now().Add(timeout)
+	var lastErr error
+	for time.Now().Before(deadline) {
+		conn, err := grpc.NewClient(
```
dweitzman: Seems like it shouldn't be necessary to re-create the client in a loop? Feels like if there's an error here it'd be more likely to be a permanent error with the parameters

Could this just make a single client and do require.Eventually() waiting for the call to InitializationStatus() to succeed?

Seems like there's also a lot of overlap here with `checkInitializationStatus()`, so an alternative would be to make that return an error instead of doing a require assertion and reuse it anywhere that needs to connect and fetch the initialization status

go/test/endtoend/bootstrap_test.go
```diff
@@ -0,0 +1,549 @@
...
+	require.NoError(t, err, "Failed to open database connection")
+
+	err = db.Ping()
+	require.NoError(t, err, "Failed to ping database")
+
+	return db
+}
+
+// startEtcd starts an etcd instance for testing
+func startEtcd(t *testing.T, dataDir string) (string, *exec.Cmd) {
```
dweitzman: Not blocking: seems like we probably have more versions of `startEtcd()` in the codebase than we need. Wonder if we can consolidate some shared test helpers for that kind of thing?

dweitzman (approved):
Lots of small comments, but happy to defer to your judgement on whether and how to resolve them

go/multipooler/manager/rpc_initialization.go
deepthi: Yes, it could. We are being lazy about this until connection pooling is in place. That will replace most of db.go.

go/test/endtoend/bootstrap_test.go
deepthi: I'm going to revert this.

go/test/endtoend/bootstrap_test.go
deepthi: I happen to agree with you that context.Background() is evil :)

go/test/endtoend/bootstrap_test.go
deepthi: Yes, I would have liked to reuse setup_test.go but it's in a different package and I didn't want to import it.
However, we already have a test_helpers.go that I'm going to use instead.

deepthi:
@dweitzman thank you for the detailed review. I have implemented most of your suggestions.

go/test/endtoend/bootstrap_test.go
deepthi: I did both refactors, and for good measure started using multipooler.rpcClient as the client, instead of creating a vanilla gRPC client.

deepthi:
Merging this to unblock others. We can debate `poolerType` and `checkReady` and incorporate any changes into a future PR.

PR #257, opened 2025-11-20
build(coverage): Report code coverage for integration test runs

dweitzman:
Output at https://coveralls.io/github/multigres/multigres

It may be theoretically possible to track lines covered by both the tests themselves and also subprocesses they execute, but it seems like it'd be a mess. You can't just use the `GOCOVERDIR` environment variable with `go test`, so getting full coverage might involve something crazy like compiling all the test binaries manually and invoking them with both `GOCOVERDIR` and `-test.coverprofile`. It's maybe not impossible, but the script to run tests would be a mess.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -94,7 +94,7 @@ jobs:
           reporter: golang-json
 
       - name: Upload coverage artifact
-        uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
         if: always()
         with:
           name: coverage-data
@@ -114,7 +114,6 @@ jobs:
           name: coverage-data
 
       - name: Upload coverage reports to Coveralls
-        if: success() || failure()
         uses: coverallsapp/github-action@643bc377ffa44ace6394b2b5d0d3950076de9f63 # v2.3.0
         with:
           file: ./coverage.txt


.github/workflows/test-integration.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           # Set environment variables for tests
           export PGCONNECT_TIMEOUT=5
           # Run tests with verbose output
-          go test -json -v ./... | tee integration-test-results.jsonl | go tool tparse -follow
+          go test -cover -covermode=atomic -coverprofile=coverage.txt -coverpkg=./... -json -v ./... | tee integration-test-results.jsonl | go tool tparse -follow
```
cuongdo, line 86: Collecting coverage data increases integration test runtime in CI from ~3m to ~4m. That's probably not an issue now but something to revisit later.

.github/workflows/test-integration.yml
```diff
@@ -92,3 +92,30 @@ jobs:
           name: Unit Tests
           path: integration-test-results.jsonl
           reporter: golang-json
+
+      - name: Upload coverage artifact
+        uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
```
cuongdo: do you specifically prefer v4.4.0 over v5.0.0?

.github/workflows/test-integration.yml
```diff
@@ -92,3 +92,30 @@ jobs:
...
+    if: always()
+
+    steps:
+      - name: Download coverage artifact
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: coverage-data
+
+      - name: Upload coverage reports to Coveralls
+        if: success() || failure()
```
cuongdo: what is the purpose of this?

dweitzman: Left over from a previous iteration. Removed

.github/workflows/test-integration.yml
dweitzman: Nope, fixed

.github/workflows/test-integration.yml
dweitzman: My follow-up proposal involves splitting this into a separate workflow, so if we want to do that then this multi-purpose workflow won't last long: https://github.com/multigres/multigres/pull/279

PR #259, opened 2025-11-20
build: verify checksums when downloading tools

dweitzman:
This is better for supply chain safety. We don't want to install an unexpected or untrusted version of etcd, etc.

Also:
- Retry a few times in case of temporary network errors
- Show better error messages when the HTTP status code is not a 2xx


.github/workflows/test-integration.yml
```diff
@@ -67,9 +67,13 @@ jobs:
 
       - name: Install etcd binary
         run: |
-          curl -L https://github.com/etcd-io/etcd/releases/download/v3.5.0/etcd-v3.5.0-linux-amd64.tar.gz | tar xz
-          sudo cp etcd-v3.5.0-linux-amd64/etcd /usr/local/bin/
-          rm -rf etcd-v3.5.0-linux-amd64
+          set -euo pipefail
+          tools/download_tool.sh etcd v3.6.4 linux-amd64 etcd.tar.gz
```
dweitzman, line 71: etcd version upgrade is to match `make`

PR #260, opened 2025-11-20
fix(test): replace counter-based port allocation with OS allocation

dweitzman:
Replaces the counter-based port allocation strategy with OS-based allocation using net.Listen(":0") to eliminate port conflicts in parallel test execution.

Changes:
- Replace GetNextPort()/GetNextEtcd2Port() with GetFreePort(t)
- Use OS to pick free ports instead of PID-based bucketing
- Add optional PeerPort to EtcdConfig for explicit etcd peer ports
- Update all test files to use new GetFreePort(t) helper
- Add ClientPort/PeerPort fields to etcdtopo.EtcdOptions
- Randomize all ports used by cluster_test. Previously it assumed that zone2 port numbers would equal zone1 port numbers plus 100

The goal here is to reduce test flakes. I'm seeing flakes related to ports in CI and one possible explanation is that the previous method of allocating that used a counter might have a higher chance of collisions if multiple tests are running in parallel.

So... I'm not sure if this will decrease the rate of flakes, but maybe it could?

Even if it doesn't, centralizing how we allocate ports for tests seems like it'll give us more flexibility and correctness even if we want to change the port selection algorithm in the future

### DIFF
--- a/go/test/endtoend/bootstrap_test.go
+++ b/go/test/endtoend/bootstrap_test.go
@@ -88,7 +88,9 @@ func TestBootstrapInitialization(t *testing.T) {
 	require.NoError(t, os.MkdirAll(etcdDataDir, 0o755))
 
 	etcdClientAddr, _ := etcdtopo.StartEtcdWithOptions(t, etcdtopo.EtcdOptions{
-		DataDir: etcdDataDir,
+		ClientPort: utils.GetFreePort(t),
+		PeerPort:   utils.GetFreePort(t),
+		DataDir:    etcdDataDir,
 	})
 
 	t.Logf("Started etcd at %s", etcdClientAddr)


go/test/utils/portallocator.go
```diff
@@ -15,51 +15,25 @@
...
-	// Use process ID to create unique port ranges for different test processes
-	// This prevents conflicts when running tests from different packages simultaneously
-	pid := os.Getpid()
-	// Give each test process enough ports for multiple zones. Tests assume that each
-	// zone can use up to 100 ports.
-	//
-	// This generates a max port number of 6700 + (99 * 500) + 200 = 56400 (< 65535)
-	basePortOffset = (pid % 100) * portsPerPid
-}
+	lis, err := net.Listen("tcp", "localhost:0")
```
rafael, line 30: oh I can see how this would behave better than our random approach. 

I like it

deepthi:
💯 I think everyone has encountered random failures with port conflicts, so this is great!

PR #262, opened 2025-11-21
feat(consensus): implement proper LSN comparison and use it to select the best candidate

deepthi:
Key Features:
- Validate LSN format before comparison
- Skip nodes with invalid LSNs gracefully (logs warning, continues evaluation)
- Only select nodes with parseable LSNs
- Use proper numeric comparison for leader selection

### DIFF
--- a/go/multiorch/coordinator/leader_appointment.go
+++ b/go/multiorch/coordinator/leader_appointment.go
@@ -209,7 +209,7 @@ func (c *Coordinator) selectCandidate(ctx context.Context, cohort []*Node) (*Nod
 			continue
 		}
 
-		cmp, err := lsn.CompareStrings(status.walPosition, bestWAL)
+		isGreater, err := lsn.Greater(status.walPosition, bestWAL)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


go/multiorch/coordinator/leader_appointment.go
```diff
@@ -195,7 +194,31 @@ func (c *Coordinator) selectCandidate(ctx context.Context, cohort []*Node) (*Nod
 			continue
 		}
 
-		if bestCandidate == nil || status.walPosition > bestWAL {
+		// Validate LSN format first
+		if _, err := lsn.Parse(status.walPosition); err != nil {
```
rafael, line 198: I've been pondering about this. Under which case could we get a response, but a bad format LSN? This seems like an invariant we can expect to be true.

What do you think if we actually return error in if any LSN is invalid? 

It seems this is the kind of error where we better bail and get an operator to understand why we would be getting LSN's in the response that we can't parse

deepthi:
@rafael pointed out that we should just use an existing library for this instead of rolling our own. I'm going to rework this PR to do so.

deepthi:
It's easier to create a new PR, closing this one.

PR #266, opened 2025-11-21
feat(multiorch): Implement Analysis Loop Infrastructure

rafael:
# Description

This PR introduces the initial analysis loop for automated health checking and recovery in Multigres. It takes inspiration from VTOrc, but made some changes on how recoveries are modeled. Also we did not bring sqlite (at least for now). 

### Core Infrastructure
- **Analysis Generator**  
  Converts raw pooler state into `ReplicationAnalysis` objects. Computes derived fields like replica counts, replication lag, and primary–replica relationships. Uses the in-memory nested map structure (database → tablegroup → shard → pooler) for efficient lookups. **NOTE: this analysis is not super robust yet, it's more like a place holder**
- **Analyzer Interface** .
  A mechanism where analyzers inspect a `ReplicationAnalysis` and return detected `Problem` objects. Keeps problem detection decoupled and easy to extend.
- **Problem Model**  
  Defines a structured representation for issues: `ProblemCode`, priority, scope (cluster vs pooler), and the associated `RecoveryAction`.
- **Recovery Action Interface**  
  Defines the contract for automated recovery operations, including timeout, locking requirements, and priority.

### Recovery Loop
- Runs on a 1-second tick: generates analyses, runs analyzers, detects problems, and triggers recoveries.
- **Problem deduplication rules:**
  - Cluster-wide problems (for example, dead primary) block all other recoveries in the shard.
  - Per-pooler problems are deduped by keeping only the highest-priority one.
- **Priority-based execution**: This is an int enum, we always solve the highest priority first.
- **Pre-recovery validation:** Re-polls affected poolers before executing a recovery to confirm the issue still exists.
- **Scope-aware refresh:** Cluster-wide problems refresh the whole shard; pooler-scoped problems only refresh the pooler and the primary.

### Discovery Enhancements 
- Refactored `refreshClusterMetadata()` to use a reusable `refreshPoolersForTarget()` helper.
- Added `forceHealthCheckShardPoolers()` to re-poll all poolers in a shard when validating recoveries.
- Added support for ignoring specific poolers (for example, skipping a known-dead primary during failover).

### Store Additions
- **`ReplicationAnalysis` **  
  A richer per-pooler structure with the fields needed for analysis, covering primary/replica state, replication topology, and computed flags.
- **DeepCopy for `PoolerHealth` **  
  Enables safe concurrent access during analysis generation. Longer term, the in-memory store should only keep values, since shared pointers make it easy to introduce subtle concurrency bugs.

## Architecture Notes

1. **Separation of concerns:**  
   Analysis generation (describe state) is separate from problem detection (interpret state) and recovery (act on state).
2. **Analyzers:**  
   New checks register implement the interface and don't need changes to the core logic.
3. **Validation before acting:**  
   Re-polls prevent recovery actions from running on stale data.
4. **Tests:**  
   Coverage includes analysis generation, grouping, deduplication, and recovery orchestration. More integration tests can be added later on. 

## What's Not Included (Intentionally Deferred)

- No analyzers yet: `DefaultAnalyzers()` currently returns an empty slice.
- No recovery implementations: the interface is defined; implementations will come in follow-up work.
- No shard locking yet: locking will be added when we introduce concurrent recoveries.

### DIFF
--- a/go/multiorch/recovery/recovery_loop.go
+++ b/go/multiorch/recovery/recovery_loop.go
@@ -302,9 +302,8 @@ func (re *Engine) recheckProblem(problem analysis.Problem) (bool, error) {
 		}
 	}
 
-	// Re-generate analysis for this specific pooler using updated store data
-	// Note: GenerateAnalysisForPooler rebuilds its internal map from the current store state,
-	// so it will see the fresh data from the re-poll above.
+	// Re-generate analysis for this specific pooler using updated store data.
+	// A new generator is created to capture the updated store state from the re-poll above.
 	generator := analysis.NewAnalysisGenerator(re.poolerStore)
 	poolerAnalysis, err := generator.GenerateAnalysisForPooler(poolerIDStr)
 	if err != nil {


go/multiorch/recovery/analysis/generator.go
```diff
@@ -0,0 +1,335 @@
...
+		// Create a deep copy to avoid concurrent access issues
+		poolersByTG[database][tableGroup][shard][poolerID] = pooler.DeepCopy()
+		return true // continue
+	})
+
+	return poolersByTG
+}
+
+// GetPoolersInTableGroup returns all pooler IDs in the same tablegroup as the given pooler.
+// This uses the cached poolersByTG map to avoid iterating the store again.
```
deepthi: This contradicts the code on line 127

go/multiorch/recovery/analysis/generator.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		if poolersByTG[database][tableGroup] == nil {
+			poolersByTG[database][tableGroup] = make(map[string]map[string]*store.PoolerHealth)
+		}
+		if poolersByTG[database][tableGroup][shard] == nil {
+			poolersByTG[database][tableGroup][shard] = make(map[string]*store.PoolerHealth)
+		}
+
+		// Create a deep copy to avoid concurrent access issues
+		poolersByTG[database][tableGroup][shard][poolerID] = pooler.DeepCopy()
```
deepthi: It may be time to import vtprotobuf so that we don't write our own DeepCopy functions all over the place.

go/multiorch/recovery/analysis/generator.go
```diff
@@ -0,0 +1,335 @@
...
+	analysis := g.generateAnalysisForPooler(poolerIDStr, pooler, poolersByTG, database, tableGroup, shard)
+
+	return analysis, nil
+}
+
+// generateAnalysisForPooler creates a ReplicationAnalysis for a single pooler.
+func (g *AnalysisGenerator) generateAnalysisForPooler(
+	poolerID string,
+	pooler *store.PoolerHealth,
+	poolersByTG PoolersByTableGroup,
```
deepthi: Is the primary purpose of grouping by TableGroup to produce metrics? My thinking is
- all recoveries will be either at pooler level or shard level
- Tools like Grafana are the right place to do aggregations, we don't need to generate aggregate metrics, we just need to label them correctly.

go/multiorch/recovery/analysis/recovery.go
```diff
@@ -0,0 +1,50 @@
...
+type RecoveryAction interface {
+	// Execute performs the recovery.
+	Execute(ctx context.Context, problem Problem) error
+
+	// Metadata returns info about this recovery.
+	Metadata() RecoveryMetadata
+
+	// RequiresLock indicates if tablegroup lock is needed.
+	RequiresLock() bool
+
```
deepthi: Why would we need a lock at the TableGroup level?

go/multiorch/recovery/analysis/types.go
```diff
@@ -0,0 +1,90 @@
...
+	// Replica problems (require healthy primary).
+	ProblemReplicaNotReplicating ProblemCode = "ReplicaNotReplicating"
+	ProblemReplicaWrongPrimary   ProblemCode = "ReplicaWrongPrimary"
+	ProblemReplicaLagging        ProblemCode = "ReplicaLagging"
+	ProblemReplicaMisconfigured  ProblemCode = "ReplicaMisconfigured"
+	ProblemReplicaIsWritable     ProblemCode = "ReplicaIsWritable"
+
+	// Non-actionable (detect but don't auto-recover).
+	ProblemPrimaryAndReplicasDead    ProblemCode = "PrimaryAndReplicasDead"
+	ProblemAllReplicasNotReplicating ProblemCode = "AllReplicasNotReplicating"
```
deepthi: This would benefit from a comment explaining why it is not recoverable.

go/multiorch/recovery/analysis/types.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// Priority defines the priority of recovery actions.
+// Higher values = higher priority (will be attempted first).
+type Priority int
+
+const (
+	// PriorityClusterBootstrap is for cluster-wide bootstrap issues like no primary exists.
+	// This is the highest priority - the cluster cannot function at all.
+	PriorityClusterBootstrap Priority = 2000
```
deepthi: So strange to see highest number == highest priority, but it makes sense for expansion if necessary.

go/multiorch/recovery/analysis/types.go
```diff
@@ -0,0 +1,90 @@
...
+	PriorityNormal Priority = 100
+)
+
+// ProblemScope indicates whether a problem affects the whole cluster or just a single pooler.
+type ProblemScope string
+
+const (
+	// ScopeClusterWide indicates the problem affects the entire shard (e.g., primary dead).
+	// Recovery requires refreshing all poolers in the shard and may involve cluster-wide operations like failover.
+	ScopeClusterWide ProblemScope = "ClusterWide"
```
deepthi:
```suggestion
	ScopeShard ProblemScope = "Shard"
```

go/multiorch/recovery/analysis/types.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// ScopeSinglePooler indicates the problem affects only a specific pooler.
+	// Recovery only requires refreshing the affected pooler and potentially the primary.
+	ScopeSinglePooler ProblemScope = "SinglePooler"
```
deepthi:
```suggestion
	ScopePooler ProblemScope = "Pooler"
```

go/multiorch/recovery/analysis/recovery.go
rafael: Good catch. This should be shard level. I will fix

go/multiorch/recovery/analysis/types.go
rafael: Makes sense, I will fix

go/multiorch/recovery/recovery_loop.go
```diff
@@ -0,0 +1,377 @@
...
+	"time"
+
+	"github.com/multigres/multigres/go/clustermetadata/topo"
+	"github.com/multigres/multigres/go/multiorch/recovery/analysis"
+	"github.com/multigres/multigres/go/multiorch/store"
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+)
+
+// TableGroupKey uniquely identifies a tablegroup.
+type TableGroupKey struct {
```
deepthi:
```suggestion
// ShardKey uniquely identifies a shard.
type ShardKey struct {
```

go/multiorch/recovery/recovery_loop.go
```diff
@@ -0,0 +1,377 @@
...
+	for _, problem := range problems {
+		if problem.Scope == analysis.ScopeClusterWide {
+			return true
+		}
+	}
+	return false
+}
+
+// smartFilterProblems applies intelligent filtering to the problem list:
+// - If there's a cluster-wide problem, return only the highest priority cluster-wide problem
```
deepthi:
```suggestion
// - If there's a shard-wide problem, return only the highest priority shard-wide problem
```

go/multiorch/recovery/recovery_loop.go
```diff
@@ -0,0 +1,377 @@
...
+// - Otherwise, deduplicate by pooler ID, keeping only the highest priority problem per pooler
+//
+// IMPORTANT: Input must already be sorted by priority (highest first).
```
deepthi: How are we guaranteeing this? It may be safer to push lines 121-124 into this function and call it `filterAndPrioritize`. I tend to be allergic to function names with `smart` in them 😏

go/multiorch/recovery/recovery_loop.go
```diff
@@ -0,0 +1,377 @@
...
+	// Post-recovery refresh
+	// If we ran a cluster-wide recovery, force refresh all poolers in the shard
+	// to ensure they have up-to-date state and prevent re-queueing the same problem.
+	if problem.Scope == analysis.ScopeClusterWide {
+		re.logger.InfoContext(re.ctx, "forcing refresh of all poolers post recovery",
+			"database", problem.Database,
+			"tablegroup", problem.TableGroup,
+			"shard", problem.Shard,
+		)
+		re.forceHealthCheckShardPoolers(context.Background(), problem.Database, problem.TableGroup, problem.Shard, nil /* poolersToIgnore */)
```
deepthi, line 249: It may be useful to store a ShardKey in the problem instead of 3 separate fields, and use it and pass it everywhere.

go/multiorch/recovery/recovery_loop.go
```diff
@@ -0,0 +1,377 @@
...
+
+// validateProblemStillExists force re-polls the pooler and re-runs analysis
+// to check if the problem still exists.
+//
+// The validation strategy depends on the problem scope:
+// - ClusterWide: Refresh shard metadata + force poll all poolers in shard (except dead ones)
+// - SinglePooler: Only refresh the affected pooler + primary pooler
+//
+// Returns (stillExists bool, error).
+func (re *Engine) validateProblemStillExists(problem analysis.Problem) (bool, error) {
```
deepthi:
```suggestion
func (re *Engine) recheckProblem(problem analysis.Problem) (bool, error) {
```

go/multiorch/store/pooler_info.go
```diff
@@ -17,19 +17,27 @@ package store
...
+	// Core pooler metadata (from topology)
+	ID             *clustermetadatapb.ID
+	Database       string
+	TableGroup     string
+	Shard          string
+	KeyRange       *clustermetadatapb.KeyRange
+	Hostname       string
+	PortMap        map[string]int32
+	TopoPoolerType clustermetadatapb.PoolerType          // Topology type (PRIMARY or REPLICA)
+	ServingStatus  clustermetadatapb.PoolerServingStatus // Current serving status
```
deepthi, line 40: Move lines 47-49 here to keep everything other than the core pooler fields alphabetical?

go/multiorch/store/pooler_info.go
```diff
@@ -17,19 +17,27 @@ package store
...
 
-	"github.com/multigres/multigres/go/pb/clustermetadata"
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 )
 
 // PoolerHealth represents runtime state of a MultiPooler instance.
 // This stores:
-// - The MultiPooler record from topology
+// - Core pooler metadata (flattened from MultiPooler proto)
```
deepthi, line 26: We discussed this and the idea of avoiding deep copies makes sense, but this is definitely more verbose. I assume we still have a pointer to the same ID, for instance, because we don't expect that to change?

go/multiorch/store/pooler_info.go
```diff
@@ -60,3 +68,38 @@ type PoolerHealth struct {
...
+		Database:       mp.Database,
+		TableGroup:     mp.TableGroup,
+		Shard:          mp.Shard,
+		KeyRange:       mp.KeyRange,
+		Hostname:       mp.Hostname,
+		PortMap:        mp.PortMap,
+		TopoPoolerType: mp.Type,
+		ServingStatus:  mp.ServingStatus,
+	}
+}
```
deepthi, line 105: This is all verbosity introduced by the flattening.

go/multiorch/recovery/analysis/generator.go
rafael: This was poorly name. The idea was grouping by shard to make it easy to perform shard lookups. I'm cleaning all this

go/multiorch/recovery/recovery_loop.go
rafael: Agree this is a good idea. Would it be ok to address this is a follow up PR?

go/multiorch/recovery/analysis/types.go
rafael: Done

go/multiorch/recovery/analysis/types.go
rafael: I updated to 10000 and left a comment on the rationale

go/multiorch/recovery/recovery_loop.go
rafael: `smart` as a name is too `smart` 🙈   . Fixed

go/multiorch/store/pooler_info.go
rafael: Correct, we do not expect the ID to change. One option here would be to introduce small Go structs that mirror the protobuf definitions, but as plain Go types. That would let us keep some structure instead of having everything completely flat, while still avoiding deep copies.

What do you think?

go/multiorch/recovery/analysis/generator.go
rafael: We talked about this. For the context of MultiOrc we are going to try to use pure golang structs to not have the need of doing deep copies. I'm in the process of doing this refactor, but that will be a follow up. That way we can remove this copy altogether.

go/multiorch/store/pooler_info.go
deepthi: hmm, that seems to be just a different kind of duplication that we'll need to keep track of. The advantage of having a pointer to a MultiPoolerInfo protobuf is that if we decide to add a field to it, we get it for free in the PoolerHealth struct. Whereas if we have a duplicate struct for it, then we need to go modify that by hand. Unless you are thinking about generating the matching structs, which is definitely a possibility.

deepthi: Btw, the current flattening also has this disadvantage.

rafael: Yeah, I agree of the downsides of the current approach as is. Using the proto makes more maintainable, but it has the issue I was trying to address. 

I was not thinking about auto generating files, but it seems like the best of both worlds is that approach.

go/multiorch/recovery/analysis/generator.go
```diff
@@ -0,0 +1,322 @@
...
+
+	// Generate analysis for this specific pooler using the cached poolersByShard.
+	// Note: If fresh data is needed (e.g., after re-polling), create a new AnalysisGenerator.
+	analysis := g.generateAnalysisForPooler(pooler, pooler.Database, pooler.TableGroup, pooler.Shard)
+
+	return analysis, nil
+}
+
+// generateAnalysisForPooler creates a ReplicationAnalysis for a single pooler.
+func (g *AnalysisGenerator) generateAnalysisForPooler(
```
rafael, line 162: In general, this is just a first pass over this analysis. This needs an iteration just focused on this. 

We can take this is as a very very naive starting point

go/multiorch/recovery/engine.go
```diff
@@ -193,6 +256,9 @@ type Engine struct {
 	// Context for shutting down loops
 	ctx    context.Context
 	cancel context.CancelFunc
+
+	// WaitGroup to track goroutines for graceful shutdown
+	wg sync.WaitGroup
```
rafael, line 261: The wait group was added to allow graceful shutdown. This was a race discovered by running the tests with `race`

go/multiorch/recovery/recovery_loop.go
```diff
@@ -0,0 +1,366 @@
...
+	// Group problems by shard
+	problemsByShard := re.groupProblemsByShard(problems)
+
+	// Process each shard independently in parallel
+	var wg sync.WaitGroup
+	for shardKey, shardProblems := range problemsByShard {
+		wg.Add(1)
+		go func(key analysis.ShardKey, problems []analysis.Problem) {
+			defer wg.Done()
+			re.processShardProblems(key, problems)
```
rafael, line 81: This is probably not thread safe. There are concurrent accesses to Problems that are not protected.

deepthi (approved):
I'm ok with merging this PR as it stands now and handling all outstanding feedback in follow up PRs.

PR #269, opened 2025-11-21
test(portallocator): stronger defense against port reuse

dweitzman:
It's not impossible for the OS to return the same ephemeral port multiple times in a row if you call GetFreePort() without actually using the port, so as better protection we can make sure we never return the same port twice within the same test case

There's still a race condition between multiple processes in the period of time between when a port is chosen and when it's actually used. That's harder to avoid, but hopefully less common as well

### DIFF
--- a/go/test/utils/portallocator_test.go
+++ b/go/test/utils/portallocator_test.go
@@ -28,7 +28,7 @@ func TestGetFreePort_NoDuplicates(t *testing.T) {
 	// Allocate many ports rapidly
 	const numPorts = 100
 	ports := make([]int, numPorts)
-	for i := 0; i < numPorts; i++ {
+	for i := range numPorts {
 		ports[i] = GetFreePort(t)
 	}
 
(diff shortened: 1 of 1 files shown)


go/test/utils/portallocator.go
```diff
@@ -16,24 +16,46 @@ package utils
...
 )
 
-// GetFreePort returns a port number that was verified free by the OS.
-// It binds to localhost:0, lets the OS choose a free port, then closes
-// the listener and returns the port number. There's a small race window
-// between closing and actual use, but this is much more reliable than
-// counter-based allocation, especially with parallel test processes.
+// portCache tracks ports currently allocated to tests to prevent duplicates.
+// Ports are added when allocated and removed via t.Cleanup() when tests complete.
+var portCache sync.Map
```
rafael, line 25: A key limitation of the previous approach was port collisions during parallel tests, since the state is not shared across test processes. This cache has the same fundamental limitation. It might reduce collisions slightly, but it will still be prone to them because the cache is not shared across parallel runs.

dweitzman: It's harder to deal with parallel process collisions, but we're getting same-process collisions even now.

Examples:
- https://github.com/multigres/multigres/actions/runs/19647929320
- https://github.com/multigres/multigres/actions/runs/19650940759

This PR should help with those, at least

dweitzman: Across processes, if we've actually used a port in a test it'll trigger a TIME_WAIT and defend against other processes using it, so the race window is between choosing a port and starting the listener for it. Cross-process is hard problem to solve without doing one of these, probably:
- Using SO_REUSEPORT on both the test and in the subprocess and using the port to trigger TIME_WAIT
- Keeping track of ports somewhere shared, like the filesystem
- Letting subprocesses allocate their own ports and report them back

PR #272, opened 2025-11-22
refactor: move go/viperutil to go/tools/viperutil

dweitzman:
This one might be worth some discussion. Previously viperutil had dependencies on go/pb for two error codes. I'm not convinced those error codes were necessary, so I removed the dependencies and I'm proposing to try putting this under go/tools

One of those error codes was FAILED_PRECONDITION when .Watch() was called twice on the same `Viper`. I think that's really more of an internal error than a FAILED_PRECONDITION. Seems like it represents a coding error on our part, not something like a system that hasn't come up yet, etc.

The other error code seems like it only impacts the text printed when starting a command fails because a flag value was bad. It doesn't feel essential to call that something that looks like a grpc INVALID_ARGUMENT, since starting a command is not a grpc.

I'm open to other perspectives, but saying that viperutil is just a library for managing flags and not something multigres-aware seems like it could help limit scope creep for that package. It's always possible to move it to go/common later if we find a compelling reason, but I'm not convinced we have a compelling reason yet

I've also moved go/viperutil/debug to go/servenv/viperdebug since it's only used by servenv and depends on go/web which will be in common


deepthi:
I agree with the basic premise that we should NOT use gRPC error codes for non-gRPC errors.

deepthi (approved):

PR #274, opened 2025-11-22
build(lint): enable "unused" linter for dead code

dweitzman:
This only catches unexported unused code. For now, this excludes the parser and protocol directories

If we want to merge this, we'll have to decide whether to remove the existing unused code or mark it with a linter exception and a comment about when/why it'll be used in the future

### DIFF
--- a/go/provisioner/local/config.go
+++ b/go/provisioner/local/config.go
@@ -321,9 +321,10 @@ func (p *localProvisioner) getServiceConfig(service string) map[string]any {
 	switch service {
 	case "etcd":
 		return map[string]any{
-			"version":  p.config.Etcd.Version,
-			"data-dir": p.config.Etcd.DataDir,
-			"port":     p.config.Etcd.Port,
+			"version":   p.config.Etcd.Version,
+			"data-dir":  p.config.Etcd.DataDir,
+			"port":      p.config.Etcd.Port,
+			"peer-port": p.config.Etcd.PeerPort,
 		}
 	case "multiadmin":
 		return map[string]any{
--- a/go/test/utils/portallocator.go
+++ b/go/test/utils/portallocator.go
@@ -16,24 +16,56 @@ package utils
 
 import (
 	"net"
+	"sync"
 	"testing"
 )
 
-// GetFreePort returns a port number that was verified free by the OS.
-// It binds to localhost:0, lets the OS choose a free port, then closes
-// the listener and returns the port number. There's a small race window
-// between closing and actual use, but this is much more reliable than
-// counter-based allocation, especially with parallel test processes.
+// portCache tracks ports currently allocated to tests to prevent duplicates.
+// Ports are added when allocated and removed via t.Cleanup() when tests complete.
+var portCache sync.Map
+
+// GetFreePort returns a port number that was verified free by the OS and not
+// currently allocated to another test. It uses a global cache to prevent the
+// same port from being returned multiple times within the same test process.
+// Ports are automatically released from the cache when the test completes via
+// t.Cleanup(), preventing memory leaks.
+//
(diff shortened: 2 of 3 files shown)


go/test/endtoend/pgctld_test.go
```diff
@@ -35,15 +34,6 @@ import (
 	"github.com/multigres/multigres/go/test/utils"
 )
 
-var (
```
deepthi, line 38: This is good, leftover from removing binary builds in tests.

go/provisioner/local/local.go
```diff
@@ -1763,23 +1762,6 @@ func (p *localProvisioner) getCellNames() ([]string, error) {
 	return names, nil
 }
 
-// getCellIndex returns the index of a cell in the list of cell names (for port calculation)
```
deepthi, line 1766: This should also be good, now that we have a better port allocator.

go/multipooler/heartbeat/writer.go
```diff
@@ -204,16 +204,6 @@ func (w *Writer) write() error {
 	return nil
 }
 
-// getWALPosition returns the current WAL LSN position
-func (w *Writer) getWALPosition(ctx context.Context) (string, error) {
```
deepthi, line 208: 👍

go/common/mterrors/code.go
```diff
@@ -75,44 +75,9 @@ func errorWithoutState(id string, code mtrpcpb.Code, short, long string) func(ar
 	}
 }
 
-func errorWithState(id string, code mtrpcpb.Code, state State, short, long string) func(args ...any) *MultigresError {
```
deepthi, line 78: I think this means we need unit tests :)
But actually, we need to look at how they are being used in Vitess and whether we expect to use them the same way.

go/common/mterrors/mterrors.go
```diff
@@ -111,12 +111,6 @@ func getLogErrStacks() bool {
 	return logErrStacks
 }
 
-func setLogErrStacks(val bool) {
```
deepthi, line 114: I'm surprised that this is unused. Don't we need it for `get` to return something useful? Feels like a missing unit test. We can expect to use it in future.

go/common/mterrors/code.go
dweitzman: Brought back this code with a comment to disable the linter

go/common/mterrors/mterrors.go
dweitzman: Ported an associated test file from vitess

go/common/rpcclient/conn_cache.go
```diff
@@ -87,14 +86,6 @@ func (cc *ConnConfig) RegisterFlags(fs *pflag.FlagSet) {
 	viperutil.BindFlags(fs, cc.cert, cc.key, cc.ca, cc.crl, cc.name)
 }
 
-// validateTLSConfig checks if TLS is configured and returns an error if so (not yet implemented).
```
rafael, line 90: 👍

go/servenv/logging.go
```diff
@@ -228,18 +228,6 @@ func fireLoggingSetupHooks(l *slog.Logger) {
 	}
 }
 
-// fireLoggingChangeHooks calls all registered logging change hooks.
```
rafael, line 231: This one, we should check what did we miss to port from vitess. I'm curious where this was being called. 

It might be useful and we are missing porting parts of servenv.

dweitzman: Doesn't seem like this code exists in vitess, since vitess is using glog?

dweitzman: This function seems to only exist in multigres: https://github.com/search?q=fireLoggingChangeHooks&type=code

PR #276, opened 2025-11-24
build(govet): Enable most of the non-default checks

dweitzman:
Seems worthwhile to enable any linters or analyzers that may provide value unless they're causing false positives or making development too slow

I've left out two non-default analyzers because they have a lot of complains:
- shadow (complains about shadowing variable names)
- fieldalignment (recommends reordering fields in structs to use less memory)

fieldalignment doesn't seem unreasonable, but if we want to do it then we should probably make a separate PR

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -71,10 +71,6 @@ linters:
     sloglint:
       context: scope
     govet:
-      disable:
-        - nilness
-        - unusedwrite
-        - loopclosure
       enable:
         # Check for missing values after append.
         - appends


go/multipooler/manager/manager.go
```diff
@@ -750,12 +750,6 @@ func (pm *MultiPoolerManager) loadConsensusTermFromDisk() {
 		pm.consensusLoaded = true
 		pm.mu.Unlock()
 
-		if err != nil {
```
dweitzman, line 753: This condition could never be true

go/parser/lexer_errors_test.go
```diff
@@ -700,35 +700,23 @@ func TestRealWorldErrorScenarios(t *testing.T) {
...
-			var lastError error
 			for {
 				token := lexer.NextToken()
 				if token.Type == EOF {
 					break
 				}
 			}
 
-			// Check for errors either as return value or in context
-			if lastError != nil {
```
dweitzman, line 712: This condition could never be true

go/multipooler/manager/manager.go
deepthi: This is why we need static analysis tools, lol

deepthi: The correct way to fix this is to remove the error return from functions that always return nil error. Which then begs the question: if `Load` can never return an error, why do we even have retries with backoff? I haven't looked deeply into this, but something changed when it was refactored to introduce `GetInconsistentCurrentTermNumber`.

dweitzman: `Load()` can return an error, it's just handled earlier with a `continue` if it's non-`nil`

deepthi: Sorry, I was mixing it up with `GetInconsistentCurrentTermNumber`, which is the one that only seems to return a nil error.

dweitzman: Ah yeah, that seems worth fixing but probably out of scope for this PR

deepthi (approved):

PR #279, opened 2025-11-24
feat(coverage): implement subprocess coverage collection

dweitzman:
## Motivation

Current test coverage only captures direct test execution, missing coverage from subprocess binaries spawned via `exec.Command()` (multipooler, multigateway, pgctld, etc.).

## Implementation

**Coverage-instrumented binaries:**
- `make build-coverage` creates instrumented binaries in `bin/cov/`
- `PrependBinToPath()` checks `GOCOVERDIR` env var to automatically use coverage binaries
- Subprocess executions inherit `GOCOVERDIR` and emit coverage on exit

**Parallel collection strategy:**
- Job 1 (in parallel): Direct test coverage (`go test -coverprofile`)
- Job 2 (in parallel): Subprocess coverage (`GOCOVERDIR` with coverage binaries)
- Job 3: Merge results with `merge-coverage.sh` and upload

**Local development:**
- `scripts/go_test_coverage.sh` runs both coverage modes in serial and merges results
- `make test-coverage` runs the full coverage collection locally using `go_test_coverage.sh`
- `scripts/merge-coverage.sh` uses `go tool gocovmerge` with filtering for yacc-generated code
- Creates unique subdirectories under `coverage/` prevent pollution between runs

**CI infrastructure:**
- New `.github/workflows/coverage.yml` with parallel jobs (no time penalty)
- Shared `.github/scripts/setup-test-environment.sh` reduces duplication among github action jobs
- Simplified `test-integration.yml` to focus on test execution only

## Key Technical Details

- Uses Go's built-in `-cover` flag with `GOCOVERDIR` for binary coverage collection
- PATH manipulation ensures binaries with coverage run (`./bin/cov/multipooler` will be before `./bin/multipooler` in `PATH`, etc.)
- Coverage merge filters out generated parser code (postgres.y, yaccpar, yacctab) to avoid overlaps that confuse the coverage tooling
- Workflows run on both PRs and main branch

## Why not run `GOCOVERDIR=... go test -cover` all at once?...

Doesn't work 😢. `go test -cover` will replace whatever `GOCOVERDIR` is already in the environment and throw away anything unexpected that appears there. It's not impossible to pre-compile all the tests by package with coverage and do `GOCOVERDIR=... ./testname -test.coverprofile=... ...` but it'd be significantly more complex and probably brittle than just doing two test runs and combining the results

## Other Notes

If we have test cases sending a KILL signal instead of TERM, they may cause subprocesses to die without having the opportunity to emit coverage data. Where possible, we should try to send a TERM first for a more graceful shutdown

### DIFF
--- a/scripts/go_test_coverage.sh
+++ b/scripts/go_test_coverage.sh
@@ -70,14 +70,19 @@ echo ""
 
 if go test -cover -covermode=atomic -coverprofile="$DIRECT_COV" -coverpkg=./... "$@"; then
   DIRECT_RESULT="✓ PASS"
+else
+  DIRECT_RESULT="✗ FAIL"
+  echo ""
+  echo "⚠️  Direct test run failed, but continuing with subprocess coverage..."
+fi
+
+# Extract coverage percentage even if tests failed (partial coverage is still useful)
+if [ -f "$DIRECT_COV" ] && [ -s "$DIRECT_COV" ]; then
   DIRECT_PCT=$(go tool cover -func="$DIRECT_COV" 2>&1 | tail -1 | grep -oE '[0-9]+\.[0-9]+%' || echo "N/A")
   echo ""
   echo "Direct coverage: $DIRECT_PCT"
 else
-  DIRECT_RESULT="✗ FAIL"
   DIRECT_PCT="N/A"
-  echo ""
-  echo "⚠️  Direct test run failed, but continuing with subprocess coverage..."
 fi
 
 echo ""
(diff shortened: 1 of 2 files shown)


scripts/go_test_coverage.sh
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env bash
+# Copyright 2025 Supabase, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
```
GuptaManan100, line 5: I think this script is unused

dweitzman: It's intended for local development, not CI. I'll add a `Makefile` target for it

dweitzman:
> And I assume there is no way to run tests with both types of coverage (main and subprocess) at the same time.

Not that I've found. Seems collecting everything simultaneously would need changes to the Go runtime.

scripts/go_test_coverage.sh
GuptaManan100: Ah makes sense!

PR #283, opened 2025-11-25
test: fix non-random port assignment for etcd in endtoend tests

dweitzman:
The `PeerPort` setting was getting dropped during transformation which falls back to the default behavior of assuming the peer port can be the client port plus one. That's not a safe assumption

Example failure on CI: https://github.com/multigres/multigres/actions/runs/19683063519/job/56381654237


rafael (approved):
Great catch. LGTM

PR #284, opened 2025-11-25
feat(multiorch): Better in-memory store for PoolerHealth

rafael:
## Summary

Replace the hand rolled `PoolerHealth` Go struct with a proto defined `PoolerHealthState` message, and introduce `ProtoStore[K, V]`, a generic store that automatically clones protobuf values on `Get` and `Set` using `proto.Clone()`.

This PR doesn't have any functional changes outside the store. It is replacing all consumers of the store with the new type. 

## Problem

The existing `Store[K, V]` protects the underlying map with a mutex, but not the objects stored inside it. When `V` is `*PoolerHealth`, every `Get(poolerID)` returns the same pointer. Multiple goroutines (health check workers, discovery loop, bookkeeping) can read and mutate the same object concurrently, which leads to data races.

## Solution

1. `PoolerHealthState` proto message (`proto/multiorchdata.proto`)  
   Defines pooler runtime state as a protobuf, embedding `MultiPooler` for topology data and `PrimaryStatus`/`StandbyReplicationStatus` for health metrics.

2. `ProtoStore[K, V proto.Message]`  
   Generic store that clones values on every `Get` and `Set` via `proto.Clone()`. Callers always receive isolated copies, so there is no need for explicit `DeepCopy()` implementations at call sites.

3. Cleanup of legacy types and helpers  
   Remove the manual `PoolerHealth` struct (`pooler_info.go`) and the reflection based deep copy helpers (`copy.go`).

## Key changes

- `store.PoolerHealth` replaced with `*multiorchdatapb.PoolerHealthState`.
- `store.NewPoolerHealthFromMultiPooler(mp)` replaced with `&multiorchdatapb.PoolerHealthState{MultiPooler: mp}`.
- Flattened fields such as `PrimaryLSN`, `ReplicaLagMillis` moved into nested messages like `PrimaryStatus.lsn` and `ReplicationStatus.lag`.
- `time.Time` fields converted to `*timestamppb.Timestamp`.

## Test plan

- All unit tests updated and passing.
- E2E bootstrap test updated and passing.

### DIFF
--- a/go/multiorch/recovery/healthcheck.go
+++ b/go/multiorch/recovery/healthcheck.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"time"
 
-	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/multigres/multigres/go/clustermetadata/topo"
@@ -120,12 +119,11 @@ func (re *Engine) pollPooler(ctx context.Context, poolerID *clustermetadata.ID, 
 
 	// Note: Poll attempts are now tracked via the poll.duration histogram with status attribute
 
-	// Mark attempt timestamp - clone the pooler to avoid race condition
+	// Mark attempt timestamp
+	// Note: pooler is already a clone from store.Get(), safe to mutate directly
 	now := time.Now()
-	updated := proto.Clone(pooler).(*multiorchdatapb.PoolerHealthState)
-	updated.LastCheckAttempted = timestamppb.New(now)
-	re.poolerStore.Set(poolerIDStr, updated)
-	pooler = updated // use updated for rest of function
+	pooler.LastCheckAttempted = timestamppb.New(now)
+	re.poolerStore.Set(poolerIDStr, pooler)
 
 	// Call Status RPC which works for both PRIMARY and REPLICA poolers
 	// Use provided context with 5 second timeout to prevent blocking forever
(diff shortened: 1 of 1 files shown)


go/multiorch/actions/bootstrap_shard.go
```diff
@@ -84,27 +84,27 @@ func (a *BootstrapShardAction) Execute(ctx context.Context, shardID string, data
 	a.logger.InfoContext(ctx, "Selected bootstrap candidate",
 		"shard", shardID,
 		"database", database,
-		"candidate", candidate.ID.Name)
+		"candidate", candidate.MultiPooler.Id.Name)
```
rafael, line 87: This is back to have a Multipooler in `PoolerHealthState` struct (no flattening)

go/multiorch/actions/scenario_detection.go
```diff
@@ -61,16 +61,16 @@ func (s InitializationScenario) String() string {
...
-	for i, node := range cohort {
+	for i, pooler := range cohort {
 		req := &multipoolermanagerdatapb.InitializationStatusRequest{}
-		status, err := rpcClient.InitializationStatus(ctx, node.ToMultiPooler(), req)
+		status, err := rpcClient.InitializationStatus(ctx, pooler.MultiPooler, req)
 		if err != nil {
 			if logger != nil {
 				logger.WarnContext(ctx, "Failed to get initialization status from node",
-					"node", node.ID.Name,
+					"node", pooler.MultiPooler.Id.Name,
```
rafael, line 73: argh - I missed updating this one to node. We are planning to remove this one soon. So leaning towards leaving for now.

go/multiorch/recovery/recovery_loop_test.go
```diff
@@ -1055,61 +1072,58 @@ func TestRecoveryLoop_PostRecoveryRefresh(t *testing.T) {
...
-		Shard:      "0",
-		Type:       clustermetadatapb.PoolerType_REPLICA,
-		Hostname:   "replica2-host",
-	})
-	replica2Pooler.IsLastCheckValid = true
-	replica2Pooler.IsUpToDate = true
-	replica2Pooler.LastSeen = time.Now()
-	engine.poolerStore.Set("multipooler-cell1-replica2-pooler", replica2Pooler)
-
 	// Record initial LastCheckAttempted times
```
rafael, line 1097: This looks a bit confusing, but it's a cleanup. The initial check times are recorded at the top of the function now.

go/multiorch/recovery/healthcheck.go
```diff
@@ -111,22 +113,17 @@ func (re *Engine) pollPooler(ctx context.Context, poolerID *clustermetadata.ID,
...
 	now := time.Now()
-	updated := store.NewPoolerHealthFromMultiPooler(pooler.ToMultiPooler())
-	updated.LastCheckAttempted = now
-	updated.LastCheckSuccessful = pooler.LastCheckSuccessful
-	updated.LastSeen = pooler.LastSeen
-	updated.IsUpToDate = pooler.IsUpToDate
-	updated.IsLastCheckValid = pooler.IsLastCheckValid
-	re.poolerStore.Set(poolerIDStr, updated)
-	pooler = updated // use updated for rest of function
+	pooler.LastCheckAttempted = timestamppb.New(now)
```
rafael, line 125: this is actually cleaner now.

dweitzman (approved):

PR #288, opened 2025-11-26
fix(test): eliminate several race conditions found by CI

dweitzman:
Fix race conditions detected by go test -race:

1. [toporeg_test.go](https://github.com/multigres/multigres/actions/runs/19685522005?pr=286): registerCallCount was accessed both inside a mutex (during increment) and outside it (in error message formatting). Fixed by replacing int+mutex with sync/atomic.Int32, which allows safe concurrent reads without holding the mutex.

2. [rpc_manager_test.go](https://github.com/multigres/multigres/actions/runs/19685457322?pr=285): setupPromoteTestManager() was assigning pm.db after starting the manager, creating a race between the assignment and goroutines reading pm.db. Fixed by:
   - Making mockDB an input parameter instead of creating it internally
   - Assigning pm.db BEFORE calling go pm.Start()
   - Adding expectStartupQueries() helper to set expectations for startup queries (pg_is_in_recovery checks) that run before tests
   
3. [rpc_consensus_test.go](https://github.com/multigres/multigres/actions/runs/19690387521): Similar to `rpc_manager_test.go`, where we can avoid a race by setting up the mockDB before startup instead of concurrently with startup

### DIFF
--- a/go/multipooler/manager/rpc_consensus_test.go
+++ b/go/multipooler/manager/rpc_consensus_test.go
@@ -37,7 +37,24 @@ import (
 )
 
 // Helper function to setup a manager with a mock database
-func setupManagerWithMockDB(t *testing.T) (*MultiPoolerManager, sqlmock.Sqlmock, string) {
+// expectPrimaryStartupQueries adds expectations for queries that happen during PRIMARY manager startup.
+// The manager is created as a PRIMARY, so:
+// 1. pg_is_in_recovery() check - returns false (primary)
+// 2. CREATE SCHEMA and CREATE TABLE queries for sidecar schema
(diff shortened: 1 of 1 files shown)


go/multipooler/manager/rpc_consensus_test.go
```diff
@@ -75,23 +92,19 @@ func setupManagerWithMockDB(t *testing.T) (*MultiPoolerManager, sqlmock.Sqlmock,
 	pm := NewMultiPoolerManager(logger, config)
 	t.Cleanup(func() { pm.Close() })
 
+	// Assign mock DB BEFORE starting the manager to avoid race conditions
+	pm.db = mockDB
+
```
deepthi, line 97: And all the db code will be obsolete when we get connection pooling :)

PR #290, opened 2025-11-26
build(lint): enable more linters (gosec, errorlint, nilerr, misspell, unconvert)

dweitzman:
Enable more linters:
- gosec: security scanning (SQL injection, command injection)
- errorlint: Go 1.13+ error handling (errors.Is/As)
- nilerr: returning nil when err != nil
- misspell: spelling mistakes in comments/strings
- unconvert: unnecessary type conversions

Code fixes:
- Use parameterized SQL query instead of Sprintf (G201)
- Use strconv.ParseInt for typesafe int32 parsing (G109)
- Add ReadHeaderTimeout to HTTP server (G114)
- Use errors.Is() for context and io.EOF comparisons
- Use %w for error wrapping in fmt.Errorf

Exclusions configured for intentional patterns:
- G115/G404/G306/G204 globally (common/acceptable)
- errorlint for mterrors (intentional direct comparison)
- unconvert for parser (generated code patterns)
- nolint comments for intentional nil returns


.golangci.yml
```diff
@@ -199,6 +206,30 @@ linters:
       - linters:
           - unused
         path: ^go/(parser|pgprotocol|tools/ruleguard)
+      # Integer overflow conversions are all over the place
```
rafael: what does this mean? 😭

dweitzman: These are places that do unsafe casting between integers, like casting between a signed and unsigned 32-bit integer, reducing from 64-bit to 32-bit, etc

I think it's a reasonable thing for a security-oriented linter to be suspicious of, but it seems far too pervasive to be worth spending time on now

PR #292, opened 2025-11-28
refactor(context): replace context.Background() with context.TODO() in lots of places

dweitzman:
In general, I don't think we should be using context.Background() except in very rare cases like a test's Cleanup() function or in a top-level main() function.
    
Most context.Background() have better alternatives that preserve tracing context and propagate timeouts from higher levels:
- CLI commands can use cmd.Context() from Cobra
- Test utilities can use t.Context() where *testing.T available, except in Cleanup()

This PR also add a ruleguard lint rule to flag unexplained context.Background() usage in non-test files. In the future test files would be included as well.


go/clustermetadata/topo/memorytopo/store.go
```diff
@@ -485,8 +485,13 @@ func (f *Factory) getOperationError(op Operation, path string) error {
 
 func init() {
 	// TODO: @rafael
```
rafael: This made me realize of this pending cleanup. I created PR for this: https://github.com/multigres/multigres/pull/312

It was a real short lived :D

PR #294, opened 2025-11-29
feat(multiorch): implement analyzers and recovery actions, and integrate with coordinator

deepthi:
## Core Features

  1. Automatic Shard Bootstrap
    - Multiorch detects uninitialized shards and automatically bootstraps them
    - Implements ShardNeedsBootstrap problem detection and BootstrapShardAction
  2. PrimaryIsDead Detection and Recovery
    - Detects when primary postgres dies (multipooler responds but postgres is down)
    - Triggers automatic leader appointment from healthy standbys
    - Prefers initialized nodes during leader election
  3. Distributed Locking for Recovery
    - Prevents concurrent recovery operations on the same shard
    - Uses LockShard from topo package
  4. Merge InitializationStatus into Status and delete it
    - We no longer error from Status if postgres is down
    - Enables multiorch to poll uninitialized poolers, and poolers where postgres is down (needed for PrimaryIsDead)
  5. Local provisioner
    - Do not start all poolers as primary/writable, they should be read_only
    - Change the order of startup so that all poolers are up before multiorch
    - Multiorch now bootstraps local cluster during provisioning

  ## Bug Fixes

  - Health check stall : Reset IsUpToDate flag to continue polling after bootstrap
  - Empty LSN in leader election : Skip nodes with empty WAL position
  - SetPrimaryConnInfo port : Use postgres port instead of grpc port
  - Port comparison in primary detection 
  - Postgres initialization status 

 ## Test Infrastructure

  - New test: TestMultiOrchDeadPrimaryRecovery - validates failover after primary dies
  - Shared helpers: Extracted multiorch_helpers.go with reusable test utilities
  - Dynamic primary detection: Tests detect which zone is primary before running write tests
  - Multigateway tests: Tests wait for multiorch to complete bootstrap before querying
  - KEEP_TEMP_DIRS support: Preserve test directories for debugging

 ## Future improvements
  - Detect when new pooler joins an already initialized shard, and wire it up properly
  - We should be able to remove SetTerm and all its usages
  - Test that recoveries work correctly with >1 multiorchs


go/multigateway/init.go
```diff
@@ -186,7 +186,7 @@ func (mg *MultiGateway) Init() {
 	multigateway := topo.NewMultiGateway(mg.serviceID.Get(), mg.cell.Get(), mg.senv.GetHostname())
 	multigateway.PortMap["grpc"] = int32(mg.grpcServer.Port())
 	multigateway.PortMap["http"] = int32(mg.senv.GetHTTPPort())
-	multigateway.PortMap["pg"] = int32(mg.pgPort.Get())
+	multigateway.PortMap["postgres"] = int32(mg.pgPort.Get())
```
deepthi, line 189: Is anyone particular about calling this pg vs postgres?

go/multiorch/coordinator/leader_appointment.go
```diff
@@ -187,15 +190,24 @@ func (c *Coordinator) selectCandidate(ctx context.Context, cohort []*multiorchda
 			"no healthy nodes available for candidate selection")
 	}
 
-	// Select node with most advanced WAL position
+	// Prefer initialized nodes over uninitialized nodes
+	// Within each group (initialized vs uninitialized), select by WAL position
 	var bestCandidate *multiorchdatapb.PoolerHealthState
 	var bestLSN pglogrepl.LSN
+	var bestIsInitialized bool
```
deepthi, line 197: This becomes necessary if we end up with ShardHasNoPrimary, and a new pooler that has not yet been initialized. We should not promote the new pooler.

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -86,7 +86,16 @@ func (a *BootstrapShardAction) Execute(ctx context.Context, shardID string, data
...
-	// Step 3: Initialize the candidate as an empty primary with term=1
+	// Step 3: Set pooler type to PRIMARY before initializing
+	changeTypeReq := &multipoolermanagerdatapb.ChangeTypeRequest{
+		PoolerType: clustermetadatapb.PoolerType_PRIMARY,
+	}
+	_, err = a.rpcClient.ChangeType(ctx, candidate.MultiPooler, changeTypeReq)
+	if err != nil {
+		return mterrors.Wrap(err, "failed to set pooler type to PRIMARY")
+	}
+
```
deepthi: This could also be done _after_  InitializeEmptyPrimary instead of before. That seems safer to me. Thoughts?

go/multiorch/recovery/analysis/appoint_leader_recovery_action.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (a *AppointLeaderRecoveryAction) Execute(ctx context.Context, problem Problem) error {
+	a.logger.InfoContext(ctx, "executing appoint leader recovery action",
+		"database", problem.Database,
+		"tablegroup", problem.TableGroup,
+		"shard", problem.Shard)
+
+	// Step 1: Acquire distributed lock for this shard
+	metadata := a.Metadata()
+	lockPath := topo.RecoveryLockPath(problem.Database, problem.TableGroup, problem.Shard)
```
deepthi: We can replace these 3 fields on problem with ShardKey and pass that around.

go/multiorch/recovery/analysis/shard_needs_bootstrap_analyzer.go
```diff
@@ -0,0 +1,95 @@
...
+	// Other analyzers (PrimaryIsDead, ShardHasNoPrimary) will handle dead primaries.
+	if poolerAnalysis.IsUnreachable {
+		return nil
+	}
+
+	// Skip primary nodes - they don't have a PrimaryPoolerID by design (they ARE the primary).
+	// A dead primary should be handled by PrimaryIsDead (detected by replicas), not by
+	// ShardNeedsBootstrap. When a primary's postgres dies, it appears as:
+	// - IsPrimary = true (it's the primary)
+	// - IsInitialized = false (postgres down)
```
deepthi: I'm doubtful about this being false when postgres is down.

go/multiorch/recovery/analysis/shard_needs_bootstrap_analyzer.go
```diff
@@ -0,0 +1,95 @@
...
+	// - PrimaryPoolerID = nil (it's the primary itself)
+	// This would incorrectly trigger ShardNeedsBootstrap if we don't skip it.
+	// Note: IsPrimary is based on MultiPooler.Type == PRIMARY from topology, which is set
+	// when the node is initialized as primary. New uninitialized nodes start with Type=PRIMARY
```
deepthi: Need to fix this comment. Tests and provisioner were previously starting all poolers as PRIMARY, but I have fixed that.

go/multiorch/recovery/analysis/types.go
```diff
@@ -58,14 +61,16 @@ const (
...
-	// Arbitrarily high priority, to leave room for other priorities.
-	PriorityShardBootstrap Priority = 10000
-
 	// PriorityEmergency is for catastrophic issues like dead primary.
-	// These must be fixed before anything else can proceed.
-	PriorityEmergency Priority = 1000
+	// This is the highest priority - a dead primary must be replaced immediately.
+	// When a primary dies, we want to elect a new leader from existing standbys,
+	// NOT re-bootstrap the entire shard.
+	PriorityEmergency Priority = 10000
```
deepthi: This was changed during debugging. I'm not quite certain whether we need to keep this change.

go/test/endtoend/cluster_test.go
```diff
@@ -906,13 +947,15 @@ func TestClusterLifecycle(t *testing.T) {
...
+		t.Log("Checking heartbeats...")
 		heartbeatsWritten, err := checkHeartbeatsWritten(multipoolerAddr)
-		require.NoError(t, err, "should be able to check heartbeats")
-		assert.True(t, heartbeatsWritten, "at least one heartbeat should have been written before cluster stopped")
-		t.Log("Heartbeats verified successfully!")
+		require.NoError(t, err, "should be able to query heartbeat table")
+		if heartbeatsWritten {
+			t.Log("Heartbeats detected")
+		} else {
+			t.Log("No heartbeats yet (normal immediately after bootstrap)")
```
deepthi: I'm skeptical about this. Maybe we should expect to have a heartbeat even immediately after bootstrap. Or wait for the heartbeat_interval and make sure it is written.

go/provisioner/local/local.go
```diff
@@ -1551,11 +1551,13 @@ func (p *localProvisioner) ProvisionDatabase(ctx context.Context, databaseName s
 
 	var results []*provisioner.ProvisionResult
 
-	// Provision services in each cell
-	for _, cellName := range cellNames {
-		fmt.Printf("=== Provisioning services in cell: %s ===\n", cellName)
+	// Provision services in phases across all cells to ensure dependencies are met.
+	// Multiorch needs all multipoolers running before bootstrap, so we start
+	// services in this order: multigateways -> multipoolers -> multiorchs
```
rafael, line 1583: Orders shouldn't matter. Why are we requiring this order?

go/clustermetadata/topo/recovery_lock.go
```diff
@@ -0,0 +1,67 @@
...
+// Use a defer with background context for unlock to ensure cleanup:
+//
+//	lockDesc, err := ts.LockShardForRecovery(ctx, db, tg, shard, "bootstrap", 15*time.Second)
+//	if err != nil { return err }
+//	defer func() {
+//	    unlockCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+//	    defer cancel()
+//	    lockDesc.Unlock(unlockCtx)
+//	}()
+func (ts *store) LockShardForRecovery(ctx context.Context, database, tableGroup, shard, purpose string, lockTimeout time.Duration) (LockDescriptor, error) {
```
rafael: I think this kind of Lock should be in the global topo.

go/clustermetadata/topo/recovery_lock.go
```diff
@@ -0,0 +1,67 @@
...
+
+package topo
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+// RecoveryLockPath returns the canonical lock path for shard-level recovery operations.
```
rafael: Thinking we should fully port `locks.go` implementation for this. That interface was battle tested. I did not port that in my initial pass, because I was not sure we were gonna need it. 

That was the one we were using in Vitess for this. 

I can help with this part.

go/multigateway/init.go
rafael: Sounds good to me

go/multiorch/coordinator/leader_appointment.go
rafael: I am not sure we should track this in this context. If a pooler is not initialized, it should never be eligible as bestCandidate. It should not participate in the selection at all or be considered for promotion.

go/multiorch/coordinator/leader_appointment.go
```diff
@@ -474,7 +508,7 @@ func (c *Coordinator) Propagate(ctx context.Context, candidate *multiorchdatapb.
 
 			setPrimaryReq := &multipoolermanagerdatapb.SetPrimaryConnInfoRequest{
 				Host:                  candidate.MultiPooler.Hostname,
-				Port:                  candidate.MultiPooler.PortMap["grpc"],
+				Port:                  candidate.MultiPooler.PortMap["pg"],
```
rafael: shouldn't this still be gRPC? or postgres?

go/multiorch/recovery/actions/bootstrap_shard.go
rafael: Agree

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -132,7 +141,7 @@ func (a *BootstrapShardAction) Execute(ctx context.Context, shardID string, data
 		"database", database,
 		"policy_name", policyName)
 
-	// Step 5: Initialize remaining nodes as standbys
+	// Step 6: Initialize remaining nodes as standbys
```
rafael: Were you able to remember why we need to do this? 

I'm still thinking that unless there is a strong reason this is not part of bootstrap.

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -157,13 +166,22 @@ func (a *BootstrapShardAction) Execute(ctx context.Context, shardID string, data
...
-	// - Node with fastest storage
-	// - Node in preferred availability zone
-	// - Node with most resources available
+	// For bootstrap in mixed initialization scenarios, prefer initialized nodes
+	// to avoid data loss. This handles cases where some nodes have data and others don't.
+	// Selection strategy:
+	// 1. First pass: look for initialized nodes
+	// 2. Second pass: if no initialized nodes, use first uninitialized node
+
+	type nodeInfo struct {
```
rafael: should we call this poolerInfo ?

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -157,13 +166,22 @@ func (a *BootstrapShardAction) Execute(ctx context.Context, shardID string, data
...
-// selectBootstrapCandidate selects the first healthy node as the bootstrap candidate
+// selectBootstrapCandidate selects a healthy node as the bootstrap candidate.
+// Prefers initialized nodes over uninitialized nodes to avoid data loss in mixed scenarios.
 func (a *BootstrapShardAction) selectBootstrapCandidate(ctx context.Context, cohort []*multiorchdatapb.PoolerHealthState) (*multiorchdatapb.PoolerHealthState, error) {
-	// For bootstrap, we just pick the first reachable node
-	// In a production system, you might want to consider factors like:
-	// - Node with fastest storage
-	// - Node in preferred availability zone
-	// - Node with most resources available
+	// For bootstrap in mixed initialization scenarios, prefer initialized nodes
```
rafael: This should be treated as an error. If we end up in a mixed state, it means something went wrong and requires manual intervention. For a bootstrap shard, the only safe state is that all nodes are uninitialized.

In other words, “all nodes are uninitialized” should be an invariant for a bootstrap shard.

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -86,7 +86,16 @@ func (a *BootstrapShardAction) Execute(ctx context.Context, shardID string, data
 		"database", database,
 		"candidate", candidate.MultiPooler.Id.Name)
 
-	// Step 3: Initialize the candidate as an empty primary with term=1
+	// Step 3: Set pooler type to PRIMARY before initializing
```
rafael: Something we discussed the other day that @dweitzman proposed:

Step 0 in this sequence should be to check a boolean in topo (similar to how we read the durability policy) to validate that this shard has not already been initialized.

We should set this boolean to true very early in the flow. If anything fails halfway through, that should require manual intervention to fix.

Rationale: we want to be extremely cautious that automated bootstrapping never wipes an existing shard by reinitializing a shard that was already initialized.

go/multiorch/recovery/analysis/analyzer.go
```diff
@@ -51,3 +54,19 @@ func SetTestAnalyzers(analyzers []Analyzer) {
...
 	defaultAnalyzers = nil
 }
+
+// globalFactory holds the global RecoveryActionFactory instance.
+// This is set during engine initialization and used by analyzers.
+var globalFactory *RecoveryActionFactory
+
+// SetRecoveryActionFactory sets the global recovery action factory.
+// This should be called during engine initialization.
+func SetRecoveryActionFactory(factory *RecoveryActionFactory) {
```
rafael, line 65: Nice. I like how this is shaping.

go/multiorch/recovery/analysis/bootstrap_recovery_action.go
```diff
@@ -0,0 +1,184 @@
+// Copyright 2025 Supabase, Inc.
```
rafael: Something seems off here. Shouldn't we be using boostrap_shard.go in actions instead of this? 

Although this has the lock shard.

go/multiorch/recovery/analysis/appoint_leader_recovery_action.go
rafael: We should need a lock for the leader appointment action.

go/multiorch/recovery/analysis/appoint_leader_recovery_action.go
```diff
@@ -0,0 +1,160 @@
+// Copyright 2025 Supabase, Inc.
```
rafael: Idem as the action for bootstrap. shouldn't this one be replaced by appoint_leader.go ?

go/multiorch/recovery/analysis/primary_is_dead_analyzer.go
```diff
@@ -0,0 +1,74 @@
...
+	if !poolerAnalysis.IsInitialized {
+		return nil
+	}
+
+	// Only trigger if:
+	// 1. A primary exists in topology (PrimaryPoolerID is set)
+	// 2. BUT the primary is unreachable (PrimaryReachable is false)
+	//
+	// This is the key difference from ShardHasNoPrimary which checks PrimaryPoolerID == nil
+	if poolerAnalysis.PrimaryPoolerID != nil && !poolerAnalysis.PrimaryReachable {
```
rafael: one small nit, we could return early if the primary is reachable. 
```
if poolerAnalysis.PrimaryPoolerID != nil && poolerAnalysis.PrimaryReachable {
  retur nil
}
```

Then have all the logic for the problem

go/multiorch/recovery/analysis/primary_is_dead_analyzer.go
```diff
@@ -0,0 +1,74 @@
...
+		factory := GetRecoveryActionFactory()
+		if factory == nil {
+			// Factory not initialized yet, skip recovery action
```
rafael: For this action, if the factory is nil, should this panic? Or should this return error? 

I wonder if the interface here should be problem or error.

go/multiorch/recovery/analysis/primary_is_dead_analyzer.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (a *PrimaryIsDeadAnalyzer) Analyze(poolerAnalysis *store.ReplicationAnalysis) []Problem {
+	// Only analyze replicas (primaries can't report themselves as dead)
+	if poolerAnalysis.IsPrimary {
+		return nil
+	}
+
+	// Skip if replica is not initialized (ShardNeedsBootstrap handles that)
+	if !poolerAnalysis.IsInitialized {
```
rafael, line 42: We should talk about this. I am on the fence about skipping here. I wonder if this analyzer should still run and report the problem, and rely on the filtering logic to ensure we do not act on it before bootstrap.

My mental model is that each problem detector should operate in isolation and report issues regardless of whether another action will eventually address them.

The filtering layer should then ensure that we do not execute recovery for this problem until the bootstrap has been handled.

go/multiorch/recovery/analysis/shard_needs_bootstrap_analyzer.go
```diff
@@ -0,0 +1,95 @@
...
+	// but will have IsInitialized=false AND PrimaryPoolerID != nil won't match below.
+	if poolerAnalysis.IsPrimary && poolerAnalysis.IsInitialized {
```
rafael: To me this is a slightly different problem. It effectively means a failed bootstrap. One idea is to have a separate analyzer (with no recovery action) that detects shards in a failed bootstrap state.

go/multiorch/recovery/analysis/shard_needs_bootstrap_analyzer.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	// Only analyze if this pooler is uninitialized
+	if poolerAnalysis.IsInitialized {
+		return nil
+	}
+
+	// If this pooler is uninitialized AND there's no primary in the shard,
+	// then the whole shard likely needs bootstrap
+	if poolerAnalysis.PrimaryPoolerID == nil {
```
rafael, line 71: In my mind, this is the only condition this analyzer needs.

go/multiorch/recovery/analysis/types.go
rafael: This seems wrong. We should revert this. If we were re-initiazling dead primaries, that seems more a problem in the analyzer than in the priorities.

go/multiorch/recovery/recovery_loop.go
```diff
@@ -210,10 +210,11 @@ func (re *Engine) attemptRecovery(problem analysis.Problem) {
...
 	}
 
-	// Acquire lock if needed
+	// Note: Locking is handled by the recovery actions themselves using
+	// topo.LockShardForRecovery(). Actions that require locking (RequiresLock() == true)
+	// acquire distributed locks internally with configurable timeout from Metadata().LockTimeout.
 	if problem.RecoveryAction.RequiresLock() {
-		// TODO: Implement shard locking
-		re.logger.DebugContext(re.ctx, "recovery action requires lock", "problem_code", problem.Code)
+		re.logger.DebugContext(re.ctx, "recovery action will acquire shard lock", "problem_code", problem.Code)
```
rafael: now that we have locking, if we the action requires lock, I think this is the place where we try to acquire the lock.

go/multiorch/recovery/recovery_loop.go
```diff
@@ -238,14 +239,26 @@ func (re *Engine) attemptRecovery(problem analysis.Problem) {
...
+	// then force health check all poolers in the shard to ensure they have up-to-date state.
 	if problem.Scope == analysis.ScopeShard {
 		re.logger.InfoContext(re.ctx, "forcing refresh of all poolers post recovery",
 			"database", problem.Database,
 			"tablegroup", problem.TableGroup,
 			"shard", problem.Shard,
 		)
+		// First refresh shard metadata from topology to pick up any type changes
+		// (e.g., standbys being set to REPLICA after bootstrap via ChangeType RPC)
+		if err := re.refreshShardMetadata(context.Background(), problem.Database, problem.TableGroup, problem.Shard, nil); err != nil {
```
rafael: we shouldn't have a context background here. 

One more thought, this also seems to be masking another subtle issue. Why `forceHealthCheckShardPoolers` is not returning the must up to date topo pooler for their own state?

go/test/endtoend/multiorch_helpers.go
```diff
@@ -0,0 +1,632 @@
+// Copyright 2025 Supabase, Inc.
```
rafael, line 1: I did not review this. Wondering if @dweitzman  could take a look at this as he has been cleaning up all these kind of helpers.

go/clustermetadata/topo/recovery_lock.go
rafael: PR for this in here: https://github.com/multigres/multigres/pull/310

go/multiorch/coordinator/leader_appointment.go
deepthi: This is a safety check. The ordinary path for a new pooler is to restore from a backup, in which case it will not be "uninitialized". The code exists to ensure exactly what you are saying we should.

go/multiorch/recovery/analysis/appoint_leader_recovery_action.go
deepthi: Good point. We should in fact NOT need a lock for leader appointment.

go/clustermetadata/topo/recovery_lock.go
deepthi: It in fact is being stored in the global topo.

go/multiorch/coordinator/leader_appointment.go
deepthi: It should NOT be gRPC, it's the replication protocol port. We can change it to postgres to match the change I made to multigateway.

go/multiorch/recovery/actions/bootstrap_shard.go
deepthi: I agree with the second statement, but just as we are asserting that multiorch should be able to bootstrap an uninitialized shard, it should also be able to repair a shard where it finds an uninitialized pooler, as long as it has sufficient healthy replicas to meet the durability policy. If we end up with 1 primary and no initialized replicas, then yes, we may have to fall back to manual intervention.

go/multiorch/recovery/recovery_loop.go
deepthi: Whether a particular action needs to take a lock should be encapsulated in the action, it doesn't need to be exposed at all. We should delete these lines altogether. I left them for info purposes, but they are probably more confusing than anything else right now.

go/multiorch/recovery/analysis/appoint_leader_recovery_action.go
rafael: Yeah, that's what I meant 😅

go/clustermetadata/topo/recovery_lock.go
rafael: Ah yes, I see it now.

rafael: Ah yes, I see it now.

go/multiorch/recovery/recovery_loop.go
rafael: That makes sense. We should then fully encapsulated in the action and maybe even remove this from the interface?
```
problem.RecoveryAction.RequiresLock()
```

It seems we are aligning this is a fully internal implementation detail

go/multiorch/recovery/actions/bootstrap_shard.go
rafael: I'm not following this. Adding note, so we can discuss it synchronously. 

In my view, the second statement applies here. To me, the code below can only run if there was a failed bootstrap before. In that case, we need manual intervention and multiorch shouldn't automatically fix it.

go/provisioner/local/local.go
deepthi: Eventually we won't, but right now we don't have an analysis and recovery that will fix up new multipoolers that start and attempt to join the cohort after bootstrap is complete.

go/multiorch/recovery/recovery_loop.go
deepthi: Removed the call to refreshShardMetadata here.

go/multiorch/recovery/actions/bootstrap_shard.go
deepthi: At the end of bootstrap the durability policy should be in force. The only way to guarantee that is to make sure sufficient number of replicas are wired up, so we wire up all of them.

go/multiorch/recovery/actions/bootstrap_shard.go
deepthi: Removed ShardHasNoPrimary analysis for now. We can revisit if we decide that is something we want to detect and handle.

go/multiorch/recovery/analysis/primary_is_dead_analyzer.go
deepthi: If the pooler has never been initialized (for whatever reason, maybe it's brand new), it won't have any primary conn info, and won't be able to detect that primary is dead, so this seems correct to me.

go/test/endtoend/cluster_test.go
deepthi: Fixed.

go/test/endtoend/multiorch_helpers.go
```diff
@@ -0,0 +1,632 @@
...
+
+// setupMultiOrchTestEnv creates a test environment with etcd and topology server
+// for multiorch integration tests.
+func setupMultiOrchTestEnv(t *testing.T, config testEnvConfig) *testEnv {
+	t.Helper()
+
+	ctx := t.Context()
+
+	// Setup test directory - use /tmp to avoid long paths that exceed Unix socket limits (103 bytes on macOS)
+	tempDir, err := os.MkdirTemp("/tmp", config.tempDirPrefix)
```
dweitzman, line 89: FYI: At some point I want to make a general helper for making temp directories with short prefixes so we don't have to keep copying this code

go/test/endtoend/multiorch_helpers.go
```diff
@@ -0,0 +1,632 @@
...
+	// Start pgctld server
+	logFile := filepath.Join(dataDir, "pgctld.log")
+	pgctldCmd := exec.Command("pgctld", "server",
+		"--pooler-dir", dataDir,
+		"--grpc-port", fmt.Sprintf("%d", pgctldGrpcPort),
+		"--pg-port", fmt.Sprintf("%d", pgPort),
+		"--log-output", logFile)
+
+	// Set MULTIGRES_TESTDATA_DIR for directory-deletion triggered cleanup
+	pgctldCmd.Env = append(os.Environ(),
```
dweitzman, line 236: FYI, I also may send out a change I have in a branch somewhere to wrap all these exec.Command() calls in a helper so that things like setting this environment variable are more fluid

go/test/endtoend/multiorch_helpers.go
```diff
@@ -0,0 +1,632 @@
...
+
+// terminateProcess gracefully terminates a process by first sending SIGTERM,
+// waiting for graceful shutdown, and only using SIGKILL if necessary.
+func terminateProcess(t *testing.T, cmd *exec.Cmd, name string, timeout time.Duration) {
+	t.Helper()
+	if cmd == nil || cmd.Process == nil {
+		return
+	}
+
+	// Try graceful shutdown with SIGTERM first
```
dweitzman, line 327: FYI: If I send out the helper for starting processes, it'll also help make graceful termination more simple and consistent

go/test/endtoend/multiorch_helpers.go
dweitzman: Added some FYI comments. There are things I want to clean up more generally, but nothing needs to be done yet in this PR

I suspect in the long term I'll also want to pull out some helpers have similar logic across multiple tests for things like waiting for multipooler to be ready, etc.

GuptaManan100 (approved):

go/multiorch/recovery/actions/appoint_leader.go
```diff
@@ -0,0 +1,144 @@
...
+var _ types.RecoveryAction = (*AppointLeaderAction)(nil)
+
+// AppointLeaderAction handles leader appointment using the coordinator's consensus protocol.
+// This action is used for both repair (mixed initialized/empty nodes) and reelect
+// (all nodes initialized) scenarios. The coordinator.AppointLeader method handles
+// both cases by selecting the most advanced node based on WAL position and running
+// the full consensus protocol to establish a new primary.
+type AppointLeaderAction struct {
+	coordinator *coordinator.Coordinator
+	poolerStore *store.ProtoStore[string, *multiorchdatapb.PoolerHealthState]
```
GuptaManan100, line 44: It would make things more readable if we added type alias for `*store.ProtoStore[string, *multiorchdatapb.PoolerHealthState]` and others like this - `type PoolerHealthStore store.ProtoStore[string, *multiorchdatapb.PoolerHealthState] 
`

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	a.logger.InfoContext(ctx, "successfully initialized primary",
+		"shard_key", problem.ShardKey.String(),
+		"primary", candidate.MultiPooler.Id.Name,
+		"backup_id", resp.BackupId)
+
+	// Set pooler type to PRIMARY after successful initialization
+	// This updates topology so other components (and tests) can observe the node as PRIMARY.
+	changeTypeReq := &multipoolermanagerdatapb.ChangeTypeRequest{
```
GuptaManan100, line 168: Both ChangeType and InitializeEmptyPrimary can be done in a single RPC call. That will prevent an additional round trip

go/multiorch/recovery/analysis/analyzer.go
```diff
@@ -51,3 +55,19 @@ func SetTestAnalyzers(analyzers []Analyzer) {
 func ResetAnalyzers() {
 	defaultAnalyzers = nil
 }
+
+// globalFactory holds the global RecoveryActionFactory instance.
+// This is set during engine initialization and used by analyzers.
+var globalFactory *RecoveryActionFactory
+
```
GuptaManan100, line 62: I don't like global variables 🫤. Could we make this field part of the orchestrator struct and pass that in the places we use this?

go/multiorch/recovery/actions/bootstrap_shard.go
deepthi: Yes, Rafa pointed this out as well and we plan to consolidate them.

go/multiorch/recovery/actions/appoint_leader.go
deepthi: Done in #332

go/multiorch/recovery/actions/bootstrap_shard.go
deepthi: Done in #332

go/multiorch/recovery/analysis/analyzer.go
deepthi: See #332

PR #305, opened 2025-12-01
fix(test): fix race condition flake in GRPC_TracePropagation test

dweitzman:
Flake: https://github.com/multigres/multigres/actions/runs/19812481846

I wasn't able to reproduce the error from CI locally, but it seems like the server span gRPC server handler can do async cleanup after sending a response but before ending its span, so the test can't assume the server span is ready just because the RPC finished.

To avoid flakes, we can use assert.Eventually to poll until the server span is ready.


rafael (approved):
LGTM

PR #307, opened 2025-12-01
fix(test): fix race condition in TestNewConn_InitialFailure

dweitzman:
CI failure: https://github.com/multigres/multigres/actions/runs/19828525214

The test asserted exactly 1 connection attempt immediately after NewWrapperConn returns, but a background retry goroutine starts on initial failure. Since the first retry has no delay (attempt 0), it could execute before the test assertion, resulting in count=2.


rafael (approved):
LGTM

PR #310, opened 2025-12-01
feat(clustermetadata): adds shard locks to topology

rafael:
## Summary
- Port topo locking infrastructure from Vitess to support distributed locks for shard operations
- Add LockNameWithTTL and TryLockName methods to the Conn interface for named locks that don't require the path to exist
- Implement LockShard and TryLockShard methods on the topo Store for coordinating shard-level operations (at the moment only required for bootstrap)

## Test plan
- Unit tests pass: go test ./go/clustermetadata/topo/...
- New tests added for LockNameWithTTL and TryLockName in test/lock.go
- Integration test for shard locking using memorytopo in shard_lock_integration_test.go

### New Conn interface methods:
  - LockNameWithTTL(ctx, dirPath, contents, ttl) - Named lock with configurable TTL (defaults to 24h if ttl=0)
  - TryLockName(ctx, dirPath, contents) - Named lock with fail-fast semantics

### DIFF
--- a/go/clustermetadata/topo/locks.go
+++ b/go/clustermetadata/topo/locks.go
@@ -217,6 +217,7 @@ func (l *Lock) unlock(ctx context.Context, lt iTopoLock, lockDescriptor LockDesc
 	// We need to still release the lock even if the parent
 	// context timed out.
 	span := trace.SpanFromContext(ctx)
+	//nolint:gocritic // Intentionally detaching from parent context - unlock must complete even if parent was cancelled
 	ctx = trace.ContextWithSpan(context.Background(), span)
 	ctx, cancel := context.WithTimeout(ctx, RemoteOperationTimeout)
 	defer cancel()


go/clustermetadata/topo/memorytopo/lock.go
```diff
@@ -121,6 +121,32 @@ func (c *conn) LockName(ctx context.Context, dirPath, contents string) (topo.Loc
...
+// LockNameWithTTL is part of the topo.Conn interface. It behaves the same as LockName
+// as TTLs are not supported in memorytopo.
+func (c *conn) LockNameWithTTL(ctx context.Context, dirPath, contents string, _ time.Duration) (topo.LockDescriptor, error) {
+	// c.factory.callstats.Add([]string{"LockNameWithTTL"}, 1)
+	return c.lock(ctx, dirPath, contents, true)
+}
+
+// TryLockName is part of the topo.Conn interface.
+func (c *conn) TryLockName(ctx context.Context, dirPath, contents string) (topo.LockDescriptor, error) {
+	// c.factory.callstats.Add([]string{"TryLockName"}, 1)
```
dweitzman: FYI, I have a branch with additional linters including detection of commented-out code at https://github.com/multigres/multigres/compare/main...dweitzman:multigres:dw-gocritic but I was planning to wait unless some of my other PRs are reviewed before making it into a PR

rafael: Makes sense. I took the opportunity to start cleaning this up. This was also a placeholder for the time we did not have metrics

go/clustermetadata/topo/shard_lock.go
```diff
@@ -0,0 +1,112 @@
...
+// - an unlock method
+// - an error if anything failed.
+//
+// We are currently only using this method to lock actions that would
+// impact each-other. Most changes of the Shard object are done by
+// UpdateShardFields, which is not locking the shard object. The
+// current list of actions that lock a shard are:
+// * all Multigres-controlled re-parenting operations:
+//   - PlannedReparentShard
+//   - EmergencyReparentShard
```
dweitzman: Does this comments still apply in multigres? I think we can do emergency recovery without depending on etcd being writeable?

go/clustermetadata/topo/memorytopo/lock.go
```diff
@@ -118,7 +112,29 @@ func (c *conn) LockWithTTL(ctx context.Context, dirPath, contents string, _ time
...
 // LockName is part of the topo.Conn interface.
 func (c *conn) LockName(ctx context.Context, dirPath, contents string) (topo.LockDescriptor, error) {
-	// c.factory.callstats.Add([]string{"LockName"}, 1)
+	return c.lock(ctx, dirPath, contents, true)
+}
+
+// LockNameWithTTL is part of the topo.Conn interface. It behaves the same as LockName
+// as TTLs are not supported in memorytopo.
+func (c *conn) LockNameWithTTL(ctx context.Context, dirPath, contents string, _ time.Duration) (topo.LockDescriptor, error) {
+	return c.lock(ctx, dirPath, contents, true)
```
dweitzman: I know this is ported from Vitess, but is there any reason not to support a TTL in the memory topo?

go/clustermetadata/topo/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+func newMetrics() *Metrics {
+	m := &Metrics{
+		meter: otel.Meter("github.com/multigres/multigres/go/clustermetadata/topo"),
+	}
+
+	var err error
+
+	// Counter for lock operations
+	m.lockOps, err = m.meter.Int64Counter(
+		"topo.lock.operations",
```
dweitzman: A recurring philosophical question I don't have the answer to: do we need topo.lock.operations if we have topo.lock.duration, since the duration technically also has a count?

My instinct is to err on the size of omitting to start with since we can always trivially recreate topo.lock.operations later by incrementing it when the duration is updated, but I don't have strong feelings either way

go/test/endtoend/etcd_topo_test.go
```diff
@@ -99,3 +101,56 @@ func testDatabaseLock(t *testing.T, ts topo.Store) {
...
+	require.NoError(t, err, "ConnForCell failed")
+
+	// TryLockName should work on a non-existent path
+	lockPath := "test_try_lock_name"
+	lockDescriptor, err := conn.TryLockName(ctx, lockPath, "test")
+	require.NoError(t, err, "TryLockName failed")
+
+	// TryLockName should fail fast when lock is held (not block)
+	_, err = conn.TryLockName(ctx, lockPath, "again")
+	require.Error(t, err, "TryLockName should fail when lock is held")
```
dweitzman, line 148: Minor observation: it's not obvious in this test that TryLock() didn't block before returning. If there's some easy way to tell that from the error code text, might be worth checking. Not a big deal if not

dweitzman (approved):
I'm hopeful we can avoid the need for a shard lock, but it seems harmless for now

go/clustermetadata/topo/shard_lock.go
```diff
@@ -0,0 +1,112 @@
...
+// * any multiorch recovery e.g
+//   - RecoverDeadPrimary
+//   - ElectNewPrimary
+//   - FixPrimary
+//
+// * operations that we don't want to conflict with re-parenting:
+//   - DeleteTablet when it's the shard's current primary
+//
+// Note: Shard locks use named locks (LockNameWithTTL) because shards don't have
+// dedicated topo entries. Named locks don't require the path to exist.
```
deepthi: I don't understand this. We do have topo entries for shards, don't we?

go/clustermetadata/topo/shard_lock.go
```diff
@@ -0,0 +1,112 @@
...
+// If no TTL is specified via WithTTL option, it defaults to NamedLockTTL (24 hours).
+func (ts *store) LockShard(ctx context.Context, database, tableGroup, shard, action string, opts ...LockOption) (context.Context, func(*error), error) {
```
deepthi: We have another proliferation of the `database, tableGroup, shard` pattern here.

go/clustermetadata/topo/locks.go
```diff
@@ -0,0 +1,373 @@
...
+	// Status is the current status of the Lock.
+	Status string
+}
+
+func init() {
+	servenv.OnParse(registerTopoLockFlags)
+}
+
+func registerTopoLockFlags(fs *pflag.FlagSet) {
+	fs.DurationVar(&RemoteOperationTimeout, "remote-operation-timeout", RemoteOperationTimeout, "time to wait for a remote operation")
```
deepthi, line 60: We have the opportunity to rename this if we want to. If we are importing this flag, we should start using it on all RPCs between multiorch and multipooler. That is how it is used in Vitess.

go/clustermetadata/topo/shard_lock.go
rafael: We don't (at least not yet), the shard info lives in the multigres schema. We are not creating shards in the topo

go/clustermetadata/topo/shard_lock.go
rafael: Fixed!

go/clustermetadata/topo/metrics.go
rafael: I think you are right. I removed this.

go/clustermetadata/topo/shard_lock.go
rafael: This is a good flag. What do you think if we introduce a Shard data structure in the clustermetadata proto? Then use that everywhere where we are trying to address an specific shard (we can deprecate ShardKey in favor of this). 

I can do that whole change in a separate PR as it will be outside the scope of this change

go/test/endtoend/etcd_topo_test.go
rafael: What do you think of the assertion below for this?

go/clustermetadata/topo/locks.go
rafael: I think this is a good idea. Should this be then more like a general timeout? It seems like this one shouldn't be in the topo package.

go/clustermetadata/topo/shard_lock.go
deepthi: Why does it need to be a proto?

go/clustermetadata/topo/locks.go
deepthi: It should also be used for calls to topo RPCs, but it can be moved out into some other common package.

go/clustermetadata/topo/shard_lock.go
deepthi: We talked about this, and we probably need a Shard proto to write into multi schema, so yes, let's go ahead and make it a proto.

PR #313, opened 2025-12-03
feat: add connection pooler and extended query-protocol pipeline from Gateway to Pooler

GuptaManan100:
### Summary
  - Add `connpool` package with mutex-protected stack, waitlist with settings matching, and connection lifecycle management
  - Add `SettingsCache` for settings interning (enables fast pointer equality comparisons)
  - Add prepared statement `Consolidator` to share underlying statements across same queries
  - Add gRPC RPCs for `PrepareStreamExecute` and `Describe`
  - Extended Query Protocol pipeline through multigateway → scatterconn → poolergateway → Pooler

  ### Key Components
  - **connpool**: Generic connection pool with 8 hash buckets for settings-based routing
  - **Settings cache**: Bounded LRU cache ensuring same settings return same pointer
  - **Consolidator**: Deduplicates prepared statements with identical SQL
  - **Connection state**: Tracks portals and reserved connections per client connection

Currently we aren't using the connection pooler. Just created it. To support the extended query protocol, we'll need to implement the client side of the protocol too which is the next thing we'll pick up.


go/tools/list/list.go
dweitzman, line 1: In the new directory structure, this sort of utility would go under go/tools/

go/common/protoutil/protoutil.go
dweitzman, line 1: In the new directory structure, this would go under go/common/protoutil

go/multipooler/connstate/connection_state.go
```diff
@@ -0,0 +1,391 @@
...
+// IsEmpty returns true if there are no variables set.
+func (s *Settings) IsEmpty() bool {
+	if s == nil {
+		return true
+	}
+	return len(s.Vars) == 0
+}
+
+// Equals checks if two Settings are equal (same variables).
+// Note: bucket numbers are NOT compared since two Settings with the same
```
dweitzman: I don't have strong feelings on this, but would it be unreasonable or awkward either rename this to something other than Equals() or split apart the settings storage from the bucket ID so this comment isn't necessary?

Feels a little unexpected to have an Equals() method that intentionally ignores fields

go/multipooler/pools/connpool/interfaces.go
```diff
@@ -0,0 +1,46 @@
...
+	// This is used by the pool for routing connections to the appropriate bucket.
+	Settings() *connstate.Settings
+
+	// IsClosed returns true if the connection has been closed.
+	IsClosed() bool
+
+	// Close closes the connection and releases associated resources.
+	Close() error
+
+	// ApplySettings applies the given settings to the connection by executing
```
dweitzman, line 38: Will the settings be applied within a transaction? Might be useful to document here whether or not reset connection needs to be manually called after apply settings fails

go/multipooler/pools/connpool/pool.go
```diff
@@ -0,0 +1,792 @@
...
+	pool.config.idleTimeout.Store(config.IdleTimeout.Nanoseconds())
+	pool.config.refreshInterval.Store(config.RefreshInterval.Nanoseconds())
+	pool.config.logWait = config.LogWait
+	pool.wait.init()
+
+	return pool
+}
+
+func (pool *Pool[C]) runWorker(close <-chan struct{}, interval time.Duration, worker func(now time.Time) bool) {
+	pool.workers.Add(1)
```
dweitzman: This could use pool.works.Go() instead of manually calling Add() and Done().

go/multipooler/pools/connpool/pool.go
```diff
@@ -0,0 +1,792 @@
...
+	pool.config.refresh = refresh
+	pool.open()
+	return pool
+}
+
+// Close shuts down the pool. No connections will be returned from Pool.Get after calling this,
+// but calling Pool.Put is still allowed. This function will not return until all of the pool's
+// connections have been returned or the default PoolCloseTimeout has elapsed.
+func (pool *Pool[C]) Close() {
+	ctx, cancel := context.WithTimeout(context.Background(), PoolCloseTimeout)
```
dweitzman: When you merge you'll get a lint error here and elsewhere in the file about the context.Background() here. If you want to leave the code unmodified you can use context.TODO(), but maybe we can just drop this method and always make clients call CloseWithContext() instead of guessing what timeout the caller will want?

go/multipooler/pools/connpool/sema_norace.go
dweitzman, line 1: Seems like the Go developers aren't super happy about people linking against this? https://go.dev/src/runtime/sema.go

Unrelated: I don't have strong sentiment on this, but what are your thoughts on whether to leave the semaphore here as a private implementation detail vs moving it under go/tools?

dweitzman (commented):
Haven't read everything, but pushing out some comments before I head out for today

go/multipooler/connstate/connection_state.go
GuptaManan100: That function was actually unused. Because if the Settings are created through the SettingsCache, then they will have the same pointer, and bucket number. Otherwise the bucket number is pretty much useless. Lol.

go/common/protoutil/protoutil.go
GuptaManan100: Fixed

go/tools/list/list.go
GuptaManan100: Fixed

go/multipooler/pools/connpool/pool.go
GuptaManan100: Fixed

go/multipooler/pools/connpool/interfaces.go
GuptaManan100: The connection pool already handles what happens when setting or resetting connection fails. Whether to apply the settings in a transaction or not isn't something the pool cares about. The implementation of the interface can take its own call. I don't see why they'd need to be applied in a transaction though

go/multipooler/pools/connpool/pool.go
GuptaManan100: I didn't get a lint error now, and we have a timeout don't we? We are adding it right there. Is the linter going to complain about the usage of context.Background() itself? Isn't this standard pattern to have a timeout on a context and use Background for the base for it? Doesn't Background have to be the eventual Parent of all contexts?

go/multipooler/pools/connpool/sema_norace.go
GuptaManan100: I think we should leave it private for now. It's specifically used in the waitlist of the connection pooler to hand off connections directly to the one waiting in the list instead of returning them and waiting for the pooler to hand it off. I don't think there are many other use cases that would want to use this. I've picked this from Vitess as is, otherwise I probably would have used channels or something else. This is an optimisation Vitess has for performance.

go/multipooler/pools/connpool/pool.go
GuptaManan100: No actually you're right, I see the lint error, but I didn't see it before 😕

dweitzman: I've updated linting to reflect the spirit of https://pkg.go.dev/context#Background: "Background returns a non-nil, empty [Context](https://pkg.go.dev/context#Context). It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests"

Some downsides to using Background() in other places:

1) It makes timeout behavior less correct while often simultaneously making code more complex and arbitrary by inventing additional, new timeouts. It should be the responsibility of a caller to know how much time is available, not the method being called. If I do `service.Init(ctxWithTimeout)` and it throws away the context to call `service.internalInitWithoutContext()` which has to make a new context, `ctxWithTimeout` is no longer respected and can be exceeded plus more code was written in the end to build a new context and timeout 

2) It makes telemetry data less useful. With context any log entry, RPC, etc. can be associated with the operation it belongs to: we could potentially see detail in a trace like that a particular failure to close a connection pool happened in the context of a particular planned reparent operation started by orchestrator. If we're not in the habit of propagating context, we miss out on a lot potential debugging power from being able to understand the bigger context in which a metric or log line occurred.

3) This is more nebulous, but preserving context across calls gives me flexibility to customize behavior in ways we might find a use for in the future. Maybe someday we'll realize we want connection pool closing to behave slightly differently if it was triggered by a particular orchestrator operation, and with propagated context all the way from the top-level RPC it'd be possible for even low-level operations to have a better understanding of their context. I'm not going to claim this is good or bad for any particular scenario, but it does create a powerful technical possibility that has potentially valid use cases that won't exist without propagating context down from the top level.

go/multipooler/pools/connpool/pooled.go
```diff
@@ -0,0 +1,74 @@
...
+	case p.Conn.IsClosed():
+		p.pool.put(nil)
+	default:
+		p.pool.put(p)
+	}
+}
+
+// Taint marks this connection as unusable and removes it from the pool.
+// The connection will be closed and a new one created when recycled.
+func (p *Pooled[C]) Taint() {
```
dweitzman, line 63: Is it important whether `Pooled` is thread-safe? Hypothetically if two people called Taint() in parallel here it could lead to `p.pool.put(nil)` being called twice instead of once

go/multipooler/pools/connpool/sema_norace.go
dweitzman: The sema code is no longer in vitess, so my preference would be to use the less-hacky approach vitess uses now unless we think there's a compelling reason why the previous approach in vitess is better than the current approach

go/multipooler/pools/connpool/pool.go
```diff
@@ -0,0 +1,789 @@
...
+	// update the idle count to match the new capacity if necessary
+	// wait for connections to be returned to the pool if we're reducing the capacity.
+	defer pool.setIdleCount()
+
+	const delay = 10 * time.Millisecond
+
+	// close connections until we're under capacity
+	for pool.active.Load() > newcap {
+		if err := ctx.Err(); err != nil {
+			return errors.New("timed out while waiting for connections to be returned to the pool")
```
dweitzman, line 741: Not at all urgent and probably out of scope for porting code, but seemed worth mentioning:

1) I think it makes more sense to close connections in put() then during setCapacity(). That'll ensure they get closed event if setCapacity() runs out of time in the context
2) I'm not convinced setCapacity() even needs to take a context and wait -- in vitess it seems like the only caller of setCapacity() is `debugenv.go` for people posting changes to settings in the admin page? That doesn't feel like enough justification to have this complexity in connection pools. People using that page aren't in the middle of some complex operation, so they don't need to know with 100% certainty that all the excess connections were killed immediately
3) If we did want to support waiting for the capacity decrease, I think it'd still be cleaner to have put() close excess connections and have some separate method or variation like SetCapacityAwaitingClose() that's capable of doing the waiting

dweitzman (approved):
Reviewing ported code is a little messy: things that looked copied I didn't necessarily read in full detail. I'd rather unblock you and if there are cleanups to do we can make them later

go/multipooler/pools/connpool/sema_norace.go
GuptaManan100: Yes, I agree with you, I'm changing the code now 😄

go/multipooler/pools/connpool/pooled.go
GuptaManan100: I don't think they need to be thread safe. Once a connection is gotten, it shouldn't be shared by different go routines. If another go routine needs a connection they should also get it from the pool. We can add a mutex if we want to make it thread safe, but I don't think it is required.

go/multipooler/pools/connpool/pool.go
GuptaManan100: Okay, I've fixed it 👍

go/multipooler/pools/connpool/pool.go
GuptaManan100: I will think about this too, and make a change in a follow up PR!

PR #314, opened 2025-12-03
refactor: eliminate most os.Exit() calls for better testability, telemetry, and cleanup

dweitzman:
I have two concerns with `os.Exit()`:
- It prevents shutdown hooks from running, so for example a buffered log entry or unexported telemetry data might be lost, or cleaning up subprocesses or unix sockets on the filesystem might be missed, etc.
- It prevents writing unit tests

There's a possible argument that `os.Exit()` might be harmless sometimes in initialization code when a binary is starting up for the first time, but I'm advocating for the position that using normal error codes and normal graceful exits is a cleaner, safer, more consistent default unless there's a compelling reason to use `os.Exit()`, and I don't think all of these uses of `os.Exit()` have a compelling reason.

What's changed in this PR:
- Refactor `servenv.Init()`, `servenv.Run()`, `topo.Open()`, and related functions to return errors instead of calling `os.Exit()`
- Update all service init files (`multiadmin`, `multigateway`, `multiorch`, `multipooler`) to propagate errors
- Update all `cmd/*/main.go` files to handle errors from `Init()` and call `os.Exit()` only at the top level
- Add `forbidigo` linter rule to prevent future `os.Exit()` calls in non-cmd packages
- Introduce `ErrorHooks` mechanism in `go/tools/event/` for graceful error handling than `Hooks`
- Use panic for programming errors in `init()` functions (unreachable code paths) instead of `os.Exit()`
- Add new tests that weren't possible before like `main_test.go` files for each service, plus tests for gRPC auth and pprof error conditions


deepthi:
Let's get Sugu to review this one.

go/common/servenv/servenv.go
```diff
@@ -166,25 +167,14 @@ func (se *ServEnv) SetListeningURL(u url.URL) {
...
 func (se *ServEnv) PopulateListeningURL(port int32) {
-	host, err := netutil.FullyQualifiedHostname()
-	if err != nil {
-		slog.Warn("Failed to get fully qualified hostname, falling back to simple hostname",
-			"error", err,
-			"note", "This may indicate DNS configuration issues but service will continue normally")
-		host, err = os.Hostname()
-		if err != nil {
-			slog.Error("os.Hostname() failed", "err", err)
-			os.Exit(1)
```
dweitzman, line 179: os.Hostname() was only used here for logging, and it seemed excessive to kill the whole process because a log-only call fails. Code's changed this to log the same hostname that's actually used as the listening URL and the os.Exit() call is removed entirely

dweitzman:
> Let's get Sugu to review this one.

cc @sougou

sougou:
No objections here.

sougou (approved):

PR #317, opened 2025-12-04
refactor(pgctld): move pgctld under go/services

dweitzman:
- Move `go/pgctld` to `go/services/pgctld` as part of directory structure refactoring
- Add `pgctld-isolation` depguard rule to enforce that only `go/services/pgctld` and `go/cmd/pgctld` can import the package
- Inline path helpers in endtoend tests to satisfy the new constraint. It doesn't seem necessary for the endtoend tests to depend on go/services/pgctld and there's already precedent for other unit tests knowing the internal directory structure used by pgctld

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -83,6 +83,17 @@ linters:
           deny:
             - pkg: "github.com/multigres/multigres/go/services/multiadmin"
               desc: "multiadmin can only be imported by its own package and cmd"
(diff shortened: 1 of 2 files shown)


dweitzman:
@rafael I realized I didn't `push` all much changes. Some things described in the PR description weren't in the code until now: the depguard rule and inlining the test path handlers

PR #319, opened 2025-12-04
build(lint): enable the noctx linter with context.TODO() as a placeholder

dweitzman:
This doesn't "fix" all the existing calls that don't intentionally clarify if they have a timeout, but it'll at least cause people in the future to consider whether a context makes sense

Cleaning up context.TODO() across the codebase will be a longer effort


dweitzman:
I'm going to close this for now -- I don't think it makes sense to enable it until we have a better solution for dealing with commands that doesn't involve sending KILL signals by default, which is what CommandContext() would do. I do have a PoC in a branch somewhere for graceful termination of commands on context cancellation, but it doesn't feel like the highest priority for using code review resources right now

PR #322, opened 2025-12-04
test(endtoend): dump service logs on multipooler test failure

dweitzman:
When multipooler end-to-end tests fail, service logs are now printed before cleanup. This helps debug CI failures where processes crash unexpectedly, such as the cascade failure in PR #319 where the multipooler died mid-test but logs were unavailable.

This PR also modifies the test setup helper to verify that processes are still running so we can give a cleaner error if a process crashed instead of just failing to send gRPC requests to it

### DIFF
--- a/go/test/endtoend/multipooler/setup_test.go
+++ b/go/test/endtoend/multipooler/setup_test.go
@@ -385,6 +385,21 @@ func (p *ProcessInstance) Stop() {
 	_ = p.Process.Wait()
 }
 
+// IsRunning checks if the process is still running.
+// Returns false if the process has exited or was never started.
+func (p *ProcessInstance) IsRunning() bool {
+	if p == nil || p.Process == nil || p.Process.Process == nil {
+		return false
+	}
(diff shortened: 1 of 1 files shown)


dweitzman:
This is inspired by https://github.com/multigres/multigres/actions/runs/19939905226?pr=319 which appears to be a flake caused by the shared test setup multipooler crashing

deepthi (approved):

rafael:
@dweitzman I think this is a great idea. One comment I had. I think we now have multiple ways of doing something similar:
- https://github.com/multigres/multigres/blob/main/go/test/endtoend/multipooler_client.go#L562
- https://github.com/multigres/multigres/blob/main/go/test/endtoend/bootstrap_test.go#L81
- https://github.com/multigres/multigres/blob/main/go/test/endtoend/multipooler/setup_test.go#L351

With this latest change, I feel we should consolidate on this new approach and delete all that other code. As dumpLogs should be all we need (I think)

PR #325, opened 2025-12-05
fix(consensus): primary should demote itself when accepting a new term

deepthi:
This was missed in the original implementation in #240.

### DIFF
--- a/go/multipooler/manager/rpc_consensus.go
+++ b/go/multipooler/manager/rpc_consensus.go
@@ -26,13 +26,15 @@ import (
 )
 
 // BeginTerm handles coordinator requests during leader appointments.
-// Accepting a new term means this node will no longer accept requests from
+// Accepting a new term means this node will not accept any new requests from
 // the current term. This is the revocation step of consensus, which applies
 // to both primaries and standbys:
 //
 //   - If this node is a primary, it must demote itself as part of revocation.
 //   - If this node is a standby, it should break replication as part of
-//     revocation (TODO: not implemented yet).
+//     revocation. However, breaking replication on all standbys before the primary
+//     is demoted will have the effect of blocking writes to the primary,
+//     so this has to be done at the proper time (TODO: not implemented yet).
 func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatapb.BeginTermRequest) (*consensusdatapb.BeginTermResponse, error) {
 	// Acquire the action lock to ensure only one consensus operation runs at a time
 	// This prevents split-brain acceptance and ensures term updates are serialized
@@ -132,7 +134,7 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
 
 	response.Accepted = true
 
-	// CRITICAL: If we were a primary and accepted a higher term, demote immediately.
+	// If we were a primary and accepted a higher term, demote immediately.
 	// This prevents split-brain by ensuring the old primary stops accepting writes
 	// before any new primary is promoted.
 	if wasPrimary && req.Term > currentTerm {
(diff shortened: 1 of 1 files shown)


go/test/endtoend/beginterm_demote_test.go
```diff
@@ -0,0 +1,240 @@
...
+	t.Logf("Created 3 empty nodes")
+
+	// Setup pgbackrest
+	env.setupPgBackRest()
+
+	// Register all nodes in topology
+	env.registerNodes()
+
+	// Start multiorch for initial bootstrap only
+	multiOrchCmd := env.startMultiOrch()
```
rafael: shouldn't we also add here a t.cleanup(terminateProcess) ?

go/test/endtoend/beginterm_demote_test.go
```diff
@@ -0,0 +1,240 @@
+// Copyright 2025 Supabase, Inc.
```
rafael: nit - on naming. Should this be: `rpc_consensus_begin_test.go`

So it is clear that this is part of the tests of that service.

Alternatively, we could add these tests to `rpc_consensus_test` that already exists.

go/multipooler/manager/rpc_consensus.go
```diff
@@ -114,6 +126,35 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
+			"old_term", currentTerm,
+			"new_term", req.Term,
+			"candidate_id", req.CandidateId.GetName())
+
+		// Use a reasonable drain timeout for demotion
+		drainTimeout := 5 * time.Second
+		demoteLSN, demoteErr := pm.demoteLocked(ctx, req.Term, drainTimeout)
+		if demoteErr != nil {
+			// Log the error but still return accepted=true
+			// The term was accepted, demotion is a best-effort cleanup
```
rafael, line 151: If the primary fails to demote, returning Accepted: true  seems problematic. The node hasn't actually completed the transition to a safe state. From the coordinator's perspective, it thinks it has a recruited node that's no longer a primary, but in reality postgres is still accepting writes.

Should we return accepted false in this case?  

The coordinator can still proceed if it has quorum from other nodes, and the failed node can be dealt with separately.

go/multipooler/manager/rpc_consensus.go
```diff
@@ -114,6 +126,35 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
+			pm.logger.ErrorContext(ctx, "Demotion failed after accepting term",
+				"error", demoteErr,
+				"term", req.Term)
+			// Don't fail the BeginTerm - the term acceptance is the critical part
```
rafael: I'm not agreeing with this, per comment above. But maybe I'm missing something

go/multipooler/manager/rpc_consensus.go
rafael, line 85: I wonder if WasPrimary should be part of the response. This could be information that multiorc uses to make decisions.

go/multipooler/manager/rpc_manager.go
```diff
@@ -909,6 +909,83 @@ func (pm *MultiPoolerManager) Demote(ctx context.Context, consensusTerm int64, d
 	}, nil
 }
 
+// demoteLocked performs the core demotion logic.
+// REQUIRES: action lock must already be held by the caller.
+// This is used by BeginTerm to demote inline without re-acquiring the lock.
+func (pm *MultiPoolerManager) demoteLocked(ctx context.Context, consensusTerm int64, drainTimeout time.Duration) (string, error) {
```
rafael: I think the Demote rpc, should also use htis method now. So we don't have duplication.

go/multipooler/manager/rpc_consensus.go
rafael: This one I did not fix in my branch. But when can follow up in a different PR

deepthi: I was doubtful about this too. So let's fix it up. It requires a bit of refactoring and moving code around so it makes sense to do another PR for it.

go/multipooler/manager/rpc_consensus.go
deepthi: Let's think about what multiorch might do with this response. We can always add it when we need it.

PR #326, opened 2025-12-05
fix(consensus): primary should demote itself when accepting a new term (CI - TEST)

rafael:
# Desc
- Iterates on feedback for PR.
- Fixes tests
- Makes sure cleanup for multipooler integration tests also resets term and leaves primary server with the correct metadata


deepthi:
Closing since #325 has been merged.

PR #327, opened 2025-12-08
feat(pgprotocol): add PostgreSQL client protocol implementation

GuptaManan100:
### Description

- Adds a PostgreSQL wire protocol client library for multipooler → PostgreSQL communication
- Implements both simple and extended query protocols with full authentication support (SCRAM-SHA-256, MD5, cleartext)
- Includes streaming result set handling and proper error propagation

### DIFF
--- a/go/multipooler/manager/rpc_consensus.go
+++ b/go/multipooler/manager/rpc_consensus.go
@@ -25,7 +25,16 @@ import (
 	consensusdatapb "github.com/multigres/multigres/go/pb/consensusdata"
 )
 
-// BeginTerm handles coordinator requests during leader appointments
+// BeginTerm handles coordinator requests during leader appointments.
+// Accepting a new term means this node will not accept any new requests from
+// the current term. This is the revocation step of consensus, which applies
+// to both primaries and standbys:
+//
+//   - If this node is a primary, it must demote itself as part of revocation.
+//   - If this node is a standby, it should break replication as part of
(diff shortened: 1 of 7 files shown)


go/pgprotocol/client/conn.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// SetConnectionState sets the connection-specific state.
+// This allows callers to store their own state per connection.
+func (c *Conn) SetConnectionState(state any) {
+	c.state = state
+}
+
+// flush flushes any buffered writes.
+func (c *Conn) flush() error {
```
dweitzman, line 221: Not a blocking comment, but for hardening I'm wondering what would stop this from potentially blocking forever if the server is slow to receive the data.

go/pgprotocol/client/extended.go
```diff
@@ -0,0 +1,834 @@
...
+			if !gotCloseComplete {
+				return fmt.Errorf("did not receive CloseComplete")
+			}
+			return nil
+
+		case protocol.MsgErrorResponse:
+			return c.handleErrorAndWaitForReady(body)
+
+		case protocol.MsgNoticeResponse:
+			// Ignore notices.
```
dweitzman, line 382: Wondering if in the future we'll need to propagate notices from multipooler all the way up to multigateway responses.

If someone writes a custom stored procedure that sends a notice, they'd probably expect that notice to arrive even though a connection pooler. Maybe there's a TODO to put here?

PR #328, opened 2025-12-08
feat(multiorc): bootstrap action should wait to discover enough poolers

rafael:
## Summary

The main change in this PR is to do not required a specific order on how components are provisioned to get the bootstrap action to work correctly. This is important both for local but for the actual kubes operator.

Tangentially  to this changes we have: 
- Add a quorum check in the bootstrap action so it waits for enough poolers before attempting shard bootstrap
- Parallelize service provisioning in the local provisioner now that bootstrap has quorum protection (this gave a significant boost in the local setup for me (from 3s to 1s).
- Make multiorch discovery, health check, and recovery intervals configurable, with 500ms defaults for local development and tests
##  Tests
 - Added relevant unit tests

### DIFF
--- a/go/multiorch/recovery/actions/bootstrap_shard.go
+++ b/go/multiorch/recovery/actions/bootstrap_shard.go
@@ -36,6 +36,9 @@ import (
 // Compile-time assertion that BootstrapShardAction implements types.RecoveryAction.
 var _ types.RecoveryAction = (*BootstrapShardAction)(nil)
 
+// DefaultStatusRPCTimeout is the default timeout for individual Status RPC calls.
+const DefaultStatusRPCTimeout = 5 * time.Second
+
 // BootstrapShardAction handles bootstrap initialization of a new shard from scratch.
 // This action assumes all nodes in the cohort are empty (uninitialized).
 // It will:
@@ -46,13 +49,14 @@ var _ types.RecoveryAction = (*BootstrapShardAction)(nil)
 // 5. Create the durability policy in the database
 // 6. Initialize remaining nodes as standbys
 type BootstrapShardAction struct {
-	rpcClient   rpcclient.MultiPoolerClient
-	poolerStore *store.ProtoStore[string, *multiorchdatapb.PoolerHealthState]
-	topoStore   topoclient.Store
-	logger      *slog.Logger
+	rpcClient        rpcclient.MultiPoolerClient
+	poolerStore      *store.ProtoStore[string, *multiorchdatapb.PoolerHealthState]
+	topoStore        topoclient.Store
+	logger           *slog.Logger
+	statusRPCTimeout time.Duration
 }
 
-// NewBootstrapShardAction creates a new bootstrap action
+// NewBootstrapShardAction creates a new bootstrap action with default settings.
 func NewBootstrapShardAction(
 	rpcClient rpcclient.MultiPoolerClient,
 	poolerStore *store.ProtoStore[string, *multiorchdatapb.PoolerHealthState],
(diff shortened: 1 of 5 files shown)


go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -349,6 +363,19 @@ func (a *BootstrapShardAction) initializeSingleStandby(ctx context.Context, node
 	return nil
 }
 
+// countReachablePoolers counts how many poolers in the cohort are reachable via RPC.
+func (a *BootstrapShardAction) countReachablePoolers(ctx context.Context, cohort []*multiorchdatapb.PoolerHealthState) int {
+	count := 0
+	for _, pooler := range cohort {
+		req := &multipoolermanagerdatapb.StatusRequest{}
+		_, err := a.rpcClient.Status(ctx, pooler.MultiPooler, req)
```
cuongdo: does this need timeout logic for handling slow/unresponsive MultiPoolers?

rafael: Good call. Addressed this and added a way for the fake_client.go to start testing this kind of failures.

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -36,6 +36,9 @@ import (
 // Compile-time assertion that BootstrapShardAction implements types.RecoveryAction.
 var _ types.RecoveryAction = (*BootstrapShardAction)(nil)
 
+// DefaultStatusRPCTimeout is the default timeout for individual Status RPC calls.
+const DefaultStatusRPCTimeout = 5 * time.Second
```
deepthi, line 40: Any scattered RPC timeouts we already have in code should be replaced with RemoteOperationTimeout.

go/multiorch/config/config.go
```diff
@@ -277,14 +292,16 @@ func (c *Config) RegisterFlags(fs *pflag.FlagSet) {
 	fs.Duration("cluster-metadata-refresh-timeout", c.DefaultClusterMetadataRefreshTimeout(), "timeout for cluster metadata refresh operation")
 	fs.Duration("pooler-health-check-interval", c.DefaultPoolerHealthCheckInterval(), "interval between health checks for a single pooler")
 	fs.Int("health-check-workers", c.DefaultHealthCheckWorkers(), "number of concurrent workers polling pooler health")
+	fs.Duration("recovery-cycle-interval", c.DefaultRecoveryCycleInterval(), "interval between recovery cycles")
```
deepthi, line 295: This needs a better description. What is this field used for? I couldn't find any usages by searching this diff.

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -349,6 +375,23 @@ func (a *BootstrapShardAction) initializeSingleStandby(ctx context.Context, node
 	return nil
 }
 
+// countReachablePoolers counts how many poolers in the cohort are reachable via RPC.
+func (a *BootstrapShardAction) countReachablePoolers(ctx context.Context, cohort []*multiorchdatapb.PoolerHealthState) int {
+	count := 0
+	for _, pooler := range cohort {
+		// Use a per-RPC timeout to prevent a single slow/unresponsive pooler from
+		// consuming the entire action timeout.
```
deepthi, line 383: We should probably do this in parallel goroutines.

go/multiorch/config/config.go
deepthi: I see where we are using it, but the first comment still stands.

PR #331, opened 2025-12-09
feat(multiorch): adds action to wire replication

rafael:
## Desc

Adds automatic replication repair to multiorch. When a replica's replication is broken or misconfigured, multiorch can now detect the problem and automatically wire the replica to the correct primary.

**Key changes:**
  - New `FixReplicationAction` that configures `primary_conninfo` and adds replicas to the synchronous standby list
  - Two new analyzers to detect replication problems
  - Recovery action metrics for observability
  - E2E test coverage

  ## Details

  ### New Recovery Action: FixReplication

  The `FixReplicationAction` handles two problem types:
  - **ReplicaNotReplicating**: Replica has no `primary_conninfo` configured
  - **ReplicaNotInStandbyList**: Replica is replicating but not in the primary's `synchronous_standby_names`

The action is fully idempotent - multiple multiorch instances can race to fix the same problem and the result will be identical. Operations are serialized by action locks at the pooler level.

  ### New Analyzers

  - `ReplicaNotReplicatingAnalyzer`: Detects replicas without replication configured
  - `ReplicaNotInStandbyListAnalyzer`: Detects replicas missing from the standby list

  ### Testing
  - Unit tests for all new components
  - E2E test (`fix_replication_test.go`) that:
    - Breaks replication by killing postgres on replica
    - Verifies multiorch detects and fixes the problem
    - Confirms data replicates after fix
    - Tests repeated breakage recovery
    - Tests standby list removal detection.
    - **NOTE**: The integration test should be moved to multiorch endtoend package to make it consistent with the latest changes in main, but I'll do that in a follow up PR.


go/test/endtoend/fix_replication_test.go
```diff
@@ -0,0 +1,296 @@
+// Copyright 2025 Supabase, Inc.
```
rafael, line 1: Ideally this test can be moved to endtoend/multiorch with some shared setup for all multiorc tests. @deepthi  I will for your PR for that. But this is a basic test that we can port there once your PR related to this lands

go/multiorch/recovery/actions/fix_replication.go
```diff
@@ -0,0 +1,356 @@
...
+		"primary", primary.MultiPooler.Id.Name,
+		"consensus_term", consensusTerm)
+
+	// Configure primary_conninfo on the replica
+	req := &multipoolermanagerdatapb.SetPrimaryConnInfoRequest{
+		Host:                  primary.MultiPooler.Hostname,
+		Port:                  primaryPort,
+		StopReplicationBefore: true,
+		StartReplicationAfter: true,
+		CurrentTerm:           consensusTerm,
```
rafael, line 180: This seemed sensible, but we should discuss. The way `SetPrimaryConnInfoRequest` works today, requires you to pass a valid term, otherwise it will reject the request.

go/multiorch/recovery/actions/bootstrap_shard_test.go
```diff
@@ -350,13 +350,14 @@ func TestBootstrapShardAction_ConfiguresSyncReplication(t *testing.T) {
 	assert.Len(t, mockClient.syncReplicationRequest.StandbyIds, 2,
 		"Should configure both standbys in the sync replication list")
 
-	// Verify the standby IDs are correct (order doesn't matter)
+	// Verify the standby IDs are valid pooler names (primary selection is non-deterministic)
```
rafael, line 353: @cuongdo  this should fix this flake: https://github.com/multigres/multigres/actions/runs/20077317166/job/57595035056?pr=333#step:6:632

PR #332, opened 2025-12-09
fix(consensus): configure synchronous replication after bootstrap if needed, and other improvements

deepthi:
1. Bootstrap now automatically configures synchronous replication on the primary after standbys are initialized
  2. `BeginTerm` Consensus Protocol Improvements
  - Demote-before-accept: Primaries now demote themselves before accepting a new term, preventing split-brain scenarios where a demoted primary continues serving writes
  - Atomic term+acceptance: Combined term update and candidate acceptance into a single file write operation for atomicity
  - Standby replication reset: Standbys now reset primary_conninfo when accepting a new term, breaking connection with the old primary
  - Added comprehensive tests for demotion failures, successful demotions, and WAL receiver checks
3. Bootstrap RPC Consolidation
  - Consolidated `ChangeType` and `CreateDurabilityPolicy` calls into `InitializeEmptyPrimary` RPC
  - Fixed error handling in `configureSynchronousReplication` to return errors instead of logging and continuing
4. Cleanup & Refactoring
  - Moved multiorch e2e tests to dedicated go/test/endtoend/multiorch/ package
  - Moved `TestMain` functions to separate main_test.go files in each test package
  - Consolidated bootstrap tests using shared `setupMultiOrchTestEnv` helper
  - Removed global factory: Replaced `globalFactory` variable with dependency injection via struct fields in analyzers
  - Added `PoolerHealthStore` type alias

### DIFF
--- a/go/multipooler/manager/rpc_consensus.go
+++ b/go/multipooler/manager/rpc_consensus.go
@@ -141,20 +141,12 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
 			var lastMsgReceiptTime *time.Time
 			err = pm.db.QueryRowContext(ctx, "SELECT last_msg_receipt_time FROM pg_stat_wal_receiver").Scan(&lastMsgReceiptTime)
 			if err != nil {
-				// No WAL receiver (disconnected standby) - update term but reject acceptance
-				pm.logger.WarnContext(ctx, "Rejecting term acceptance: standby has no WAL receiver",
-					"term", req.Term,
-					"error", err)
-				// Still update term if higher, but don't accept
-				if req.Term > currentTerm {
-					if err := pm.validateAndUpdateTerm(ctx, req.Term, false); err != nil {
-						return nil, fmt.Errorf("failed to update term: %w", err)
-					}
-					response.Term = req.Term
-				}
-				return response, nil
-			}
-			if lastMsgReceiptTime != nil {
+				// No WAL receiver (disconnected standby) - this is EXPECTED during failover
+				// when the primary just died. Don't reject - proceed with acceptance.
+				// The standby's data may still be recent even without an active WAL receiver.
+				pm.logger.InfoContext(ctx, "Standby has no WAL receiver, proceeding with term acceptance",
+					"term", req.Term)
+			} else if lastMsgReceiptTime != nil {
 				timeSinceLastMessage := time.Since(*lastMsgReceiptTime)
 				if timeSinceLastMessage > 30*time.Second {
 					// We're too far behind in replication, update term but don't accept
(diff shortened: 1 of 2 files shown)


go/multiorch/coordinator/sync_replication.go
```diff
@@ -0,0 +1,138 @@
...
+	"log/slog"
+
+	"github.com/multigres/multigres/go/common/mterrors"
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
+	multiorchdatapb "github.com/multigres/multigres/go/pb/multiorchdata"
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+)
+
+// BuildSyncReplicationConfig creates synchronous replication configuration based on the quorum policy.
```
rafael, line 28: I did not review this, assuming this is just the rename.

go/multipooler/manager/rpc_consensus.go
```diff
@@ -101,66 +102,101 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
 	}
 
+	// Atomically update term AND accept candidate in single file write
+	// This combines what used to be two separate file writes
+	if err := cs.UpdateTermAndAcceptCandidate(ctx, req.Term, req.CandidateId); err != nil {
+		return nil, fmt.Errorf("failed to update term and accept candidate: %w", err)
+	}
+
+	response.Accepted = true
+
```
rafael, line 199: small nit - we should probably refactor the logic here between primary / standby, as there quite a few if / else depedning on the type.

Seems simpler / better if we check this early in the method and delegate to different helpers depending on the type

go/multiorch/recovery/actions/bootstrap_shard.go
```diff
@@ -189,38 +193,10 @@ func (a *BootstrapShardAction) Execute(ctx context.Context, problem types.Proble
...
 
-	a.logger.InfoContext(ctx, "successfully initialized primary",
+	a.logger.InfoContext(ctx, "successfully initialized primary with durability policy",
 		"shard_key", problem.ShardKey.String(),
 		"primary", candidate.MultiPooler.Id.Name,
-		"backup_id", resp.BackupId)
-
-	// Set pooler type to PRIMARY after successful initialization
-	// This updates topology so other components (and tests) can observe the node as PRIMARY.
-	changeTypeReq := &multipoolermanagerdatapb.ChangeTypeRequest{
```
rafael, line 199: This was pushed down to rpc_initialization

go/multipooler/manager/rpc_initialization.go
```diff
@@ -124,6 +125,19 @@ func (pm *MultiPoolerManager) InitializeEmptyPrimary(ctx context.Context, req *m
 	}
 	pm.logger.InfoContext(ctx, "Initial backup created", "backup_id", backupID)
 
+	// Set pooler type to PRIMARY
```
rafael, line 128: we should update unit tests to assert this happening

PR #334, opened 2025-12-09
feat(pgprotocol): add SCRAM-SHA-256 auth library with passthrough support

dweitzman:
## Summary

Implements SCRAM-SHA-256 authentication for PostgreSQL protocol connections, enabling Multigres to verify client credentials and extract SCRAM keys for passthrough authentication to backend servers.

This PR includes both server-side and client-side SCRAM, with the client protocol code (`client/scram.go`) refactored to use the shared library.

## Key Features

- **Core SCRAM-SHA-256 protocol**: Full RFC 5802 compliant implementation
- **ClientKey extraction**: Recovers ClientKey from client's proof for SCRAM passthrough authentication
  - Enables multigateway to authenticate clients, then use extracted keys to authenticate as that user to PostgreSQL
  - Eliminates need to store plaintext passwords
- **Client-side passthrough**: `SCRAMClient` supports both password and key-based authentication modes
- **Security hardening**: 
  - Enforces minimum 4096 PBKDF2 iterations
  - Requires minimum 8-byte salts
  - Constant-time comparisons to prevent timing attacks
  - Proper error distinction (authentication failures vs programming errors)
- **context.Context support**: Allows timeout/cancellation during credential lookup

## Architecture

- `authenticator.go` - Stateful server-side SCRAM authenticator
- `scram_client.go` - Client-side SCRAM with password and passthrough modes
- `scram_crypto.go` - RFC 5802 cryptographic primitives
- `password.go` - PostgreSQL password hash parsing/validation
- `scram.go` - SCRAM message parsing and generation
- `doc.go` - Package documentation and design rationale
- `client/scram.go` - Thin wrapper for protocol I/O (refactored to use shared library)

## Why Custom Implementation?

Existing Go SCRAM libraries (xdg-go/scram, lib/pq, jackc/pgx) don't support our critical requirement: ClientKey extraction for passthrough authentication. See [doc.go](go/pgprotocol/scram/doc.go#L38-L54) for detailed comparison.

## What's NOT Included (future work)

- Integration with multigateway/multipooler (wiring up passthrough flow)
- Credential caching in multigateway

### DIFF
--- a/go/pgprotocol/scram/scram_crypto_test.go
+++ b/go/pgprotocol/scram/scram_crypto_test.go
@@ -205,65 +205,6 @@ func TestComputeServerSignature(t *testing.T) {
 	})
 }
 
-func TestVerifyClientProof(t *testing.T) {
-	t.Run("valid proof is verified", func(t *testing.T) {
-		// Simulate a full SCRAM exchange
-		password := "pencil"
-		salt := []byte("testsalt")
-		iterations := 4096
-		authMessage := "n=user,r=clientnonce,r=clientnonce+servernonce,s=dGVzdHNhbHQ=,i=4096,c=biws,r=clientnonce+servernonce"
-
-		// Server side: compute stored values from password
-		saltedPassword := ComputeSaltedPassword(password, salt, iterations)
-		clientKey := ComputeClientKey(saltedPassword)
-		storedKey := ComputeStoredKey(clientKey)
-		serverKey := ComputeServerKey(saltedPassword)
-
-		// Client side: compute proof
-		clientSignature := ComputeClientSignature(storedKey, authMessage)
-		clientProof, err := computeClientProof(clientKey, clientSignature)
-		require.NoError(t, err)
-
-		// Server side: verify proof
-		_, err = ExtractAndVerifyClientProof(storedKey, authMessage, clientProof)
-		assert.NoError(t, err)
-
-		// Also compute server signature for mutual auth
-		serverSignature := ComputeServerSignature(serverKey, authMessage)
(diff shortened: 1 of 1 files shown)


dweitzman:
Leaving as draft until I finish reviewing all of AI-written bits, but it's probably mostly reasonable

go/pgprotocol/scram/authenticator.go
```diff
@@ -0,0 +1,275 @@
...
+)
+
+// Sentinel errors for SCRAM authentication.
+var (
+	// ErrUserNotFound indicates the user does not exist.
+	ErrUserNotFound = errors.New("user not found")
+
+	// ErrAuthenticationFailed indicates the password proof was invalid.
+	ErrAuthenticationFailed = errors.New("authentication failed")
+)
```
GuptaManan100, line 31: This needs to be changed to mterrors package, but that cleanup has to happen everywhere, so not something to worry about now, but just pointing out  so that we don't miss it.

GuptaManan100 (approved):
The server side isn't integrated into the server code, so that's pending but otherwise the implementation looks complete! This is great!

GuptaManan100:
Don't know about the passthrough though for authentication, that's part of a much larger discussion on authentication and authorisation on MultiPooler level and how we want to handle it

dweitzman:
> Don't know about the passthrough though for authentication, that's part of a much larger discussion on authentication and authorisation on MultiPooler level and how we want to handle it

Seems like passthrough could at least let us easily keep parity with pgbouncer for the short term in a world with separate pools per user, so I'm imagining we'll use it until a better idea comes along

PR #335, opened 2025-12-10
docs: developer docs, initial content

sougou:
This is the initial content, partly inspired by https://github.com/multigres/multigres/pull/82 by @rafael.

I'll need to copy a few more things. Some pages have TODOs. For example, we need to add more stuff to the architecture. We also need to expand `contributing.md` (@dweitzman).

I thought I'll get this in for now, and we can iterate forward.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -7,7 +7,7 @@ This document provides instructions for building and testing the Multigres proje
 Multigres is organized as a monorepo with all the core application code located in the `go/` directory. The repository contains multiple components including:
 
 - **multigateway** - Gateway service for routing and load balancing
-- **multiorch** - Orchestration service for cluster management  
+- **multiorch** - Orchestration service for cluster management
 - **multipooler** - Connection pooling service
 - **pgctld** - PostgreSQL control daemon
 - **multigres** - Main CLI tool
@@ -17,6 +17,7 @@ Each component has its own main entry point under `go/cmd/` and shared libraries
 ## Prerequisites
 
 You need to install:
+
 - Go (version 1.25 or later)
 - PostgreSQL (we have been working with version 17.6)
 
@@ -77,7 +78,6 @@ make clean
 ```
 
 This does not remove the generated protobuf files. As per go standards, these generated files are checked into the repository.
-
 
 ## Protocol Buffers
 
--- a/docs/teamwork.md
+++ b/docs/teamwork.md
@@ -4,11 +4,11 @@ Our goal is to build a thriving community around Multigres. So, we welcome parti
 
 ## Discussions
 
-If you have questions or would like to start a discussion about a topic, you can start one [here](https://github.com/multigres/multigres/discussions). We look forward to your participation.
+If you have questions or would like to start a discussion about a topic, you can [start a discussion on GitHub](https://github.com/multigres/multigres/discussions). We look forward to your participation.
 
 ## Issues
 
-If you see something broken, you can file an issue [here](https://github.com/multigres/multigres/issues).
+If you see something broken, you can [file an issue on GitHub](https://github.com/multigres/multigres/issues).
 
 ## Proposals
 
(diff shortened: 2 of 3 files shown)


docs/building.md
```diff
@@ -0,0 +1,90 @@
...
+
+Multigres is organized as a monorepo with all the core application code located in the `go/` directory. The repository contains multiple components including:
+
+- **multigateway** - Gateway service for routing and load balancing
+- **multiorch** - Orchestration service for cluster management
+- **multipooler** - Connection pooling service
+- **pgctld** - PostgreSQL control daemon
+- **multigres** - Main CLI tool
+
+Each component has its own main entry point under `go/cmd/` and shared libraries are organized in the other subdirectories of `go/`.
```
rafael, line 15: maybe we can mention go/common here

rafael (commented):
This makes sense. Agree with all the content here. We can keep iterating

docs/building.md
sougou: I have a separate place-holder for code structure under `docs/architecture.md`. That'll be a better place right?

dweitzman: There's going to be some overlap between what we want to tell humans and what we want to tell claude

We may be able to have `.claude/docs/architecture.md` include this document, but we don't have to decide that now

docs/contributing.md
```diff
@@ -0,0 +1,29 @@
+# Contributing
+
+We have started welcoming minor bug fixes at this point. At some point, we will also open the project up for major contributions.
+
+## Preparing to contribute
+
+Multigres is a complex project. For making major contributions, a good deal of learning is required depending on the part of the project you'd like to contribute to.
```
dweitzman, line 7: I'm not proposing different wording, but from a philosophy perspective I prefer to think of Multigres as the simplest solution we can find for a set of problems that are complex. If it was equally true or useful to say something like we're ambitious in scope and deal with a complex problem space, I'd prefer something more like that vs saying we're complex.

docs/teamwork.md
```diff
@@ -0,0 +1,19 @@
+# Working with us
```
dweitzman, line 1: Another subtle philosophical thing: maybe this could be something more like "How we collaborate" or "How we work" instead of "Working with us"?

To me, the "us" distinction feels like it creates more distance between the core team and new or outside contributors

docs/workflow.md
```diff
@@ -0,0 +1,122 @@
...
+
+As you work in a package, you can run just the unit tests for that package by running `go test` from within that package.
+
+When you're ready to test the whole system, run the full test suite with `make test` from the root of the Git tree.
+If you haven't installed all dependencies for `make test`, you can rely on the CI test results as well.
+These results will be present on your pull request.
+
+## Committing your work
+
+When running `git commit` use the `-s` option to add a Signed-off-by line.
```
dweitzman, line 74: Somewhere near the commits or PR content here it's probably worth mentioning that we use the conventional commits format for PRs?

docs/workflow.md
```diff
@@ -0,0 +1,122 @@
...
+(new-feature) $ git push
+```
+
+Then go to the [repository page](https://github.com/multigres/multigres) and it
+should prompt you to create a Pull Request from a branch you recently pushed.
+You can also [choose a branch manually](https://github.com/multigres/multigres/compare).
+
+## Addressing Changes
+
+If you need to make changes in response to the reviewer's comments, just make
```
dweitzman, line 108: Something for us to eventual figure out: do we have strong opinions on rebasing vs merging for existing PRs?

Personally I also gravitate toward merging since rebasing mucks with history, but I don't know if other folks on the team have strong preferences either way

cuongdo: Once a PR has started the review process, I strongly prefer merges. I haven't found a satisfactory code review solution for branches that have been rebased after review comments have been made.

PR #336, opened 2025-12-10
test(multiorch): harden bootstrap test with explicit primary/standby tracking

dweitzman:
Track which pooler was actually initialized as primary and verify the
standby list correctly excludes it. The previous test only checked that
standby names were valid poolers, not that they excluded the selected
primary.

- Add `initializedPrimaries` tracking to testRPCClient
- Verify exactly one primary was initialized
- Verify standbys are the two non-primary poolers


dweitzman:
This was already fixed by another PR. On merge it seems like this approach is a little more thorough, so I'm keeping the PR open in case we want to add the small improvements

dweitzman:
Changed the PR title and description to reflect that it's no longer fixing a flake, just hardening some test assertions

PR #337, opened 2025-12-11
feat: implement per-user connection pooling with extended query protocol support

GuptaManan100:
###  Summary

This PR introduces a comprehensive per-user connection pooling architecture in the multipooler service, replacing the previous `sql.DB`-based approach with a custom pooling system that supports the PostgreSQL extended query protocol. As part of this PR, we wire up the connection pooler, and the client side protocol that were previously written. We also wire up the `PreparedStatementConsolidator` and `SettingsCache` in the ConnPoolManager.

### Key Changes
- **Per-user connection pools**: Each PostgreSQL user gets dedicated pools (regular and reserved) that authenticate directly as that user, enabling proper Row-Level Security (RLS) support
- **Three pool types**:
  - `AdminPool`: Shared pool for control plane operations (query cancellation, connection termination)
  - `RegularPool`: Per-user pools for stateless query execution with settings-based bucket routing
  - `ReservedPool`: Per-user pools for transactions and operations requiring connection affinity
- **Extended query protocol**: Full support for Parse/Bind/Execute flow with prepared statements and portals
- **ConnectionPoolManager**: Unified manager orchestrating all pools with lazy user pool creation

### Test Plan
- [x] Existing multipooler tests pass
- [x] New end-to-end tests for extended query protocol (`multigateway_pg_test.go`)
- [x] Integration tests with actual PostgreSQL instances
- [ ] ~NULL handling test skipped (known issue)~

## Next Steps
1. **Fix NULL handling**: The test for NULL value handling has been skipped due to a known issue. This needs investigation and a fix to properly handle NULL values in query results.

2. **Migrate remaining `sql.DB` usage**: Other parts of the codebase still use `sql.DB` directly. These should be migrated to use the new `connpoolmanager` for consistency and to benefit from the connection pooling features.

3. **Add unit tests**: The new `connpoolmanager` and `connpool` packages need comprehensive unit tests to ensure reliability and make future refactoring safer.

### DIFF
--- a/go/multigateway/discovery.go
+++ b/go/multigateway/discovery.go
@@ -203,11 +203,6 @@ func (pd *PoolerDiscovery) processPoolerChange(watchData *topoclient.WatchDataRe
 	_, existed := pd.poolers[poolerID]
 	pd.poolers[poolerID] = pooler
 	pd.lastRefresh = time.Now()
-	// TODO: Remove this. Currently a hack, poolers not registering as the correct
-	// type in the topo, so making do with this for now.
-	if pooler.Type == clustermetadatapb.PoolerType_UNKNOWN {
-		pooler.Type = clustermetadatapb.PoolerType_PRIMARY
-	}
 
 	if !existed {
 		pd.logger.Info("New pooler discovered",
@@ -260,6 +255,21 @@ func (pd *PoolerDiscovery) GetPooler(target *query.Target) *clustermetadatapb.Mu
 		targetType = clustermetadatapb.PoolerType_PRIMARY
 	}
 
+	// Debug: Log all discovered poolers
+	pd.logger.Debug("GetPooler called - listing all discovered poolers",
+		"target_tablegroup", target.TableGroup,
+		"target_shard", target.Shard,
+		"target_pooler_type", targetType.String(),
+		"total_poolers", len(pd.poolers))
+	for i, pooler := range pd.poolers {
+		pd.logger.Debug("discovered pooler",
+			"index", i,
+			"pooler_id", topoclient.MultiPoolerIDString(pooler.Id),
+			"tablegroup", pooler.TableGroup,
+			"shard", pooler.Shard,
+			"type", pooler.Type.String())
+	}
+
 	// Find matching pooler
 	for _, pooler := range pd.poolers {
 		// TableGroup must match
@@ -278,6 +288,11 @@ func (pd *PoolerDiscovery) GetPooler(target *query.Target) *clustermetadatapb.Mu
 		}
 
 		// Found a match!
+		pd.logger.Debug("selected pooler for target",
+			"pooler_id", topoclient.MultiPoolerIDString(pooler.Id),
+			"pooler_type", pooler.Type.String(),
+			"tablegroup", pooler.TableGroup,
+			"shard", pooler.Shard)
 		return proto.Clone(pooler.MultiPooler).(*clustermetadatapb.MultiPooler)
 	}
 
--- a/go/pgprotocol/server/conn.go
+++ b/go/pgprotocol/server/conn.go
@@ -405,9 +405,11 @@ func (c *Conn) handleQuery() error {
 		return nil
 	})
 	if err != nil {
-		// Send error response.
+		// Send error response with the actual error in the message for better visibility.
+		// lib/pq and other clients often only show the message field, not the detail field.
 		c.logger.Error("query execution failed", "query", queryStr, "error", err)
-		if err := c.writeErrorResponse("ERROR", "42000", "query execution failed", err.Error(), ""); err != nil {
+		errMsg := fmt.Sprintf("query execution failed: %v", err)
+		if err := c.writeErrorResponse("ERROR", "42000", errMsg, "", ""); err != nil {
 			return err
 		}
 	}


docs/query_serving/connection_pooling.md
dweitzman, line 1: Should I review this file? Not sure if it's a temporary artifact from/for Claude

.github/linters/.markdownlint.json
dweitzman, line 1: Maybe this could be split into a separate PR? This PR feels interesting enough already without extra cleanups added

go/parser/ast/oids.go
```diff
@@ -185,3 +185,198 @@ const (
...
+	// Character types
+	case CHAROID:
+		return "CHAR"
+	case NAMEOID:
+		return "NAME"
+	case TEXTOID:
+		return "TEXT"
+	case VARCHAROID:
+		return "VARCHAR"
+	case 1042: // bpchar
```
dweitzman: Is there a reason this one is a hard-coded number instead of a pre-declared constant?

go/parser/ast/oids.go
```diff
@@ -185,3 +185,198 @@ const (
...
+		return "DATE"
+	case TIMEOID:
+		return "TIME"
+	case TIMESTAMPOID:
+		return "TIMESTAMP"
+	case TIMESTAMPTZOID:
+		return "TIMESTAMPTZ"
+	case TIMETZOID:
+		return "TIMETZ"
+	case 1186: // interval
```
dweitzman: Same question about hard-coded OID vs pre-declared constant

go/provisioner/local/local.go
```diff
@@ -798,6 +801,8 @@ func (p *localProvisioner) provisionMultipooler(ctx context.Context, req *provis
 		"--pg-port", fmt.Sprintf("%d", pgPort),
 		"--hostname", "localhost",
 		"--pgbackrest-stanza", "multigres",
+		"--connpool-admin-password", "postgres", // Password created in initializePgctldDirectories
```
dweitzman, line 804: cc @rafael re: thoughts on passwords, since https://github.com/multigres/multigres/pull/342 relates to that

I don't know if a password for the unix socket is useful, but if it is then setting it via command line argument in production doesn't feel particularly secure since it'd leak via things like `ps` output.

go/multipooler/grpcpoolerservice/service.go
```diff
@@ -84,3 +84,61 @@ func (s *poolerService) ExecuteQuery(ctx context.Context, req *multipoolerpb.Exe
...
+		func(ctx context.Context, result *querypb.QueryResult) error {
+			// Send the result back to the client
+			response := &multipoolerpb.PortalStreamExecuteResponse{
+				Result: result,
+			}
+			return stream.Send(response)
+		},
+	)
+	if err != nil {
+		return err
```
dweitzman, line 136: Just to confirm: would it be possible here to have both an error and a reserved connection ID? I'm wondering if this `return err` being above the `if reservedState.ReservedConnectionId > 0 {` is ok

go/multipooler/connpoolmanager/manager.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil, fmt.Errorf("manager is closed")
+	}
+
+	// Check if pool already exists
+	if pool, ok := m.userPools[user]; ok {
+		return pool, nil
+	}
+
+	// Check max users limit
+	if maxUsers := m.config.MaxUsers(); maxUsers > 0 && int64(len(m.userPools)) >= maxUsers {
```
dweitzman, line 142: Might also be worthwhile to garbage collect user pools after a long enough period of inactivity to reduce the odds of reaching the user pool maximum? Could be a TODO()

go/multipooler/connpoolmanager/user_pool.go
```diff
@@ -0,0 +1,179 @@
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.closed {
+		return
+	}
+	p.closed = true
+
+	// Close reserved pool first (it has its own internal regular pool)
+	if p.reservedPool != nil {
```
dweitzman: I don't think these regular pools should ever be nil?

go/multipooler/connstate/connection_state.go
```diff
@@ -19,38 +19,36 @@ import (
...
-// and portals.
+// This includes all state modifiers like session settings and prepared statements.
 //
 // All methods are thread-safe.
 type ConnectionState struct {
 	// mu protects all mutable fields in this struct.
 	mu sync.Mutex
 
+	// User is the current role set via SET ROLE.
+	// This is NOT used for pool bucket routing - only for RLS enforcement.
```
dweitzman: The phrasing of this comment feels inaccurate, since RLS is not the only feature that can use roles as a factor when granting access to things. Other objects like stored procedures, tables, etc. can be role-protected.

Maybe it's enough to say that this is similar to a setting but tracked differently and not using for bucket routing?

go/multipooler/executor/executor.go
```diff
@@ -14,164 +14,170 @@
...
-		maxRows = options.MaxRows
+		settings = options.SessionSettings
 	}
 
-	// Execute the query and stream results
-	return e.executeQuery(ctx, sql, maxRows)
+	// Get a connection from the pool for this user
+	conn, err := e.poolManager.GetRegularConnWithSettings(ctx, settings, user)
+	if err != nil {
+		e.logger.ErrorContext(ctx, "failed to get connection", "error", err, "user", user)
```
dweitzman: My general instinct would be to say that whoever handles an error should log it and avoid logging before returning, to avoid double-logging

go/multipooler/executor/executor.go
```diff
@@ -14,164 +14,170 @@
...
+		reservedConn, _ := e.poolManager.GetReservedConn(int64(options.ReservedConnectionId), user)
+		if reservedConn == nil {
+			return fmt.Errorf("reserved connection %d not found for user %s", options.ReservedConnectionId, user)
+		}
 
-// Close closes the executor and releases resources.
-func (e *Executor) Close(ctx context.Context) error {
-	if !e.isOpen.Swap(false) {
+		if err := reservedConn.QueryStreaming(ctx, sql, callback); err != nil {
+			e.logger.ErrorContext(ctx, "query execution failed", "error", err, "query", sql)
```
dweitzman: Same comment re: logging + returning errors, here and elsewhere

go/multipooler/executor/executor.go
```diff
@@ -180,169 +186,265 @@ func (e *Executor) PortalStreamExecute(
...
+	// We use connId 0 since we just need the canonical name mapping
+	// Get the canonical prepared statement info
+	psi := e.consolidator.GetPreparedStatementInfo(0, stmt.Name)
+	if psi == nil {
+		// Add to consolidator to get/create canonical name
+		if err := e.consolidator.AddPreparedStatement(0, stmt.Name, stmt.Query, stmt.ParamTypes); err != nil {
+			return "", fmt.Errorf("failed to consolidate prepared statement: %w", err)
+		}
+		// This will be not-nil because we just added it.
+		psi = e.consolidator.GetPreparedStatementInfo(0, stmt.Name)
```
dweitzman: Would it make sense for AddPreparedStatement() to return the statement so it's not necessary to call Get() immediately afterwards and assume it's non-nil?

go/multipooler/pools/reserved/reserved_pool.go
```diff
@@ -0,0 +1,377 @@
...
+	p.mu.Unlock()
+
+	// Get a regular connection from the pool.
+	pooled, err := p.conns.GetWithSettings(ctx, settings)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get connection: %w", err)
+	}
+
+	// Generate unique ID.
+	connID := p.lastID.Add(1)
```
dweitzman, line 157: Do we need to worry about conflicts between these connection IDs and connections created with the same ID after a multipooler restart? If we always start at lastID "1" and share these IDs with multigateway it seems hypothetically possible to run into a conflict without detecting it

If that is a concern, seems like we could either use randomized IDs or do something like put some of the pooler's start timestamp or some other unique token into the connection ID when communicating with multigateway

go/pgprotocol/client/extended.go
```diff
@@ -227,6 +233,250 @@ func (c *Conn) PrepareAndExecute(ctx context.Context, queryStr string, params []
...
+			return err
+		}
+		batchedRows = nil
+		batchedSize = 0
+		return nil
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
```
dweitzman: Probably worth adding a TODO() here about the known bug

go/pgprotocol/client/query.go
```diff
@@ -97,41 +98,46 @@ func (c *Conn) writeQueryMessage(queryStr string) error {
...
-		}
-
-		// Read message.
+		// Read message. We don't check context here because we MUST drain to
+		// ReadyForQuery to keep the connection in a clean state.
 		msgType, body, err := c.readMessage()
 		if err != nil {
+			// If we had a callback error, return that instead since it's more relevant.
+			if callbackErr != nil {
+				return callbackErr
```
dweitzman: Does a callback error terminate the connection? Wondering if this `return` statement could leave a connection with unconsumed result sets

go/test/endtoend/multigateway_pg_test.go
```diff
@@ -190,3 +191,314 @@ func TestMultiGateway_PostgreSQLConnection(t *testing.T) {
...
+	if testing.Short() {
+		t.Skip("skipping ExtendedQueryProtocol test in short mode")
+	}
+	if utils.ShouldSkipRealPostgres() {
+		t.Skip("PostgreSQL binaries not found, skipping cluster lifecycle tests")
+	}
+
+	// Setup full test cluster with all services
+	cluster := setupTestCluster(t)
+	t.Cleanup(cluster.Cleanup)
```
dweitzman, line 207: You don't need to change this, but I'm making a note to myself: setupTestCluster() should return `cleanup` as a return value instead of hiding it under `cluster` to make it harder to forget to call it

go/test/endtoend/multigateway_pg_test.go
```diff
@@ -190,3 +191,314 @@ func TestMultiGateway_PostgreSQLConnection(t *testing.T) {
...
+
+	// Connect using pgx (which uses Extended Query Protocol by default)
+	connStr := fmt.Sprintf("host=localhost port=%d user=postgres password=postgres dbname=postgres sslmode=disable",
+		cluster.PortConfig.Zones[0].MultigatewayPGPort)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
dweitzman: This should use `t.Context()` instead of `context.Background()`. I would have expected lint to catch that

go/test/endtoend/multipooler_client.go
```diff
@@ -144,6 +144,50 @@ func IsPrimary(addr string) (bool, error) {
...
+	conn, err := grpc.NewClient(addr, grpccommon.LocalClientDialOptions()...)
+	if err != nil {
+		return clustermetadatapb.PoolerType_UNKNOWN, fmt.Errorf("failed to create client: %w", err)
+	}
+	defer conn.Close()
+
+	client := multipoolermanagerpb.NewMultiPoolerManagerClient(conn)
+	deadline := time.Now().Add(timeout)
+
+	for time.Now().Before(deadline) {
```
dweitzman: Feels like a lot of this is duplicating what `require.Eventually()` or `assert.Eventually()` already handle

docs/query_serving/connection_pooling.md
GuptaManan100: Yes, please review it. I've added it to document our approach and why along with future steps

.github/linters/.markdownlint.json
GuptaManan100: I had to do that to even get the design doc past the linter. The big table that I created was failing the 80 character line check.

go/multipooler/connpoolmanager/manager.go
GuptaManan100: All of that is a todo. Even dynamic user pools is a todo. This PR got big enough already.

go/pgprotocol/client/query.go
GuptaManan100: Fixed this, and handled context cancellation correctly.

go/pgprotocol/client/extended.go
GuptaManan100: Fixed this, and handled context cancellation properly

go/parser/ast/oids.go
GuptaManan100: Fixed

go/parser/ast/oids.go
GuptaManan100: Fixed

go/multipooler/grpcpoolerservice/service.go
GuptaManan100: Yes, both don't happen together. It's either or. We can order the checks in either way

go/multipooler/connpoolmanager/user_pool.go
GuptaManan100: Fixed

go/multipooler/connstate/connection_state.go
GuptaManan100: Fixed

go/multipooler/executor/executor.go
GuptaManan100: Fixed

go/multipooler/executor/executor.go
GuptaManan100: Fixed

go/multipooler/executor/executor.go
GuptaManan100: Fixed

go/test/endtoend/multigateway_pg_test.go
GuptaManan100: That would be better for sure

go/test/endtoend/multigateway_pg_test.go
GuptaManan100: Fixed

go/test/endtoend/multipooler_client.go
GuptaManan100: Fixed

GuptaManan100: I copied a function lower down the file, and changed a few lines

go/multipooler/pools/admin/admin_conn.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// Close closes the underlying connection.
+func (c *Conn) Close() error {
+	return c.conn.Close()
+}
+
+// ApplySettings panics because admin connections don't support settings.
+// This should never be called - admin connections are always "clean".
+func (c *Conn) ApplySettings(_ context.Context, _ *connstate.Settings) error {
```
sougou, line 68: Is this needed to satisfy some interface?

sougou (approved):
There is a lot to fix forward. So, this is good to go if tests pass.

go/multipooler/pools/admin/admin_conn.go
GuptaManan100: Yes, the connection interface requires it

go/multipooler/pools/reserved/reserved_pool.go
GuptaManan100: Gonna mark this as a todo and leave it at that for now

GuptaManan100: Addressed in https://github.com/multigres/multigres/pull/390

PR #338, opened 2025-12-11
feat(multipooler): undemote implementation

rafael:
## Summary
  Implement the `UndoDemote` RPC to allow a demoted primary to be restored back to primary state. This is useful when a demotion needs to be reverted before any standby has been promoted.

  Key changes:
  - Add `restartPostgresAsPrimary()` in manager to restart PostgreSQL without standby.signal
  - Implement `UndoDemote` RPC that restarts PostgreSQL as primary and updates topology
  - Update pgctld restart command to remove standby.signal when `AsStandby=false`
  - Add `WasAlreadyPrimary` and `LsnPosition` fields to `UndoDemoteResponse` proto
  - Add unit tests for UndoDemote (already primary, connection failures, restart failures)
  - Enhance e2e test to verify replication works after undo demote

  The implementation is idempotent - calling UndoDemote on an already-primary node returns success with `WasAlreadyPrimary=true`.

## Tests
- Added relevant unit and integration tests

### DIFF
--- a/go/multipooler/manager/rpc_manager.go
+++ b/go/multipooler/manager/rpc_manager.go
@@ -874,8 +874,9 @@ func (pm *MultiPoolerManager) demoteLocked(ctx context.Context, consensusTerm in
 //  2. No other node accepted writes (follows from #1)
 //  3. The old primary's data directory hasn't been modified as a standby
 //
-// It is expected from the caller that these invariants will be respected.
-func (pm *MultiPoolerManager) UndoDemote(ctx context.Context) (*multipoolermanagerdatapb.UndoDemoteResponse, error) {
+// The consensusTerm parameter ensures safety by verifying the term hasn't changed
+// since demotion. If the term changed, another node may have been promoted.
+func (pm *MultiPoolerManager) UndoDemote(ctx context.Context, consensusTerm int64) (*multipoolermanagerdatapb.UndoDemoteResponse, error) {
 	if err := pm.checkReady(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


go/multipooler/manager/rpc_manager.go
```diff
@@ -932,21 +932,116 @@ func (pm *MultiPoolerManager) demoteLocked(ctx context.Context, consensusTerm in
...
+			return nil, err
+		}
+		pm.logger.InfoContext(ctx, "Already running as primary, undo demotion not needed (idempotent)")
+		return &multipoolermanagerdatapb.UndoDemoteResponse{
+			WasAlreadyPrimary: true,
+			LsnPosition:       lsn,
+		}, nil
+	}
+
+	// TODO: What could be safety guardrails we add here? Should we check the timeline
```
dweitzman: Should this verify that the term number hasn't incremented since the demote?

go/multipooler/manager/rpc_manager_test.go
```diff
@@ -1228,3 +1217,151 @@ func TestReplicationStatus(t *testing.T) {
...
+
+// =============================================================================
+// UndoDemote Tests
+// =============================================================================
+
+// setupUndoDemoteTestManager creates a manager configured as a demoted primary
+// (topology shows REPLICA) for undo demotion tests.
+// Returns the manager and the mock pgctld service for configuring restart behavior.
+func setupUndoDemoteTestManager(t *testing.T, mockDB *sql.DB) (*MultiPoolerManager, *testutil.MockPgCtldService) {
+	ctx := context.Background()
```
dweitzman: Could probably be `t.Context()`?

go/multipooler/manager/rpc_manager_test.go
```diff
@@ -1228,3 +1217,151 @@ func TestReplicationStatus(t *testing.T) {
...
+	resp, err := pm.UndoDemote(ctx)
+	require.Error(t, err)
+	require.Nil(t, resp)
+	assert.Contains(t, err.Error(), "failed to check if primary")
+
+	require.NoError(t, mock.ExpectationsWereMet())
+}
+
+// TestUndoDemote_RestartFailed tests error handling when pgctld restart fails.
+func TestUndoDemote_RestartFailed(t *testing.T) {
```
dweitzman, line 1355: It seems like these tests are for error scenarios, but I'm not spotting a test for the success scenario

rafael: Yeah, this was intentional. For the success test in unit tests, it gets tricky with the mocks. I added a comment around this.

deepthi:
Closed after discussion. We will revisit this when we implement "Planned failovers".

PR #340, opened 2025-12-11
fix(telemetry): reduce log noise by lowering init/shutdown to debug

dweitzman:
OpenTelemetry init and shutdown messages are operational details that don't need to appear at INFO level in normal operation.


dweitzman:
> can you also take care of this message too?
> 
> Shutting down OpenTelemetry
> 
> Thanks!

Good catch, done!

PR #341, opened 2025-12-11
build(lint): add depguard isolation rules for gateway, orch, pooler

dweitzman:
Prevent incoming dependencies on multigateway, multiorch, and multipooler packages. These services should only be imported by their own package and cmd, similar to the existing rules for pgctld and multiadmin.

Each rule has a TODO to move the service under go/services/ when timing is opportune to minimize merge conflicts.


rafael (approved):
LGTM

PR #342, opened 2025-12-11
feat(pgtctld): improves how initial password is used and provided

rafael:
## Summary

Refactors password file handling in pgctld to use a convention-based approach instead of passing the path as a parameter through RPCs and CLI flags.

**Before**: The --pg-pwfile flag and pg_pwfile proto field were passed around between commands and RPCs, which was awkward since the password is only needed during init.

**After**: pgctld looks for the password file at a conventional location (<pooler-dir>/pgpassword.txt). If the file exists, it uses it; otherwise falls back to the PGPASSWORD environment variable if set.

Note: This is our own convention. PostgreSQL's initdb does not use PGPASSWORD directly—it only supports --pwfile. We use PGPASSWORD as a fallback because it's the standard way users provide passwords to PostgreSQL client tools, so if a user has it set for authentication purposes, we assume they also want it used as the initial password during init.

## Changes

  - Remove --pg-pwfile CLI flag from pgctld init
  - Remove pg_pwfile field from InitDataDirRequest proto message
  - Add PostgresPasswordFile(poolerDir) helper that returns the conventional path
  - Update resolvePassword() to check the conventional location
  - Simplify initializeDataDir() and setPostgresPassword() to derive paths from poolerDir
  - Update local provisioner to copy source password file content to the conventional location during setup

## Tests
- No real functional changes, updated tests to not use the flag.
- Added integration test in setup_cluster to validate that once the cluster is initialized, the password is configured correctly and we can connect via the TPC connection using the password.

### DIFF
--- a/go/test/endtoend/cluster_test.go
+++ b/go/test/endtoend/cluster_test.go
@@ -790,6 +790,28 @@ func testPostgreSQLConnection(t *testing.T, tempDir string, port int, zone strin
 	require.NoError(t, err, "PostgreSQL connection failed on port %d (Zone %s): %s", port, zone, string(output))
 
 	t.Logf("Zone %s PostgreSQL (port %d) is responding correctly", zone, port)
+
+	// Also test TCP connection with password to validate password was set correctly
+	// The default password is "postgres" (set by the local provisioner at pgpassword.txt)
+	testPostgreSQLTCPConnection(t, port, zone)
+}
+
+// testPostgreSQLTCPConnection tests TCP connection with password authentication.
+// This validates that the password file convention is working correctly.
+func testPostgreSQLTCPConnection(t *testing.T, port int, zone string) {
+	t.Helper()
+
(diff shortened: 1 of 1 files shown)


go/test/endtoend/cluster_test.go
```diff
@@ -247,7 +247,7 @@ func createTestConfigWithPorts(tempDir string, portConfig *testPortConfig) (stri
 				Timeout:        30,
 				LogLevel:       "info",
 				PoolerDir:      local.GeneratePoolerDir(tempDir, serviceID),
-				PgPwfile:       filepath.Join(local.GeneratePoolerDir(tempDir, serviceID), "pgctld.pwfile"),
+				// PgPwfile not set - provisioner will create pgpassword.txt with default "postgres" password
```
dweitzman, line 250: Not sure this comment is necessary

dweitzman (approved):

deepthi:
Note that operator will need to create the pgpassword.txt file via the pooler pod's init container.

Issue #320, opened 2025-12-04
why strict version pinning of these dependencies in the dockerfile?


teaglebuilt:
I believe there are newer versions than this in the alpine linux repository https://github.com/multigres/multigres/blob/841b09b4ef0a97807005b991c29d9dacb4103ac8/Dockerfile#L9

Is there a reason for these strict versions?

sougou:
@fernando-villalba

teaglebuilt:
```dockerfile
RUN apk add --no-cache git make bash
```

jchrys:
`golang:1.25-alpine` update(Alpine 3.22 -> 3.23) broke the builds due to version mismatches. I think pinning the base image and removing the package pins is the right fix here. I can open a PR for this. :D

PR #345, opened 2025-12-11
build(Dockerfile): remove alpine pinned dependencies

teaglebuilt:
Change in reference to #320

- unpin alpine requirements in dockerfile and use specific base image version


fernando-villalba:
@teaglebuilt thank you for your contribution. The pinned versions were added for two reasons:

1. Our linter would fail if we didn't do that.
2. It's good to be able to see exactly what versions are being used, makes the image more deterministic. So we will not be incorporating this change at this point.

sougou:
Closing this PR as it's not necessary.

PR #346, opened 2025-12-12
feat(local): change default config to generate three node cluster

rafael:
# Summary
- Updates the local config to generate a three node cluster. Now that we have a full bootstrap and multiorch detecting primary failures, this will make it more interesting to test multiorch operations in local.
- Fixes an issue where were not passing a lock path to pgbackrest

### DIFF
--- a/go/test/endtoend/cluster_test.go
+++ b/go/test/endtoend/cluster_test.go
@@ -874,7 +874,7 @@ func TestClusterLifecycle(t *testing.T) {
 
 		cells, ok := cellsRaw.([]any)
 		require.True(t, ok, "cells config should be a slice")
-		require.Len(t, cells, 3, "should have exactly 3 cells")
+		require.Len(t, cells, 2, "should have exactly 2 cells")
 
 		// Get the first cell (for backward compatibility)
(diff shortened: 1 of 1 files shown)


go/provisioner/local/config.go
```diff
@@ -476,6 +518,19 @@ func (p *localProvisioner) GeneratePgBackRestConfigs() error {
 			})
 		}
 
+		// Create per-pooler spool and lock directories inside backup repo
```
rafael, line 521: @cuongdo  could you double check this makes sense? After thinking a bit more about this, I think the spool and the lock should be per pooler. But I'm not 100% sure

cuongdo: spool directories can be separate. I need to understand the lock directory better though

cuongdo: It looks like lock directories can be per-pooler for identical operations. So, this is fine for bootstrap. For normal operation, the ActionLock will prevent problematic concurrent access to the backup_location.

rafael: That makes sense

PR #347, opened 2025-12-12
feat(pgprotocol): add unix socket support to client

dweitzman:
Add Unix socket support to pgprotocol/client. When Host starts with "/", the client dials a Unix socket at {host}/.s.PGSQL.{port} instead of TCP.

This matches PostgreSQL's libpq behavior and enables local connections without TCP overhead.


GuptaManan100:
Closing since https://github.com/multigres/multigres/pull/337 also has this change.

PR #350, opened 2025-12-14
docs: expand on architecture

sougou:


docs/architecture.md
```diff
@@ -2,9 +2,127 @@
...
+### Opinionated
+
+We will focus on functional completeness and ensure that all the features gel together well.
+To achieve this, we will make tooling decisions that are opinionated.
+For example, we'll use etcd for the toposerver.
+Other similar tools like zookeeper or consul will not be supported.
+
+## Components
+
+Multigres is designed to run with a minimal set of loosely coupled components.
```
rafael: This doesn't seem clear to me

docs/architecture.md
```diff
@@ -2,9 +2,127 @@
...
+Other than these necessary differences, we will support all other constructs as drop-in replacements.
+
+### Approachability
+
+We intend to make Multigres approachable by making it easy to launch, and by providing intuitive command line parameters with reasonable defaults.
+The price to pay for this will be a loss of flexibility. We believe that this is a worthy tradeoff.
+
+### Opinionated
+
+We will focus on functional completeness and ensure that all the features gel together well.
```
dweitzman: Super minor thought: "work together well" could be more accessible language than "gel together well" for non-native English speakers

docs/architecture.md
```diff
@@ -2,9 +2,127 @@
...
+
+The Operator is a Kubernetes Operator. Its primary responsibility is to provision resources for a cluster and bring up all the required Multigres components.
+
+The Operator does not have any secondary responsibilities.
+
+The Operator has no secondary responsibilities in order to ease the development of future operators to allow for Multigres to be deployed in non-Kubernetes environments.
+
+## Metadata
+
+### Toposervers
```
dweitzman, line 78: I know the ship has probably sailed on this and this PR isn't the place to discuss it, but FWIW I still don't like the name toposerver. Something like "discovery service" feels more accurate to me, since the primary purpose is for registering and discover the addresses of nodes.

docs/architecture.md
```diff
@@ -2,9 +2,127 @@
...
+### mterrors
+
+mterrors provides a unified error handling framework.
+
+## Code Structure
+
+Every go package contains a detailed `README.md` file that explains what the package is meant for. The following directories are noteworthy:
+
+- `go/cmd` contains the main programs for all the binaries
+- `go/<component>` (eg `multigateway`) contains component specific logic
```
dweitzman, line 127: Note: For services, I'm still in the process of moving them all under go/services/multiadmin, go/services/multigateway, etc.

I only paused to minimize merge conflicts while people were in the middle of big PRs

dweitzman (approved):

docs/architecture.md
sougou: How about: Although Multigres is a distributed system, we work hard at minimizing the number of moving parts.

docs/architecture.md
sougou: Coincidentally, I was thinking the same thing. Too late now.

docs/architecture.md
sougou: Yeah. I think we'll need to keep this part updated whenever we make major changes.

PR #359, opened 2025-12-16
feat(multipooler): add GetAuthCredentials gRPC endpoint for SCRAM authentication

dweitzman:
Adds a gRPC method on multipooler for getting a user's password hash, similar to what `auth_query` would to for pgbouncer
h

This will be used by a follow-up PR to add SCRAM support to multigate in the wire protocol. The wire protocol code will depend on this code to fetch password hashes.

## Summary
- Add `GetAuthCredentials` RPC to multipooler for fetching SCRAM-SHA-256 password hashes from PostgreSQL
- Add `PoolerHashProvider` in multigateway that implements `scram.PasswordHashProvider` interface

## Changes

**Proto** (`proto/multipoolerservice.proto`)
- Add `GetAuthCredentials` RPC with request/response messages
- Returns `codes.NotFound` if user doesn't exist
- Returns empty `scram_hash` if user exists but has no password

**Multipooler** (`go/multipooler/grpcpoolerservice/`)
- Implement `GetAuthCredentials` that queries `pg_authid` for the user's password hash
- Validate request fields (username, database required)
- Return appropriate gRPC error codes

**Multigateway** (`go/multigateway/auth/`)
- Add `PoolerHashProvider` that calls multipooler's `GetAuthCredentials`
- Implements `scram.PasswordHashProvider` interface for SCRAM authentication
- Translates `NotFound` errors to `scram.ErrUserNotFound`

**Tests**
- Unit tests for validation and error handling
- End-to-end tests that create real PostgreSQL users and verify the returned SCRAM hash matches the password


go/multipooler/grpcpoolerservice/service.go
```diff
@@ -85,6 +89,61 @@ func (s *poolerService) ExecuteQuery(ctx context.Context, req *multipoolerpb.Exe
...
+	if err != nil {
+		return nil, status.Errorf(codes.Unavailable, "executor not initialized: %v", err)
+	}
+
+	// Query pg_authid for the user's password hash.
+	// Note: This requires superuser access to read pg_authid.
+	query := fmt.Sprintf("SELECT rolpassword FROM pg_catalog.pg_authid WHERE rolname = '%s'",
+		escapeString(req.Username))
+
+	options := &querypb.ExecuteOptions{MaxRows: 1}
```
GuptaManan100, line 119: Just as a comment, not important, but MaxRows aren't respected in the ExecuteQuery code. If you want to enforce it, add a `LIMIT 1` to the query.

`MaxRows` are only used in running PreparedStatements.

We control what ExecuteOptions we send from Multigateway, that is the primary user of the `query service` interface.

PR #360, opened 2025-12-16
fix(multiorch): Primary failure detection logic

rafael:
# Description

- This PR updates the primary failover detection logic to make sure it can distinguish from postgres being down vs just the multipooler process.
- It also performs a cosmetic change around how we check for skip postgres tests in integration tests.

## Tests
- Added unit tests and end-to-end tests for the failover detection logic.


go/test/endtoend/multipooler/backup_test.go
```diff
@@ -467,9 +469,11 @@ func TestBackup_CreateListAndRestore(t *testing.T) {
 }
 
 func TestBackup_ValidationErrors(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping end-to-end tests in short mode")
+	skip, err := utils.ShouldSkipRealPostgres()
```
dweitzman: Hmm, I wonder if the best balance between clarity and brevity is just to do the normal go approach and just use a clear message?...

```
func TestBackup_ValidationErrors(t *testing.T) {
  if testing.Short() {
    t.Skip("Short tests don't start real Postgres servers")
  }

  // ... the rest of the test
}
```

Sorry for being inconsistent about this.

dweitzman: Also, how would you feel about splitting the change to how we skip into its own separate PR?

rafael: That sounds good!

go/test/endtoend/multiorch/dead_primary_recovery_test.go
```diff
@@ -115,3 +115,154 @@ func TestDeadPrimaryRecovery(t *testing.T) {
...
+
+	// Kill the multipooler process on the primary (but leave Postgres running)
+	t.Logf("Killing multipooler on primary node %s (postgres stays running)", primaryNode.name)
+	killMultipooler(t, primaryNode)
+
+	// Wait for multiorch to detect the pooler is down and run several recovery cycles.
+	// Tests configure pooler-health-check-interval and recovery-cycle-interval to 500ms,
+	// so 3 seconds allows ~6 cycles which is plenty to detect and process the state.
+	t.Logf("Waiting for multiorch to process the pooler down state...")
+	time.Sleep(3 * time.Second)
```
dweitzman, line 194: Not a blocking comment, but for faster tests in the future maybe we should have a synchronous HTTP request or something to force multiorch to run recovery on-demand during endtoend tests instead of just waiting

rafael: good idea

PR #361, opened 2025-12-16
fix(pgprotocol/client): reject cleartext and MD5 password authentication

dweitzman:
Remove support for insecure authentication methods to prevent sending passwords over weak protocols. The client now only supports SCRAM-SHA-256 authentication.

### DIFF
--- a/go/pgprotocol/client/startup_test.go
+++ b/go/pgprotocol/client/startup_test.go
@@ -1,0 +1,80 @@
+// Copyright 2025 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


GuptaManan100 (commented):
All for using scram-sha-256 as the default. Don't know if we want to get rid of the md5 and cleartext altogether

deepthi:
> All for using scram-sha-256 as the default. Don't know if we want to get rid of the md5 and cleartext altogether

Yes we do.

deepthi (approved):

dweitzman:
> Should we not add a test that asserts that the server in fact rejects md5 passwords?

This change is only for the client, not the server protocol (the server protocol doesn't understand md5 passwords), but I've added test coverage for the client changes in this PR

PR #363, opened 2025-12-16
fix(tests): use ReadyPGPort to connect to multigateway with PRIMARY access

dweitzman:
The TestMultiGateway_ExtendedQueryProtocol test is flaking because it hardcoded connection to Zone 0's multigateway, but after bootstrap the PRIMARY pooler could be in any zone. Each multigateway only discovers poolers in its own zone.

In the future we should make all gates allow connecting to the primary pooler even across zones, but this is probably the quickest way to resolve the test flake.

### DIFF
--- a/go/test/endtoend/multigateway_pg_test.go
+++ b/go/test/endtoend/multigateway_pg_test.go
@@ -40,24 +40,12 @@ func TestMultiGateway_PostgreSQLConnection(t *testing.T) {
 	}
 
 	// Setup full test cluster with all services (includes waiting for bootstrap)
-	cluster := setupTestCluster(t)
-	t.Cleanup(cluster.Cleanup)
-
-	// Find a multigateway that has access to the PRIMARY pooler.
-	// Each multigateway only discovers poolers in its own zone, and the PRIMARY
-	// may be in any zone after bootstrap. Try all zones to find one that works.
-	pgPorts := []int{
-		cluster.PortConfig.Zones[0].MultigatewayPGPort,
-		cluster.PortConfig.Zones[1].MultigatewayPGPort,
-	}
-	findCtx, findCancel := context.WithTimeout(t.Context(), 30*time.Second)
-	defer findCancel()
-	readyPort, err := findReadyMultigateway(t, findCtx, pgPorts)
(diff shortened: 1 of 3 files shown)


go/test/endtoend/multigateway_pg_test.go
```diff
@@ -40,24 +40,12 @@ func TestMultiGateway_PostgreSQLConnection(t *testing.T) {
...
-	pgPorts := []int{
-		cluster.PortConfig.Zones[0].MultigatewayPGPort,
-		cluster.PortConfig.Zones[1].MultigatewayPGPort,
-	}
-	findCtx, findCancel := context.WithTimeout(t.Context(), 30*time.Second)
-	defer findCancel()
-	readyPort, err := findReadyMultigateway(t, findCtx, pgPorts)
-	require.NoError(t, err, "should find a ready multigateway")
+	cluster, cleanup := setupTestCluster(t)
+	t.Cleanup(cleanup)
```
dweitzman, line 44: This test was already fixed

go/test/endtoend/multigateway_pg_test.go
```diff
@@ -203,12 +191,12 @@ func TestMultiGateway_ExtendedQueryProtocol(t *testing.T) {
...
 	// Setup full test cluster with all services
-	cluster := setupTestCluster(t)
-	t.Cleanup(cluster.Cleanup)
+	cluster, cleanup := setupTestCluster(t)
+	t.Cleanup(cleanup)
 
 	// Connect using pgx (which uses Extended Query Protocol by default)
 	connStr := fmt.Sprintf("host=localhost port=%d user=postgres password=postgres dbname=postgres sslmode=disable",
-		cluster.PortConfig.Zones[0].MultigatewayPGPort)
+		cluster.ReadyPGPort)
```
dweitzman, line 199: This test didn't have the fix

cuongdo (approved):
thanks for fixing this. I just ran into this locally

dweitzman:
As a follow-up, https://github.com/multigres/multigres/pull/365 updates multigate behavior to support cross-zone routing to primary instead of trying to identify which zone has the primary and connecting to that gate in tests.

PR #365, opened 2025-12-16
feat(multigateway): implement minimal version of cross-zone PRIMARY routing

dweitzman:
Enable any multigateway to route write queries to the PRIMARY pooler regardless of which zone/cell it's in.

This is a follow-up to https://github.com/multigres/multigres/pull/363 which resolved a flake by working around not having this behavior. With this change we can make tests simpler so they don't have to guess which multigate to connect to if they want to target PRIMARY.

Changes:
- Add primaryPooler field to PoolerDiscovery for cross-zone PRIMARY tracking
- Modify Start() to discover all cells and launch a watch per cell
- Add watchCell() method for per-cell pooler watching
- Update processInitialPoolers() to call considerPrimaryCandidate() serially after releasing lock, ensuring primaryPooler is set before function returns
- Keep goroutine approach in processPoolerChange() for non-blocking updates
- Modify GetPooler() to use primaryPooler for PRIMARY requests (cross-zone) while keeping REPLICA routing local-cell only

Test changes:
- Simplify findReadyMultigateway() to waitForMultigatewayReady() with single port
- Remove outdated comment about multigateways only discovering local poolers
- Update tests to use Zone 0's multigateway directly
- Add tests for new behavior in multigateway discovery

Future phases will add verification logic (reachability, term numbers, ConsensusStatus RPCs) to considerPrimaryCandidate().

### DIFF
--- a/go/multigateway/discovery.go
+++ b/go/multigateway/discovery.go
@@ -291,16 +291,15 @@ func (pd *PoolerDiscovery) processPoolerChange(cell string, watchData *topoclien
 }
 
 // considerPrimaryCandidate evaluates a pooler that claims to be PRIMARY.
-// Phase 1: Trust etcd and accept the candidate.
-// Future phases: Verify reachability, check term numbers, confirm agreement.
+// Currently trusts etcd - if etcd says it's PRIMARY, accept it.
 func (pd *PoolerDiscovery) considerPrimaryCandidate(cell string, candidate *topoclient.MultiPoolerInfo) {
-	// TODO(phase3): Add verification logic:
+	// TODO: Add verification logic.
 	// 1. Verify pooler is reachable (gRPC dial with timeout)
 	// 2. Call ConsensusStatus to confirm it agrees it's PRIMARY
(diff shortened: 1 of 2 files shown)


go/multigateway/discovery.go
```diff
@@ -64,71 +67,94 @@ func NewPoolerDiscovery(ctx context.Context, topoStore topoclient.Store, cell st
 }
 
 // Start begins the discovery process using topology watch.
+// It watches the local cell for all poolers, and remote cells for PRIMARY poolers only.
 func (pd *PoolerDiscovery) Start() {
-	pd.wg.Go(func() {
-		pd.logger.Info("Starting pooler discovery with topology watch", "cell", pd.cell)
+	// Discover all cells
+	allCells, err := pd.topoStore.GetCellNames(pd.ctx)
```
rafael, line 73: We don't want gateways to connect to all cells. Our idea is that this information comes from the shard metadata in the multischema, but this is not in place yet.

dweitzman:
Deepthi will look at this stuff soon, so closing for now

PR #370, opened 2025-12-17
feat(multigateway): add SCRAM-SHA-256 server authentication

dweitzman:
- Implements SCRAM-SHA-256 authentication for multigateway PostgreSQL connections
- Fetches real user credentials from multipooler via gRPC `GetAuthCredentials` endpoint
- Uses dynamic pooler discovery for authentication, selecting an available pooler at auth time
- Caches gRPC service clients alongside query connections for efficient reuse
- Accepts empty username in SCRAM client-first-message with fallback to startup message username (PostgreSQL/pgx compatibility)
- Adds `TrustAuthProvider` interface to pgprotocol/server for so unit tests can use trust authentication, similar to what happens over a unix socket locally

This might disrupt people's dev workflows since it'll start enforcing that passwords exist and have the right value when connecting to multigateway

At some point we should make the grpc Client lifetime align with discovery.go's awareness of poolers, but I don't want to refactor things right now to avoid causing conflits for upcoming work to have multigate track the primary pooler across zone

### DIFF
--- a/go/multipooler/pools/admin/admin_conn.go
+++ b/go/multipooler/pools/admin/admin_conn.go
@@ -19,12 +19,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/multigres/multigres/go/common/pgprotocol/client"
 	"github.com/multigres/multigres/go/multipooler/connstate"
 	"github.com/multigres/multigres/go/multipooler/pools/connpool"
+	"github.com/multigres/multigres/go/parser/ast"
 )
 
 // DefaultCancelTimeout is the default timeout for cancel/terminate operations.
@@ -86,14 +86,10 @@ func (c *Conn) ResetSettings(_ context.Context) error {
 // Returns an error if the user doesn't exist or if the query fails.
 //
 // This is used during authentication to retrieve password hashes for SCRAM-SHA-256 verification.
-// It works even during bootstrap before the executor is fully initialized.
 func (c *Conn) GetRolPassword(ctx context.Context, username string) (string, error) {
 	// Query pg_authid for the user's password hash.
-	// Note: This requires superuser access to read pg_authid.
-	// We don't use parameterized queries here because admin connections don't support
-	// the extended query protocol - they're meant for simple administrative operations.
 	sql := fmt.Sprintf("SELECT rolpassword FROM pg_catalog.pg_authid WHERE rolname = %s LIMIT 1",
-		quoteStringLiteral(username))
+		ast.QuoteStringLiteral(username))
 
 	results, err := c.conn.Query(ctx, sql)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


go/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -302,6 +330,43 @@ func (pg *PoolerGateway) Close(ctx context.Context) error {
...
+// getPoolerServiceClient returns a MultiPoolerServiceClient for the given database.
+// This can be used for authentication or other admin operations.
+// It finds any available pooler and returns a client connected to it.
+func (pg *PoolerGateway) getPoolerServiceClient(ctx context.Context, database string) (multipoolerpb.MultiPoolerServiceClient, error) {
+	// Find any pooler - for authentication we just need access to pg_authid
+	// which is available from any pooler connected to this database.
+	// Use PRIMARY type as it's always available.
+	target := &query.Target{
+		TableGroup: "default", // TODO: Make configurable or discover from database
+		PoolerType: clustermetadatapb.PoolerType_PRIMARY,
```
dweitzman, line 348: I'm imagining in the long term that:
- password hashes can come from any primary or replica within a database
- we might get updates through something more like a cache + streaming RPC for invalidation instead of fetching password hashes on demand at connection time

At connection time we don't yet know if the user wants a primary or replica, so assuming primary here is "correct" but until we fix primary routing across zones this could cause issues for people testing in a zone without a primary

deepthi:
> This might disrupt people's dev workflows since it'll start enforcing that passwords exist and have the right value when connecting to multigateway

What exactly do we need to change in local workflows?

dweitzman:
> > This might disrupt people's dev workflows since it'll start enforcing that passwords exist and have the right value when connecting to multigateway
> 
> What exactly do we need to change in local workflows?

A password will be required to connect. Previously we had `trust` auth so a password could be omitted

dweitzman:
> Looks good to me. Just that there quite a few wrappers and interfaces. Not really sure if they're all needed or if we can clean a couple up, but otherwise all good 👍

`PoolerSystemDiscoverer` might be able to go away after the proposed introduction of `PoolerConnection` in https://github.com/multigres/multigres/pull/516

haritabh17:
> > > This might disrupt people's dev workflows since it'll start enforcing that passwords exist and have the right value when connecting to multigateway
> > 
> > 
> > What exactly do we need to change in local workflows?
> 
> A password will be required to connect. Previously we had `trust` auth so a password could be omitted

Maybe provide a dev setup script that sets default passwords
Or add a `--dev-mode` flag to multigateway that enables trust auth for localhost connections.

Rest LGTM.

dweitzman:
> > > > This might disrupt people's dev workflows since it'll start enforcing that passwords exist and have the right value when connecting to multigateway
> > > 
> > > 
> > > What exactly do we need to change in local workflows?
> > 
> > 
> > A password will be required to connect. Previously we had `trust` auth so a password could be omitted
> 
> Maybe provide a dev setup script that sets default passwords Or add a `--dev-mode` flag to multigateway that enables trust auth for localhost connections.
> 
> Rest LGTM.

Dev already sets default passwords, but it might be worth updating the `psql` command that gets printed out for how to connect. I'll do a test and make sure that the local provisioner and demo/k8s print commands that'll actually work

dweitzman:
Added changes to:
- Propagate the username down to multipooler so it uses the correct connection pool
- Print the full psql command with password when advising how to connect during `cluster start` or for the k8s demo

dweitzman:
More changes:
- Needed to use an admin connection for fetching password hashes
- Fixed some tests that were passing in the wrong password

haritabh17 (approved):

PR #371, opened 2025-12-19
refactor(multipooler): migrate heartbeat package to use executor connection pool

GuptaManan100:
## Description                                                                                                                                   
- Introduces `InternalQuerier` interface in the executor package for internal queries using connection pooling                   
- Migrates heartbeat package from `*sql.DB` to use the executor's pooled connections with "postgres" user
- Simplifies heartbeat writer by replacing PID tracking/kill logic with context cancellation                                     
                                                                                                                                   
## Test plan
- [x] All heartbeat tests pass

### DIFF
--- a/go/multipooler/manager/manager.go
+++ b/go/multipooler/manager/manager.go
@@ -358,10 +358,15 @@ func (pm *MultiPoolerManager) closeConnectionsLocked() error {
 	}
 
 	if pm.db != nil {
-		if err := pm.db.Close(); err != nil {
-			return err
-		}
+		err := pm.db.Close()
+		// Always nil out pm.db to allow reconnection. When postgres is down,
+		// pm.db may have idle broken connections, and Close() can return errors
+		// like "broken pipe". We must nil out pm.db regardless so that
+		// connectDB() will create a fresh connection on the next Open().
 		pm.db = nil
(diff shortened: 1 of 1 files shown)


go/multipooler/heartbeat/writer.go
```diff
@@ -136,22 +130,30 @@ func (w *Writer) enableWrites() {
...
-	ctx, cancel := context.WithCancel(context.TODO())
-	go func() {
-		w.ticks.Stop()
-		cancel()
-	}()
-	w.killWritesUntilStopped(ctx)
+	// Cancel any ongoing write to unblock it
+	w.writeMu.Lock()
+	if w.writeCancel != nil {
+		w.writeCancel()
```
cuongdo, line 144: does this cancel require making a separate connection to Postgres (like it does in lib/pq)? if so, is the pool guaranteed to have a connection available for that?

GuptaManan100: Yep, it manages this through the admin pool. The pool starts with 5 connections (configurable, 5 is the default), and the pool doesn't support any long lived operations. It only supports killing queries and connections, none of which are blocking. So we're good, even if there is no connection now, one will be available very soon.

PR #373, opened 2025-12-19
refactor(endtoend): Improvements end to end

rafael:
# Summary
- Extract end-to-end test infrastructure into a reusable `shardsetup` package.
- Refactor `multipooler` and `multiorch` tests to use shared infrastructure.
- Add validation to ensure `multipooler` tests maintain the 2-node assumption.

## Context
The `multipooler` and `multiorch` test suites had duplicated infrastructure for spinning up `etcd`, `pgctld`, `multipooler`, and `multiorch` processes. Both maintained their own copies of process management, cleanup, and replication setup code.

The old design also hardcoded a 2-node topology (`PrimaryMultipooler`, `StandbyMultipooler` fields), which does not match how `multiorch` actually works. It elects a primary dynamically from N candidates.

The main idea of this change is to provide a consistent way to bring up a shard in end-to-end tests. Where the setup can be shared between tests, the framework resets the cluster state between test cases (this is what we were already doing for the `multipooler` tests). For `multiorch`, where tests can be more destructive, we create shards in isolation that live for the duration of the test. This part is not ideal, but it is the safest approach for now.

## Approach
The new `shardsetup` package provides:
- N-node support: `WithMultipoolerCount(3)` creates 3 equivalent nodes; the primary is discovered after bootstrap.
- Shared vs isolated setups: `SharedSetupManager` for test reuse, `NewIsolated()` for destructive tests that kill primaries.
- Backward compatibility: `MultipoolerTestSetup` wrapper preserves the old field-based access for existing `multipooler` tests.

The `multipooler` tests continue using a 2-node setup with a guard that fails fast if misconfigured. Tests needing 3+ nodes use `shardsetup` directly.

## Test plan
I've run the tests multiple times with this and haven't been able to see flakes. 
- `go test -v ./go/test/endtoend/... -count=10`

### DIFF
--- a/go/test/endtoend/multiorch/bootstrap_test.go
+++ b/go/test/endtoend/multiorch/bootstrap_test.go
@@ -155,6 +155,11 @@ func TestBootstrapInitialization(t *testing.T) {
 		require.True(t, ok, "requiredCount should be a number")
 		assert.Equal(t, float64(2), requiredCount)
 
+		// Verify Description
+		description, ok := quorumRule["description"].(string)
+		require.True(t, ok, "description should be a string")
+		assert.Equal(t, "Any 2 nodes must acknowledge", description)
+
 		t.Logf("Verified durability policy: policy_name=%s, quorum_type=ANY_N, required_count=%d",
 			policyName, int(requiredCount))
 	})


go/test/endtoend/multipooler/setup_test.go
```diff
@@ -12,1146 +12,232 @@
...
-	"github.com/multigres/multigres/go/test/utils"
-	"github.com/multigres/multigres/go/tools/pathutil"
+	"github.com/multigres/multigres/go/test/endtoend/shardsetup"
 
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
-	consensuspb "github.com/multigres/multigres/go/pb/consensus"
-	consensusdatapb "github.com/multigres/multigres/go/pb/consensusdata"
 	multipoolermanagerpb "github.com/multigres/multigres/go/pb/multipoolermanager"
 	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
-	"github.com/multigres/multigres/go/pb/pgctldservice"
```
rafael, line 49: most of the code here was refactored in to the shardsetup package

go/test/endtoend/multipooler_client.go
```diff
@@ -60,21 +60,6 @@ func NewMultiPoolerTestClient(addr string) (*MultiPoolerTestClient, error) {
 
 	client := multipoolerpb.NewMultiPoolerServiceClient(conn)
 
-	// Test the connection by making a simple RPC call with a short timeout
-	ctx, cancel := context.WithTimeout(context.TODO(), 2*time.Second)
-	defer cancel()
-
-	// Try to make a basic ExecuteQuery call to test connectivity
-	_, err = client.ExecuteQuery(ctx, &multipoolerpb.ExecuteQueryRequest{
```
rafael, line 68: I removed this check because this client is now use in a generic way where the manager could be ready, but the pooler is not serving queries yet. So this code breaks the client on those contexts

go/test/endtoend/multiorch/bootstrap_test.go
```diff
@@ -22,122 +22,123 @@ package multiorch
...
 
 	// Verify nodes are completely uninitialized (no data directory, no postgres running)
 	// Multiorch will detect these as needing bootstrap and call InitializeEmptyPrimary
-	// which will create the data directory, configure archive mode, and start postgres
-	for i, node := range nodes {
-		status := checkInitializationStatus(t, node)
-		t.Logf("Node %d (%s) Status: IsInitialized=%v, HasDataDirectory=%v, PostgresRunning=%v, PostgresRole=%s, PoolerType=%s",
-			i, node.name, status.IsInitialized, status.HasDataDirectory, status.PostgresRunning, status.PostgresRole, status.PoolerType)
+	for name, inst := range setup.Multipoolers {
+		client, err := shardsetup.NewMultipoolerClient(inst.Multipooler.GrpcPort)
```
dweitzman: Should the created multipooler client be closed at some ponit?

go/test/endtoend/multiorch/bootstrap_test.go
```diff
@@ -22,122 +22,123 @@ package multiorch
...
+		require.NoError(t, err, "should connect to %s", name)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
dweitzman: I like the `utils.WithTimeout()` helper you added at some point for this kind of timeout, although the helper code itself could use `t.Context()` instead of `context.Background()`

I might make a PR to try to use `utils.WithTimeout()` more widely

go/test/endtoend/multiorch/bootstrap_test.go
```diff
@@ -22,122 +22,123 @@ package multiorch
...
-	env.setupPgBackRest()
-	env.registerNodes()
+		client.Close()
+	}
 
-	// Start multiorch and wait for bootstrap
-	env.startMultiOrch()
+	// Create and start multiorch to trigger bootstrap
+	watchTargets := []string{"postgres/default/0-inf"}
+	mo := setup.CreateMultiOrchInstance(t, "test-multiorch", setup.CellName, watchTargets)
```
dweitzman: Will we always want to call terminate gracefully at the end of a test? If so, I wonder if it's make sense for CreateMultiOrchInstance() to return both the multiorch and also a cleanup function to make it harder for people to forget to call cleanup

go/test/endtoend/multiorch/bootstrap_test.go
```diff
@@ -154,62 +155,65 @@ func TestBootstrapInitialization(t *testing.T) {
 		require.True(t, ok, "requiredCount should be a number")
 		assert.Equal(t, float64(2), requiredCount)
 
-		// Verify Description
```
dweitzman, line 157: Was this assertion not needed?

go/test/endtoend/shardsetup/process.go
```diff
@@ -0,0 +1,407 @@
...
+	// Start the multipooler server
+	p.Process = exec.Command(p.Binary, args...)
+
+	// Set MULTIGRES_TESTDATA_DIR for directory-deletion triggered cleanup
+	p.Process.Env = append(p.Environment,
+		"MULTIGRES_TESTDATA_DIR="+filepath.Dir(p.DataDir),
+	)
+
+	t.Logf("Running multipooler command: %v", p.Process.Args)
+	return p.waitForStartup(t, 15*time.Second, 30)
```
dweitzman, line 149: Not a blocker for now, but I wonder if it might make sense for the start function to not do a blocking wait for the startup to succeed so that it's possible to start multiple processes in parallel and wait on them afterwards

go/test/endtoend/shardsetup/setup.go
```diff
@@ -0,0 +1,1146 @@
...
+	for _, opt := range opts {
+		opt(config)
+	}
+
+	if config.MultipoolerCount < 1 {
+		t.Fatalf("MultipoolerCount must be at least 1, got %d", config.MultipoolerCount)
+	}
+
+	// Set the PATH so our binaries can be found
+	if err := pathutil.PrependBinToPath(); err != nil {
```
dweitzman: TestMain() also calls `PrependBinToPath()`, which feels like a more reasonable place to me unless this once was using `sync.Once`

Can we assume or assert this will only be called when the test main function is in use and the binaries are already in PATH?

go/test/endtoend/shardsetup/setup.go
```diff
@@ -0,0 +1,1146 @@
...
+func waitForShardBootstrap(t *testing.T, setup *ShardSetup) (string, error) {
+	t.Helper()
+
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+	timeout := time.After(30 * time.Second)
+
+	for {
+		select {
+		case <-timeout:
```
dweitzman: Not a blocking comment, but you could use WithTimeout() on t.Context() here and wait for `ctx.Done()` instead of `time.After`.

go/test/endtoend/shardsetup/setup.go
```diff
@@ -0,0 +1,1146 @@
...
+		defer queryCancel()
+		inRecovery, err := QueryStringValue(queryCtx, client.Pooler, "SELECT pg_is_in_recovery()")
+		return err == nil && inRecovery == "t"
+	}, 10*time.Second, 100*time.Millisecond, "primary should be in recovery mode after demotion")
+
+	t.Log("Primary demoted successfully")
+}
+
+// GetClient returns a MultipoolerClient for the specified multipooler instance.
+func (s *ShardSetup) GetClient(t *testing.T, name string) *MultipoolerClient {
```
dweitzman: Maybe this should be called `NewClient` to make it clear the client was constructed and may need to be closed?

go/test/endtoend/shardsetup/setup.go
```diff
@@ -0,0 +1,1146 @@
...
+	}
+
+	client, err := NewMultipoolerClient(inst.Multipooler.GrpcPort)
+	require.NoError(t, err, "failed to connect to %s", name)
+
+	return client
+}
+
+// GetPrimaryClient returns a MultipoolerClient for the primary instance.
+func (s *ShardSetup) GetPrimaryClient(t *testing.T) *MultipoolerClient {
```
dweitzman: Similar comment re: naming: could be `NewPrimaryClient()` for clarity

go/test/endtoend/shardsetup/client.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if len(resp.Rows) == 0 || len(resp.Rows[0].Values) == 0 {
+		return "", nil
+	}
+	return string(resp.Rows[0].Values[0]), nil
+}
+
+// ValidateGUCValue queries a GUC and returns an error if it doesn't match the expected value.
+// Follows the pattern from multipooler/setup_test.go:validateGUCValue.
+func ValidateGUCValue(ctx context.Context, client *endtoend.MultiPoolerTestClient, gucName, expected, instanceName string) error {
```
dweitzman: This file doesn't feel like the most obvious location to me for SaveGUCs() and RestoreGUCs()

go/test/endtoend/shardsetup/setup.go
```diff
@@ -0,0 +1,1146 @@
...
+			LockPath:        lockPath,
+			RetentionFull:   2,
+		}
+		if err := pgbackrest.WriteConfigFile(configPath, backupCfg); err != nil {
+			t.Fatalf("failed to write pgbackrest config for %s: %v", inst.Name, err)
+		}
+		t.Logf("Created pgbackrest config for %s (stanza: %s)", inst.Name, stanza)
+
+		// Start multipooler
+		if err := multipooler.Start(t); err != nil {
```
dweitzman, line 605: Long term, if Start() is going to block for starting to finish then it feels like all the `Start()` functions should be called in parallel in a wait group

If Start() becomes non-blocking then the waiting could happen later

go/test/endtoend/shardsetup/client.go
rafael: I moved all these to cleanup.go

go/test/endtoend/shardsetup/setup.go
rafael: I added assertion here, removed the PrependBinToPath

go/test/endtoend/multiorch/bootstrap_test.go
rafael: I  addressed this

go/test/endtoend/multiorch/bootstrap_test.go
rafael: It does happen, but I refactor this bit of code to make it more explicit.

PR #374, opened 2025-12-22
feat: Move remaining sql.DB usages to use the internalQuerier instead

GuptaManan100:
### Description

This PR is the spiritual successor of https://github.com/multigres/multigres/pull/371. It replaces remaining `sql.DB` usages in multipooler with the `internalQuerier` interface for consistent database access.

Key changes:
- Removes direct `sql.DB` dependency from manager package (deletes db.go). We only use `sql.DB` for testing now.
- Adds structured Result type for query results with proper row scanning
- Updates replication, consensus, and multischema code to use internal querier

### DIFF
--- a/go/multipooler/executor/mock/mock_querier.go
+++ b/go/multipooler/executor/mock/mock_querier.go
@@ -103,6 +103,35 @@ func (m *QueryService) AddQueryPatternOnce(pattern string, result *query.QueryRe
 	})
 }
 
+// AddQueryPatternOnceWithError adds a query pattern that returns an error and is consumed after the first match.
+func (m *QueryService) AddQueryPatternOnceWithError(pattern string, err error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.patterns = append(m.patterns, queryPattern{
+		pattern:     regexp.MustCompile(pattern),
+		err:         err,
+		consumeOnce: true,
+	})
+}
+
+// ExpectationsWereMet returns an error if any consumeOnce patterns were not matched.
+// This is useful for verifying that all expected queries were executed.
+func (m *QueryService) ExpectationsWereMet() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	var unmet []string
+	for _, p := range m.patterns {
+		if p.consumeOnce {
+			unmet = append(unmet, p.pattern.String())
+		}
+	}
+	if len(unmet) > 0 {
+		return fmt.Errorf("expected queries were not executed: %v", unmet)
+	}
+	return nil
+}
+
 // Query implements executor.InternalQueryService.
 func (m *QueryService) Query(ctx context.Context, queryStr string) (*query.QueryResult, error) {
 	m.mu.Lock()
--- a/go/multipooler/manager/pg_multischema_test.go
+++ b/go/multipooler/manager/pg_multischema_test.go
@@ -80,13 +80,13 @@ func TestCreateSidecarSchema(t *testing.T) {
 			name:       "successful schema creation for default tablegroup",
 			tableGroup: constants.DefaultTableGroup,
 			setupMock: func(m *mock.QueryService) {
-				m.AddQueryPattern("CREATE SCHEMA IF NOT EXISTS multigres", mock.MakeQueryResult(nil, nil))
-				m.AddQueryPattern("CREATE TABLE IF NOT EXISTS multigres.heartbeat", mock.MakeQueryResult(nil, nil))
-				m.AddQueryPattern("CREATE TABLE IF NOT EXISTS multigres.durability_policy", mock.MakeQueryResult(nil, nil))
-				m.AddQueryPattern("CREATE INDEX IF NOT EXISTS idx_durability_policy_active", mock.MakeQueryResult(nil, nil))
-				m.AddQueryPattern("CREATE TABLE IF NOT EXISTS multigres.tablegroup", mock.MakeQueryResult(nil, nil))
-				m.AddQueryPattern("CREATE TABLE IF NOT EXISTS multigres.tablegroup_table", mock.MakeQueryResult(nil, nil))
-				m.AddQueryPattern("CREATE TABLE IF NOT EXISTS multigres.shard", mock.MakeQueryResult(nil, nil))
+				m.AddQueryPatternOnce("CREATE SCHEMA IF NOT EXISTS multigres", mock.MakeQueryResult(nil, nil))
+				m.AddQueryPatternOnce("CREATE TABLE IF NOT EXISTS multigres.heartbeat", mock.MakeQueryResult(nil, nil))
+				m.AddQueryPatternOnce("CREATE TABLE IF NOT EXISTS multigres.durability_policy", mock.MakeQueryResult(nil, nil))
+				m.AddQueryPatternOnce("CREATE INDEX IF NOT EXISTS idx_durability_policy_active", mock.MakeQueryResult(nil, nil))
+				m.AddQueryPatternOnce("CREATE TABLE IF NOT EXISTS multigres.tablegroup", mock.MakeQueryResult(nil, nil))
+				m.AddQueryPatternOnce("CREATE TABLE IF NOT EXISTS multigres.tablegroup_table", mock.MakeQueryResult(nil, nil))
+				m.AddQueryPatternOnce("CREATE TABLE IF NOT EXISTS multigres.shard", mock.MakeQueryResult(nil, nil))
 			},
 			expectError: false,
 		},
@@ -94,7 +94,7 @@ func TestCreateSidecarSchema(t *testing.T) {
 			name:       "schema creation fails",
 			tableGroup: constants.DefaultTableGroup,
 			setupMock: func(m *mock.QueryService) {
-				m.AddQueryPatternWithError("CREATE SCHEMA IF NOT EXISTS multigres", fmt.Errorf("permission denied"))
+				m.AddQueryPatternOnceWithError("CREATE SCHEMA IF NOT EXISTS multigres", fmt.Errorf("permission denied"))
 			},
 			expectError:   true,
 			errorContains: "failed to create multigres schema",
@@ -103,8 +103,8 @@ func TestCreateSidecarSchema(t *testing.T) {
 			name:       "heartbeat table creation fails",
 			tableGroup: constants.DefaultTableGroup,
 			setupMock: func(m *mock.QueryService) {
-				m.AddQueryPattern("CREATE SCHEMA IF NOT EXISTS multigres", mock.MakeQueryResult(nil, nil))
-				m.AddQueryPatternWithError("CREATE TABLE IF NOT EXISTS multigres.heartbeat", fmt.Errorf("table creation failed"))
+				m.AddQueryPatternOnce("CREATE SCHEMA IF NOT EXISTS multigres", mock.MakeQueryResult(nil, nil))
+				m.AddQueryPatternOnceWithError("CREATE TABLE IF NOT EXISTS multigres.heartbeat", fmt.Errorf("table creation failed"))
 			},
 			expectError:   true,
 			errorContains: "failed to create heartbeat table",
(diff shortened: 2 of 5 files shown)


go/multipooler/executor/result.go
```diff
@@ -0,0 +1,167 @@
...
+// Returns an error if the result has no rows.
+func ScanSingleRow(result *query.QueryResult, dests ...any) error {
+	if result == nil || len(result.Rows) == 0 {
+		return fmt.Errorf("no rows in result")
+	}
+	return ScanRow(result.Rows[0], dests...)
+}
+
+// scanValue scans a single value into the destination.
+func scanValue(val []byte, dest any) error {
```
rafael, line 55: could be as fast follow, but we should extensively unit test this file

go/multipooler/executor/result.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		*d = v
+	case *float64:
+		v, err := strconv.ParseFloat(s, 64)
+		if err != nil {
+			return fmt.Errorf("cannot parse %q as float64: %w", s, err)
+		}
+		*d = v
+	case *time.Time:
+		// Try common PostgreSQL timestamp formats
```
rafael: Is there a canonical reference for this to make sure we are not missing any ?

go/multipooler/manager/pg_multischema_test.go
```diff
@@ -19,131 +19,126 @@ import (
...
-					WillReturnResult(sqlmock.NewResult(0, 0))
-				mock.ExpectExec(regexp.QuoteMeta("CREATE INDEX IF NOT EXISTS idx_durability_policy_active")).
-					WillReturnResult(sqlmock.NewResult(0, 0))
-				mock.ExpectExec(regexp.QuoteMeta("CREATE TABLE IF NOT EXISTS multigres.tablegroup")).
-					WillReturnResult(sqlmock.NewResult(0, 0))
-				mock.ExpectExec(regexp.QuoteMeta("CREATE TABLE IF NOT EXISTS multigres.table")).
-					WillReturnResult(sqlmock.NewResult(0, 0))
-				mock.ExpectExec(regexp.QuoteMeta("CREATE TABLE IF NOT EXISTS multigres.shard")).
-					WillReturnResult(sqlmock.NewResult(0, 0))
+			setupMock: func(m *mock.Querier) {
```
rafael: This is a nit - but I like the other interface better. I think it was easier to read what was going on with the Mock.

go/multipooler/manager/pg_multischema_test.go
```diff
@@ -257,8 +232,6 @@ func TestInsertDurabilityPolicy(t *testing.T) {
 			} else {
 				assert.NoError(t, err)
 			}
-
-			assert.NoError(t, mock.ExpectationsWereMet())
```
rafael, line 261: what's the equivalent of this in the new mock implementation?

go/multipooler/manager/pg_replication.go
```diff
@@ -71,129 +71,144 @@ func formatStandbyList(standbyIDs []*clustermetadatapb.ID) string {
...
-	)
+	result, err := pm.query(queryCtx, sql)
 	if err != nil {
 		return nil, mterrors.Wrap(err, "failed to query replication status")
 	}
 
-	status := &multipoolermanagerdatapb.StandbyReplicationStatus{
-		IsWalReplayPaused:   isPaused,
-		WalReplayPauseState: pauseState,
+	if result == nil || len(result.Rows) == 0 {
```
rafael: nit - It would be nice to also have the same interface where you can scan multiple values in the same request It is less verbose.

go/multipooler/manager/pg_replication_test.go
```diff
@@ -683,118 +680,94 @@ func TestApplyReplaceOperation(t *testing.T) {
...
 			name: "primary server - not in recovery",
-			setupMock: func(mock sqlmock.Sqlmock) {
-				rows := sqlmock.NewRows([]string{"pg_is_in_recovery"}).AddRow(false)
-				mock.ExpectQuery("SELECT pg_is_in_recovery\\(\\)").WillReturnRows(rows)
+			setupMock: func(m *mock.Querier) {
+				m.AddQueryPattern("SELECT pg_is_in_recovery", mock.MakeQueryResult([]string{"pg_is_in_recovery"}, [][]any{{"f"}}))
 			},
 			expectError:  false,
-			expectResult: true,
+			expectResult: false,
```
rafael, line 696: why this change?

GuptaManan100: Previously the test was for `isPrimary`. It's now for `isInRecovery`. They're the same except the results are exactly the opposite

go/multipooler/executor/mock/mock_querier.go
```diff
@@ -90,22 +91,43 @@ func (m *Querier) AddQueryPatternWithContextCallback(pattern string, result *que
...
+	defer m.mu.Unlock()
+	m.patterns = append(m.patterns, queryPattern{
+		pattern:     regexp.MustCompile(pattern),
+		result:      result,
+		consumeOnce: true,
+	})
+}
+
 // Query implements executor.InternalQuerier.
 func (m *Querier) Query(ctx context.Context, queryStr string) (*query.QueryResult, error) {
```
deepthi: Querier and InternalQuerier are somewhat awkward to say. Can they be renamed to MockQueryService and InternalQueryService?

go/multipooler/manager/manager.go
```diff
@@ -1064,8 +1032,8 @@ func (pm *MultiPoolerManager) getActiveWriteConnections(ctx context.Context) ([]
 	// have the query pool. Thinking that we should have a
 	// specific user for the write pool and we can kill all connections
 	// associated with that user.
-	query := `
-		SELECT COALESCE(array_agg(pid), ARRAY[]::integer[])
+	sql := `
+		SELECT pid
```
deepthi, line 1054: 💯

go/multipooler/manager/pg_replication.go
```diff
@@ -71,129 +71,144 @@ func formatStandbyList(standbyIDs []*clustermetadatapb.ID) string {
...
 	}
 
-	if replayLsn.Valid {
-		status.LastReplayLsn = replayLsn.String
+	row := result.Rows[0]
+	status := &multipoolermanagerdatapb.StandbyReplicationStatus{}
+
+	// Get replay LSN (may be NULL)
+	if replayLsn, err := executor.GetString(row, 0); err == nil && replayLsn != "" {
+		status.LastReplayLsn = replayLsn
```
deepthi: Checking for non-empty string changes the meaning of this check. Previously we were checking for NULL via `.Valid`
```
type NullString struct {
	String string
	Valid  bool // Valid is true if String is not NULL
}
```
If we want to get `''` when the LSN is NULL, we have to 
```
SELECT COALESCE(pg_last_wal_receive_lsn()::text, '')
```

go/multipooler/manager/pg_replication.go
```diff
@@ -71,129 +71,144 @@ func formatStandbyList(standbyIDs []*clustermetadatapb.ID) string {
...
 
+	// Get pause state (string)
+	if pauseState, err := executor.GetString(row, 3); err == nil {
+		status.WalReplayPauseState = pauseState
+	}
+
+	// Get last xact time (may be NULL)
+	if lastXactTime, err := executor.GetString(row, 4); err == nil && lastXactTime != "" {
+		status.LastXactReplayTimestamp = lastXactTime
+	}
```
deepthi: Similar to LSNs above: NULL and empty string are not the same.
`::text` preserves NULL-ness, it does not convert NULL to empty string.

go/multipooler/manager/rpc_manager.go
```diff
@@ -101,13 +101,13 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoLocked(ctx context.Context, host
 	}
 
 	// Guardrail: Check if the PostgreSQL instance is in recovery (standby mode)
-	isPrimary, err := pm.isPrimary(ctx)
+	inRecovery, err := pm.isInRecovery(ctx)
```
deepthi: Why did this need to change? Code is more readable without negations.

go/multipooler/manager/rpc_manager.go
```diff
@@ -264,26 +264,21 @@ func (pm *MultiPoolerManager) Status(ctx context.Context) (*multipoolermanagerda
...
-	}
-
 	// Get WAL position (ignore errors, just return empty string)
 	walPosition, _ := pm.getWALPosition(ctx)
 	poolerStatus.WalPosition = walPosition
 
-	// Database is connected - try to get detailed status based on PostgreSQL role
-	isPrimary, err := pm.isPrimary(ctx)
+	// Try to get detailed status based on PostgreSQL role
+	inRecovery, err := pm.isInRecovery(ctx)
```
deepthi: Same comment here about `isPrimary` vs `!inRecovery`

go/multipooler/executor/result.go
GuptaManan100: Added

go/multipooler/executor/result.go
GuptaManan100: Added the reference to the docs

go/multipooler/executor/mock/mock_querier.go
GuptaManan100: Fixed

go/multipooler/manager/pg_replication.go
GuptaManan100: Fixed

go/multipooler/manager/pg_replication.go
GuptaManan100: Fixed

go/multipooler/manager/pg_replication.go
GuptaManan100: Fixed

go/multipooler/manager/rpc_manager.go
GuptaManan100: Fixed

go/multipooler/manager/rpc_manager.go
GuptaManan100: Fixed. It was one of those things that Claude messed up.

go/multipooler/executor/internal_query.go
```diff
@@ -52,3 +59,25 @@ func (e *Executor) Query(ctx context.Context, queryStr string) (*query.QueryResu
...
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Recycle()
+
+	results, err := conn.Conn.QueryArgs(ctx, sql, args...)
+	if err != nil {
+		return nil, err
+	}
+	if len(results) == 0 {
```
dweitzman: For defense in depth, would it make sense for this to return some kind of internal error if the results length is greater than 1?

go/multipooler/manager/rpc_consensus.go
```diff
@@ -136,18 +136,21 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
 				pm.logger.InfoContext(ctx, "Standby has no WAL receiver, proceeding with term acceptance",
 					"term", req.Term)
-			} else if timeStr, _ := executor.GetString(result.Rows[0], 0); timeStr != "" {
-				// Parse the timestamp
-				t, parseErr := time.Parse("2006-01-02 15:04:05.999999-07", timeStr)
-				if parseErr == nil {
-					lastMsgReceiptTime = &t
+			} else {
+				queryErr = executor.ScanSingleRow(result, &lastMsgReceiptTime)
+				if queryErr != nil {
```
deepthi, line 148: I'm curious, how does this happen? No error from executing the query, but an error from parsing the result?

deepthi (approved):

go/multipooler/executor/result_test.go
```diff
@@ -0,0 +1,603 @@
...
+}
+
+func TestScanValue_Bool(t *testing.T) {
+	tests := []struct {
+		name        string
+		input       string
+		expected    bool
+		expectError bool
+	}{
+		{"t", "t", true, false},
```
dweitzman: For these error test cases, might be nice to have the expected error message instead of just a boolean for whether or not an error happened. Same comment for other similar test cases in this file

dweitzman (approved):

go/multipooler/manager/rpc_consensus.go
GuptaManan100: It shouldn't happen, but can if parsing the output doesn't match the type requested

go/multipooler/executor/internal_query.go
GuptaManan100: Fixed

go/multipooler/executor/result_test.go
GuptaManan100: Fixed

go/multipooler/manager/pg_multischema_test.go
GuptaManan100: FIxed! It is now similar

go/multipooler/manager/pg_multischema_test.go
GuptaManan100: Fixed

PR #376, opened 2025-12-22
refactor: extract service names and PostgreSQL defaults as constants

dweitzman:
### Motivation

I noticed that it's tricky to identify code that's using `"postgres"` as a username, database name, executable name, etc. which makes refactorings and behavior changes harder, so using pre-declared constants feels like a better way to link the hard-coded strings to their purpose

This PR goes a little beyond that and enables `goconst` to discourage repeating the same string constants multiple times in a package and I also pulled out constants for service names like `multipooler`, `multiadmin`, etc. For those changes the value added seems less obvious, but it's at least some level of defense against accidental typos finding their way into metrics and things. I'm not going to push for the service constants if other folks don't think it's worth doing.

### Details

- Add centralized constants for service names (`ServiceMultigateway`, `ServiceMultipooler`, etc.) and PostgreSQL defaults (`DefaultPostgresUser`, `DefaultPostgresDatabase`) in `go/common/constants/`
- Replace string literals across the codebase with typed constants to enable safer refactoring
- ~~Enable `goconst` linter~~ — commented out due to unexplained CI/local inconsistency (CI reports issues that can't be reproduced locally or in Docker Linux with identical versions)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -25,7 +25,11 @@ linters:
     - errcheck
     - errorlint
     - forbidigo
-    - goconst
+    # TODO: Enable goconst - disabled because it reports different results in CI vs local.
+    # CI (Linux) detects issues that cannot be reproduced locally (macOS) or in Docker (Linux),
+    # even with the same golangci-lint version and config. Until this is resolved, developers
+    # can't validate goconst locally before pushing.
+    # - goconst
     - gocritic
     - gosec
     - govet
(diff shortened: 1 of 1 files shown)


misc/hooks/golangci-lint
```diff
@@ -67,6 +67,9 @@ fi
 # Get unique directories of the Go files
 gopackages=$(echo "$gofiles" | xargs -n1 dirname | sort -u | paste -sd ' ' -)
 
+# Verify config is valid (catches invalid options that 'run' silently ignores)
+golangci-lint config verify
```
dweitzman, line 71: Sneaking in a bonus check to catch in the future during `git commit` if `.golangci.yml` has validation issues that `golangci-lint` would silently ignore in a normal run

deepthi (approved):

PR #384, opened 2025-12-23
feat(multigateway): add multi-cell pooler discovery

dweitzman:
Refactor pooler discovery to watch all cells for multipoolers, enabling cross-cell primary routing while maintaining local cell affinity for replicas.

- Rename PoolerDiscovery to CellPoolerDiscovery (single cell watcher)
- Add GlobalPoolerDiscovery to orchestrate per-cell watchers
- Implement cell affinity: replicas prefer local cell, primaries search all cells
- Update admin status page to show per-cell pooler lists with last refresh
- Simplify e2e tests since any gateway can now find primary in any zone


deepthi (approved):
Some of this will clearly change when we implement continuous health checking of poolers from gateways, but this looks good enough for now.
cc @GuptaManan100

PR #385, opened 2025-12-23
fix(provisioner): use absolute paths for pgbackrest and pq driver

dweitzman:
pgbackrest requires absolute paths for all directory configurations, and the pq driver requires paths starting with '/' for Unix sockets. Add EnsureAbs utility to pathutil and use it in config generation and db connection setup to ensure paths are absolute.

The motivation for this is that in local testing I was getting bootstrap failures that left poolers stuck without being assigned a replica type. These changes seemed to help, for me at least.


deepthi:
@dweitzman how much of this is obsolete given the changes made in #438?

PR #387, opened 2025-12-24
test(multipooler): add unit tests for connection pool packages

GuptaManan100:
### Description
- Add `fakepgserver` package that speaks PostgreSQL wire protocol for testing pools without a real database
- Add unit tests for `admin`, `regular`, and `reserved` pool packages

### DIFF
--- a/go/multipooler/pools/admin/admin_test.go
+++ b/go/multipooler/pools/admin/admin_test.go
@@ -83,6 +83,9 @@ func TestPool_TerminateBackend(t *testing.T) {
 	success, err := pool.TerminateBackend(ctx, 12345)
 	require.NoError(t, err)
 	assert.True(t, success)
+
+	// Verify the query was actually executed.
+	server.VerifyAllPatternsUsedOrFail()
 }
 
 func TestPool_TerminateBackend_NotFound(t *testing.T) {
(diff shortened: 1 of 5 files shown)


go/multipooler/pools/admin/admin_test.go
```diff
@@ -0,0 +1,228 @@
...
+		[][]any{{"t"}}, // "t" = true in PostgreSQL
+	))
+
+	pool := newTestPool(t, server)
+	defer pool.Close()
+
+	ctx := context.Background()
+
+	// Terminate a backend.
+	success, err := pool.TerminateBackend(ctx, 12345)
```
dweitzman, line 83: I don't think the fake is verifying that pg_terminate_backend() was actually called, so it seems like this test would pass even if TerminateBackend() was a no-op?

Could/should the fake server add some kind of t.Cleanup() check to make sure all registered query patterns were actually used?

dweitzman: Another potential approach could be `terminateQueries = s.AddQueryPattern(...)` returning some kind of `QueryMatcher` struct that saves all the matching queries and lets you do `assert.Equal(t, 1, termateQueries.Count())` or something like that at the end of the test

GuptaManan100: Yes, that's an excellent idea. And you're right it's not testing we use the patterns. I don't know how I missed this, but thank you for catching it

GuptaManan100: Fixed

PR #388, opened 2025-12-24
feat(multipooler): pass primary pooler ID in SetPrimaryConnInfo and store it in standby pooler's manager struct

deepthi:
There are two use cases for this feature
1. pgBackRest config will be generated by each pooler for itself. In production configs, it only needs the host, and the data path is by convention. However, in local setups where each pooler's data path is prefixed by the pooler id, we need the primary pooler id in order to infer its data path and add it as pg2-path into the config.
2. We have not yet implemented streaming health checks between multigateway and multipooler. When we do, we will be using the primary pooler ID in that feature.

### DIFF
--- a/go/multipooler/manager/rpc_manager.go
+++ b/go/multipooler/manager/rpc_manager.go
@@ -90,7 +90,12 @@ func (pm *MultiPoolerManager) SetPrimaryConnInfo(ctx context.Context, primary *c
 	var port int32
 	if primary != nil {
 		host = primary.Hostname
-		port = primary.PortMap["postgres"]
+		var ok bool
+		port, ok = primary.PortMap["postgres"]
+		if !ok {
+			return mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT,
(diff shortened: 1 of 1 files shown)


proto/multipoolermanagerdata.proto
```diff
@@ -88,23 +88,21 @@ message StartReplicationResponse {
...
 // SetPrimaryConnInfo sets the primary connection info for a standby server
 message SetPrimaryConnInfoRequest {
-  // Primary server hostname
-  string host = 1;
-
-  // Primary server port
-  int32 port = 2;
+  // Primary pooler metadata. Contains id (cell+name), hostname, and port_map.
+  // The postgres port is extracted from port_map["postgres"].
+  clustermetadata.MultiPooler primary = 1;
```
deepthi, line 93: It was cleaner to replace host/port with a full MultiPooler struct, than to just add another field here for the primary ID. Passing 3 args instead of one opened the door to inconsistencies from callers, which were detected through testing.

PR #390, opened 2025-12-25
fix(multipooler): prevent reserved connection ID collisions after restart

GuptaManan100:
### Description

Fixes an issue where reserved connection IDs could collide after multipooler restarts, potentially causing queries to execute on the wrong connection.

**The problem:** Connection IDs were generated sequentially starting from 1. After a multipooler restart, the counter would reset, and new connections could receive the same IDs as connections from before the restart. If a multigateway was still holding a reference to an old connection ID, it could inadvertently route queries to a different connection.

**The fix:** Initialize the connection ID counter (`lastID`) with the current Unix nanosecond timestamp when creating the pool. This ensures IDs from different pool instances never overlap, as each restart starts from a later timestamp.

Addresses https://github.com/multigres/multigres/pull/337#discussion_r2620745461

### DIFF
--- a/go/multipooler/pools/reserved/reserved_pool.go
+++ b/go/multipooler/pools/reserved/reserved_pool.go
@@ -62,7 +62,8 @@ type Pool struct {
 	// active tracks reserved connections by their unique ID.
 	active map[int64]*Conn
 
-	// lastID generates unique connection IDs.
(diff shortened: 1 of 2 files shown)


dweitzman (approved):
Seems sufficient to mitigate the risk while keeping mostly sequential IDs which may be helpful when debugging

PR #392, opened 2025-12-25
fix(heartbeat): resolve deadlock between Writer.Close and timer callback

GuptaManan100:
### Description
- Fix deadlock in heartbeat Writer that caused `TestReplTrackerMakePrimary` to timeout after 10 minutes in race detection tests
- Change `isOpen` from `bool` to `atomic.Bool` to allow lock-free reads from the timer callback

### Problem
A deadlock occurred between `Close()` and the timer callback `writeHeartbeat()`:
1. `Close()` acquires `w.mu` and calls `ticks.Stop()`
2. `ticks.Stop()` blocks waiting for the in-flight callback to complete
3. The callback calls `IsOpen()` which tries to acquire `w.mu` → deadlock

## Solution
Make `isOpen` an `atomic.Bool` so that:
- Writes still require holding `mu` (preserving synchronization for Open/Close)
- Reads can proceed without the mutex, allowing the callback to check the closed state and return early

### DIFF
--- a/go/common/fakepgserver/handler.go
+++ b/go/common/fakepgserver/handler.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/multigres/multigres/go/common/sqltypes"
 	"github.com/multigres/multigres/go/pb/query"
 	"github.com/multigres/multigres/go/pgprotocol/server"
 )
@@ -49,7 +50,7 @@ type portal struct {
 }
 
 // HandleQuery handles a simple query protocol message.
-func (h *fakeHandler) HandleQuery(ctx context.Context, conn *server.Conn, queryStr string, callback func(context.Context, *query.QueryResult) error) error {
+func (h *fakeHandler) HandleQuery(ctx context.Context, conn *server.Conn, queryStr string, callback func(context.Context, *sqltypes.Result) error) error {
 	result, err := h.server.handleQuery(queryStr)
 	if err != nil {
 		return err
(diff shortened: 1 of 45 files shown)


dweitzman:
It feels like the synchronization & cancellation strategy used by writer.go is unusual and non-idiomatic and that might be a factor in why there was a deadlock

Curious to hear your thoughts on alternative proposal at https://github.com/multigres/multigres/pull/394 to change how synchronization and cancellation are handled for `writer.go` to hopefully make the whole thing easier to read/understand/trust.

cuongdo:
@sougou suggested that we port the heartbeat writer synchronization strategy from Vitess, which is what I did. Apparently, it was created after some trouble with deadlocks. He may have more context.

dweitzman:
> @sougou suggested that we port the heartbeat writer synchronization strategy from Vitess, which is what I did. Apparently, it was created after some trouble with deadlocks. He may have more context.

Sugu would be the expert, so let's see what he thinks. To me it feels like the Vitess approach is sort of unusual and complex without necessarily handling every theoretical edge case.

For example, in Vitess it seems theoretically possible for the heartbeat writer to try to write a heartbeat after `Writer.Close()` has returned if a goroutine freezes up for 10 seconds. It's very unlikely in practice, but here's how it could happen: `writeHeartbeat()` could get just past the point where `w.appPool.Get()` succeeds, then get delayed for 10 seconds while `Writer.Close()` tries to close the appPool. Closing an app pool only waits up to 10 seconds for borrowed connections to be returned, at which point `Writer.Close()` can return while `writeHeartbeat()` is still holding onto a connection. Then the `writeHearbeat` goroutine could un-freeze and finally write a heartbeat even though `Writer.Close()` previously returned.

In the simplification I'm proposing it would be impossible for delayed heartbeat writes to happen after Close() and it's easy to understand why without reading anything other than `writer.go` (no need to consult timer.go, connection pool code, etc.): both `writeHeartbeat()` and `Writer.Close()` briefly hold a mutex to check or change `isOpen`, and `writeHearbeat` also does `waitGroup.Add(1)` while holding that same lock if the Writer is not yet closed and decides to start a new heartbeat write attempt.

GuptaManan100:
I agree with you @dweitzman, we should do the simpler, and correct approach. My solution was to make the `isOpen` boolean an `atomic.Bool` value, so that it can be checked by `write` without holding the lock, but other than that we agree that we want `write` to check `isOpen` is false.

One of the problems with a mutex lock was that `Close` gets the mutex and then tries to stop the timer. This is a blocking operation. If then a timer tick happens and `writeHearbeat` runs it gets blocked on the `mu.Lock` call, causing a deadlock. That is what this PR is fixing.

PR #394, opened 2025-12-26
refactor(heartbeat): more idiomatic synchronization in writer and reader

dweitzman:
Replace `timer.NewTimer` with `timer.PeriodicRunner`, a context-aware periodic execution utility that feels more like idiomatic Go.

Migrating heartbeat reader/writer to use`timer.PeriodicRunner` makes both shorter and eliminates the need for several mutexes.

### DIFF
--- a/go/tools/timer/periodic_runner.go
+++ b/go/tools/timer/periodic_runner.go
@@ -174,6 +174,13 @@ func (r *PeriodicRunner) execute() {
 		return
 	}
 
+	// Opportunistically stop if parent context is cancelled
+	if r.ctx.Err() != nil {
+		r.mu.Unlock()
+		r.Stop()
+		return
+	}
+
 	// Track this execution so Stop() can wait for it to complete
 	r.wg.Add(1)
 	defer r.wg.Done()
--- a/go/tools/timer/periodic_runner_test.go
+++ b/go/tools/timer/periodic_runner_test.go
@@ -419,3 +419,61 @@ func TestPeriodicRunnerConcurrentStartStop(t *testing.T) {
 
 	runner.Stop()
 }
+
+// TestPeriodicRunnerStopsOnParentContextCancellation verifies that the runner
+// automatically stops when the parent context is cancelled.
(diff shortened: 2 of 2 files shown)


go/multipooler/heartbeat/writer.go
```diff
@@ -64,105 +56,57 @@ func NewWriter(queryService executor.InternalQueryService, logger *slog.Logger,
...
 }
 
 // writeHeartbeat updates the heartbeat row with the current time in nanoseconds.
-func (w *Writer) writeHeartbeat() {
-	if !w.IsOpen() {
-		return
-	}
-	if err := w.write(); err != nil {
-		w.recordError(err)
+func (w *Writer) writeHeartbeat(ctx context.Context) {
```
rafael, line 91: why remove the logic of `w.IsOpen`? That part is not obvious to me.

go/tools/timer/periodic_runner.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	r.ctx = nil
+	r.cancel = nil
+	r.callback = nil
+
+	r.mu.Unlock()
+
+	// Wait for any in-flight callback to complete
+	r.wg.Wait()
```
rafael, line 140: oh this is cool.

go/multipooler/heartbeat/writer.go
dweitzman: The simplified invariant here is that the writer is open if and only if the PeriodicRunner is started. This avoids the need for the writer to try to follow along with whether or not the runner is running and stay in sync.

If you look higher up, the implementation of Writer.isOpen() delegates:

```
func (w *Writer) IsOpen() bool {
	return w.runner.Running()
}
```

dweitzman: If the question is more, "Why don't we need to check isOpen() here?" it's because `writeHeartbeat` is only called by the PeriodicRunner which is responsible for ensuring it can't be running before `periodicRunner.Start()` begins or after `periodicRunner.Stop()` returns

PR #396, opened 2025-12-26
feat(multiorch): detect and repair when a dead primary comes back

deepthi:
# Summary: Stale Primary and Diverged Timeline Repair

##  Goal

  After a failover, when an old primary comes back online with diverged WAL (transactions that weren't replicated), automatically detect and repair the node using pg_rewind.

##  What Was Implemented 

  1. Protocol Changes:
  - Added DRAINED to PoolerType enum in clustermetadata.proto

  2. RPC Client Methods:
  - Added RewindToSource to MultiPoolerClient interface
  - Implemented in gRPC client and FakeClient
  - go/multipooler/manager/rpc_manager.go - RewindToSource RPC that proxies to pgctld

  3. Recovery Actions:
  - handling diverged timeline with pg_rewind in fix_replication.go

  4. Stale Primary Detection & Demotion:
  - ProblemStalePrimary problem code
  - StalePrimaryAnalyzer - Detects when multiple PRIMARYs exist in same shard
  - DemoteStalePrimaryAction - Demotes the primary with lower consensus term
  - Fields added to ReplicationAnalysis: OtherPrimaryInShard, OtherPrimaryTerm, ConsensusTerm

  5. Bug Fixes:
  - fix(demote): Demote should not update term unless the actual demote succeeds
  - fix(analysis): Use strict < for term comparison to prevent double demotion
  - feat(healthcheck): Fetch consensus status for stale primary detection

  6. Tests:
  - Unit tests for StalePrimaryAnalyzer
  - Unit tests for resetting synchronous_standby_names as part of demote
  - E2E test TestDemoteStalePrimary. This also creates a diverged timeline and fixes it.

### DIFF
--- a/go/common/topoclient/locks.go
+++ b/go/common/topoclient/locks.go
@@ -205,7 +205,7 @@ func (l *Lock) unlock(ctx context.Context, ts *store, lt iTopoLock, lockDescript
 	// Detach from the parent timeout, but preserve the trace span.
 	// We need to still release the lock even if the parent context timed out.
 	ctx = context.WithoutCancel(ctx)
-	ctx, cancel := context.WithTimeout(ctx, ts.getRemoteOperationTimeout())
+	ctx, cancel := context.WithTimeout(ctx, ts.GetRemoteOperationTimeout())
 	defer cancel()
 
 	ctx, unlockSpan := telemetry.Tracer().Start(ctx, "TopoServer.Unlock",
--- a/go/common/topoclient/store.go
+++ b/go/common/topoclient/store.go
@@ -210,6 +210,10 @@ type Store interface {
 	// See shard_lock.go for full documentation.
 	TryLockShard(ctx context.Context, shardKey types.ShardKey, action string) (context.Context, func(*error), error)
 
+	// GetRemoteOperationTimeout returns the configured timeout for remote operations.
+	// This should be used for RPCs and database operations that should use a shorter timeout than the parent context.
+	GetRemoteOperationTimeout() time.Duration
+
 	// Resource cleanup
 	io.Closer
 }
@@ -264,8 +268,9 @@ func (ts *store) getLockTimeout() time.Duration {
 	return ts.config.GetLockTimeout()
 }
 
-// getRemoteOperationTimeout returns the configured remote operation timeout.
-func (ts *store) getRemoteOperationTimeout() time.Duration {
+// GetRemoteOperationTimeout returns the configured remote operation timeout.
+// This should be used for RPCs and database operations that should use a shorter timeout than the parent context.
+func (ts *store) GetRemoteOperationTimeout() time.Duration {
 	return ts.config.GetRemoteOperationTimeout()
 }
 
--- a/go/multiorch/coordinator/leader_appointment.go
+++ b/go/multiorch/coordinator/leader_appointment.go
@@ -314,9 +314,13 @@ func (c *Coordinator) recruitNodes(ctx context.Context, cohort []*multiorchdatap
 }
 
 // Propagate implements stage 5 of the consensus protocol:
-// - Promote the candidate to primary
-// - Configure standbys to replicate from the new primary
-// - Configure synchronous replication based on quorum policy
+// - Configure standbys to replicate from the candidate (before promotion)
+// - Promote the candidate to primary (includes sync replication configuration)
+//
+// Standbys are configured BEFORE promotion to avoid a deadlock: the Promote RPC
+// configures sync replication and writes leadership history, which blocks until
+// at least one standby acknowledges. If standbys aren't configured to replicate
+// yet, the write blocks forever.
 func (c *Coordinator) Propagate(ctx context.Context, candidate *multiorchdatapb.PoolerHealthState, standbys []*multiorchdatapb.PoolerHealthState, term int64, quorumRule *clustermetadatapb.QuorumRule, reason string, cohort []*multiorchdatapb.PoolerHealthState, recruited []*multiorchdatapb.PoolerHealthState) error {
 	// Build synchronous replication configuration based on quorum policy
 	syncConfig, err := BuildSyncReplicationConfig(c.logger, quorumRule, standbys, candidate)
@@ -376,6 +380,50 @@ func (c *Coordinator) Propagate(ctx context.Context, candidate *multiorchdatapb.
 		}
 	}
 
+	// Configure standbys to replicate from the candidate BEFORE promoting.
+	// This ensures standbys are ready to connect when sync replication is configured.
+	// Without this, the Promote call can deadlock: it configures sync replication and
+	// tries to write leadership history, but blocks waiting for standby acknowledgment.
+	// The standbys can't acknowledge because they haven't been told to replicate yet.
+	var wg sync.WaitGroup
+	errChan := make(chan error, len(standbys))
+
+	for _, standby := range standbys {
+		wg.Add(1)
+		go func(s *multiorchdatapb.PoolerHealthState) {
+			defer wg.Done()
+			c.logger.InfoContext(ctx, "Configuring standby replication before promotion",
+				"standby", s.MultiPooler.Id.Name,
+				"primary", candidate.MultiPooler.Id.Name)
+
+			setPrimaryReq := &multipoolermanagerdatapb.SetPrimaryConnInfoRequest{
+				Primary:               candidate.MultiPooler,
+				CurrentTerm:           term,
+				StopReplicationBefore: false,
+				StartReplicationAfter: true, // Start streaming immediately so sync replication can proceed
+				Force:                 false,
+			}
+			if _, err := c.rpcClient.SetPrimaryConnInfo(ctx, s.MultiPooler, setPrimaryReq); err != nil {
+				errChan <- mterrors.Wrapf(err, "failed to configure standby %s", s.MultiPooler.Id.Name)
+				return
+			}
+
+			c.logger.InfoContext(ctx, "Standby configured successfully", "standby", s.MultiPooler.Id.Name)
+		}(standby)
+	}
+
+	wg.Wait()
+	close(errChan)
+
+	// Check for errors - log but don't fail, standbys can be fixed later
+	var standbyErrs []error
+	for err := range errChan {
+		standbyErrs = append(standbyErrs, err)
+	}
+	for _, err := range standbyErrs {
+		c.logger.WarnContext(ctx, "Standby configuration failed", "error", err)
+	}
+
 	// Get coordinator ID as a string
 	coordinatorIDStr := topoclient.ClusterIDString(c.GetCoordinatorID())
 
@@ -395,52 +443,6 @@ func (c *Coordinator) Propagate(ctx context.Context, candidate *multiorchdatapb.
 	}
 
 	c.logger.InfoContext(ctx, "Candidate promoted successfully", "node", candidate.MultiPooler.Id.Name)
-
-	// Configure standbys to replicate from the new primary
-	var wg sync.WaitGroup
-	errChan := make(chan error, len(standbys))
-
-	for _, standby := range standbys {
-		wg.Add(1)
-		go func(s *multiorchdatapb.PoolerHealthState) {
-			defer wg.Done()
-			c.logger.InfoContext(ctx, "Configuring standby replication",
-				"standby", s.MultiPooler.Id.Name,
-				"primary", candidate.MultiPooler.Id.Name)
-
-			setPrimaryReq := &multipoolermanagerdatapb.SetPrimaryConnInfoRequest{
-				Primary:               candidate.MultiPooler,
-				CurrentTerm:           term,
-				StopReplicationBefore: false,
-				StartReplicationAfter: false,
-				Force:                 false,
-			}
-			if _, err := c.rpcClient.SetPrimaryConnInfo(ctx, s.MultiPooler, setPrimaryReq); err != nil {
-				errChan <- mterrors.Wrapf(err, "failed to configure standby %s", s.MultiPooler.Id.Name)
-				return
-			}
-
-			c.logger.InfoContext(ctx, "Standby configured successfully", "standby", s.MultiPooler.Id.Name)
-		}(standby)
-	}
-
-	wg.Wait()
-	close(errChan)
-
-	// Check for errors
-	var errs []error
-	for err := range errChan {
-		errs = append(errs, err)
-	}
-
-	if len(errs) > 0 {
-		// Log all errors but continue - partial success is acceptable
-		for _, err := range errs {
-			c.logger.WarnContext(ctx, "Standby configuration failed", "error", err)
-		}
-		// Don't fail the whole operation if some standbys failed
-		// The primary is established, standbys can be fixed later
-	}
 
 	return nil
 }
--- a/go/multipooler/manager/pg_multischema.go
+++ b/go/multipooler/manager/pg_multischema.go
@@ -327,6 +327,9 @@ func (pm *MultiPoolerManager) insertDurabilityPolicy(ctx context.Context, policy
 }
 
 // insertLeadershipHistory inserts a leadership history record into the leadership_history table.
+// This operation uses the remote-operation-timeout and will fail if it cannot complete within
+// that time. A timeout typically indicates that synchronous replication is not functioning
+// (no standbys are connected to acknowledge the write).
 func (pm *MultiPoolerManager) insertLeadershipHistory(ctx context.Context, termNumber int64, leaderID, coordinatorID, walPosition, reason string, cohortMembers, acceptedMembers []string) error {
 	pm.logger.InfoContext(ctx, "Inserting leadership history",
 		"term", termNumber,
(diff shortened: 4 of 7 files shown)


go/common/pgutil/timeline.go
```diff
@@ -0,0 +1,69 @@
...
+	ParentTimelineID int64
+	// ForkLSN is the LSN where the fork occurred.
+	ForkLSN string
+}
+
+// ParseTimelineHistory parses the content of a PostgreSQL timeline history file.
+// The format is one entry per line: <parent_tli>\t<switch_lsn>\t<reason>
+// Lines starting with # are comments and are ignored.
+// The timelineID parameter is the timeline whose history file this is.
+func ParseTimelineHistory(timelineID int64, content []byte) []TimelineForkPoint {
```
rafael: I don't  see calls to this function

go/multiorch/recovery/actions/demote_stale_primary.go
```diff
@@ -0,0 +1,225 @@
...
+func (a *DemoteStalePrimaryAction) RequiresHealthyPrimary() bool {
+	// We're demoting a primary, so we can't require a healthy primary
+	return false
+}
+
+// Execute demotes the stale primary using the Demote RPC with the correct primary's term.
+// This is safer than BeginTerm because:
+// 1. We use the correct primary's term (not a new term), avoiding term inconsistency
+// 2. The stale primary accepts term >= its current term and demotes
+// 3. Both primaries end up with the same term (no term inconsistency)
```
rafael, line 94: nice. This makes sense to me

go/multiorch/recovery/actions/demote_stale_primary.go
```diff
@@ -0,0 +1,225 @@
...
+		return mterrors.Wrap(err, "Demote RPC failed")
+	}
+
+	a.logger.InfoContext(ctx, "stale primary demoted successfully",
+		"stale_primary", poolerIDStr,
+		"lsn_position", demoteResp.LsnPosition)
+
+	// Update topology to reflect the demotion
+	// The pooler's type in topology should be changed to REPLICA
+	_, err = a.topoStore.UpdateMultiPoolerFields(ctx, problem.PoolerID, func(mp *clustermetadatapb.MultiPooler) error {
```
rafael: This happens already in demote. So we should be able to remove this

go/multiorch/recovery/actions/demote_stale_primary.go
```diff
@@ -0,0 +1,225 @@
...
+	a.logger.InfoContext(ctx, "demote stale primary action completed",
+		"shard_key", problem.ShardKey.String(),
+		"demoted_primary", poolerIDStr)
+
+	return nil
+}
+
+// findCorrectPrimaryTerm finds the correct primary in the shard and returns its term.
+// The correct primary is the one with the higher consensus term.
+func (a *DemoteStalePrimaryAction) findCorrectPrimaryTerm(ctx context.Context, shardKey commontypes.ShardKey, stalePrimaryIDStr string) (int64, error) {
```
rafael: I think we need to do a pass and a cleanup of how we the current term across all actions. But we can come back to this

proto/clustermetadata.proto
```diff
@@ -177,15 +177,18 @@ message KeyRange {
 
 // PoolerType represents the type of a given MultiPooler.
 enum PoolerType {
-  // UNKNOWN this is not a valid type.
```
rafael, line 182: Why this change?  I think we should either keep with the convention that UNKNOWN is the zero for uninitialized  values.  

Also it seems we are not being consistent. Some enums have:
* UNKOWN 
* Others [TypeName]Unknown

We should agree on one way and keep it consistent across all enum

go/multiorch/recovery/actions/demote_stale_primary.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	a.logger.InfoContext(ctx, "stale primary demoted successfully",
+		"stale_primary", poolerIDStr,
+		"lsn_position", demoteResp.LsnPosition)
+
+	// Update topology to reflect the demotion
+	// The pooler's type in topology should be changed to REPLICA
+	_, err = a.topoStore.UpdateMultiPoolerFields(ctx, problem.PoolerID, func(mp *clustermetadatapb.MultiPooler) error {
+		mp.Type = clustermetadatapb.PoolerType_REPLICA
```
rafael: Something that I'm thinking. In this context, I think the default behavior for the Demote action is to make sure the pooler is left in a drained state. It is possible that we need to fix the timeline.

It seems that it could pro-actively check if it also needs to rewind. Instead of being a separate action. 

So the flow I'm thinking:
- Demote should allow to leave a pooler drained.
- This action checks if it needs rewind. If it does, it perform that action.
- After that it can update the status to be serving again.

proto/clustermetadata.proto
```diff
@@ -177,15 +177,18 @@ message KeyRange {
...
+  POOLER_TYPE_UNSPECIFIED = 0;
 
   // PRIMARY is the primary server for the shard. Only PRIMARY allows DMLs.
   PRIMARY = 1;
 
   // REPLICA replicates from leader. It is used to read only traffic
   REPLICA = 2;
+
+  // POOLER_TYPE_DRAINED is used for poolers that are temporarily removed from serving traffic
+  POOLER_TYPE_DRAINED = 3;
```
rafael: I don't think drained should be a type.  This is already a state in the serving status.

go/multiorch/recovery/actions/demote_stale_primary.go
```diff
@@ -0,0 +1,225 @@
...
+	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
+	multiorchdatapb "github.com/multigres/multigres/go/pb/multiorchdata"
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+)
+
+// Compile-time assertion that DemoteStalePrimaryAction implements types.RecoveryAction.
+var _ types.RecoveryAction = (*DemoteStalePrimaryAction)(nil)
+
+// DefaultDrainTimeout is the default time to wait for connections to drain during demotion.
+const DefaultDrainTimeout = 30 * time.Second
```
rafael: We are not using this right now?

go/multiorch/recovery/actions/demote_stale_primary.go
rafael: ctx is not used

go/multiorch/recovery/actions/fix_replication.go
```diff
@@ -436,6 +445,98 @@ func (a *FixReplicationAction) Priority() types.Priority {
 	return types.PriorityHigh
 }
 
+// fixDivergedTimeline repairs a replica that has a diverged timeline using pg_rewind.
+//
+// The flow is:
+// 1. Run pg_rewind --dry-run via multipooler to check if repair is feasible
```
rafael: why doing in this check? In our context it seems we want to do it and if the actual run fails, the outcome is the same for us.

proto/multipoolermanagerdata.proto
```diff
@@ -710,3 +710,26 @@ message CreateDurabilityPolicyRequest {
...
+// =============================================================================
+// PgRewind APIs
+// =============================================================================
+
+// RewindToSourceRequest requests pg_rewind to synchronize with a source server.
+// This stops PostgreSQL, runs pg_rewind, and restarts PostgreSQL.
+message RewindToSourceRequest {
+  // Source server connection string for pg_rewind (--source-server)
+  // Format: "host=<hostname> port=<port> user=<user> dbname=<dbname>"
+  string source_server = 1;
```
rafael: should this just be a multipooler id? Then the source is derived from that ?

go/multiorch/recovery/actions/fix_replication.go
```diff
@@ -436,6 +445,98 @@ func (a *FixReplicationAction) Priority() types.Priority {
...
+	// 2. Run actual pg_rewind (this stops PG, rewinds, and starts PG)
+	a.logger.InfoContext(ctx, "running pg_rewind",
+		"replica", replica.MultiPooler.Id.Name,
+		"source", sourceConnStr)
+
+	rewindReq := &multipoolermanagerdatapb.RewindToSourceRequest{
+		SourceServer: sourceConnStr,
+		DryRun:       false,
+	}
+	rewindResp, err := a.rpcClient.RewindToSource(ctx, replica.MultiPooler, rewindReq)
```
rafael: This should probably have a big timeout

go/multiorch/recovery/analysis/generator.go
```diff
@@ -369,6 +380,11 @@ func (g *AnalysisGenerator) populatePrimaryInfo(
 	// When the primary pooler is down but Postgres is still running, replicas remain connected
 	// and we should NOT trigger failover. Instead, the operator should restart the pooler process.
 	analysis.ReplicasConnectedToPrimary = g.allReplicasConnectedToPrimary(primary, poolers)
+
+	// Check for timeline divergence between replica and primary
+	if divergedProblem := g.checkTimelineDivergence(primary, replica); divergedProblem != nil {
```
rafael: Note to talk about this

go/multiorch/recovery/analysis/generator.go
```diff
@@ -198,6 +204,9 @@ func (g *AnalysisGenerator) generateAnalysisForPooler(
 
 		// Aggregate replica stats
 		g.aggregateReplicaStats(pooler, analysis, shardKey)
+
+		// Check for stale primary: look for other PRIMARYs in the same shard
+		g.detectOtherPrimary(analysis, shardKey, pooler)
```
rafael, line 208: note to talk about this.

go/multiorch/recovery/analysis/generator.go
```diff
@@ -480,3 +496,85 @@ func (g *AnalysisGenerator) isReplicaConnectedToPrimary(
...
+
+	if primary.ConsensusStatus != nil {
+		primaryTL = primary.ConsensusStatus.TimelineInfo
+	}
+	if replica.ConsensusStatus != nil {
+		replicaTL = replica.ConsensusStatus.TimelineInfo
+	}
+
+	if primaryTL == nil || replicaTL == nil {
+		return nil // Can't determine divergence without timeline info
```
rafael: I think we should be logging warning / error here.

go/multiorch/recovery/analysis/generator.go
```diff
@@ -480,3 +496,85 @@ func (g *AnalysisGenerator) isReplicaConnectedToPrimary(
...
+// A timeline is diverged if the replica is on an older timeline and has
+// WAL past the point where the primary's timeline forked.
+func (g *AnalysisGenerator) checkTimelineDivergence(
+	primary *multiorchdatapb.PoolerHealthState,
+	replica *multiorchdatapb.PoolerHealthState,
+) *multiorchdatapb.ShardProblem {
+	// Get timeline info from both nodes
+	var primaryTL *consensusdatapb.TimelineInfo
+	var replicaTL *consensusdatapb.TimelineInfo
+
```
rafael: Should we have a sanity check that primary and replica are not the same multipooler?

go/multiorch/recovery/analysis/generator.go
```diff
@@ -480,3 +496,85 @@ func (g *AnalysisGenerator) isReplicaConnectedToPrimary(
...
+	if replica.ConsensusStatus != nil {
+		replicaTL = replica.ConsensusStatus.TimelineInfo
+	}
+
+	if primaryTL == nil || replicaTL == nil {
+		return nil // Can't determine divergence without timeline info
+	}
+
+	// If replica is on same or newer timeline, not diverged
+	if replicaTL.TimelineId >= primaryTL.TimelineId {
```
rafael: how can the replica be in a newer timeline than the primary? That also seems like an error

go/multiorch/recovery/analysis/stale_primary_analyzer.go
```diff
@@ -0,0 +1,91 @@
...
+	if poolerAnalysis.OtherPrimaryInShard == nil {
+		return nil, nil
+	}
+
+	// Multiple primaries detected! Determine which one is stale.
+	// The one with the lower consensus term is stale and should be demoted.
+	// If terms are equal, neither node demotes - this is an unusual state that
+	// should be resolved through normal failover or manual intervention.
+	// Using strict < prevents both nodes from demoting simultaneously.
+	isThisNodeStale := poolerAnalysis.ConsensusTerm < poolerAnalysis.OtherPrimaryTerm
```
rafael, line 66: note to talk about this

go/multipooler/manager/pg_replication.go
```diff
@@ -690,6 +705,40 @@ func (pm *MultiPoolerManager) getSynchronousReplicationConfig(ctx context.Contex
...
+	// Reload configuration to apply changes immediately
+	if err := pm.exec(execCtx, "SELECT pg_reload_conf()"); err != nil {
+		pm.logger.WarnContext(ctx, "Failed to reload configuration for demotion", "error", err)
+		return mterrors.Wrap(err, "failed to reload configuration for demotion")
+	}
+
+	pm.logger.InfoContext(ctx, "Successfully cleared synchronous replication for demotion")
+	return nil
+}
+
```
rafael, line 741: Do you think is worth adding unit tests for this one?

go/common/pgutil/timeline.go
deepthi: I'm going to delete this because we are going to use `pg_rewind dry-run` to do the diverged timeline detection instead.

proto/clustermetadata.proto
deepthi: 🤦‍♀️ I'll fix them up to be consistent. And yes, it should be UNKNOWN.

go/multiorch/recovery/actions/demote_stale_primary.go
deepthi: We have to wait until we have serving state transitions implemented in the Pooler before we can invoke them.

proto/clustermetadata.proto
deepthi: For now I have reserved 4 in serving status and commented out the DRAINED enum value. This has to wait for state transitions too.

go/multiorch/recovery/analysis/generator.go
deepthi: Removed this analysis

go/multiorch/recovery/analysis/generator.go
deepthi: no longer relevant.

go/multiorch/recovery/analysis/generator.go
deepthi: No longer relevant.

go/multiorch/recovery/analysis/generator.go
deepthi: No longer relevant.

go/multiorch/recovery/actions/demote_stale_primary.go
deepthi: Removing it.

go/multiorch/recovery/actions/fix_replication.go
deepthi: We'll have to tune this. Typical scenarios should not involve more than a handful of transactions.

go/multiorch/recovery/actions/fix_replication.go
deepthi: yes, it's debatable whether it is worth doing the dry-run, when we are going to run pg_rewind anyway. Running pg_rewind and failing is the same as not running it, we would have to throw away this replica and bootstrap a new one.

go/multiorch/recovery/analysis/generator.go
```diff
@@ -480,3 +489,46 @@ func (g *AnalysisGenerator) isReplicaConnectedToPrimary(
...
+		poolerType := pooler.PoolerType
+		if poolerType == clustermetadatapb.PoolerType_UNKNOWN && pooler.MultiPooler != nil {
+			poolerType = pooler.MultiPooler.Type
+		}
+
+		if poolerType == clustermetadatapb.PoolerType_PRIMARY {
+			// Found another PRIMARY - one of them is stale!
+			analysis.OtherPrimaryInShard = pooler.MultiPooler.Id
+			if pooler.ConsensusStatus != nil {
+				analysis.OtherPrimaryTerm = pooler.ConsensusStatus.CurrentTerm
```
rafael, line 529: not a blocker. I think we discussed making this a list. As it is theoretically possible that there could be even more than one.

proto/clustermetadata.proto
```diff
@@ -188,6 +188,9 @@ enum PoolerType {
 
   // REPLICA replicates from leader. It is used to read only traffic
   REPLICA = 2;
+
+  // DRAINED is used for poolers that are temporarily removed from serving traffic
+  DRAINED = 3;
```
rafael, line 193: I'm still more on the camp that drained should be in the serving status. But we can come back to this when we rehaul the whole pooler state management.

proto/clustermetadata.proto
```diff
@@ -213,7 +216,10 @@ enum PoolerServingStatus {
   // the tools processes may need to run queries), but it's not used
   // to route queries from Multigres users. In this state,
   // this multipooler is dedicated to the process that uses it.
-  DRAINED = 4;
+  // DRAINED = 4;
+  // (TODO): This enum value conflicts with DRAINED in PoolerType.
+  // We need to reconcile them once we have serving state transitions in Pooler.
+  reserved 4;
```
rafael, line 222: do we need to mark it as reserved? I think we can still just remove it

go/multiorch/recovery/analysis/generator.go
deepthi: Good point.

deepthi:
> There is that extra issue with discussed offline

The extra issue is that setting up replication as part of stale primary recovery was failing. Debugging this led to the discovery of two bugs which have now been fixed.

I tested both PrimaryIsDead recovery and StalePrimary recovery manually on a local cluster and verified that everything is configured properly with these fixes. Also added e2e tests which would have caught this issue.

PR #397, opened 2025-12-27
feat(multiorch): store leadership history in each shard

deepthi:
# Leadership History Feature

##  Goal

  Implement an audit trail for leader elections in multigres, tracking who became leader, when, why, and which coordinator orchestrated it.

##  What Was Implemented

  1. Schema - multigres.leadership_history table
  - Append-only audit table created during bootstrap/initialization
  - Fields: term_number, leader_id, coordinator_id, wal_position, reason, cohort_members, accepted_members, created_at
  - Index on term_number DESC for efficient queries

  2. Protobuf Changes - Extended PromoteRequest
  - Added fields: reason, coordinator_id, cohort_members, accepted_members
  - Also added coordinator_id to InitializeEmptyPrimaryRequest for bootstrap

  3. Multipooler Changes
  - Promote() writes leadership history after successful promotion
  - InitializeEmptyPrimary() writes leadership history during bootstrap
  - New insertLeadershipHistory() function handles the SQL insert

  4. Multiorch Coordinator Changes
  - AppointLeader() populates the new PromoteRequest fields
  - BootstrapShard() passes coordinator_id to InitializeEmptyPrimaryRequest
  - Tracks cohort members and accepted members during BeginTerm consensus

  5. --service-id Flag for Multiorch
  - Added configurable service ID (previously random 8-char string)
  - Makes coordinator_id in audit logs deterministic and human-readable

  6. E2E Tests
  - bootstrap_test.go - Verifies leadership_history written during shard bootstrap
  - dead_primary_recovery_test.go - Verifies leadership_history written during failover

### DIFF
--- a/go/multipooler/manager/rpc_initialization.go
+++ b/go/multipooler/manager/rpc_initialization.go
@@ -275,6 +275,11 @@ func (pm *MultiPoolerManager) InitializeAsStandby(ctx context.Context, req *mult
 		}
 	}
 
+	// 6. Set pooler type to REPLICA
+	if err := pm.changeTypeLocked(ctx, clustermetadatapb.PoolerType_REPLICA); err != nil {
+		return nil, mterrors.Wrap(err, "failed to set pooler type")
+	}
+
 	// Mark as initialized after successful standby initialization.
 	// This sets the cached boolean and writes the marker file.
 	if err := pm.setInitialized(); err != nil {
--- a/go/multiorch/recovery/actions/bootstrap_shard.go
+++ b/go/multiorch/recovery/actions/bootstrap_shard.go
@@ -355,14 +355,8 @@ func (a *BootstrapShardAction) initializeSingleStandby(ctx context.Context, node
 		return fmt.Errorf("initialization failed: %s", resp.ErrorMessage)
 	}
 
-	// Set pooler type to REPLICA after InitializeAsStandby
-	changeTypeReq := &multipoolermanagerdatapb.ChangeTypeRequest{
-		PoolerType: clustermetadatapb.PoolerType_REPLICA,
-	}
-	_, err = a.rpcClient.ChangeType(ctx, node.MultiPooler, changeTypeReq)
-	if err != nil {
-		return fmt.Errorf("failed to set pooler type: %w", err)
-	}
+	// Note: InitializeAsStandby now sets the pooler type to REPLICA internally,
+	// matching the behavior of InitializeEmptyPrimary which sets type to PRIMARY.
 
 	return nil
 }


go/multiorch/coordinator/leader_appointment.go
```diff
@@ -360,11 +360,36 @@ func (c *Coordinator) Propagate(ctx context.Context, candidate *multiorchdatapb.
...
+	for _, pooler := range recruited {
+		if pooler.MultiPooler != nil && pooler.MultiPooler.Id != nil {
+			acceptedMembers = append(acceptedMembers, pooler.MultiPooler.Id.Name)
+		}
+	}
+
+	// Get coordinator ID as a string
+	coordinatorIDStr := ""
+	if c.coordinatorID != nil {
+		coordinatorIDStr = c.coordinatorID.Name
```
rafael: Three nits:
* I think we can use: `GetCoordinatorID()` here. 
* I wonder if we should add a helper to generate a string of a cluster id. We could add it to `id.go` in the topo client. 
* I wonder if we should validate the existence of coordinatorId in the constructor of the coordinator. So methods don't have to be adding guardrails against this. It seems that this should be an invariant we can guarantee at the moment of creation.

go/multiorch/coordinator/coordinator.go
```diff
@@ -94,7 +94,14 @@ func (c *Coordinator) AppointLeader(ctx context.Context, shardID string, cohort
 
 	// Stage 6: Propagate (setup replication within shard)
 	c.logger.InfoContext(ctx, "Stage 6: Propagating replication", "shard", shardID)
-	if err := c.Propagate(ctx, candidate, standbys, term, quorumRule); err != nil {
+
+	// Reconstruct the recruited list (nodes that accepted the term)
```
rafael: under which scenarios would the cohort be different than recruited? Maybe good to add a comment around that.

go/multiorch/recovery/analysis/action_factory.go
```diff
@@ -53,7 +53,13 @@ func NewRecoveryActionFactory(
 
 // NewBootstrapShardAction creates a bootstrap shard action.
 func (f *RecoveryActionFactory) NewBootstrapShardAction() types.RecoveryAction {
-	return actions.NewBootstrapShardAction(f.rpcClient, f.poolerStore, f.topoStore, f.logger)
+	coordinatorID := ""
+	if f.coordinator != nil {
```
rafael: nit: similar to the one above.

Passing coordinatorID string to NewBootstrapShardAction feels odd to me.  Wondering passing the *coordinator.Coordinator instead, consistent with how AppointLeaderAction works.

Also, NewRecoveryActionFactory could validate that coordinator != nil upfront (other actions like AppointLeaderAction already assume this implicitly), which would eliminate the nil checks when extracting the ID.

go/multipooler/manager/rpc_initialization.go
```diff
@@ -138,6 +138,27 @@ func (pm *MultiPoolerManager) InitializeEmptyPrimary(ctx context.Context, req *m
...
+	}
+
+	// Write leadership history record for bootstrap
+	leaderID := generateApplicationName(pm.serviceID)
+	coordinatorID := req.CoordinatorId
+	reason := "ShardNeedsBootstrap"
+	cohortMembers := []string{leaderID} // Only the initial primary during bootstrap
+	acceptedMembers := []string{leaderID}
+
+	if err := pm.insertLeadershipHistory(ctx, req.ConsensusTerm, leaderID, coordinatorID, finalLSN, reason, cohortMembers, acceptedMembers); err != nil {
```
rafael, line 155: we could add a unit test, to validate that error here doesn't fail the whole action

go/multipooler/manager/rpc_manager.go
```diff
@@ -996,15 +996,24 @@ func (pm *MultiPoolerManager) Promote(ctx context.Context, consensusTerm int64,
 		return nil, err
 	}
 
-	// TODO: Populate consensus metadata tables.
```
rafael, line 999: 🥳

go/multipooler/manager/rpc_manager.go
```diff
@@ -996,15 +996,24 @@ func (pm *MultiPoolerManager) Promote(ctx context.Context, consensusTerm int64,
...
 	// Get final LSN position
 	finalLSN, err := pm.getPrimaryLSN(ctx)
 	if err != nil {
 		pm.logger.ErrorContext(ctx, "Failed to get final LSN", "error", err)
 		return nil, err
 	}
 
+	// Write leadership history record
+	leaderID := generateApplicationName(pm.serviceID)
+	if reason != "" && coordinatorID != "" {
```
rafael: I wonder if we should log even if these are nil. Maybe populate it with unknown and it's something to debug. But better to keep the paper trail.

go/test/endtoend/multiorch/bootstrap_test.go
```diff
@@ -90,9 +91,13 @@ func TestBootstrapInitialization(t *testing.T) {
 	require.NotEmpty(t, primaryName, "Expected multiorch to bootstrap shard automatically")
 	setup.PrimaryName = primaryName
 
-	// Wait for at least 1 standby to initialize
+	// Wait for both standbys to initialize (needed for sync replication with ANY_2)
 	t.Log("Waiting for standbys to complete initialization...")
-	waitForStandbysInitialized(t, setup, 1, 60*time.Second)
+	waitForStandbysInitialized(t, setup, 2, 60*time.Second)
```
rafael: * I think these can be shorter timeouts. 
* Wonder if we can fold the logic of `waitForSyncReplicationConfigured` inside `waitForStandbysInitialized`. In this context, initialization, also includes wiring up replication

deepthi:
All feedback has been addressed.
This PR exposed a flaky test - the newly added CLI commands in cluster_test.go. There were two reasons for the flakiness
- pooler type has not yet been changed to REPLICA before we run the CLI commands. This has been fixed by waiting for the right condition.
- pooler type was being set to REPLICA by bootstrap as a separate RPC call instead of in InitializeAsStandby. This is not the same as what we are doing for PRIMARY, where everything happens in one RPC. This has also been fixed.

deepthi:
Test failures are from flaky tests. I'm going to merge this PR and open a new one to try and fix the failing tests.

PR #401, opened 2025-12-29
fix(test): retry cluster setup on port conflicts

dweitzman:
When running `go test ./go/test/endtoend/...`, multiple packages execute in parallel as separate processes, each with their own port allocation cache. This can cause TOCTOU races where two packages allocate the same port nearly simultaneously, and one fails with "port already in use".

Add a retry loop in setupTestCluster() that detects port conflict errors, cleans up any partially started processes, reallocates ports, and retries up to 3 times before failing.

Here's an example of a flake because of a port already in use, which feels like it's probably happening due to a race with other test processes on the same container: https://github.com/multigres/multigres/actions/runs/20576665319/attempts/1?pr=397


rafael:
This makes sense. Stamping but requires updating as we moved this to a different package. Probably a similar fix is required for shardsetup? I'm not sure

PR #413, opened 2025-12-30
feat(telemetry): add grpccommon.NewClient() for consistent gRPC client instrumentation

dweitzman:
- Add grpccommon.NewClient() wrapper that automatically includes OTel stats handler for all gRPC client connections
- Add ruleguard rule to enforce using the wrapper instead of direct grpc.NewClient() calls (excludes test files and grpccommon itself)
- Update all production gRPC client creation sites to use the new wrapper
- Avoid duplicate OTel handler in LocalClientDialOptions()

This ensures all gRPC client metrics are automatically instrumented, enabling visibility into outbound RPC calls in Prometheus/Grafana.


deepthi (approved):
LGTM. I checked the demo PR to see if any of the MultiAdmin changes there need this, and it didn't look like it.

PR #414, opened 2025-12-30
feat(telemetry): add OpenTelemetry spans for recovery loop operations

dweitzman:
Add distributed tracing spans to the recovery loop to track:
- recovery/cycle: top-level span for each recovery cycle with problem count
- recovery/attempt: span for each recovery attempt with shard context, problem code, pooler ID, action name, and priority

This enables tracing recovery operations and correlating recovery actions with their outcomes.

recovery/cycle seems like the sort of span that's mainly interesting in local demos, or in prod with low sampling rates just have a sense of how long the loop typically takes

recovery/attempt seems like a more useful span that could more justifiably have a higher sampling rate in prod, since it should be for actions which are infrequent but usually interesting

### DIFF
--- a/go/multiorch/recovery/recovery_loop_test.go
+++ b/go/multiorch/recovery/recovery_loop_test.go
@@ -1574,45 +1574,42 @@ func TestRecoveryLoop_TracingSpans(t *testing.T) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


deepthi (approved):
Nice work!

PR #416, opened 2025-12-30
feat(multiadmin): http multiadmin grpc gateway (take 2)

rafael:


proto/google/api/http.proto
```diff
@@ -0,0 +1,370 @@
+// Copyright 2025 Google LLC
```
rafael, line 1: This is from the instructions in:  https://github.com/grpc-ecosystem/grpc-gateway?tab=readme-ov-file#2-with-custom-annotations

tools/setup_build_tools.sh
```diff
@@ -415,6 +415,8 @@ install_pgbackrest_macos() {
 install_go_plugins() {
   # Reinstall protoc-gen-go and protoc-gen-go-grpc
   GOBIN=$MTROOT/bin go install google.golang.org/protobuf/cmd/protoc-gen-go google.golang.org/grpc/cmd/protoc-gen-go-grpc
```
rafael, line 417: should we pin the protoc-gen-go ?

go/services/multiadmin/init.go
```diff
@@ -98,11 +102,25 @@ func (ma *MultiAdmin) Init() error {
...
+
+			// Set up grpc-gateway for REST API
+			gwmux := runtime.NewServeMux(
+				runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{
+					MarshalOptions:   protojson.MarshalOptions{EmitUnpopulated: true, UseProtoNames: true},
+					UnmarshalOptions: protojson.UnmarshalOptions{DiscardUnknown: true},
+				}),
+			)
+			//nolint:gocritic // grpc-gateway handler registration runs at startup before request context is available
+			if err := multiadminpb.RegisterMultiAdminServiceHandlerServer(context.Background(), gwmux, ma.adminServer); err != nil {
```
dweitzman, line 118: FYI, `cobra.Command` creates a context so one exists, it just may not be propagated here through function calls

tools/setup_build_tools.sh
dweitzman: I think the versions come from go.mod?

dweitzman (approved):

PR #441, opened 2026-01-03
feat(logging): add logging to make it easier to diagnose flaky tests

deepthi:
- Add progress logging to postgres startup wait
- tail logs if startup times out
- detect crash during startup
- enhance error logging while restarting postgres as standby

Bonus: there is a test failure on main caused by merging an out-of-date PR. That is fixed in this PR as well.


go/cmd/pgctld/command/start.go
```diff
@@ -271,28 +271,96 @@ func startPostgreSQLWithConfig(logger *slog.Logger, config *pgctld.PostgresCtlCo
...
 	// Wait for PostgreSQL to be ready using pg_isready
-	return waitForPostgreSQLWithConfig(config)
+	return waitForPostgreSQLWithConfig(logger, config)
 }
 
-func waitForPostgreSQLWithConfig(config *pgctld.PostgresCtlConfig) error {
-	// Try to connect using pg_isready
+// readLogTail reads the last N lines from the PostgreSQL log file for diagnostics
+func readLogTail(logPath string, lines int) string {
+	content, err := os.ReadFile(logPath)
```
dweitzman, line 279: Not sure how well this will scale in practice if log files get big, but that feels like a future problem

PR #443, opened 2026-01-03
fix(multiorch): fix nil coordinator panic in TestRecoveryLoop_TracingSpans

dweitzman:
Due to a merge conflict the test was passing nil as the coordinator parameter to NewEngine(), which caused a panic in NewRecoveryActionFactory. Updated to use newTestCoordinator() like all other tests in the package.

Also syncs go.mod/go.sum after running make build-all && go mod tidy.


dweitzman:
This was already fixed in https://github.com/multigres/multigres/pull/441

PR #444, opened 2026-01-04
feat(pghba): Override default configs with command line

sougou:
This change allows you to override the default postgres config and pg hba configs with command line flags.
If the files cannot be loaded at start time, pgctld errors out. Upon successful load it replaces the default templates with the contents of the provided files.

I've also reverted the pghba unsafe change. For the demo, I'm now providing the override as a config map.

In the future, we should write a watcher thread that can regenerate the files and restart posttgres when the config maps get updated.

### DIFF
--- a/go/cmd/pgctld/command/root.go
+++ b/go/cmd/pgctld/command/root.go
@@ -35,17 +35,18 @@ import (
 
 // PgCtlCommand holds the configuration for pgctld commands
 type PgCtlCommand struct {
-	reg               *viperutil.Registry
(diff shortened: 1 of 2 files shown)


deepthi:
> In the future, we should write a watcher thread that can regenerate the files and restart posttgres when the config maps get updated.

This is an interesting requirement. Viper will reload config when the flag value changes, but here the flag value is the same (file name), it's the contents of the file that change!

deepthi (approved):

PR #445, opened 2026-01-04
feat(pooler): default pooler type to REPLICA on startup, and change bootstrap to work with this

deepthi:
## Use case
We might have to add new replicas to an already running multigres deployment. Starting them as UNKNOWN pooler type doesn't work.
We have to support both use cases:
- Start a 3-node cluster which are all REPLICA, multiorch should correctly bootstrap them.
- Add a new node which is REPLICA, it should restore from backup and then multiorch should setup replication

## Solution
  - Default pooler type to REPLICA, allowing new poolers to register as REPLICA type instead of UNKNOWN
  - Change bootstrap detection to use HasDataDirectory (presence of PG_VERSION file) instead of checking PoolerType != REPLICA
  - Change auto-restore to skip when there is no backup yet OR data directory already exists

  This enables poolers to start as REPLICA type while still correctly detecting fresh poolers that need bootstrap. The presence of a data directory is a more reliable signal for "was ever initialized" than pooler type.

## Test plan
  - Unit tests added for new bootstrap detection behavior
  - Existing unit tests updated to explicitly set `HasDataDirectory`
  - All endtoend tests should still pass

### DIFF
--- a/go/multipooler/init.go
+++ b/go/multipooler/init.go
@@ -56,7 +56,6 @@ type MultiPooler struct {
 	pgBackRestKeyFile  viperutil.Value[string]
 	pgBackRestCAFile   viperutil.Value[string]
 	pgBackRestPort     viperutil.Value[int]
-	initialPoolerType  viperutil.Value[string]
 	// GrpcServer is the grpc server
 	grpcServer *servenv.GrpcServer
 	// Senv is the serving environment
@@ -157,11 +156,6 @@ func NewMultiPooler(telemetry *telemetry.Telemetry) *MultiPooler {
 			FlagName: "pgbackrest-port",
 			Dynamic:  false,
 		}),
-		initialPoolerType: viperutil.Configure(reg, "initial-pooler-type", viperutil.Options[string]{
-			Default:  "REPLICA",
-			FlagName: "initial-pooler-type",
-			Dynamic:  false,
-		}),
 		grpcServer:     servenv.NewGrpcServer(reg),
 		senv:           servenv.NewServEnvWithConfig(reg, servenv.NewLogger(reg, telemetry), viperutil.NewViperConfig(reg), telemetry),
 		telemetry:      telemetry,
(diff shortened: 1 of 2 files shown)


go/multipooler/init.go
```diff
@@ -277,6 +285,17 @@ func (mp *MultiPooler) Init(startCtx context.Context) error {
 	multipooler.ServingStatus = clustermetadatapb.PoolerServingStatus_NOT_SERVING
 	multipooler.PoolerDir = mp.poolerDir.Get()
 
+	// Set initial pooler type based on flag (only affects first-time registration)
+	switch mp.initialPoolerType.Get() {
+	case "REPLICA":
+		multipooler.Type = clustermetadatapb.PoolerType_REPLICA
+	case "UNKNOWN":
+		multipooler.Type = clustermetadatapb.PoolerType_UNKNOWN
```
deepthi: We could remove this and disallow starting as UNKNOWN. I can't think of a case where we would actually want UNKNOWN.

deepthi: Technically we don't need this flag at all right now and can simply default to REPLICA.
@Rafael @sougou WDYT?

sougou: Agree. I don't see a use case where we'll ever need UNKNOWN.

go/multipooler/init.go
sougou, line 1: Let's not introduce this flag yet. We can start with REPLICA as the only value for now.

go/multipooler/manager/rpc_backup.go
```diff
@@ -875,6 +883,14 @@ func (pm *MultiPoolerManager) tryAutoRestoreOnce(ctx context.Context) (success b
...
 		return false, false
 	}
+	if len(backups) == 0 {
+		// No backups at all - this is a fresh cluster. Bootstrap will handle
+		// initialization, so don't retry auto-restore. This prevents a race
+		// where auto-restore picks up a backup created during bootstrap before
+		// bootstrap can initialize this pooler as a standby.
+		pm.logger.InfoContext(ctx, "Auto-restore skipped: no backups exist for shard (fresh cluster, bootstrap will initialize)")
+		return false, true // done, don't retry
+	}
```
deepthi, line 893: Pointing out for posterity that this is the key change that allowed things to work properly when multiorch bootstrap is racing with pooler startup.

PR #451, opened 2026-01-05
fix(auto-restore): makes sure we have a term set after restoring

rafael:
# Description
* When auto-restoring, we were not initializing the consensus term. This created issues down the road because multiorch can't fix replication.
* This is a quick fix for our current demo, but we will come back to this and refactor the code such that:
   - Bootstrap ends the moment we take the bakcup.
   - There is only one path for auto-restore.

### DIFF
--- a/go/multipooler/manager/rpc_backup.go
+++ b/go/multipooler/manager/rpc_backup.go
@@ -929,7 +929,7 @@ func (pm *MultiPoolerManager) tryAutoRestoreOnce(ctx context.Context) (success b
 	// validateAndUpdateTerm will automatically update our term when multiorc fixes replication.
 	var term int64 = 0
 	if pm.consensusState != nil {
-		pm.logger.InfoContext(ctx, "Loading default consensus term for auto-restore")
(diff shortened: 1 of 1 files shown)


go/multipooler/manager/rpc_backup.go
```diff
@@ -924,7 +924,24 @@ func (pm *MultiPoolerManager) tryAutoRestoreOnce(ctx context.Context) (success b
 		return false, false
 	}
 
-	pm.logger.InfoContext(ctx, "Auto-restore: completed successfully", "backup_id", latestBackup.BackupId)
+	// Set consensus term after restore.
+	// Use term=1 as the default for bootstrap scenarios. If the cluster is at a higher term,
+	// validateAndUpdateTerm will automatically update our term when multiorc fixes replication.
+	const defaultBootstrapTerm int64 = 1
```
dweitzman: Maybe add a TODO or something about the follow-ups you mention in the description?

deepthi (approved):

PR #456, opened 2026-01-06
test(queryserving): add PostgreSQL transaction behavior tests

GuptaManan100:
### Description
- Add new `queryserving` test package in `go/test/endtoend/`
- Add `TestTransactionScenarios` covering PostgreSQL transaction semantics
- The intent is to understand how Postgres behaves to be able to duplicate that behaviour for MultiGres. At that point, we'll run the same tests against MultiGateway instead of going to the Postgres server directly.

### Tests
**Multi-Statement Handling (6 tests)**
- Successful multi-statement with implicit transaction
- Failure rolls back entire implicit transaction
- BEGIN transitions implicit to explicit transaction
- Failure after BEGIN without prior COMMIT rolls back everything
- Failure before BEGIN rolls back everything
- Multiple BEGIN/COMMIT blocks - only committed data survives errors

**Autocommit Behavior (1 test)**
- Single statements auto-commit independently

**DDL Transactions (4 tests)**
- DDL is fully transactional (can be rolled back)
- DDL rolls back with DML failure in implicit transaction
- DDL survives in committed explicit transaction
- CREATE DATABASE cannot run inside transaction block

**Explicit Transactions (4 tests)**
- BEGIN/COMMIT workflow
- BEGIN/ROLLBACK workflow
- SAVEPOINT and ROLLBACK TO SAVEPOINT
- Aborted transaction state (error code 25P02)

### DIFF
--- a/go/test/endtoend/queryserving/transaction_test.go
+++ b/go/test/endtoend/queryserving/transaction_test.go
@@ -23,7 +23,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/multigres/multigres/go/pgprotocol/client"
+	"github.com/multigres/multigres/go/common/pgprotocol/client"
 	"github.com/multigres/multigres/go/test/endtoend/shardsetup"
 	"github.com/multigres/multigres/go/test/utils"
 )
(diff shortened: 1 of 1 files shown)


go/test/endtoend/queryserving/transaction_test.go
```diff
@@ -0,0 +1,582 @@
...
+
+// multiStatementTestCases returns test cases for multi-statement behavior (Section 1).
+func multiStatementTestCases() []transactionTestCase {
+	return []transactionTestCase{
+		{
+			// Example 1: Successful Multi-Statement (Implicit Transaction)
+			// All statements execute within TBLOCK_IMPLICIT_INPROGRESS and commit together.
+			name: "Example1_SuccessfulMultiStatement",
+			testFunc: func(ctx context.Context, t *testing.T, conn *client.Conn) error {
+				_, err := conn.Query(ctx, "DROP TABLE IF EXISTS users_ex1")
```
dweitzman, line 145: Seems like these tests could potentially leave behind data that accidentally impacts other tests

Might be worth trying to have a simple approach to knowing what tables might be created during tests and making sure they're always dropped between test cases?

dweitzman (approved):

PR #459, opened 2026-01-06
feat(webadmin): Improves topology view of the cluster

rafael:
# Desc
- Adds polling to the cluster state.
- Exposes pooler status via the RPC interface so we can display more information.
- Animate some of the edges depending on the actual state of the cluster
# Loom
- Loom showing how this looks: https://www.loom.com/share/86b4027d94e64b78afac7d4d0895eb30

### DIFF
--- a/web/multiadmin/components/topology-graph.css
+++ b/web/multiadmin/components/topology-graph.css
@@ -1,0 +1,50 @@
+/* Animated dashed red border for disconnected nodes */
+@keyframes marching-ants {
+  0% {
+    background-position: 0 0, 100% 0, 0 100%, 0 0;
+  }
+  100% {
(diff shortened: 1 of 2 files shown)


web/multiadmin/lib/api/client.ts
```diff
@@ -193,6 +195,14 @@ export class MultiAdminClient {
       `/api/v1/backups${query ? `?${query}` : ""}`
     );
   }
+
+  // Pooler Status operations
+
+  async getPoolerStatus(poolerId: ID): Promise<GetPoolerStatusResponse> {
```
dweitzman, line 201: We should be able to generate the typescript types and possibly the client from the proto3, as a future option

rafael: oh cool. I'll look into this after the demo

PR #460, opened 2026-01-06
feat(multiorch): bootstrap simplification

rafael:
## Description

* This simplifies the bootstrap process by not explicitly initializing standbys as part of it.
* We have already run into a few races where `InitializeAsStandby` conflicts with the auto restore logic. We discussed making this idempotent, but leaned toward a simpler approach:

  * **Bootstrap ends** the moment we choose a primary.
  * **Standby MultiPoolers** organically restore from backup when they detect one exists.
  * **MultiOrch** will organically configure replicas once they show up.

## Tests

* Integration tests have good coverage and validate that this approach works correctly.
* I updated a few unit tests that were asserting that the code called `InitializeAsStandby`.
* I think there is a race in `IsInitialized` where it can return `true` before the consensus term is set. For now:
  * I updated the integration test to use `require.Eventually` and validate the state converges.
  * I created a ticket to audit `IsInitialized` behavior.

### DIFF
--- a/go/test/endtoend/multiorch/bootstrap_test.go
+++ b/go/test/endtoend/multiorch/bootstrap_test.go
@@ -197,20 +197,43 @@ func TestBootstrapInitialization(t *testing.T) {
 	})
 
 	t.Run("verify consensus term", func(t *testing.T) {
-		// All initialized nodes should have consensus term = 1
-		for name, inst := range setup.Multipoolers {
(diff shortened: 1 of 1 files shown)


dweitzman (approved):
Makes sense to me

deepthi (approved):
Tested manually. LGTM.

PR #465, opened 2026-01-07
feat(telemetry): add multigres.pooler.id attributes to multipooler gRPC client spans

dweitzman:
Implements functional options pattern for gRPC client telemetry configuration, allowing callers to set OpenTelemetry span attributes like peer.service without duplicating stats handlers.

Changes:
- grpccommon: Add ClientOption interface with WithMultipoolerTarget() and WithDialOptions()
- grpccommon: Deprecate NewClient() in favor of NewClientWithOptions()
- grpccommon: Upgrade semconv from v1.26.0 to v1.37.0
- rpcclient: Thread pooler ID through connection cache (getOrDial → newDial)
- Add WithMultipoolerTarget() helper that sets multigres.pooler.id
- rpcclient: Use WithMultipoolerTarget for multipooler connections
- poolergateway: Use WithMultipoolerTarget for direct connections

This makes it easier to identify which specific multipooler instance a span is communicating with in distributed traces.

### DIFF
--- a/go/tools/grpccommon/options.go
+++ b/go/tools/grpccommon/options.go
@@ -20,7 +20,6 @@ import (
 	"github.com/spf13/pflag"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"go.opentelemetry.io/otel/attribute"
-	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 )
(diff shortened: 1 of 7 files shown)


go/tools/grpccommon/options.go
```diff
@@ -66,12 +68,82 @@ func LocalClientDialOptions() []grpc.DialOption {
...
+
+type funcOption func(*clientConfig)
+
+func (f funcOption) apply(c *clientConfig) {
+	f(c)
+}
+
 // NewClient creates a gRPC client connection with OpenTelemetry instrumentation.
-// Use this instead of grpc.NewClient() directly to ensure consistent telemetry.
+// Deprecated: Use NewClientWithOptions for custom telemetry configuration.
```
deepthi: Do we plan to remove all calls to the deprecated func and then delete it?

deepthi (approved):

go/tools/grpccommon/options.go
dweitzman: I'll make the change now in this PR, just need to wrap DialOption within a ClientOption for existing callers

PR #466, opened 2026-01-07
fix(multiorch): fix replication should check properly for success

deepthi:
- It is not enough to check for pg_last_wal_receive_lsn != "", we must check the actual status of wal_receiver
- pg_rewind was not getting called at all because of the above bug, and that code path needed a bunch of fixes.


config/postgres/template.cnf
```diff
@@ -87,6 +87,7 @@ wal_buffers = {{.WalBuffers}}			# min 32kB, -1 sets based on shared_buffers
 					# (change requires restart)
```
rafael, line 87: both of these settings are required for pg_rewind

rafael (approved):
LGTM

PR #489, opened 2026-01-08
test(testutil): fix data races in MockPgCtldService

dweitzman:
Add mutex protection to prevent concurrent access to call tracking slices and configuration fields when multiple goroutines call RPC methods simultaneously.

Fixes TestReplicationStatus race conditions:
- https://github.com/multigres/multigres/actions/runs/20825569457
- https://github.com/multigres/multigres/actions/runs/20825983165

Verified with -race across 300 test iterations.


rafael:
I fixed this in a PR that was merged earlier. we can close it

PR #493, opened 2026-01-09
fix(rewind): implement a separate DemoteStalePrimary RPC and do the rewind there

deepthi:
# Description
Get the DemoteStalePrimary action working properly with pg_rewind
- Previous behavior
   - Only act after the Stale Primary postgres is back
   - Then do a Demote followed by FixReplication. 

The problem with this is that it would work for the first failover, but the second or third failover among the same 3 servers would fail. After learning more about how things work with pg_rewind, we realized that we should never start an old primary postgres without first doing rewind. This PR implements that by creating a new RPC that can be used by the StalePrimary recovery which does everything in the right order.


go/multipooler/manager/manager.go
```diff
@@ -1088,20 +1088,22 @@ func (pm *MultiPoolerManager) restartPostgresAsStandby(ctx context.Context, stat
...
-		pm.isOpen = true
-		pm.logger.InfoContext(ctx, "Manager connections reopened (monitor NOT restarted)")
+	pm.logger.InfoContext(ctx, "Closing stale connections after postgres restart")
+	pm.closeConnectionsLocked()
+
+	pm.logger.InfoContext(ctx, "Reopening manager connections after restart")
+	pm.ctx, pm.cancel = context.WithCancel(context.TODO())
+	if err := pm.openConnectionsLocked(); err != nil {
+		pm.mu.Unlock()
+		pm.logger.ErrorContext(ctx, "Failed to reopen connections after restart", "error", err)
```
sougou: Use pm.reopenConnections instead.

go/multipooler/manager/manager.go
```diff
@@ -1079,10 +1088,28 @@ func (pm *MultiPoolerManager) restartPostgresAsStandby(ctx context.Context, stat
...
+		return mterrors.Wrap(err, "failed to reopen connections")
+	}
+	pm.isOpen = true
```
sougou: Only Open and Close should change pm.isOpen. pm.reopen doesn't change isOpen state.

kind_demo/multiorch.log
sougou, line 1: Are these log files checked in intentionally?

go/multipooler/manager/rpc_manager.go
```diff
@@ -950,6 +992,130 @@ func (pm *MultiPoolerManager) UndoDemote(ctx context.Context) error {
...
+	pm.logger.InfoContext(ctx, "DemoteStalePrimary RPC called",
+		"source", source.Id.Name,
+		"consensus_term", consensusTerm)
+
+	// Acquire the action lock to ensure only one mutation runs at a time
+	ctx, err := pm.actionLock.Acquire(ctx, "DemoteStalePrimary")
+	if err != nil {
+		return nil, mterrors.Wrap(err, "failed to acquire action lock")
+	}
+	defer pm.actionLock.Release(ctx)
```
sougou, line 987: You can call pm.disableMontiorInternal->defer pm.enableMonitorInternal here. Will that work?

go/multipooler/manager/rpc_manager.go
```diff
@@ -950,6 +992,130 @@ func (pm *MultiPoolerManager) UndoDemote(ctx context.Context) error {
...
+
+	// Validate the term
+	if err := pm.validateTerm(ctx, consensusTerm, force); err != nil {
```
sougou, line 998: validateTerm doesn't seem to change the term. We also need to update the term if we're going to accept material requests.

sougou (changes requested):
Suggested some fixes including how to handle enable/disable of monitorpg.
I didn't go into timeline divergence. I trust your approach there.

go/multipooler/manager/rpc_manager.go
rafael: This is ok. Because in this context we assume the term will be stale. The end state here is a disconnected replica. Once multiorch fixes replication, it will also update the term.

rafael: Actually it's not ok. We need to update. I missed that we are now wiring replication. I fixed it

PR #494, opened 2026-01-09
feat(demo): Improvements to launch script

rafael:
# Desc
- Splits launch scripts between infra / and multigres cluster.
- Adds automatic port forwarding script.
- Adds script to test failovers.
# Test
- Tested in my local kind cluster

### DIFF
--- a/kind_demo/teardown.sh
+++ b/kind_demo/teardown.sh
@@ -15,7 +15,9 @@
 
 set -ex
 
-# Kind cluster demo teardown also deletes data.
+# Kind cluster demo teardown - removes everything including data
+# This script calls teardown-multigres-cluster.sh and teardown-infrastructure.sh,
+# then deletes the Kind cluster and data directory
 
 # Ensure we're in the kind_demo directory.
 # This is because we use a relative path name to the data files.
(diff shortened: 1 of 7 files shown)


kind_demo/teardown.sh
```diff
@@ -24,8 +26,37 @@ if [[ $(basename "$PWD") != "kind_demo" ]]; then
...
+  ./teardown-infrastructure.sh
+fi
+
+# Delete the Kind cluster
+echo "Deleting Kind cluster..."
 kind delete cluster --name=multidemo
+
+# Clean up data directory
+echo "Cleaning up data directory..."
 rm -rf data/*
```
dweitzman, line 47: Minor comment: if we wanted this script to be idempotent, I think this line would need to change -- for me it currently fails if data/* is already deleted

kind_demo/port-forward.sh
```diff
@@ -0,0 +1,200 @@
...
+        echo "  psql --host=localhost --port=15433 -U postgres -d postgres  (zone1-0)"
+        echo "  psql --host=localhost --port=15434 -U postgres -d postgres  (zone1-1)"
+        echo "  psql --host=localhost --port=15435 -U postgres -d postgres  (zone1-2)"
+        echo ""
+        echo "To stop: $0 stop"
+        echo "To check status: $0 status"
+        echo ""
+        ;;
+
+    stop)
```
dweitzman, line 176: I wonder if the port-forwarding approach here is more complex than it needs to be. Doesn't kubectl port-forward automatically stop when the underlying k8s resource goes away?

Would it be crazy to automatically start port-forwarding as a background process whenever the cluster goes up and rely on port-forwarding stopping on its own with the pods get restarted or stopped?

kind_demo/failover-test.py
dweitzman, line 1: Is this script something you're thinking we'd keep in the repo for a long time? Feels like it'd make sense to default to writing traffic or chaos simulation tools in Go if they're going to be part of this repo

If this is intended to be short-lived, maybe a draft PR would be fine?

PR #497, opened 2026-01-10
feat(demo): simplification to start scripts

rafael:
# Desc
- Addresses feedback from https://github.com/multigres/multigres/pull/494
- Minor improvements / simplification to start / teardown scripts for the demo

### DIFF
--- a/kind_demo/launch-infra.sh
+++ b/kind_demo/launch-infra.sh
@@ -113,4 +113,4 @@ echo "  (multipooler, multiorch, multigateway)"
 echo ""
 
 # Start infrastructure port-forwards
-./port-forward-infrastructure.sh
+./port-forward-infra.sh


kind_demo/teardown-infrastructure.sh
dweitzman, line 1: Are the teardown scripts supposed to be deleted in this PR?

dweitzman: Are we saying deleting the k8s cluster is the only kind of teardown?

rafael: Yeah, I think that was actually the previous version. Simpler and in fact less error prone. I've seen cases where the infra fails to bootstrap and the previous script expects resources to exists otherwise fails. 

What do you think?

PR #500, opened 2026-01-12
fix(tests):  Fixes several flakes

rafael:
# Description
- Fixes some timeouts that make tests flaky
- Also fixes some issues around disabling / enabling pg monitor that was also causing flakes.
- I still see other flakes, but this seems way better. Going to merge this and keep iterating

### DIFF
--- a/go/cmd/pgctld/command/init.go
+++ b/go/cmd/pgctld/command/init.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 
 	"github.com/multigres/multigres/go/services/pgctld"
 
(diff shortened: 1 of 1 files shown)


go/cmd/pgctld/command/start.go
```diff
@@ -299,68 +299,79 @@ func waitForPostgreSQLWithConfig(logger *slog.Logger, config *pgctld.PostgresCtl
...
-			if !isProcessRunning(pid) {
-				// PID file exists but process is gone - crashed
-				logTail := readLogTail(logPath, 20)
-				logger.Error("PostgreSQL process crashed during startup",
-					"pid", pid,
-					"attempt", i,
-					"postgresql_log_tail", logTail,
-				)
-				return fmt.Errorf("PostgreSQL process (PID %d) crashed during startup (check postgresql.log)", pid)
+	ticker := time.NewTicker(1 * time.Second)
```
dweitzman, line 302: Could use retry.New() here to help with the retries vs the timeout, but I don't have strong feelings either way.

dweitzman (approved):

go/cmd/pgctld/command/start.go
rafael: I tried this change and for some reason tests started to fail consistently in my local. I will try just this change in a different PR

go/cmd/pgctld/command/init.go
```diff
@@ -175,31 +147,50 @@ func initializeDataDir(logger *slog.Logger, poolerDir string, pgUser string) err
...
+	// Capture both stdout and stderr to include in error messages
+	output, err := cmd.CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("failed to resolve password: %w", err)
+		return fmt.Errorf("initdb failed: %w\nOutput: %s", err, string(output))
 	}
-	// Start PostgreSQL temporarily in single-user mode to set password
-	// Use the configured user in single-user mode with trust auth to set the password
-	// Set password_encryption to scram-sha-256 to ensure SCRAM encoding
-	cmd := exec.Command("postgres", "--single", "-D", dataDir, pgUser)
```
rafael, line 195: I was seeing a flake in CI like this:
```
  {"time":"2026-01-12T20:41:34.471980353Z","level":"ERROR","msg":"recovery action                                         
  failed","problem_code":"ShardNeedsBootstrap","pooler_id":"multipooler-test-cell-pooler-1","error":"failed to            
  initialize empty primary: rpc error: code = Unknown desc = failed to initialize data directory: rpc error: code =       
  Unknown desc = failed to initialize data directory: failed to initialize data directory: failed to set user             
  password: failed to set postgres password: exit status                                                                  
  1","trace_id":"24883e89818c3e0b447d65cfd42eb36f","span_id":"b1c84ed1d2134d80"}                                                                                                                          ```
```

I think it was probably timing out to start pg in single mode.  A simple (and I think better way) is to actually set the password as part of initdb.  I think this both simplifies the code and removes this potential source of errors

PR #501, opened 2026-01-12
feat(observability): reorganize Grafana dashboard to use service selector

dweitzman:
Reorganize the Grafana dashboard with one panel for http & grpc monitoring where you use the service selector at the top to choose what service(s) you want to filter on. Service-specific metrics like connection pool stats are still in service-specific panels.

Previously we only displayed HTTP metrics for multigateway, for example, but it makes sense to me that we'd want to be able to see them for any service like multiadmin, etc.

To make this work, the service selector query also had to be fixed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,6 +33,7 @@ require (
 	go.opentelemetry.io/contrib/exporters/autoexport v0.64.0
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.63.0
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.64.0
+	go.opentelemetry.io/contrib/samplers/probability/consistent v0.33.0
 	go.opentelemetry.io/otel v1.39.0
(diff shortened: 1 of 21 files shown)


kind_demo/observability/grafana-dashboard.json
```diff
@@ -1048,7 +906,7 @@
         "multi": true,
         "name": "service",
         "options": [],
-        "query": "label_values(up, job)",
+        "query": "label_values({__name__=~\"http_server_request_duration_seconds|rpc_server_duration_milliseconds\"}, job)",
```
dweitzman, line 909: The service selector dropdown wasn't working previously because "up" is a metric prometheus creates when you fetch metrics by pulling, but the demo is configured to push instead of pulling metrics so this needed to be changed to other metrics that exist. `http_server_request_duration_seconds` + `rpc_server_duration_milliseconds` should cause any service that responds to either http or grpc to appear in the dropdown options

dweitzman:
We can revisit this, but for now my attention is on other things

PR #504, opened 2026-01-13
fix(multipooler): wait for topology before initialization to prevent pgbackrest hang

dweitzman:
InitializeEmptyPrimary and InitializeAsStandby were being called before topology finished loading, causing backupLocation to be empty. This resulted in invalid pgbackrest configs with empty repo1-path, causing stanza-create to hang indefinitely and preventing bootstrap completion.

Add checkReady() guards to both RPCs to wait for topology, and add defensive validation in initPgBackRest to fail fast with a clear error if backupLocation is missing.


rafael:
argh - The code coverage, keeps having some flaky tests. Going to merge this one

PR #505, opened 2026-01-13
feat(e2e): Migrate gateway tests to use sharedsetup

rafael:
# Desc
- This PR looks bigger than it actually is. It is mostly moving some files around.
- The main change is that the multigateway tests start using the more consistent shardsetup. This also removes one full cluster bootstrap from the tests. 
- It also moves `cluster_test` to it's own package. This test is in reality testing the local provisioner. Moved this one to it's own package as the `test_main.go` for that one is slightly different and it was conflicting with the shared setup.

### DIFF
--- a/go/test/endtoend/shardsetup/setup.go
+++ b/go/test/endtoend/shardsetup/setup.go
@@ -1077,8 +1077,7 @@ func (s *ShardSetup) DemotePrimary(t *testing.T) {
 	require.NoError(t, err, "failed to connect to primary")
 	defer client.Close()
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-	defer cancel()
+	ctx := utils.WithTimeout(t, 20*time.Second)
 
 	// Demote using the Demote RPC with term 1 (clean state starts at term 1)
 	_, err = client.Manager.Demote(ctx, &multipoolermanagerdatapb.DemoteRequest{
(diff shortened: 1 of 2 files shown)


go/test/endtoend/shardsetup/setup.go
```diff
@@ -1045,8 +1077,7 @@ func (s *ShardSetup) DemotePrimary(t *testing.T) {
 	require.NoError(t, err, "failed to connect to primary")
 	defer client.Close()
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-	defer cancel()
+	ctx := utils.WithTimeout(t, 20*time.Second)
```
rafael, line 1080: I saw a flake around this, so increased the timeout and also changed it to use the utils package

go/test/endtoend/shardsetup/multigateway_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	_, err = db.ExecContext(ctx, fmt.Sprintf("INSERT INTO %s (id, value) VALUES (1, 'test')", tableName))
+	require.NoError(t, err, "failed to insert via multigateway")
+
+	var value string
+	err = db.QueryRowContext(ctx, fmt.Sprintf("SELECT value FROM %s WHERE id = 1", tableName)).Scan(&value)
+	require.NoError(t, err, "failed to select via multigateway")
+	assert.Equal(t, "test", value)
+
+	_, err = db.ExecContext(ctx, "DROP TABLE "+tableName)
```
dweitzman, line 88: If one of the require statements above fails this drop might not happen

Maybe out of scope for this particular PR: I think there are some options when getting the shared test setup to say which tables to drop afterwards, but I wonder if it would make the most sense for the per-test setup helpers to just drop any table that isn't in the multigres schema

PR #507, opened 2026-01-13
fix(multipooler): adds term after restore / ensure monitoring is disabled

rafael:
## Summary                                                                                                              
                                                                                                                          
Fixes an issue where the consensus term was not properly loaded after restoring from backup. This could create issues in the future.                                                  
                                                                                                                          
## Changes                                                                                                              
                                                                                                                          
### Core Fix: Term Loading After Restore                                                                                
  - Moved consensus term loading from `restoreAndStartPostgres` to `restoreFromBackupLocked` in `rpc_backup.go:367-380`   
  - Term is now loaded AFTER PostgreSQL starts, ensuring the restored term file is properly read                          
  - Previous implementation only loaded during original restore, but not if that happened via the RPC.                                       
                                                                                                                          
### Monitor Management Improvements                                                                                     
  - Added `PausePostgresMonitor()` helper method in `manager.go:1813-1839`                                                
  - Returns a resume function for clean defer-based monitor state restoration                                             
  - Updated `Demote`, `DemoteStalePrimary`, and `RewindToSource` to use the new pattern                                   
  - `RestoreFromBackup` now pauses monitoring during restore to prevent interference                                      
                                                                                                                          
### Test Improvements                                                                                                   
  - Added comprehensive term verification in `TestBackup_CreateListAndRestore`                                            
    - Updates term to higher value before restore                                                                         
    - Verifies term is restored to backup's original value (1)                                                            
    - Uses restored term when configuring replication                                                                     
  - Replaced all `time.Sleep` calls with `require.Eventually` for reliability                                             
  - Changed polling interval from 1s to 100ms for faster test completion                                                  
  - Simplified test output (removed step enumeration)

### DIFF
--- a/go/multipooler/manager/rpc_initialization_test.go
+++ b/go/multipooler/manager/rpc_initialization_test.go
@@ -596,33 +596,6 @@ func TestMonitorPostgres_WaitsForReady(t *testing.T) {
 	assert.True(t, mockPgctld.startCalled, "Should attempt to start when ready")
 }
 
-func TestMonitorPostgres_ExitsOnContextCancel(t *testing.T) {
-	ctx, cancel := context.WithCancel(context.Background())
-	cancel() // Cancel immediately
-
-	readyChan := make(chan struct{})
-	close(readyChan) // Ready immediately
-
-	mockPgctld := &mockPgctldClient{
-		statusResponse: &pgctldpb.StatusResponse{
-			Status: pgctldpb.ServerStatus_STOPPED,
-		},
-	}
-
-	pm := &MultiPoolerManager{
-		logger:       slog.Default(),
-		readyChan:    readyChan,
-		pgctldClient: mockPgctld,
-		state:        ManagerStateReady,
-	}
-
-	// Call iteration with cancelled context - should handle gracefully
-	pm.monitorPostgresIteration(ctx)
-
-	// Should still complete without panic (context cancellation handled by pgctld client)
-	// Note: Actual behavior depends on how pgctld client handles cancelled context
-}
-
 func TestMonitorPostgres_HandlesRunningPostgres(t *testing.T) {
 	ctx := context.Background()
 


go/multipooler/manager/rpc_backup.go
```diff
@@ -329,10 +329,15 @@ func (pm *MultiPoolerManager) RestoreFromBackup(ctx context.Context, backupID st
 	}
 	defer pm.actionLock.Release(ctx)
 
+	// Pause monitoring during restore to prevent interference
+	resumeMonitor := pm.PausePostgresMonitor()
```
dweitzman: This makes me a little nervous if in the future multiple methods are trying to pause and resume the postgres monitor and end up fighting with each other.

In terms of synchronization I'm also not sure that this is ensuring the monitor is definitely stopped before it returns? We'd need to use something like the custom timer package instead of just retry.New(), although I have a proposal to replace that package with what feels like a more go-idiomatic alternative in https://github.com/multigres/multigres/pull/394

sougou: I think we rely on action lock for preventing such issues. I don't think it's a strong model, but it may be good enough. We can add an action lock check in enable & disable.

dweitzman: If this is assuming an action lock is held then we should make sure it takes a context argument and asserts that the action lock is held by the current call stack

go/multipooler/manager/rpc_initialization.go
```diff
@@ -50,6 +50,13 @@ func (pm *MultiPoolerManager) InitializeEmptyPrimary(ctx context.Context, req *m
 	}
 	defer pm.actionLock.Release(ctx)
 
+	// Pause monitoring during initialization to prevent interference
```
rafael, line 53: This is another race that we need to prevent. Saw it one in one the CI flakes:
```
2026-01-14T17:08:17.6711187Z {"time":"2026-01-14T17:06:30.462344488Z","level":"INFO","msg":"InitializeEmptyPrimary called","shard":"0-inf","term":1,"trace_id":"dfdedc11cfb8c9837f1dd34244effbc9","span_id":"cbaad37a11eb3fcd"}
2026-01-14T17:08:17.6712456Z {"time":"2026-01-14T17:06:30.46238793Z","level":"INFO","msg":"Initializing data directory","shard":"0-inf","trace_id":"dfdedc11cfb8c9837f1dd34244effbc9","span_id":"cbaad37a11eb3fcd"}
2026-01-14T17:08:17.6713439Z {"time":"2026-01-14T17:06:30.753407235Z","level":"INFO","msg":"MonitorPostgres: PostgreSQL initialized but not running, starting PostgreSQL"}
```
The monitor tried to start PG, while InitializeEmptyPrimary was still running

go/test/endtoend/shardsetup/setup.go
```diff
@@ -1032,6 +1032,19 @@ func (s *ShardSetup) breakReplication(t *testing.T, ctx context.Context) {
 
 		_, _ = client.Pooler.ExecuteQuery(ctx, "ALTER SYSTEM RESET primary_conninfo", 0)
 		_, _ = client.Pooler.ExecuteQuery(ctx, "SELECT pg_reload_conf()", 0)
+
+		// Wait for primary_conninfo to be cleared and WAL receiver to stop
+		// pg_reload_conf() is async, so we need to wait for changes to take effect
+		require.Eventually(t, func() bool {
```
rafael, line 1038: This should fix this flake:
```
  ❌ TestShardSetup_WithoutReplication                                                                                    
  shardsetup_test.go:217: SetupTest: Cleared synchronous_standby_names on primary pooler-2                                
  shardsetup_test.go:217: SetupTest: Cleared primary_conninfo on standby pooler-1                                         
  shardsetup_test.go:217: SetupTest: Cleared primary_conninfo on standby pooler-3                                         
  shardsetup_test.go:228:                                                                                                 
  Error Trace:                                                                                                            
  /home/runner/work/multigres/multigres/go/test/endtoend/shardsetup/shardsetup_test.go:228                                
  Error:          Not equal:                                                                                              
  expected: ""                                                                                                            
  actual  : "host=localhost port=42869 user=postgres application_name=test-cell_pooler-3"                                 
                                                                                                                          
  Diff:                                                                                                                   
  --- Expected                                                                                                            
  +++ Actual                                                                                                              
  @@ -1 +1 @@                                                                                                             
  -                                                                                                                       
  +host=localhost port=42869 user=postgres application_name=test-cell_pooler-3                                            
  Test:           TestShardSetup_WithoutReplication                                                                       
  Messages:       pooler-3 primary_conninfo should be empty with WithoutReplication()                                     
  shardsetup_test.go:233:                                                                                                 
  Error Trace:                                                                                                            
  /home/runner/work/multigres/multigres/go/test/endtoend/shardsetup/shardsetup_test.go:233                                
  Error:          Should be empty, but was [0xc0002664b0]                                                                 
  Test:           TestShardSetup_WithoutReplication                                                                       
  Messages:       pooler-3 should have no WAL receiver with WithoutReplication()   
```

go/multipooler/manager/rpc_initialization_test.go
```diff
@@ -572,77 +564,67 @@ func TestStartPostgres_StartFails(t *testing.T) {
...
-		// Success
-	case <-time.After(1 * time.Second):
-		t.Fatal("MonitorPostgres should exit after context cancel")
-	}
+	// Call iteration again when ready - should proceed and attempt to start
+	pm.monitorPostgresIteration(ctx)
+	assert.True(t, mockPgctld.startCalled, "Should attempt to start when ready")
 }
 
 func TestMonitorPostgres_ExitsOnContextCancel(t *testing.T) {
```
rafael, line 599: This test was not longer relevant as is part of the runner

PR #509, opened 2026-01-13
feat(multipooler): better logging for backups

rafael:
# Desc 

I've been chasing down test flakes in CI related to bootstrap and backup interactions.  This PR adds better observability for backup/restore operations:                                                                                                                                                                 
  - Periodic progress logging (every 10s) for long-running pgbackrest commands                                            
  - Log backup_location at operation start to catch path mismatches                                                       
  - Extract progress logging into reusable helper                                                                         

Previously, when backups hung or were killed, the logs were silent. Now we'll see clear progress.

### DIFF
--- a/go/multipooler/manager/rpc_backup.go
+++ b/go/multipooler/manager/rpc_backup.go
@@ -876,7 +876,6 @@ func (pm *MultiPoolerManager) runLongCommand(ctx context.Context, cmd *exec.Cmd,
 	// Run command in main goroutine
 	output, err := cmd.CombinedOutput()
 
-	// Cancel logging goroutine immediately to prevent race with completion log
 	cancelLog()
 
 	// Log completion


go/multipooler/manager/rpc_backup_test.go
```diff
@@ -36,13 +37,13 @@ import (
 )
 
 // createTestManager creates a minimal MultiPoolerManager for testing
-func createTestManager(poolerDir, stanzaName, tableGroup, shard string, poolerType clustermetadatapb.PoolerType) *MultiPoolerManager {
-	return createTestManagerWithBackupLocation(poolerDir, stanzaName, tableGroup, shard, poolerType, "/tmp/backups")
+func createTestManager(poolerDir, tableGroup, shard string, poolerType clustermetadatapb.PoolerType) *MultiPoolerManager {
```
rafael, line 40: The parameter was not being used

go/multipooler/manager/rpc_backup.go
```diff
@@ -838,3 +843,51 @@ func (pm *MultiPoolerManager) GetBackupByJobId(ctx context.Context, jobID string
...
+	pm.logger.InfoContext(ctx, "Starting command", "operation", operationName)
+
+	// Run command in goroutine
+	done := make(chan struct{})
+	var output []byte
+	var cmdErr error
+
+	go func() {
+		defer close(done)
+		output, cmdErr = cmd.CombinedOutput()
```
dweitzman: I propose reversing the goroutine vs the non-goroutine here, for a few reasons. One reason is that exec.Cmd() isn't officially thread-safe, so calling CombinedOutput() from a different goroutine feels potentially risky. Could trigger an error in the race detector or some kind of issue due to not having memory barriers from a lock?

I also wonder if the code would also end up simpler from being reversed to something more in the spirit of this:

```
runLongCommand() ([]byte, error) {
  doneCtx, cancelDoneCtx = ...
  go func() {
    // loop and log the time until cancelCtx is cancelled
  }();

  defer cancelDoneCtx()
  return cmd.CombinedOutput()
}
```

dweitzman (approved):

PR #516, opened 2026-01-14
feat(multigateway): add LoadBalancer for pooler selection and connection management

dweitzman:
Replace per-tablegroup discovery with a centralized LoadBalancer that:
- Maintains PoolerConnection instances for discovered multipoolers
- Handles connection lifecycle (create on add, close on remove)
- Implements selection logic (PRIMARY vs REPLICA, cell locality)
- Integrates with existing discovery via LoadBalancerListener

Key components:
- LoadBalancer: Manages pooler connections and selection
- PoolerConnection: Wraps gRPC connection with pooler metadata
- LoadBalancerListener: Bridges discovery events to LoadBalancer

Selection logic:
- PRIMARY: Returns the primary pooler (any cell)
- REPLICA: Prefers local cell, falls back to remote cells
- UNKNOWN type defaults to PRIMARY

This simplifies multigateway architecture by moving pooler selection from discovery into LoadBalancer.

Note: Current implementation uses basic type-based selection without health checks. Health streaming support will be added in follow-up PRs.


docs/query_serving/load_balancing.md
```diff
@@ -0,0 +1,184 @@
...
+- **Health-aware**: Skip poolers with high error rates or slow responses
+- **Weighted selection**: Route more traffic to poolers with more capacity
+
+### Health Streaming Integration
+
+When health streaming is added, LoadBalancer can avoid routing to unhealthy poolers.
+
+### Connection Affinity
+
+For queries requiring session state (temp tables, session variables), route all queries for a session to the same connection. Current design uses shared connections without affinity.
```
GuptaManan100: This is not going to be required. Scatter conn expclitity asks for a connection to a specific pooler-id for the things that need connection affinity. LoadBalancer won't need to worry about it.

go/services/multigateway/planner/variable_set_stmt.go
```diff
@@ -61,7 +62,7 @@ func (p *Planner) planVariableSetStmt(
 		"value", value)
 
 	// 1. Route: Send to PostgreSQL for validation and execution
-	route := engine.NewRoute(p.defaultTableGroup, "", sql)
+	route := engine.NewRoute(p.defaultTableGroup, constants.DefaultShard, sql)
```
GuptaManan100, line 64: Nice, I should have had had this constant from the get go ❤️

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -0,0 +1,261 @@
...
+// GetConnection returns a PoolerConnection matching the target specification.
+// Returns an error immediately if no suitable connection is available (fail-fast).
+//
+// Selection logic:
+// - For PRIMARY: returns the primary pooler (any cell)
+// - For REPLICA: prefers local cell, falls back to other cells
+//
+// This method never waits or retries. Retry logic should be implemented
+// at a higher level in the stack (e.g., PoolerGateway).
+func (lb *LoadBalancer) GetConnection(target *query.Target) (*PoolerConnection, error) {
```
GuptaManan100, line 134: This is great, but we'd need a method to also get a connection for a given poolerID for reserved connections

go/services/multigateway/poolergateway/pooler_gateway.go
GuptaManan100, line 88: Could we also implement this as part of the PR?

go/services/multigateway/discovery.go
GuptaManan100, line 370: We might need to call our onDeleted here too 😄

docs/query_serving/load_balancing.md
dweitzman: Removed

go/services/multigateway/discovery.go
dweitzman: Done

go/services/multigateway/poolergateway/pooler_gateway.go
dweitzman: Done

go/services/multigateway/poolergateway/load_balancer.go
dweitzman: Done

go/services/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -340,36 +230,25 @@ var _ Gateway = (*PoolerGateway)(nil)
 func (pg *PoolerGateway) getSystemServiceClient(ctx context.Context, database string) (multipoolerpb.MultiPoolerServiceClient, error) {
 	// Find any pooler - for authentication we just need access to pg_authid
 	// which is available from any pooler connected to this database.
-	// Try PRIMARY first, fall back to REPLICA if not found in this cell.
-	// TODO: Once multigateway discovers poolers across all cells, we can
-	// remove the REPLICA fallback and always use PRIMARY.
+	// Try PRIMARY first, fall back to REPLICA if not found.
 	target := &query.Target{
 		TableGroup: "default", // TODO: Make configurable or discover from database
+		Shard:      "0-inf",   // TODO: Use proper shard constant
```
GuptaManan100: Shouldn't we use the Default Shard constant we added?

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -0,0 +1,290 @@
...
+		return poolerType == clustermetadatapb.PoolerType_REPLICA
+	default:
+		return false
+	}
+}
+
+// poolerIDString returns the string ID for a pooler.
+func poolerIDString(id *clustermetadatapb.ID) string {
+	return fmt.Sprintf("%s/%s", id.GetCell(), id.GetName())
+}
```
GuptaManan100: We should use `topoclient.MultiPoolerIDString()` instead so that the way of writing a poolerID is consistent across all use cases.

go/services/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -66,209 +48,122 @@ type Gateway interface {
...
 }
 
 func (pg *PoolerGateway) getQueryServiceForTarget(ctx context.Context, target *query.Target) (queryservice.QueryService, error) {
-	pooler := pg.discovery.GetPooler(target)
-	if pooler == nil {
-		return nil, fmt.Errorf("no pooler found for target: tablegroup=%s, shard=%s, type=%s",
-			target.TableGroup, target.Shard, target.PoolerType.String())
+	conn, err := pg.loadBalancer.GetConnection(target)
+	if err != nil {
+		return nil, err
```
GuptaManan100: nit: We removed its usage in a few places but left it in others. The value this function provided was it was a singular way of getting the query service for a target, but it doesn't serve a purpose now in my opinion.

go/services/multigateway/discovery.go
```diff
@@ -531,40 +512,48 @@ func (gd *GlobalPoolerDiscovery) Stop() {
...
+	// Replay current state to the new listener
+	gd.mu.Lock()
+	for _, watcher := range gd.cellWatchers {
+		watcher.mu.Lock()
+		for _, pooler := range watcher.poolers {
+			listener.OnPoolerChanged(pooler.MultiPooler)
 		}
+		watcher.mu.Unlock()
 	}
+	gd.mu.Unlock()
```
GuptaManan100, line 642: RegisterListener acquires `listenersMu`, then `gd.mu`, then `watcher.mu`. But `notifyPoolerChanged/notifyPoolerRemoved` are called from CellPoolerDiscovery while holding `pd.mu`, then acquire `listenersMu`. This is a potential lock ordering inversion: `listenersMu` → `gd.mu` → `pd.mu` vs `pd.mu` → `listenersMu`. This could deadlock if the timing is right. We should look at all the mutexes, and make sure they are acquired in the correct order, and we should document that.

dweitzman: Very good catch! My proposed solution for this to say that notifications and adding listeners to the list are async operations that can be appended to a list to have a deterministic ordering

A separate gorountine will process the notifications and listener appends later, but that gorountine doesn't need to lock any non-notification or non-listener state so there's no risk of lock conflicts.

go/services/multigateway/poolergateway/load_balancer_test.go
```diff
@@ -0,0 +1,575 @@
...
+	lb := NewLoadBalancer("zone1", logger)
+
+	// Add stable poolers that won't be removed
+	const numStablePoolers = 5
+	for i := range numStablePoolers {
+		pooler := createTestMultiPooler(
+			fmt.Sprintf("stable-pooler-%d", i),
+			"zone1",
+			constants.DefaultTableGroup,
+			"0",
```
GuptaManan100, line 281: nit: This should also use the default shard constant.  Also these stable poolers will never match the targets cause the default shard is `0-inf` and not `0`.

GuptaManan100 (approved):
Rest LGTM!

go/services/multigateway/poolergateway/load_balancer_test.go
dweitzman: Fixed

go/services/multigateway/poolergateway/pooler_gateway.go
dweitzman: Removed getQueryServiceForTarget()

go/services/multigateway/poolergateway/pooler_gateway.go
dweitzman: Fixed

PR #522, opened 2026-01-16
fix(multipooler): fix small bug in archive command configuration for multipooler

rafael:
# Desc
* The archive command in the multipooler was not setting the correct path for the pgbackrest archive command. The following log was failing before this change:
```
2026-01-16 13:12:11.705 P00   INFO: archive-get command end: completed successfully (3ms)
2026-01-16 13:12:11.718 P00   INFO: archive-get command begin 2.57.0: [000000010000000000000002, pg_wal/RECOVERYXLOG] --config=~multigres/multigres_local/data/pooler_n6t8hvgl/pgbackrest/pgbackrest.conf --exec-id=19445-9f7300d9 --lock-path=~/multigres/multigres_local/data/pooler_n6t8hvgl/pgbackrest/lock --log-level-console=info --log-level-file=detail --log-path=~/sandboxes/multigres/multigres_local/data/pooler_n6t8hvgl/pgbackrest/logs --log-subprocess --pg1-path=~/sandboxessandboxes/multigres/multigres_local/data/pooler_n6t8hvgl/pg_data --repo1-path=~/sandboxessandboxes/multigres/multigres_local/data/backups/postgres/default/0-inf --stanza=multigres
```

# Test
* Validated locally and opened added integration test.

### DIFF
--- a/go/multipooler/manager/rpc_initialization.go
+++ b/go/multipooler/manager/rpc_initialization.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/multigres/multigres/go/common/mterrors"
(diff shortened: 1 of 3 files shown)


go/multipooler/manager/rpc_initialization.go
```diff
@@ -442,7 +471,7 @@ func (pm *MultiPoolerManager) configureArchiveMode(ctx context.Context) error {
 	// Following the pattern from test/endtoend/multipooler/setup_test.go:479-498
 	archiveConfig := fmt.Sprintf(`
 # Archive mode for pgbackrest backups
-archive_mode = on
+archive_mode = 'on'
```
cuongdo, line 474: this seems harmless, but does this need quotes around it?

rafael: I added this because when we read it back, it includes the quotes. The test was failing because what we generate doesn’t include the quotes, but when we read it back, they were present. So thought it seemed harmless to add it here.

PR #531, opened 2026-01-23
feat(pooler): change pooler to be authoritative source of truth

sougou:
Previously, the pooler tried to respect previously saved settings in the topo. This was leading to confusion about who is the authoritative source of state. In this change, the pooler always overwrites the topo with its current state.
* Removed the cached state from topo.
* Pooler always updates its current state and publishes itself to the topo.
* Simplified MultiPoolerManger initialization. The Config had duplicate fields that were duplicated from the MultiPooler record.
* I had to add some hacks to PostgresMonitor to get the cluster to function properly. This change is incomplete and just enough to allow for this refactor to be completed.
* I also had to disable some tests, which will need to be resurrected after corresponding changes are made in multiorch.

### DIFF
--- a/go/multipooler/manager/manager_test.go
+++ b/go/multipooler/manager/manager_test.go
@@ -160,7 +160,7 @@ func TestManagerState_CancellationDuringLoad(t *testing.T) {
 }
 
 func TestManagerState_RetryUntilSuccess(t *testing.T) {
-	t.Skip("Need to change to test new retry behavior after implementation")
+	t.Skip("TODO(sougou): Need to change to test new retry behavior after implementation")
 	ctx := t.Context()
 	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
 	ts, factory := memorytopo.NewServerAndFactory(ctx, "zone1")
--- a/go/multipooler/manager/rpc_initialization_test.go
+++ b/go/multipooler/manager/rpc_initialization_test.go
@@ -347,7 +347,11 @@ func TestTakeRemedialAction_PostgresRunning(t *testing.T) {
 	ctx := context.Background()
 
 	pm := &MultiPoolerManager{
-		logger: slog.Default(),
+		logger:     slog.Default(),
+		actionLock: NewActionLock(),
+		multipooler: &clustermetadatapb.MultiPooler{
+			Type: clustermetadatapb.PoolerType_REPLICA,
+		},
 	}
 
 	state := postgresState{
@@ -436,7 +440,11 @@ func TestTakeRemedialAction_LogDeduplication(t *testing.T) {
 	ctx := context.Background()
 
 	pm := &MultiPoolerManager{
-		logger: slog.Default(),
+		logger:     slog.Default(),
+		actionLock: NewActionLock(),
+		multipooler: &clustermetadatapb.MultiPooler{
+			Type: clustermetadatapb.PoolerType_REPLICA,
+		},
 	}
 
 	state := postgresState{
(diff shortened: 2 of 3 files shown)


go/multipooler/manager/config.go
```diff
@@ -18,20 +18,12 @@ package manager
...
 	"github.com/multigres/multigres/go/common/topoclient"
 	"github.com/multigres/multigres/go/multipooler/connpoolmanager"
-
-	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 )
 
 // Config holds configuration for the MultiPoolerManager
 type Config struct {
 	SocketFilePath      string
-	PoolerDir           string
```
rafael, line 28: I think this makes sense. I think the reason this was here is that it doens't change after start, so then it was safe to not read it with a lock in the multipooler

go/multipooler/manager/manager.go
```diff
@@ -91,8 +91,10 @@ type MultiPoolerManager struct {
 	//   to match the manager's state.
 	mu sync.Mutex
 
-	isOpen          bool
-	multipooler     *topoclient.MultiPoolerInfo
+	isOpen bool
+	// all fields other than Type and ServingStatus never change after initialization.
```
rafael, line 95: The comment is a bit confusing to me. It's not obvious why we don't need to hold a lock if Type and ServingStatus can change

go/multipooler/manager/manager.go
```diff
@@ -679,16 +632,15 @@ func (pm *MultiPoolerManager) loadMultiPoolerFromTopo() {
 		}
 
 		ctx, cancel := context.WithTimeout(pm.ctx, 5*time.Second)
-		mp, err := pm.topoClient.GetMultiPooler(ctx, pm.serviceID)
+		_, err := pm.topoClient.GetMultiPooler(ctx, pm.serviceID)
```
rafael, line 635: I think no longer need to make this call here. This method now became, load backup location. Loading the multipooler record is not longer needed. 

I see the TODO, so we can address this later, but flagging to check how you are thinking about it.

go/multipooler/manager/manager.go
```diff
@@ -1591,6 +1544,38 @@ func (pm *MultiPoolerManager) takeRemedialAction(ctx context.Context, currentSta
 			pm.logger.InfoContext(ctx, "MonitorPostgres: PostgreSQL is running")
 			pm.pgMonitorLastLoggedReason = reasonPostgresRunning
 		}
+		if currentState.isPrimary {
+			pm.logger.InfoContext(ctx, "MonitorPostgres: PostgreSQL is running and primary")
```
rafael, line 1548: Let's talk about this.

go/multipooler/manager/rpc_backup.go
```diff
@@ -139,9 +140,9 @@ func (pm *MultiPoolerManager) initPgBackRest(ctx context.Context, mode PgBackRes
 
 		Repo1Path: pm.backupLocation,
 
-		Pg1SocketPath: filepath.Join(pm.config.PoolerDir, "pg_sockets"),
-		Pg1Port:       pm.config.PgPort,
-		Pg1Path:       filepath.Join(pm.config.PoolerDir, "pg_data"),
+		Pg1SocketPath: filepath.Join(pm.multipooler.PoolerDir, "pg_sockets"),
+		Pg1Port:       func() int { return int(pm.multipooler.PortMap["postgres"]) }(),
```
rafael, line 144: why call it in a anonymous function?

sougou: Dang. AI slop slipped through my eyes...

go/multipooler/manager/manager.go
sougou: Yeah. The clean fix for this is big. I've created https://github.com/multigres/multigres/issues/532.

go/multipooler/manager/manager.go
sougou: As discussed, I agree that it's safer to use a lock for all fields to avoid future bugs.

PR #533, opened 2026-01-23
feat(multiorch): improvements to multiorch and dead primary recovery

rafael:
# Desc

This PR enhances the multiorch consensus and recovery loops with two efficiency and liveness improvements:

1. **Configurable primary failover grace period**: adds a tunable delay with jitter before executing primary failover. This de synchronizes coordinators so they are less likely to act on the same failure simultaneously, reducing contention that can otherwise stall recovery progress.

2. **Pre vote mechanism**: adds a lightweight pre flight check before starting a leader appointment. If there are not enough healthy, initialized poolers to plausibly reach quorum, we skip the attempt. We also back off when we observe recent term acceptance from another coordinator, treating it as a signal that an election is already in progress and giving it time to complete instead of starting a competing election.

The goal here is to reduce unnecessary work and improve liveness under concurrent failures by lowering coordinator contention.

### Pre Vote Mechanism

Adds a pre flight check before attempting leader appointment. The pre vote validates that:

* There are enough healthy, initialized poolers to plausibly reach quorum.
* Poolers expose consensus term metadata required for coordination decisions.
* No recent term acceptance has occurred from another coordinator, which is used as a signal that a different coordinator is already driving a new term. 

When recent coordinator activity is detected within a short window (about 10 seconds), the coordinator backs off instead of starting a competing election attempt. 

### Grace Period for Primary Failover

Introduces deadline based grace period tracking for primary failover:

* Base delay plus randomized jitter (defaults: 4 seconds base, up to 8 seconds jitter)
* While the system is healthy, the grace period deadline continuously resets
* When a primary failover condition is detected, the deadline freezes and the countdown begins
* The recovery action only executes once the grace period expires

This functions similarly to an election timeout: it de synchronizes multiple coordinators so they do not repeatedly collide while trying to fix the same problem. In practice, this reduces the probability of livelock patterns where coordinators continuously interrupt each other’s progress by repeatedly attempting overlapping recovery operations.

## Configuration

New dynamic flags:

* `--primary-failover-grace-period-base=4s`
* `--primary-failover-grace-period-max-jitter=8s`

Environment variables:

* `MT_PRIMARY_FAILOVER_GRACE_PERIOD_BASE`
* `MT_PRIMARY_FAILOVER_GRACE_PERIOD_MAX_JITTER`

---

## Testing
* All existing tests pass
* Updated dead primary failover to use 3 multiorchs 

## Next Steps

* I was discussing with Sugu that the backoff when a new term started should happen across all actions and not just the primary failover. I will follow up on that change in a new PR.
* Related to above, DemoteStalePrimary still has some issues. I've seen a few flakes where sometimes it hangs. But this PR is already sufficiently large. I will address those as part of other PR's

### DIFF
--- a/go/test/endtoend/multiorch/dead_primary_recovery_test.go
+++ b/go/test/endtoend/multiorch/dead_primary_recovery_test.go
@@ -150,7 +150,7 @@ func TestDeadPrimaryRecovery(t *testing.T) {
 		t.Logf("New primary %s is on term %d", newPrimaryName, newPrimaryTerm)
 
 		// Wait for killed multipooler to rejoin as standby (always wait, even on last iteration)
-		waitForNodeToRejoinAsStandby(t, setup, currentPrimaryName, newPrimaryName, newPrimaryTerm, 15*time.Second)
+		waitForNodeToRejoinAsStandby(t, setup, currentPrimaryName, newPrimaryName, newPrimaryTerm, 30*time.Second)
 
 		// Ensure monitoring is disabled on all multipoolers (multiorch recovery might have re-enabled it)
 		t.Logf("Re-disabling monitoring on all multipoolers after failover %d...", i+1)


demo/k8s/failover-test.py
```diff
@@ -83,8 +83,9 @@ def disable_postgres_monitoring_on_all_poolers():
         log_info(f"  Disabling monitoring on: {pooler_name} (cell={cell}, service_id={service_id})")
         try:
             result = subprocess.run(
-                [MULTIGRES_BIN, "disablepostgresmonitor",
+                [MULTIGRES_BIN, "setpostgresmonitor",
```
rafael: This was broken after a refactor we did in: https://github.com/multigres/multigres/pull/520

demo/local/failover-test.py
```diff
@@ -0,0 +1,775 @@
+#!/usr/bin/env python3
```
rafael: This is a version of the script that works with local. I don't mind not being DRY here. I feel this throw away code

go/cmd/multigres/command/topo/getpoolers.go
```diff
@@ -64,8 +64,13 @@ func runGetPoolers(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to get poolers: %w", err)
 	}
 
-	// Convert to JSON and output
-	jsonData, err := json.MarshalIndent(response, "", "  ")
+	// Convert to JSON and output using protojson to properly render enums as strings
+	marshaler := protojson.MarshalOptions{
```
rafael, line 68: This makes consistent with how we return json from the API. 

I will audit other calls and fix in a different PR. Changed these two here, because I needed them for the failover script

go/services/multiorch/coordinator/leader_appointment.go
```diff
@@ -84,61 +83,31 @@ func (c *Coordinator) BeginTerm(ctx context.Context, shardID string, cohort []*m
...
-// discoverMaxTerm queries all nodes in parallel to find the maximum consensus term.
-func (c *Coordinator) discoverMaxTerm(ctx context.Context, cohort []*multiorchdatapb.PoolerHealthState) (int64, error) {
-	type result struct {
-		term int64
-		err  error
-	}
-
-	results := make(chan result, len(cohort))
-	var wg sync.WaitGroup
+// discoverMaxTerm finds the maximum consensus term from cached health state.
```
rafael, line 89: I think this is better moving forward. Doing these RPC's every time, when we already have this info in the pooler state, seems inefficient

go/cmd/multigres/command/pooler/status.go
```diff
@@ -87,8 +87,13 @@ func runGetPoolerStatus(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("GetPoolerStatus RPC failed: %w", err)
 	}
 
-	// Output the response in JSON format
-	jsonData, err := json.MarshalIndent(response, "", "  ")
+	// Output the response in JSON format using protojson to properly render enums as strings
```
rafael, line 90: This makes it consistent with what we return from the multiadmin API

demo/k8s/failover-test.go
```diff
@@ -0,0 +1,824 @@
+// Copyright 2025 Supabase, Inc.
```
rafael, line 1: Just a port of the failover test script to go.

go/services/multiorch/recovery/analysis/generator.go
```diff
@@ -188,7 +188,6 @@ func (g *AnalysisGenerator) generateAnalysisForPooler(
 
 	// Compute staleness
 	analysis.IsStale = !pooler.IsUpToDate
-	analysis.IsUnreachable = !pooler.IsLastCheckValid
```
rafael, line 191: This is a cleanup. I thought it was confusing to have two fields where one is the negation of the other

go/services/multiorch/recovery/analysis/generator.go
```diff
@@ -231,7 +230,7 @@ func (g *AnalysisGenerator) generateAnalysisForPooler(
 		}
 
 		// Lookup primary info
-		g.populatePrimaryInfo(analysis, shardKey, pooler)
+		g.populatePrimaryInfo(analysis, shardKey)
```
rafael, line 233: this variable was not being used

go/multipooler/manager/rpc_manager.go
```diff
@@ -287,7 +287,7 @@ func (pm *MultiPoolerManager) Status(ctx context.Context) (*multipoolermanagerda
 
 	// Get consensus term if available (use inconsistent read for monitoring)
 	if pm.consensusState != nil {
-		term, err := pm.consensusState.GetInconsistentCurrentTermNumber()
```
deepthi, line 290: Are we still using the old function anywhere?

go/services/multiorch/config/config.go
```diff
@@ -220,6 +222,18 @@ func NewConfig(reg *viperutil.Registry) *Config {
 			Dynamic:  true,
 			EnvVars:  []string{"MT_RECOVERY_CYCLE_INTERVAL"},
 		}),
+		primaryFailoverGracePeriodBase: viperutil.Configure(reg, "primary-failover-grace-period-base", viperutil.Options[time.Duration]{
+			Default:  4 * time.Second,
+			FlagName: "primary-failover-grace-period-base",
+			Dynamic:  true,
+			EnvVars:  []string{"MT_PRIMARY_FAILOVER_GRACE_PERIOD_BASE"},
```
deepthi, line 229: Why do all multiorch options have env variables associated with them? I don't see this on multipooler. Maybe we need to add them there?

go/services/multiorch/recovery/recovery_grace_period.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// NewRecoveryGracePeriodTracker creates a new deadline tracker.
+// By default, uses a random seed for jitter generation and slog.Default() for logging.
+func NewRecoveryGracePeriodTracker(ctx context.Context, config *config.Config, opts ...RecoveryGracePeriodTrackerOption) *RecoveryGracePeriodTracker {
+	dt := &RecoveryGracePeriodTracker{
+		ctx:       ctx,
+		config:    config,
+		logger:    slog.Default(),
+		deadlines: make(map[gracePeriodKey]time.Time),
```
dweitzman, line 85: Should old entries from deadlines be cleared eventually if they're not relevant anymore? Like after a pooler ID is decomissioned, for example.

rafael: I thought about this, but my intuition is that it was a rare edge case and not that much memory.  Something that we can add if we ever see it becoming a problem

go/services/multiorch/config/config.go
rafael: Good call. I'll create a ticket for this. I see in Mutipooler some have them and some don't. We should make it consistent.

PR #542, opened 2026-01-27
feat: DemoteStalePrimary recovery should work when postgres is killed

deepthi:
##  Overview

This PR enhances the recovery flow for dead primary scenarios. Prior to this PR, we already had to ability to detect and recover in a scenario where PostgreSQL is killed with SIGKILL on a primary node. However, we were not able to successfully repair the old primary and have it rejoin the cluster. We now add the ability to automatically recover the killed node as a standby through crash recovery and pg_rewind.

##  Problem Solved

  **Before**: When a primary's PostgreSQL process was killed (simulating database crash or OOM kill), the `DemoteStalePrimary` recovery flow would fail because:
  1. The manager was being completely closed when stopping postgres, preventing subsequent operations
  2. Crash recovery with `postgres --single` would hang indefinitely
  3. Restarting postgres as standby created a chicken-and-egg deadlock (needs connection to configure replication, but needs replication configured to accept connections)
  4. Health checks during consensus caused spurious errors during postgres restarts

  **After**: Killed primaries automatically rejoin the cluster as standbys within ~20 seconds of a new primary election.

##  Key Changes

  1. Crash Recovery Infrastructure

  Files: go/cmd/pgctld/command/crash_recovery.go, go/cmd/pgctld/command/server.go, proto/pgctldservice.proto

  - New CrashRecovery RPC in pgctld service to detect and repair inconsistent postgres state
  - Checks for presence of postmaster.pid (indicates unclean shutdown)
  - Uses postgres --single mode for safe recovery without network startup
  - State detection: dirty_shutdown, needs_crash_recovery, clean

  2. Postgres Restart as standby

  - Solves chicken-and-egg problem:
    - Standby postgres needs WAL stream to reach consistent recovery state
    - Can't reach consistent state without primary_conninfo configured
    - Can't configure primary_conninfo without database connection
    - Solution: run pg_rewind with -R so that the primary_conninfo is written to the config file as part of rewind

  3. Fixed Connection Pool Closure Bug (This is a bug regardless of the feature in this PR)

  File: go/multipooler/manager/rpc_manager.go

in `stopPostgresIfRunning`:
Replace
```
 pm.Close()  // Closes entire manager, breaks subsequent operations
```
with
  ```
  pm.mu.Lock()
  if pm.replTracker != nil {
      pm.replTracker.Close()
      pm.replTracker = nil
  }
  if pm.connPoolMgr != nil {
      pm.connPoolMgr.Close()
  }
  pm.mu.Unlock()
```
  - Now closes ONLY connection pools, not entire manager
  - Allows postgres to stop cleanly (no waiting for connections)
  - Manager stays operational for subsequent operations (topology updates, replication config)

  4. Optimized BeginTerm Health Checks 

  Files: go/multipooler/manager/rpc_consensus.go, go/multipooler/manager/rpc_consensus_test.go

  Moved health check to occur AFTER rejection logic:
 ```
 // Check consensus rules FIRST (lines 77-87)
  if req.Term < currentTerm {
      return response, nil  // Reject old term - no postgres needed
  }
  if req.Term == currentTerm && alreadyAcceptedDifferentCandidate {
      return response, nil  // Reject duplicate - no postgres needed
  }

  // THEN check postgres health (line 94)
  if _, err = pm.query(ctx, "SELECT 1"); err != nil {
      return nil, fmt.Errorf("postgres unhealthy, cannot accept new term: %w", err)
  }
```
  **Benefits**:
  - Don't need postgres healthy to reject based on consensus rules
  - Prevents spurious "postgres unhealthy" errors during restarts
  - Only check health when actually accepting terms (where we need to demote/query)
---
  DemoteStalePrimary Flow (After All Fixes)

  1. DemoteStalePrimary called by multiorch
    ↓
  2. stopPostgresIfRunning()
     - Close connection pools (NOT entire manager) 
     - Stop postgres via pgctld
    ↓
  3. runPgRewind()
     - CrashRecovery RPC (detects/repairs inconsistent state) 
     - pg_rewind -R if timelines diverged
    ↓
  4. restartAsStandbyAfterRewind()
     - Postgres reads primary_conninfo on startup
     - Begins streaming WAL immediately
    ↓
  5. Wait for postgres with exponential backoff 
     - Retry: 100ms, 200ms, 400ms, 800ms, 1s...
     - Max 30s timeout
    ↓
  6. resetSynchronousReplication() + topology update 
     - Manager still operational, DB queries work
    ↓
  7. Node successfully rejoins as standby


go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Start the command
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("failed to start postgres --single: %w", err)
+	}
+
+	// Give postgres a brief moment to attach to stdin before closing it.
+	// Without this, there's a race where stdin.Close() can execute before
+	// postgres has attached, causing postgres to wait indefinitely for input.
+	time.Sleep(50 * time.Millisecond)
```
dweitzman: This makes me nervous. Should there be a WithTimeout() on the context as extra protection against the postgres process hanging indefinitely?

go/multipooler/manager/rpc_manager.go
```diff
@@ -897,6 +899,49 @@ func (pm *MultiPoolerManager) demoteLocked(ctx context.Context, consensusTerm in
...
 
+	// Wait for postgres to accept connections after restart
+	// restartPostgresAsStandby uses skip_wait=true, so postgres may not be ready immediately
+	// Use exponential backoff with 10-second timeout
+	pm.logger.InfoContext(ctx, "Waiting for PostgreSQL to accept connections after demotion")
+
+	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	r := retry.New(100*time.Millisecond, 1*time.Second)
```
dweitzman: Since this is a local operation with no risk of thundering herd issues, the initial wait and retry cap could potentially be lower, like 1ms and 100ms or something like that. Will defer to your judgement

go/multipooler/manager/rpc_manager.go
```diff
@@ -897,6 +899,49 @@ func (pm *MultiPoolerManager) demoteLocked(ctx context.Context, consensusTerm in
...
+				"last_error", lastErr,
+				"timeout_error", err)
+			// Don't fail the demotion - postgres will eventually be ready
+			// The cleanup or subsequent operations will retry
+			break
+		}
+
+		// Check if postgres is ready by querying pg_is_in_recovery
+		queryCtx, queryCancel := context.WithTimeout(ctx, 500*time.Millisecond)
+		result, queryErr := pm.query(queryCtx, "SELECT pg_is_in_recovery()")
```
dweitzman: Should this call the existing pm.isInRecovery() method?

go/multipooler/manager/rpc_manager.go
```diff
@@ -1003,30 +1048,90 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
...
+		lastAttempt = attempt
+		if err != nil {
+			// Context timeout or cancellation
+			msg := fmt.Sprintf("postgres did not accept connections after restart (timeout after %d attempts, last error: %v)", attempt, lastErr)
+			return nil, mterrors.Wrap(err, msg)
+		}
+
+		// Check if postgres is ready by querying pg_is_in_recovery
+		queryCtx, queryCancel := context.WithTimeout(ctx, 500*time.Millisecond)
+		result, queryErr := pm.query(queryCtx, "SELECT pg_is_in_recovery()")
```
dweitzman: Similar comments about using the existing isInRecovery() method and potentially decreasing the timings on this retry loop

go/multipooler/manager/rpc_manager.go
```diff
@@ -1420,10 +1525,18 @@ func (pm *MultiPoolerManager) stopPostgresIfRunning(ctx context.Context) error {
...
 	pm.logger.InfoContext(ctx, "Stopping postgres if running")
 
-	// Close manager to release connections
-	if err := pm.Close(); err != nil {
-		pm.logger.WarnContext(ctx, "Failed to close manager", "error", err)
+	// Close ONLY connection pools to release database connections.
+	// This allows postgres to stop cleanly without waiting for connections,
+	// but keeps the manager operational for subsequent operations.
+	pm.mu.Lock()
+	if pm.replTracker != nil {
```
dweitzman: Should this reuse the existing closeConnectionsLocked() method? Seems like it closes the same things

proto/pgctldservice.proto
```diff
@@ -213,3 +222,22 @@ message PgRewindResponse {
...
+message CrashRecoveryRequest {
+  // No parameters needed - operates on the configured data directory
+}
+
+message CrashRecoveryResponse {
+  // Whether recovery was needed and performed
+  bool recovery_performed = 1;
+
+  // Database cluster state before recovery (e.g., "in production", "shut down")
+  string state_before = 2;
```
dweitzman: Should these return enums instead of strings for better type safety?

proto/pgctldservice.proto
```diff
@@ -49,6 +49,10 @@ service PgCtld {
   // PgRewind rewinds a PostgreSQL data directory to an earlier point in the timeline
   // This is used to resynchronize a server that diverged from the primary after a failback
   rpc PgRewind(PgRewindRequest) returns (PgRewindResponse);
+
+  // CrashRecovery performs crash recovery in single-user mode if needed
+  // This is used before pg_rewind when PostgreSQL was killed uncleanly (e.g., SIGKILL)
+  rpc CrashRecovery(CrashRecoveryRequest) returns (CrashRecoveryResponse);
```
dweitzman, line 56: FYI, if https://github.com/multigres/multigres/pull/570 gets approved and merged before this PR,  it'll cause a minor merge conflict for this method that won't be detected before merging or rebasing

go/multipooler/manager/rpc_manager.go
```diff
@@ -1034,30 +1035,81 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
...
-		return nil, mterrors.Wrap(err, "failed to restart as standby")
-	}
-
-	if err := pm.resetSynchronousReplication(ctx); err != nil {
-		pm.logger.WarnContext(ctx, "Failed to reset synchronous replication", "error", err)
-	}
-
-	pm.logger.InfoContext(ctx, "Configuring replication to source primary",
+	// Configure primary_conninfo BEFORE restarting postgres
+	// After pg_rewind, postgres needs to stream WAL from the primary to reach consistent
```
rafael: I don't understand this change. This was working before without it. In my mental model there is no chicken egg here. Postgres should be able to start (but is not connected), then it starts streaming and catches up. 

what changed?

go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -0,0 +1,138 @@
...
+	clusterState := stringToClusterState(clusterStateStr)
+
+	// For PostgreSQL 17+, check database cluster state
+	// States that indicate need for crash recovery:
+	// - "in production" - was running when killed
+	// - "shutting down" - was shutting down when killed
+	// - "in crash recovery" - already in crash recovery
+	// Clean states: "shut down", "shut down in recovery"
+
+	needsRecovery := clusterState == pb.DatabaseClusterState_DATABASE_CLUSTER_STATE_IN_PRODUCTION ||
```
rafael, line 57: how about unknown ? It seems we should add a guardrail that if we don't understand the state, we should error. Right now it will silently assume we don't need recovery

go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -0,0 +1,138 @@
...
+	"os/exec"
+	"strings"
+	"time"
+
+	pb "github.com/multigres/multigres/go/pb/pgctldservice"
+)
+
+// needsCrashRecovery checks if PostgreSQL requires crash recovery.
+// Returns (needsRecovery bool, clusterState enum, error)
+func needsCrashRecovery(ctx context.Context, logger *slog.Logger, poolerDir string) (bool, pb.DatabaseClusterState, error) {
```
rafael, line 31: It seems we can also inspect the postmaster.pid. And do something like:
* missing file => clean signal (exists=false)
* file exists + live process => running=true
* file exists + dead process => stale=true

go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -0,0 +1,138 @@
...
+		return pb.DatabaseClusterState_DATABASE_CLUSTER_STATE_UNKNOWN
+	}
+}
+
+// runCrashRecovery performs crash recovery in single-user mode.
+// This runs postgres --single to complete crash recovery, then exits cleanly.
+func runCrashRecovery(ctx context.Context, logger *slog.Logger, poolerDir string) error {
+	logger.InfoContext(ctx, "Starting single-user crash recovery")
+
+	crashCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
rafael, line 109: We should think about timeouts here. The overall RPC is gonna be 15 seconds. In this context we should probably be generous. I wouldn't be surprised if of larger databases is actually larger that 30 seconds

go/cmd/pgctld/command/crash_recovery.go
rafael: This also makes me nervous. Claude thinks something like this would work:
```
        // Write a newline to stdin. postgres --single reads commands line-by-line,
        // so this acts as a no-op command. Unlike time.Sleep, this naturally blocks
        // until postgres has attached to the pipe and consumed the input.
        if _, err := io.WriteString(stdin, "\n"); err != nil {
                return fmt.Errorf("failed to write to postgres stdin: %w", err)
        }
```

go/cmd/pgctld/command/restart.go
```diff
@@ -170,7 +170,7 @@ func (r *PgCtlRestartCmd) runRestart(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	result, err := RestartPostgreSQLWithResult(r.pgCtlCmd.lg.GetLogger(), config, r.mode.Get(), r.asStandby.Get())
+	result, err := RestartPostgreSQLWithResult(r.pgCtlCmd.lg.GetLogger(), config, r.mode.Get(), r.asStandby.Get(), false)
```
rafael: nit - could we add `/* wait */ `

go/cmd/pgctld/command/server.go
```diff
@@ -431,3 +431,49 @@ func (s *PgCtldService) PgRewind(ctx context.Context, req *pb.PgRewindRequest) (
...
+	s.logger.InfoContext(ctx, "Database requires crash recovery, running single-user recovery",
+		"state_before", stateBefore.String())
+
+	// Run crash recovery
+	if err := runCrashRecovery(ctx, s.logger, s.poolerDir); err != nil {
+		return nil, fmt.Errorf("crash recovery failed: %w", err)
+	}
+
+	// Verify recovery completed successfully
+	_, stateAfter, err := needsCrashRecovery(ctx, s.logger, s.poolerDir)
```
rafael: as a triple check invariant. Should we look a needRecovery here and validate that returns false?

go/multipooler/manager/manager.go
```diff
@@ -989,40 +989,34 @@ func (pm *MultiPoolerManager) restartPostgresAsStandby(ctx context.Context, stat
...
 	pm.logger.InfoContext(ctx, "Restarting PostgreSQL as standby")
-	// Call pgctld to restart as standby
-	// This will create standby.signal and restart the server
 	req := &pgctldpb.RestartRequest{
 		Mode:      "fast",
 		Timeout:   nil, // Use default timeout
 		Port:      0,   // Use default port
 		ExtraArgs: nil,
 		AsStandby: true, // Create standby.signal before restart
+		SkipWait:  true, // Don't wait for ready - postgres needs primary_conninfo configured
```
rafael: Same flag as the other comment. I feel that if this was true, many things would be broken already. Something doesn't add up

go/multipooler/manager/manager.go
```diff
@@ -989,40 +989,34 @@ func (pm *MultiPoolerManager) restartPostgresAsStandby(ctx context.Context, stat
...
 
-	// Verify server is in recovery mode (standby)
-	inRecovery, err := pm.isInRecovery(ctx)
-	if err != nil {
-		return mterrors.Wrap(err, "failed to verify standby status")
+	// Reopen connections after restart
+	// Note: reopenConnections() just configures the connection pool manager - it doesn't
+	// validate that postgres is accepting connections. That's intentional here because
+	// postgres (in standby mode after pg_rewind) won't accept connections until we
+	// configure primary_conninfo, which happens in the next step (setPrimaryConnInfo).
```
rafael: Oh ok, this comment seems to explain it. But I guess still doesn't fully explain it, because `pg_rewind` was working before

go/multipooler/manager/rpc_consensus_test.go
```diff
@@ -1164,9 +1164,8 @@ func TestDemoteStalePrimary_UpdatesConsensusTerm(t *testing.T) {
 				m.AddQueryPattern("SELECT pg_is_in_recovery", mock.MakeQueryResult([]string{"pg_is_in_recovery"}, [][]any{{"t"}}))
 				m.AddQueryPatternOnce("ALTER SYSTEM SET synchronous_standby_names = ''", mock.MakeQueryResult(nil, nil))
 
-				// setPrimaryConnInfoLocked queries (after reopen)
-				m.AddQueryPatternOnce("ALTER SYSTEM SET primary_conninfo", mock.MakeQueryResult(nil, nil))
-				m.AddQueryPattern("SELECT pg_reload_conf", mock.MakeQueryResult([]string{"pg_reload_conf"}, [][]any{{true}}))
+				// Note: primary_conninfo is now written directly to postgresql.auto.conf file
```
rafael: if we end up removing this, feel like this comments can be removed

go/multipooler/manager/rpc_manager.go
```diff
@@ -1034,30 +1035,81 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
...
+	}
+
+	// Wait for postgres to accept connections and verify replication is active
+	// Use exponential backoff with 30-second timeout
+	pm.logger.InfoContext(ctx, "Waiting for PostgreSQL to accept connections")
+
+	waitCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	r := retry.New(10*time.Millisecond, 200*time.Millisecond)
```
rafael: I'm not sure we need this. It seems we need to move it above `resetSynchronousReplication` otherwise, that would happen before connection being ready.

go/multipooler/manager/rpc_manager.go
deepthi: This was never working. Remember that we were not able to get pg_rewind to work properly with a killed old primary back in early January because of the recovery.signal / standby.signal issue. Now that we have figured out how to get that working, it leads to the next problem: a pg_rewinded old primary not able to complete recovery.

go/cmd/pgctld/command/crash_recovery.go
deepthi: When this change is committed, we should remove the reference to the timeout from the PR description.

go/cmd/pgctld/command/crash_recovery.go
deepthi: Good point. We had also talked about pg_rewind potentially needing more time than 15-30 seconds.

go/multipooler/manager/manager.go
deepthi: See my earlier comment :)

go/cmd/pgctld/command/crash_recovery.go
deepthi: I have addressed this. Tests seem to pass.

go/cmd/pgctld/command/crash_recovery.go
dweitzman: If crash recovery takes more than 15 seconds, maybe we should let it keep going instead of killing the process?

I think crash recovery only saves progress as often as the checkpoint interval, so if we have a 15 or 30 second RPC timeout and apply that to the crash recovery process, if that's not enough time to checkpoint we might never be able to complete crash recovery

Maybe there's an argument to be made in the long term that crash recovery is not just an operation that we can quickly complete during a single RPC, but more like a state we can enter and check progress on?

go/cmd/pgctld/command/crash_recovery.go
dweitzman: Done, sort of. Let me know if what I did aligns with what you were thinking

go/cmd/pgctld/command/crash_recovery.go
dweitzman: Added code above the detect unknown and unspecified

go/cmd/pgctld/command/server.go
dweitzman: Done

go/cmd/pgctld/command/crash_recovery.go
deepthi: Yes, we should make a note of this. It should become part of postgres state management. Unfortunately, pg_rewind is not something pooler/pgctld can run autonomously because it requires knowing that the primary is that you should compare against.

go/multipooler/manager/rpc_manager.go
deepthi: Rafa pointed me to pg_rewind docs which actually say that _it_ will do the crash recovery. We don't know why that doesn't work with our invocation. Something to come back to and figure out.

PR #545, opened 2026-01-28
test(pg_regress): add e2e test to run pg_regress against multigres cluster

haritabh17:
## Summary
This PR adds end-to-end testing infrastructure that validates multigres compatibility by running PostgreSQL's regression test suite against a multigres cluster. The test clones PostgreSQL 17.6 source, builds it, spins up a 2-node multigres cluster with multigateway, and runs regression tests through the multigateway endpoint to verify compatibility.

## Running the Test
```
cd go/test/endtoend/pgregresstest
go test -v -timeout 60m
```
The first run clones and builds PostgreSQL (~200MB source, ~500MB builds), while subsequent runs reuse the cached source. The cache location can be customized via MULTIGRES_PG_CACHE_DIR environment variable.

> Note: This test is not yet integrated into the GitHub CI pipeline. It requires standard build tools (make, gcc, git)

## Test Plan
Run the test locally and verify it completes successfully. It will be added to CI pipeline in later iterations to have the compatibility report

### DIFF
--- a/go/test/endtoend/pgregresstest/README.md
+++ b/go/test/endtoend/pgregresstest/README.md
@@ -1,0 +1,212 @@
+# PostgreSQL Regression Tests
+
+This test package validates multigres compatibility by running PostgreSQL's official regression test suite against a multigres cluster.
+
+## Overview
+
+The test performs the following steps:
+
+1. **Checkout PostgreSQL source**: Clones PostgreSQL 17.6 (REL_17_6) from GitHub
+2. **Build with make**: Compiles PostgreSQL using traditional ./configure and make
+3. **Spin up multigres cluster**: Creates a 2-node cluster with multigateway
+4. **Run regression tests**: Executes specific PostgreSQL tests through multigateway using make installcheck-tests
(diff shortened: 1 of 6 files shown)


go/test/endtoend/pgregresstest/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+1. Clear build cache: `rm -rf /tmp/multigres_pg_cache/builds`
+2. Free up space (source ~200MB, builds ~500MB)
+3. Use custom cache location: `export MULTIGRES_PG_CACHE_DIR=~/multigres_cache`
+
+## Architecture
+
+### Component Flow
+
+<!-- markdownlint-disable MD013 -->
```
GuptaManan100, line 149: Why are we disabling the linting lol

haritabh17: 😁 Copied it from reference like - https://github.com/multigres/multigres/blob/main/docs/query_serving/connection_pooling.md?plain=1#L13  - to keep the linter happy

PR #546, opened 2026-01-28
feat(multipooler): dynamic fair share pool allocation

GuptaManan100:
### Description
Replace static per-user connection pool sizes with dynamic fair share allocation that automatically rebalances capacity across users based on actual demand. Previously, each user got a fixed number of connections configured at startup. When the number of users exceeded what `max_connections` could support at that fixed size, new users would error or wait indefinitely. Now, the multipooler distributes the global connection budget fairly using max-min fairness, rebalancing periodically in the background as users come and go.                                                                       
                                                                                                                                         
 ### Key changes 
- **Lock-free hot path:** Replace `sync.Mutex` with `atomic.Pointer` + copy-on-write for pool lookups.                                                                                                 
- **Demand tracking:** Per-user `DemandTracker` with sliding window buckets samples peak `requested` connections (not just borrowed) to capture true demand including queued waiters                                                                                                      
- **Fair share allocator:** Resource-agnostic max-min fairness algorithm with two instances - one for regular pools, one for reserved pools - ensuring independent allocation for different workload types                                                                                  
- **Non-blocking resize:** `SetCapacity` propagated through UserPool → regular/reserved → connpool layers without blocking on borrowed connections                                                                  
- **Background rebalancer:** Periodic goroutine that collects demand, runs fair share allocation, applies new capacities, and garbage collects inactive user pools                                                                                                                                
- **Configuration:** Six new flags (`--connpool-global-capacity`, `--connpool-reserved-ratio`, `--connpool-rebalance-interval`, `--connpool-demand-window`, `--connpool-demand-sample-interval`, `--connpool-inactive-timeout`); per-user capacity flags removed since the rebalancer manages capacity dynamically


docs/query_serving/connection_pooling.md
haritabh17: While at it, lets update it to reflect that this is used for auth related queries as well, or something that is more generic so that is clear that its not only used for term_backend.

docs/query_serving/connection_pooling.md
```diff
@@ -22,18 +26,25 @@ isolation while supporting Row-Level Security (RLS) policies.
...
+│  ┌────────────────────────────────┐  ┌────────────────────────────────┐          │
+│  │  FairShareAllocator (Regular)  │  │  FairShareAllocator (Reserved) │          │
+│  │  Capacity: 400 (80% of 500)    │  │  Capacity: 100 (20% of 500)    │          │
+│  └────────────────────────────────┘  └────────────────────────────────┘          │
+│                                                                                  │
+│  userPoolsSnapshot (atomic pointer - lock-free reads)                            │
 │  ┌──────────────────────────────────────────────────────────────────────────┐    │
 │  │  UserPool["alice"]                                                       │    │
 │  │  ┌──────────────────┐   ┌──────────────────────────┐                     │    │
 │  │  │   RegularPool    │   │      ReservedPool        │                     │    │
```
haritabh17, line 40: Given the usecase, should we call them -- pinned / unpinned conn pools? `Reserved` means like a separate pool that only roles with special privs can use... E.g. reserved pools in PG's context

docs/query_serving/connection_pooling.md
```diff
@@ -22,18 +26,25 @@ isolation while supporting Row-Level Security (RLS) policies.
...
 │  │  │   (user: alice)  │   │      (user: alice)       │                     │    │
+│  │  │   Capacity: 80   │   │      Capacity: 20        │                     │    │
```
haritabh17: With the FS allocator, `alice.regularPool` can potentially grow to [0, 400] and `alice.reservedPool` can grow to [0, 100] independently right? So at a point in time, it could be so that `regular = 15` and `reserved = 75`...

So lets depict that way then to clarify that:
1. The 80:20 split is at the global level and not per user level
2. Per user pools will scale independently under the global limits

docs/query_serving/connection_pooling.md
```diff
@@ -327,73 +537,77 @@ type PoolManager interface {
...
 
-**Admin Pool Flags:**
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--connpool-global-capacity` | 100 | Total PostgreSQL connections to manage |
+| `--connpool-reserved-ratio` | 0.2 | Fraction of global capacity for reserved pools |
+| `--connpool-rebalance-interval` | 10s | How often to run rebalancing |
+| `--connpool-demand-window` | 30s | Sliding window for peak demand tracking |
+| `--connpool-demand-sample-interval` | 100ms | How often to sample pool demand |
+| `--connpool-inactive-timeout` | 5m | Remove user pools after this inactivity |
```
haritabh17: Can they be re-configured by something like a file change followed by SIGHUP?

go/multipooler/connpoolmanager/demand_tracker.go
```diff
@@ -0,0 +1,180 @@
...
+	// The tracker keeps demand history for this duration.
+	// Number of buckets = WindowDuration / PollInterval.
+	WindowDuration time.Duration
+
+	// PollInterval is how often the rebalancer polls for demand.
+	// This determines bucket granularity.
+	PollInterval time.Duration
+
+	// SampleInterval is how often to sample the pool's requested count.
+	SampleInterval time.Duration
```
haritabh17: The description for these 2 look very similar, lets clarify a bit more to differentiate among them, maybe we can call `PollInterval` as `SingleBucketInterval`

go/multipooler/connpoolmanager/demand_tracker.go
```diff
@@ -0,0 +1,180 @@
...
+
+// NewDemandTracker creates a new DemandTracker with sliding window buckets.
+// It starts a background goroutine that samples demand at the configured interval.
+func NewDemandTracker(ctx context.Context, config *DemandTrackerConfig) *DemandTracker {
+	windowDuration := config.WindowDuration
+	pollInterval := config.PollInterval
+	sampleInterval := config.SampleInterval
+
+	// Calculate number of buckets
+	numBuckets := max(int(windowDuration/pollInterval), 1)
```
haritabh17: Should we also assert that we can collect atleast k samples (say 3) per bucket? If not then die with `InvalidParam` kind of error?

go/multipooler/connpoolmanager/demand_tracker.go
```diff
@@ -0,0 +1,180 @@
...
+		return
+	}
+
+	current := d.sampler()
+	bucketIdx := d.current.Load()
+
+	// Atomically update max: CAS loop
+	for {
+		old := d.buckets[bucketIdx].Load()
+		if current <= old {
```
haritabh17: Interesting! I imagined this would take running mean or some other aggregate... But if max also works well, then good; simpler.

go/multipooler/connpoolmanager/manager.go
```diff
@@ -116,81 +163,127 @@ func (m *Manager) buildClientConfig(user, password string) *client.Config {
...
+	if m.closed.Load() {
+		return nil, errors.New("manager is closed")
 	}
 
 	// Check max users limit
 	// TODO: Consider garbage collecting user pools after a period of inactivity
 	// to reduce the odds of reaching the user pool maximum.
-	if maxUsers := m.config.MaxUsers(); maxUsers > 0 && int64(len(m.userPools)) >= maxUsers {
+	currentPools := *pools
+	if maxUsers := m.config.MaxUsers(); maxUsers > 0 && int64(len(currentPools)) >= maxUsers {
```
haritabh17: With the FS allocator, do we really need `maxUsers` limit? Could we simply check if `initial_capacity` worth of connections are available and if not then trigger GC, otherwise allocate from pool?

go/multipooler/connpoolmanager/manager.go
```diff
@@ -116,81 +163,127 @@ func (m *Manager) buildClientConfig(user, password string) *client.Config {
...
 			IdleTimeout:     m.config.UserRegularIdleTimeout(),
 			MaxLifetime:     m.config.UserRegularMaxLifetime(),
 			ConnectionCount: m.metrics.RegularConnCount(),
 			Logger:          m.logger,
 		},
 		ReservedPoolConfig: &connpool.Config{
 			Name:            "reserved:" + user,
-			Capacity:        m.config.UserReservedCapacity(),
-			MaxIdleCount:    m.config.UserReservedMaxIdle(),
+			Capacity:        initialUserPoolCapacity,
```
haritabh17: nit: Does it make sense to have this heuristically be `min((initialUserPoolCapacity * adminPoolRatio), 1)` ??

go/multipooler/connpoolmanager/rebalancer.go
```diff
@@ -0,0 +1,163 @@
...
+		return
+	}
+
+	now := time.Now().UnixNano()
+	cutoff := now - inactiveTimeout.Nanoseconds()
+
+	// Find inactive pools
+	var inactiveUsers []string
+	for user, pool := range *pools {
+		if pool.LastActivity() < cutoff {
```
haritabh17, line 119: Refactoring idea / curiosity: would it be simpler to model the GC as rebalance-to-0... So something like have the `rebalance` function delete the entry if the demand is 0 (for config controller, we cleanup only if demand is 0, k consecutive times).

go/multipooler/connpoolmanager/rebalancer.go
```diff
@@ -0,0 +1,163 @@
...
+		m.logger.InfoContext(ctx, "garbage collected inactive user pool",
+			"user", user,
+			"inactive_duration", time.Duration(now-pool.LastActivity()))
+	}
+
+	if closedCount > 0 {
+		m.userPoolsSnapshot.Store(&newPools)
+		m.logger.InfoContext(ctx, "garbage collection complete",
+			"removed_pools", closedCount,
+			"remaining_pools", len(newPools))
```
haritabh17, line 168: Foreshadowing that adding some more logs here could be beneficial -- e.g. each pool capacity, total free conns. Which user pools were closed etc.

go/multipooler/connpoolmanager/user_pool.go
```diff
@@ -97,45 +112,102 @@ func NewUserPool(ctx context.Context, config *UserPoolConfig) *UserPool {
 		},
 	})
 
+	// Create demand trackers for rebalancer
+	// DemandTrackerConfig: WindowDuration, PollInterval (rebalance interval), SampleInterval, Sampler
+	var regularDemandTracker, reservedDemandTracker *DemandTracker
+	if config.DemandWindow > 0 && config.DemandSampleInterval > 0 && config.RebalanceInterval > 0 {
```
haritabh17: nit: this should perhaps be fatal? Also coming from within `NewDemandTracker` only and not here??

go/multipooler/connpoolmanager/user_pool.go
```diff
@@ -97,45 +112,102 @@ func NewUserPool(ctx context.Context, config *UserPoolConfig) *UserPool {
...
+	return p.lastActivity.Load()
+}
+
+// RegularDemand returns the peak demand for regular connections.
+// Returns 0 if demand tracking is not enabled.
+func (p *UserPool) RegularDemand() int64 {
+	if p.regularDemandTracker == nil {
+		return 0
+	}
+	return p.regularDemandTracker.GetPeakAndRotate()
```
haritabh17, line 190: Did not understand why this needs to rotate.

docs/query_serving/connection_pooling.md
GuptaManan100: Fixed

docs/query_serving/connection_pooling.md
GuptaManan100: Fixed

docs/query_serving/connection_pooling.md
GuptaManan100: No, these values are used to initialise components. Whenever we want them to change on SIGHUP, we'd need to write code to catch that and re-initialize the said components.

go/multipooler/connpoolmanager/demand_tracker.go
GuptaManan100: It's not about working well. You want to know the peak demand, so max is the only correct aggregate for the max-min fairness algorithm. 😄

go/multipooler/connpoolmanager/manager.go
GuptaManan100: Removed

go/multipooler/connpoolmanager/manager.go
GuptaManan100: Good point, fixed

go/multipooler/connpoolmanager/rebalancer.go
GuptaManan100: I thought that would be more complicated. This seems easier

go/multipooler/connpoolmanager/demand_tracker.go
GuptaManan100: I made a different refactor. I am making the pool keep track of the peek, instead of sampling over and over

go/multipooler/connpoolmanager/rebalancer.go
GuptaManan100: Done!

go/multipooler/connpoolmanager/user_pool.go
GuptaManan100: Added explanation in the demand tracker.

go/multipooler/connpoolmanager/user_pool.go
GuptaManan100: Fixed

go/multipooler/connpoolmanager/demand_tracker.go
GuptaManan100: I don't think that's necessary. Its valid to want to run with a single bucket, and the code already ensures, you don't have less than 1.

docs/query_serving/connection_pooling.md
GuptaManan100: I don't have a strong opinion. If we want to change this we should do a separate refactor only PR, makes it easier to review.

docs/query_serving/connection_pooling.md
```diff
@@ -52,25 +67,28 @@ isolation while supporting Row-Level Security (RLS) policies.
...
 - Small pool size (default: 5 connections)
 - Connects as PostgreSQL superuser (default: `postgres`)
 - Shared across all users
-- Used exclusively for `pg_terminate_backend()` operations
+- Used for privileged operations that user connections cannot perform
 
 **Use Cases:**
 
-- Canceling long-running queries
+- Authentication queries (e.g., password verification via `pg_shadow`)
```
haritabh17: nit: Lets use `pg_authid` as that is the currently used one

docs/query_serving/connection_pooling.md
```diff
@@ -327,73 +581,79 @@ type PoolManager interface {
...
+### Dynamic Allocation Flags
 
-**Admin Pool Flags:**
+These flags control how pool capacities are distributed across users:
+
+| Flag                            | Default | Description                                    |
+| ------------------------------- | ------- | ---------------------------------------------- |
+| `--connpool-global-capacity`    | 100     | Total PostgreSQL connections to manage         |
+| `--connpool-reserved-ratio`     | 0.2     | Fraction of global capacity for reserved pools |
+| `--connpool-rebalance-interval` | 10s     | How often to run rebalancing                   |
```
haritabh17, line 592: Might this be better to tie with some engine timeout GUC, e.g. `connection_timeout` or `query_timeout / 2` as default?

docs/query_serving/connection_pooling.md
```diff
@@ -327,73 +581,79 @@ type PoolManager interface {
...
-| `--connpool-user-regular-max-lifetime` | 1h      | Maximum lifetime before recycling               |
+Pool capacities are managed automatically by the rebalancer. New user pools start
+with an initial capacity of 10 connections and are adjusted within seconds based on
+demand. These flags control timeout behavior only:
 
-**Per-User Reserved Pool Flags:**
+| Flag                                          | Default | Description                                     |
+| --------------------------------------------- | ------- | ----------------------------------------------- |
+| `--connpool-user-regular-idle-timeout`        | 5m      | Idle timeout before closing regular connections |
+| `--connpool-user-regular-max-lifetime`        | 1h      | Maximum lifetime before recycling               |
```
haritabh17, line 621: IIUC, even if a connection is being actively reused again and again, it will get recycled in 1h, is this correct? Why is this needed? Somehow linked with starvation? I was guessing `idle-timeout` would be sufficient

docs/query_serving/connection_pooling.md
```diff
@@ -327,73 +581,79 @@ type PoolManager interface {
...
+| `--connpool-user-reserved-inactivity-timeout` | 30s     | Inactivity timeout for reserved connections     |
```
haritabh17, line 622: Having this separately settable could result in ambiguous UX, e.g. consider the engine timeouts here - https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-IDLE-IN-TRANSACTION-SESSION-TIMEOUT - now if user has set some value there but pooler decided to kill the underlying connection.

go/multipooler/connpoolmanager/fair_share_allocator.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	// If there's remaining capacity after all demands are met, distribute it evenly
+	// among all users for burst headroom. This ensures users can handle sudden
+	// traffic spikes without waiting for the next rebalance cycle.
+	if remaining > 0 {
+		extraPerUser := remaining / int64(numUsers)
+		if extraPerUser > 0 {
+			for user := range allocs {
+				allocs[user] += extraPerUser
```
haritabh17, line 137: If we allocate rest of all conns to existing users, does this mean a new user now has to wait for conns to get available from one of the existing pools at the next recycle interval??

docs/query_serving/connection_pooling.md
GuptaManan100: The rebalance interval is about how often to redistribute pool capacity across users based on demand - it's unrelated to query execution timeouts. A shorter interval (e.g., 5s) means faster adaptation to changing demand but more frequent allocator work; a longer interval (e.g., 30s) means more stable allocations but slower response to load changes.
Database timeouts like connection_timeout or query_timeout control how long individual operations can take, which is orthogonal to when we should re-examine the global capacity distribution. Tying them together would create confusing coupling - changing your query timeout shouldn't affect how often pool capacities are rebalanced.

docs/query_serving/connection_pooling.md
GuptaManan100: Correct - even actively used connections get recycled after max-lifetime. This serves several purposes:

1. Memory leak protection: PostgreSQL connections can accumulate memory over time (plan cache, temp tables, etc.). Periodic recycling prevents unbounded growth.
2. Credential rotation: If database credentials change, max-lifetime ensures old connections eventually cycle out.
3. Defensive hygiene: Long-lived connections can accumulate stale state or hit edge-case bugs in client libraries

These were the reasons to do the same in Vitess.

Idle-timeout only handles connections sitting unused in the pool. Max-lifetime handles the "busy" case where a connection is constantly in use. However we still only close it when the connection is returned to the pool.

docs/query_serving/connection_pooling.md
GuptaManan100: Good point. The pooler's `reserved-inactivity-timeout` and PostgreSQL's `idle_in_transaction_session_timeout` do overlap in purpose. We need to ensure the pooler's timeout is shorter than PostgreSQL's to prevent the backend from killing connections unexpectedly. If PostgreSQL kills first, we get a broken connection in the pool.

Options:
1. Document the relationship: Recommend setting reserved-inactivity-timeout < idle_in_transaction_session_timeout                      
2. Default to reading the GUC: Query the server's setting and default to 50-80% of that value                                          
3. Accept the current design: Let operators configure both independently since they may want pooler-side enforcement regardless of server settings                                                                                                                        

I lean toward (1) for now - documenting the best practice - and considering (2) as a future enhancement.

go/multipooler/connpoolmanager/fair_share_allocator.go
GuptaManan100: No, if the users don't need the connections, they won't open against postgres. We give a pool of 10 to a new user, so if the connections are free, they can open them. Essentially we overprovision in some sense, until the next cycle runs.

PR #547, opened 2026-01-28
feat(copycmd): add support for copy from stdin command

haritabh17:
## Summary
This PR implements PostgreSQL's COPY FROM STDIN command in multigateway, enabling efficient bulk data loading through bidirectional gRPC streaming.

## Implementation Overview
The COPY protocol is implemented as a three-phase state machine that coordinates between the client, multigateway, and multipooler:

**Phase 1 - Initiation:** When a client sends `COPY table FROM STDIN`, multigateway establishes a bidirectional gRPC stream with multipooler. The multipooler reserves a dedicated PostgreSQL connection and initiates the COPY operation, then sends back a `CopyInResponse` message through multigateway to the client, signaling readiness to receive data.

**Phase 2 - Data Streaming:** The client streams `CopyData` messages containing table rows. Multigateway forwards each chunk immediately to multipooler with zero buffering, and multipooler writes the data directly to PostgreSQL's COPY input stream. This continues until the client signals completion.

**Phase 3 - Finalization:** When the client sends `CopyDone`, multipooler closes the PostgreSQL COPY stream and returns the final `CommandComplete` message with the row count back to the client. If an error occurs at any point, a `CopyFail` message aborts the operation cleanly.

The implementation follows the existing planner-primitive architecture pattern. A new `CopyStatement` primitive handles `COPY FROM STDIN` using bidirectional streaming, while COPY FROM/TO file operations pass through directly to PostgreSQL. COPY FROM/TO PROGRAM is rejected for security (arbitrary command execution risk).

Connection state tracking was extended with a new `CopyState` struct that maintains the reserved connection ID, pooler ID, format information, and gRPC stream handle throughout the COPY operation. This ensures proper cleanup even when errors occur during any phase.

The wire protocol implementation adds methods for reading and writing COPY-specific messages (`CopyInResponse`, `CopyData`, `CopyDone`, `CopyFail`) on both the server side (client-facing) and client side (PostgreSQL-facing).

## Example Usage

```
# Connect to multigateway
psql -h localhost -p <port> -U postgres

# Create a test table
CREATE TABLE users (id INT, name TEXT, email TEXT);

# Load data with COPY FROM STDIN (TEXT format)
COPY users FROM STDIN;
1	Alice	alice@example.com
2	Bob	bob@example.com
3	Charlie	charlie@example.com
\.

# Verify the data
SELECT * FROM users;
-- Returns:
--  id |  name   |       email
-- ----+---------+-------------------
--   1 | Alice   | alice@example.com
--   2 | Bob     | bob@example.com
--   3 | Charlie | charlie@example.com

# CSV format with header
COPY users FROM STDIN CSV HEADER;
id,name,email
4,Diana,diana@example.com
5,Eve,eve@example.com
\.

# With custom delimiter
COPY users FROM STDIN CSV DELIMITER '|';
6|Frank|frank@example.com
7|Grace|grace@example.com
\.
```

### DIFF
--- a/go/services/multigateway/planner/variable_set_stmt.go
+++ b/go/services/multigateway/planner/variable_set_stmt.go
@@ -30,7 +30,6 @@ func (p *Planner) planVariableSetStmt(
 	stmt *ast.VariableSetStmt,
 	conn *server.Conn,
 ) (*engine.Plan, error) {
-	// TODO: implement when adding support for block transactions.
 	// Just pass through to PostgreSQL
 	if stmt.IsLocal {
 		p.logger.Debug("SET LOCAL detected, passing through",


go/multigateway/handler/connection_state.go
```diff
@@ -196,3 +210,98 @@ func (m *MultiGatewayConnectionState) RestoreSessionSettings(settings map[string
...
+
+	// Preserve existing stream if one was already set
+	existingStream := stream
+	if m.CopyState != nil && m.CopyState.Stream != nil {
+		existingStream = m.CopyState.Stream
+	}
+
+	m.CopyState = &CopyState{
+		Query:          query,
+		ReservedConnID: reservedConnID,
```
GuptaManan100: Shard state has the reserved conn id, we should use that. Don't need two fields to store the same information.

go/multigateway/planner/copy_stmt.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/multigres/multigres/go/parser/ast"
+)
+
+// planCopyStmt plans COPY commands.
+// Supports COPY FROM STDIN (streaming), COPY FROM/TO file (pass-through).
+// Rejects COPY FROM/TO PROGRAM for security. COPY TO STDOUT not yet supported.
+func (p *Planner) planCopyStmt(
+	sql string,
+	stmt *ast.CopyStmt,
+	conn *server.Conn,
```
GuptaManan100: conn seems to be unused

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -267,6 +269,242 @@ func (sc *ScatterConn) Describe(
...
 
+// --- COPY FROM STDIN methods ---
+
+// CopyInitiate initiates a COPY FROM STDIN operation using bidirectional streaming.
+// Returns: reservedConnID, poolerID, format, columnFormats, error
+func (sc *ScatterConn) CopyInitiate(
+	ctx context.Context,
+	conn *server.Conn,
+	queryStr string,
+	callback func(ctx context.Context, result *sqltypes.Result) error,
```
GuptaManan100, line 283: Look at the other function implementations, they all provide the tablet group, shard etc information so that scatterConn doesn't worry about it.
Right now we're sending defaults but for the future the information should flow through from the planner and executor into scatter conn

GuptaManan100: Once a target is built, we can use this code - 
```go
ss := state.GetMatchingShardState(target)
	// If we have a reserved connection, we have to ensure
	// we are routing the query to the pooler where we got the reserved
	// connection from. If a reparent happened, then we will get an error
	// back.
	if ss != nil && ss.ReservedConnectionId != 0 {
		eo.ReservedConnectionId = uint64(ss.ReservedConnectionId)
		qs, err = sc.gateway.QueryServiceByID(ctx, ss.PoolerID, target)
	}
```

to use a reserved connection id we already have

go/multigateway/handler/connection_state.go
GuptaManan100: The reason is that you might already have a reserved connection because a transaction is ongoing. So you should reuse that and not get another reserved connection.

go/services/multigateway/engine/engine.go
```diff
@@ -98,6 +99,44 @@ type IExecute interface {
...
+
+	// --- COPY FROM STDIN methods (called by CopyStatement primitive) ---
+
+	// CopyInitiate initiates a COPY FROM STDIN operation using bidirectional streaming.
+	// Returns: reservedConnID, poolerID, format, columnFormats, error
+	CopyInitiate(
+		ctx context.Context,
+		conn *server.Conn,
+		queryStr string,
+		callback func(ctx context.Context, result *sqltypes.Result) error,
```
GuptaManan100, line 116: All of these will need to taken in the target and shard as pointed out ☝️, and the reserved connection id they'd store internally into the state without needing to return it.

go/multigateway/engine/copy_statement.go
```diff
@@ -0,0 +1,206 @@
...
+
+// abort sends CopyFail to pooler.
+func (c *CopyStatement) abort(
+	ctx context.Context,
+	exec IExecute,
+	conn *server.Conn,
+	state *handler.MultiGatewayConnectionState,
+) {
+	reservedConnID, poolerID := state.GetCopyReservedConn()
+	_ = exec.CopyAbort(ctx, conn, reservedConnID, poolerID)
```
GuptaManan100: None of these should require sending the reservation id, scatter conn will find if there is a reserved conn based on the state in the shardState

proto/multipoolerservice.proto
```diff
@@ -51,6 +51,11 @@ service MultiPoolerService {
   // an up-to-date cache of all password hashes and by real-time updates from multipooler
   // when any credentials change.
   rpc GetAuthCredentials(GetAuthCredentialsRequest) returns (GetAuthCredentialsResponse);
+
+  // BidirectionalExecute handles bidirectional streaming operations (e.g., COPY commands).
+  // The gateway sends the initial command and then streams data/messages.
+  // The pooler responds with protocol-specific messages and final result.
+  rpc BidirectionalExecute(stream BidirectionalExecuteRequest) returns (stream BidirectionalExecuteResponse);
```
GuptaManan100: Added to the wrong service 😄. We should add it to the queryservice

go/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -61,6 +62,9 @@ type Gateway interface {
 
 	// QueryServiceByID returns a QueryService
 	QueryServiceByID(ctx context.Context, id *clustermetadatapb.ID, target *query.Target) (queryservice.QueryService, error)
+
+	// GetBidirectionalExecuteStream creates a bidirectional stream for commands requiring streaming (e.g., COPY)
+	GetBidirectionalExecuteStream(ctx context.Context, target *query.Target) (multipoolerservice.MultiPoolerService_BidirectionalExecuteClient, error)
```
GuptaManan100: Why do we need a new client? Shouldn't we just add another RPC to the queryservice interface already present?
We can call it `CopyFromFileStream` or something like that which is a bidirectional rpc.

go/multipooler/grpcpoolerservice/service.go
```diff
@@ -200,3 +201,142 @@ func (s *poolerService) PortalStreamExecute(req *multipoolerpb.PortalStreamExecu
...
+	exec, err := s.pooler.Executor()
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "executor not initialized: %v", err)
+	}
+
+	// Cast to concrete executor type to access COPY methods
+	poolerExecutor, ok := exec.(*executor.Executor)
+	if !ok {
+		return status.Error(codes.Internal, "executor is not of expected type")
+	}
```
GuptaManan100: This might not be required if its on the interface?

go/multipooler/executor/executor.go
```diff
@@ -426,6 +426,216 @@ func (e *Executor) ensurePrepared(ctx context.Context, conn *regular.Conn, stmt
 	return canonicalName, nil
 }
 
+// HandleCopyInitiate handles the INITIATE phase of a COPY operation
+// It sends the COPY command to PostgreSQL and reads the CopyInResponse or CopyOutResponse
+// Returns: reservedConnID, poolerID, format, error
+// Note: This is a stub - full implementation pending access to raw PostgreSQL client methods
```
GuptaManan100: Is it really?

go/multipooler/pools/regular/regular_conn.go
```diff
@@ -389,3 +389,9 @@ func execWithContextCancel[T any](c *Conn, ctx context.Context, op func() (T, er
 		return res.val, res.err
 	}
 }
+
+// ClientConn returns the underlying PostgreSQL client connection.
+// This is used for low-level protocol operations like COPY.
+func (c *Conn) ClientConn() *client.Conn {
+	return c.conn
+}
```
GuptaManan100: This is generally not a good idea. Just export the underlying functions you want to call. Keeps the abstraction intact.

go/multigateway/planner/copy_stmt.go
haritabh17: Right, removed.

go/multipooler/executor/executor.go
haritabh17: Ha! pending delete of WIP comment 😄 , removed

go/multigateway/engine/copy_statement.go
haritabh17: Ack, updated

go/services/multigateway/engine/engine.go
haritabh17: Ack, updated

go/multigateway/handler/connection_state.go
haritabh17: Right, removed. Makes the copy state much lighter now. Pretty much everything is already there in state.

go/multigateway/poolergateway/pooler_gateway.go
haritabh17: Ack, thats better, updated

go/services/multigateway/scatterconn/scatter_conn.go
haritabh17: Ack, updated

go/multipooler/pools/regular/regular_conn.go
haritabh17: Ack, updated

go/multipooler/grpcpoolerservice/service.go
haritabh17: Yupp, updated.

proto/multipoolerservice.proto
haritabh17: Did not understand this 😕

go/common/protoutil/protoutil.go
```diff
@@ -62,7 +62,7 @@ func TargetEquals(t1, t2 *query.Target) bool {
 	if t1 == nil && t2 == nil {
 		return true
 	}
-	if t1 != nil || t2 != nil {
+	if (t1 == nil) != (t2 == nil) {
```
GuptaManan100: Wow, I don't know how I messed this up before 😮. We can just do a `t1 == nil || t2 == nil` check. Cause we've already checked both aren't nil together.

go/services/multigateway/handler/connection_state.go
```diff
@@ -129,6 +129,24 @@ func (m *MultiGatewayConnectionState) StoreReservedConnection(target *query.Targ
 	m.ShardStates = append(m.ShardStates, ss)
 }
 
+// ClearReservedConnection removes a reserved connection for a given target.
+// This should be called when a reserved connection is released (e.g., after COPY completes).
+func (m *MultiGatewayConnectionState) ClearReservedConnection(target *query.Target) {
```
GuptaManan100, line 134: Perfect 👍

go/multigateway/planner/variable_set_stmt.go
```diff
@@ -30,7 +30,7 @@ func (p *Planner) planVariableSetStmt(
 	stmt *ast.VariableSetStmt,
 	conn *server.Conn,
 ) (*engine.Plan, error) {
-	// SET LOCAL is transaction-scoped, not tracked in Phase 1
+	// TODO: implement when adding support for block transactions.
```
GuptaManan100: I don't know if we'll need to do anything even then 😆. Cause they're in the transaction scope, they'll just be gone when the transaction aborts 🤔

go/multigateway/poolergateway/grpc_query_service.go
```diff
@@ -43,6 +44,12 @@ type grpcQueryService struct {
 
 	// poolerID for logging
 	poolerID string
+
+	// copyStreamsMu protects copyStreams map
+	copyStreamsMu sync.Mutex
+
+	// copyStreams maps reserved connection IDs to active bidirectional streams for COPY operations
+	copyStreams map[uint64]multipoolerservice.MultiPoolerService_BidirectionalExecuteClient
```
GuptaManan100: Just a question, I think we're already doing it, but confirming, that we cleanup the stream when the copy ends (both successfully or on an error)

proto/multipoolerservice.proto
GuptaManan100: Forget this, I put this comment in the wrong place. You've already fixed it.

GuptaManan100: Are we intending to use the BiDirectionalExecute for things other than Copy? If not, then we should just call it CopyBiDiExecute, or something like that

go/common/protoutil/protoutil.go
haritabh17: Ack, updated

go/multigateway/planner/variable_set_stmt.go
haritabh17: That is true. Let me remove this.

go/multigateway/poolergateway/grpc_query_service.go
haritabh17: Found a place where it was not closed -- i.e. in the `CopyReady` function if an error occurs after the gRPC stream is created but before it's added to the map, then the gRPC stream is opened but never closed:
```
  stream, err := g.client.BidirectionalExecute(ctx)  // line 293 - stream created
  // ... any error occurring here
  g.copyStreams[resp.ReservedConnectionId] = stream  // line 341 - only added on success
```
These error paths don't close the stream or add it to the map for later cleanup. The gRPC stream will eventually be garbage collected, but it's not being explicitly closed on error, which could hold resources longer than necessary. 

Fixing and adding test for it.

proto/multipoolerservice.proto
haritabh17: Ack, renaming, can make generic when needed

PR #551, opened 2026-01-28
fix: remove unnecessary comments

dweitzman:


go/multipooler/grpcpoolerservice/service.go
```diff
@@ -103,19 +103,15 @@ func (s *poolerService) GetAuthCredentials(ctx context.Context, req *multipooler
...
 	}
 
-	// Get the pool manager from the pooler.
 	poolManager := s.pooler.PoolManager()
 	if poolManager == nil {
 		return nil, status.Error(codes.Unavailable, "pool manager not initialized")
 	}
 
-	// Get an admin connection. This works even before the executor is initialized,
-	// which is critical for authentication during bootstrap.
```
deepthi, line 118: This one seems useful - at least the part where we want it to work before the executor is initialized.

haritabh17: Agree. `bootstrap` was the confusing bit for me as I mixed it up with the bootstrap phase of PG (initdb), but here bootstrap means _any_ connection establishment phase... So we can potentially keep the comment and just reword it to clarify that part.

dweitzman: Changed the comment to this:

```
	// An admin connection:
	// - has permission to read password hashes
	// - also avoids a chicken-egg scenario of needing to create and use a role-specific connection
	//   to figure out if the caller should have access to that role-specific connection.
```

PR #556, opened 2026-01-30
feat(multipooler): Refactor demote behavior

rafael:
# Desc

Per our discussion, this change makes sure that during an emergency Demote, when a multipooler receives a request, it stops itself. It is expected that this postgres will have a timeline divergence and we will run `pg_rewind`. The core of this PR is making that change.

To make code clearer, `Demote` is renamed to `EmergencyDemote`. That reflects that current reality. We will eventually add a `GracefulFailover` that will be the flow we use when demoting in a graceful way. 

# Tests
- It added relevant integration test to cover the scenario where a primary receives a BeginTerm and it correctly demotes itself. We assert that multi-orch performs the failover and get the dead primary rewinded correctly.

### DIFF
--- a/go/multipooler/manager/rpc_manager.go
+++ b/go/multipooler/manager/rpc_manager.go
@@ -796,6 +796,8 @@ func (pm *MultiPoolerManager) EmergencyDemote(ctx context.Context, consensusTerm
 	defer pm.actionLock.Release(ctx)
 
 	// Permanently disable monitoring to prevent accidental postgres restart after emergency demotion
+	// TODO: This is not a long-term solution. The disableMonitor() approach will likely be
+	// replaced with proper state management in follow-up work to PR #550.
 	pm.disableMonitorInternal()
 
 	// Validate the term but DON'T update yet. We only update the term AFTER
--- a/go/test/endtoend/multipooler/main_test.go
+++ b/go/test/endtoend/multipooler/main_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/multigres/multigres/go/test/endtoend/shardsetup"
 
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
 )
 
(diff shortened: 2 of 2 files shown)


go/test/endtoend/shardsetup/shardsetup_test.go
```diff
@@ -100,80 +100,6 @@ func TestShardSetup_ThreeNodeCluster(t *testing.T) {
...
-
-	// Verify initial state - primary is not in recovery
-	primaryClient := setup.NewPrimaryClient(t)
-	inRecovery, err := QueryStringValue(ctx, primaryClient.Pooler, "SELECT pg_is_in_recovery()")
-	require.NoError(t, err)
-	require.Equal(t, "f", inRecovery, "primary should start as primary (not in recovery)")
-	primaryClient.Close()
-
-	// Demote the primary
-	setup.DemotePrimary(t)
```
rafael, line 120: The whole purpose of this test was to test this `method` in case we wanted it to use it other tests. This is not being used anywhere and the current semantic (emergency demote), shouldn't be used as part of other tests. 

For now, simpler to just remove.

sougou (approved):

go/multipooler/manager/manager.go
```diff
@@ -976,6 +976,30 @@ func (pm *MultiPoolerManager) runCheckpointAsync(ctx context.Context) chan error
 	return checkpointDone
 }
 
+// stopPostgresForEmergencyDemote stops PostgreSQL during emergency demotion without restarting.
+// This is used when a primary needs to step down immediately during consensus term changes.
+// The node will be left in a stopped state and will require pg_rewind to rejoin the cluster.
+func (pm *MultiPoolerManager) stopPostgresForEmergencyDemote(ctx context.Context, state *demotionState) error {
```
dweitzman, line 982: What will stop postgres monitor from restarting the pooler? Seems like we'll need to put something in memory so postgres monitor knows not to try restarting and also persist it to disk so restarting multipooler won't lose the in-memory state?

go/test/endtoend/multipooler/rpc_consensus_test.go
```diff
@@ -211,6 +212,20 @@ func TestConsensus_BeginTerm(t *testing.T) {
...
+		defer pgctldClient.Close()
+
+		require.Eventually(t, func() bool {
+			statusResp, err := pgctldClient.Status(context.Background(), &pgctldpb.StatusRequest{})
+			return err == nil && statusResp.Status == pgctldpb.ServerStatus_STOPPED
+		}, 10*time.Second, 1*time.Second, "PostgreSQL should be stopped after emergency demotion from BeginTerm on pooler: %s", setup.PrimaryMultipooler.Name)
+		t.Log("Confirmed: PostgreSQL stopped after emergency demotion")
+
+		// BeginTerm on primary triggers emergency demotion - restore to working state
+		restoreAfterEmergencyDemotion(t, setup, setup.PrimaryPgctld, setup.PrimaryMultipooler, setup.PrimaryMultipooler.Name)
```
dweitzman: Wondering if this should be in a t.Cleanup() earlier so it'll still run even if the Eventually() times out

go/multipooler/manager/manager.go
rafael: Good catch. Today Postgres monitor would try to restart unless we explicitly suppress it. I made one incremental change: emergency demote now calls disableMonitor so the monitor loop stops attempting restarts in process.

Persisting that intent across a multipooler restart is still missing, and I propose we leave it out of scope for this PR. It should be addressed as part of the broader postgres monitor and pooler state management improvements.

dweitzman: I'm ok with putting real state into a future PR. Maybe put a TODO on the disableMonitor() that it's not a long-term solution?

The `disableMonitor()` method will likely be removed in a follow-up to https://github.com/multigres/multigres/pull/550, so I'll have to figure out something then if we haven't dealt with it already

PR #565, opened 2026-02-01
feat(multipooler): adds primary term to consensus data

rafael:
# Add `primary_term` tracking to consensus state

## Problem

The current consensus implementation conflates **term acceptance** (agreeing to participate in a term) with **primary authority** (actually being the PostgreSQL primary for a term). This creates an edge case where a node can accept a newer term while still unwinding its previous primary role, leading to ambiguity about which term a node is *actually* primary for.

### Edge case scenario

1. Term 5: **Node A** is the PostgreSQL primary; **Node B** is a replica.
2. A coordinator starts an election for term 6.
3. **Node A** accepts term 6 and persists `term_number = 6`, but it is still PostgreSQL primary for term 5 while it unwinds demotion.
4. The election succeeds elsewhere and **Node B** is promoted for term 6. Node B reports `role = primary` and `term_number = 6`.
5. For a period of time, both nodes can report `role = primary` with `term_number = 6`, even though only Node B is the true primary for term 6.
6. Without `primary_term`, consumers cannot distinguish:
   * **Node A**: primary role is leftover from term 5
   * **Node B**: primary authority for term 6

This does not break failover today because discovery chooses the most advanced candidate based on WAL position (LSN). However, it can break other operations such as `DemoteStalePrimary`: if two multipoolers both claim to be primary, we lack a reliable way to determine which one is the true, most advanced primary.

## Solution
Track two distinct term numbers per node:
1. **Consensus term (`term_number`)**: the highest term the node has agreed to participate in (term acceptance, voting).
2. **Primary term (`primary_term`)**: the specific term for which the node actually became primary and is executing primary operations.

This design was anticipated in our [Generalization of Consensus series - Part 7: The Paxos way](https://multigres.com/blog/generalized-consensus-part7#the-paxos-way), where we discussed Paxos's separation of "proposal number" and "value's term." Just as Paxos tracks both the term a node has agreed to participate in and the term under which a value was actually accepted, we separate term acceptance from primary authority. As we noted in the blog: *"That's why we need to add a new variable to track the value's term"* and *"The value's term must be persisted for the same reasons why the node's term must be persisted."*

When a node reports its status, it must include:
* Whether it is currently a primary (role)
* The specific primary term it is primary for (`primary_term`)

**Example:** If a node has accepted term 6 but is still primary for term 5, it reports:
* `term_number = 6` (accepted participation in term 6)
* `primary_term = 5` (still executing primary authority for term 5)
* `role = primary`

This distinction allows consumers to determine which node holds true primary authority for a given term, resolving the ambiguity during role transitions.
## Testing

* Relevant tests were added to both unit and e2e tests.

No migration needed. This is a new field that defaults to `0` (meaning “never been primary”), which is the correct state for all existing nodes.

## Future work

This change enables future improvements to `DemoteStalePrimary` and other operations that need to distinguish between:

* Poolers that have accepted a term (voting participants)
* Poolers that are actually primary for a term (authority holders)

### DIFF
--- a/go/multipooler/manager/manager.go
+++ b/go/multipooler/manager/manager.go
@@ -226,6 +226,9 @@ func NewMultiPoolerManagerWithTimeout(logger *slog.Logger, multiPooler *clusterm
 		cancel: cancel,
 	}
 
+	// Consensus state is always available; it will be loaded when needed.
+	pm.consensusState = NewConsensusState(pm.multipooler.PoolerDir, pm.serviceID)
+
 	// Create the query service controller with the pool manager
 	pm.qsc = poolerserver.NewQueryPoolerServer(logger, connPoolMgr)
 
--- a/go/multipooler/manager/rpc_initialization.go
+++ b/go/multipooler/manager/rpc_initialization.go
@@ -136,7 +136,7 @@ func (pm *MultiPoolerManager) InitializeEmptyPrimary(ctx context.Context, req *m
 		return nil, mterrors.Wrap(err, "failed to set pooler type")
 	}
 
-	// Record primary term during propagation
+	// Set primary term during bootstrap initialization
 	if pm.consensusState != nil {
 		if err := pm.consensusState.SetPrimaryTerm(ctx, req.ConsensusTerm, false /* force */); err != nil {
 			return nil, mterrors.Wrap(err, "failed to set primary term")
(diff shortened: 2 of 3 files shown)


go/multipooler/manager/rpc_consensus.go
```diff
@@ -233,7 +241,7 @@ func (pm *MultiPoolerManager) ConsensusStatus(ctx context.Context, req *consensu
 	walPosition := &consensusdatapb.WALPosition{
 		Timestamp: timestamppb.New(time.Now()),
 	}
-	role := "replica"
+	role := "unknown"
```
rafael, line 244: This is not related to this PR, but I noticed that we were defaulting to replica when we can't contact postgres. That seems misleading. For now updated it to unknown. This should probably be a proto enum. But we can follow up on that.

PR #566, opened 2026-02-01
feat(consensus): BeginTerm explicit actions

rafael:
## Description
The following PR adds the foundation to fix some gaps we have in our logic when wiring new replicas. Today, when adding a fresh replica to a cluster, MultiOrch adds it to the cluster, but there is no guarantee that all other MultiOrch instances know about the existence of this new replica. This creates a data safety edge case:

1. A replica may have already received WAL that includes committed writes
2. Other MultiOrch coordinators have not yet discovered the replica  
3. If a failover happens, MultiOrch could elect a primary without considering that replica
4. This risks data loss if a commit only reached this new replica as the acker for the primary

## Root Cause

There is a way to solve this problem within our generalized consensus approach. The limitation comes from our BeginTerm implementation performing multiple actions at the same time. This implementation decision is efficient for electing new primaries, but it creates limitations for the edge case above.

### The Gap: No Way to Recruit for Other Operations

The limitation is that BeginTerm *only* supports failover. We have no way to recruit nodes for other consensus operations.

From our consensus generalization ([Part 6](https://multigres.com/blog/generalized-consensus-part6)):

> "The coordinator does not need to specify a reason at the time of recruitment. 
> It can choose what to do with the recruited nodes at a later time."

**The consensus protocol allows:**
```
1. Recruit nodes into a term
2. Perform some operation with those recruited nodes (now or later)
```

**Current implementation only allows:**
```
1. Recruit nodes + immediately perform failover operation
```

This gap prevents us from implementing operations that need to:
- Recruit nodes for discovery
- Make decisions durable across the cohort
- Execute those decisions as separate steps

## Solution Overview

This PR is **part 1** of a two-part solution to enable safe replica discovery and wiring:

### Part 1: BeginTerm Actions (This PR)

Add an `action` parameter to decouple recruitment from specific operations:
```
BeginTerm(term, action=REVOKE_PREVIOUS)  // Current behavior: recruit + failover (preserved!)
BeginTerm(term, action=NO_ACTION)         // New: recruit only, action comes later
```

**Benefits:**
- Preserves efficient single-RPC failover path (REVOKE_PREVIOUS)
- Enables recruitment for other operations (NO_ACTION)
- Makes protocol intentions explicit in code
- Aligns with consensus theory: recruitment is separate from action

### Part 2: Fix Replication Safety (Future Work)

Use NO_ACTION recruitment to safely wire replicas:
```
1. MultiOrch: BeginTerm(term, NO_ACTION)         // Recruit replica for discovery
2. MultiOrch: [Win term]
3. MultiOrch: [Persist replica in standby list]  // Make decision durable in cohort
4. MultiOrch: Wire replication                   // Execute decision safely
```

Once the decision to add a new replica to the cohort is made durable, we can ensure other MultiOrch coordinators will discover this during a failover. Stay tuned for how we're actually going to implement that, as it's outside the scope of this PR. But we have a plan!

## Testing

- Added relevant unit and integration tests

### DIFF
--- a/go/test/endtoend/multipooler/main_test.go
+++ b/go/test/endtoend/multipooler/main_test.go
@@ -97,7 +97,7 @@ func restoreAfterEmergencyDemotion(t *testing.T, setup *MultipoolerTestSetup, pg
 	// Step 3: Restart multipooler so it picks up the topology change
 	t.Logf("Restarting multipooler %s to pick up topology change...", multipoolerName)
 	multipooler.Stop()
-	err = multipooler.Start(t)
+	err = multipooler.Start(restartCtx, t)
 	require.NoError(t, err, "Multipooler should restart successfully: %s", multipoolerName)
 
 	// Wait for manager to be ready


go/multipooler/manager/rpc_consensus.go
```diff
@@ -71,141 +121,104 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
-				queryErr = executor.ScanSingleRow(result, &lastMsgReceiptTime)
-				if queryErr != nil {
-					// No WAL receiver (disconnected standby) - this is EXPECTED during failover
-					// when the primary just died. Don't reject - proceed with acceptance.
-					// The standby's data may still be recent even without an active WAL receiver.
-					pm.logger.InfoContext(ctx, "Standby has no WAL receiver, proceeding with term acceptance",
-						"term", req.Term)
-				}
-			}
-			if lastMsgReceiptTime != nil {
```
rafael, line 157: I simplified all this code as I thought this was adding complexity and unpredictable behavior. For instance, if the instance was lagging we were not accepting, but if there was no wal receiver, we accept the term.

go/test/endtoend/multipooler/rpc_manager_api_test.go
```diff
@@ -106,21 +108,13 @@ func TestPrimaryStatus(t *testing.T) {
 	waitForManagerReady(t, setup, setup.StandbyMultipooler)
 
 	// Create shared clients for all subtests
-	primaryConn, err := grpc.NewClient(
-		fmt.Sprintf("localhost:%d", setup.PrimaryMultipooler.GrpcPort),
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
-	)
+	primaryClient, err := shardsetup.NewMultipoolerClient(setup.PrimaryMultipooler.GrpcPort)
```
rafael, line 111: this is just cleanup, to use the client from the shardsetup.

go/multipooler/executor/internal_user_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Supabase, Inc.
+// Copyright 2026 Supabase, Inc.
```
dweitzman, line 1: Was this a bug in the copyright updater? It's only supposed to fix the year on newly-added fixed

go/multipooler/manager/rpc_consensus.go
```diff
@@ -48,21 +50,69 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
+				"error", err)
+			// Return current term with accepted=false
+			pm.mu.Lock()
+			cs := pm.consensusState
+			pm.mu.Unlock()
+			currentTerm, termErr := cs.GetCurrentTermNumber(ctx)
+			if termErr != nil {
+				currentTerm = 0
+			}
+			return &consensusdatapb.BeginTermResponse{
```
dweitzman, line 104: Not a blocking comment, just a thought: I wonder if we'll eventually want to return a richer error code or message explaining why a begin term wasn't accepted

go/multipooler/manager/rpc_consensus.go
```diff
@@ -71,141 +121,104 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
-		if demoteErr != nil {
-			// Demotion failed - do NOT accept the term
-			// Return accepted=false so coordinator knows this node couldn't safely step down
-			pm.logger.ErrorContext(ctx, "Demotion failed, rejecting term acceptance",
-				"error", demoteErr,
-				"term", req.Term)
-			// TODO: we should attempt to undo the demote.
-			// UndoDemote has not been implemented yet
-			return response, nil
+		demoteResp, err := pm.emergencyDemoteLocked(ctx, term, drainTimeout)
```
dweitzman, line 202: Will a graceful / non-emergency demote go through this same codepath in the future? Wondering if this should have a TODO mentioning that non-emergency demote still needs to be implemented

go/multipooler/manager/rpc_consensus.go
```diff
@@ -71,141 +121,104 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
...
-					if req.Term > currentTerm {
-						if err := pm.validateAndUpdateTerm(ctx, req.Term, false); err != nil {
-							return nil, fmt.Errorf("failed to update term: %w", err)
-						}
-						response.Term = req.Term
-					}
-					return response, nil
-				}
-			}
+		// Revoke standby: pause replication, then check if caught up
```
dweitzman: The comment says "check if caught up" but I don't see that in the code

go/multipooler/manager/rpc_consensus_test.go
```diff
@@ -245,20 +268,25 @@ func TestBeginTerm(t *testing.T) {
...
+				m.AddQueryPatternOnce("SELECT pg_reload_conf", mock.MakeQueryResult(nil, nil))
 			},
 			expectedAccepted:                    true,
 			expectedTerm:                        10,
 			expectedAcceptedTermFromCoordinator: "new-candidate",
 			description:                         "Primary should accept term after successful demotion (idempotent case - already demoted)",
 		},
 		{
-			name: "StandbyAcceptsTermWhenNoWALReceiver",
+			name:   "StandbyAcceptsTermWhenNoWALReceiver",
```
dweitzman: Feels like the name of this test case isn't accurate anymore?

go/multipooler/manager/rpc_consensus_test.go
```diff
@@ -312,6 +343,113 @@ func TestBeginTerm(t *testing.T) {
...
+				// DO NOT add SELECT 1 expectation - let it fail
+			},
+			expectedError:                       false,
+			expectedAccepted:                    false, // TODO(FUTURE): Once we separate voting term from primary term, this should be TRUE
+			expectedTerm:                        5,     // Should remain at current term since we rejected
+			expectedAcceptedTermFromCoordinator: "",    // Should not accept new coordinator
+			description:                         "Primary with postgres down should accept term once we separate voting term from primary term",
+		},
+		{
+			name:   "StandbyWithPostgresDownRejectsTerm_TemporaryBehavior",
```
dweitzman: This seems like a duplicate of the test case above?

go/multipooler/executor/internal_user_test.go
rafael: Good catch. I think so. I don't recall updating this manually

go/multipooler/manager/rpc_consensus.go
rafael: I added a comment. I think graceful should be a complete different codepath.

PR #568, opened 2026-02-02
feat(claude): add local cluster skill

rafael:
# Desc
- After testing this locally, this seems to be working nicely. Bringing it back


.claude/skills/mt-local-cluster/SKILL.md
dweitzman, line 1: Would `SKILL.md` be a more conventional name for this than `Skill.md`?

dweitzman (approved):

PR #569, opened 2026-02-02
feat(multiorch): Start using periodic runner in multiorch

rafael:
# Desc
- Starts migrating multiorch runners, to use the generic PeriodicRunner implementation

# Test
- We have enough coverage for this. So I think we don't need new tests for this

### DIFF
--- a/go/services/multiorch/recovery/engine.go
+++ b/go/services/multiorch/recovery/engine.go
@@ -245,13 +245,13 @@ type Engine struct {
 	// Config reloader for dynamic updates (only shardWatchTargets is dynamic)
 	reloadConfig func() []string
 
-	// Periodic runner for bookkeeping tasks
+	// Periodic runners for background tasks
 	bookkeepingRunner *timer.PeriodicRunner
+	recoveryRunner    *timer.PeriodicRunner
 
 	// Goroutine management - prevent pile-up of concurrent operations
 	metadataRefreshInProgress atomic.Bool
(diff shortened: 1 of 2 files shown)


go/services/multiorch/recovery/engine.go
```diff
@@ -364,7 +369,22 @@ func (re *Engine) Start() error {
...
+	// Start bookkeeping runner
+	re.bookkeepingRunner.Start(func(ctx context.Context) {
+		re.runBookkeeping()
+	}, nil)
+
+	// Start recovery runner with dynamic interval support
+	re.recoveryRunner.Start(func(ctx context.Context) {
+		// Check if interval changed (dynamic config)
+		newInterval := re.config.GetRecoveryCycleInterval()
+		if re.recoveryRunner.UpdateInterval(newInterval) {
```
dweitzman, line 381: Super nit: might make more sense to check and update the interval after calling performRecoveryCycle(), to react faster to changes?

go/services/multiorch/recovery/engine.go
```diff
@@ -364,7 +369,22 @@ func (re *Engine) Start() error {
...
+			re.logger.InfoContext(re.shutdownCtx, "recovery cycle interval changed", "interval", newInterval)
+		}
+		re.performRecoveryCycle()
```
dweitzman, line 384: Nit: could propagate the context through here instead of having performRecoveryCycle() assume it should use the shutdown context?

PR #572, opened 2026-02-02
feat(multipooler): periodic runner refactor part II

rafael:
# Desc
* Follow up from: https://github.com/multigres/multigres/pulls?q=is:pr+author:rafael+is:closed
* It migrates the remainder of loops to use the periodic runner.

### DIFF
--- a/go/services/multiorch/recovery/engine.go
+++ b/go/services/multiorch/recovery/engine.go
@@ -19,7 +19,6 @@ import (
 	"log/slog"
 	"slices"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/multigres/multigres/go/common/rpcclient"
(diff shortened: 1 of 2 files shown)


go/services/multiorch/recovery/engine.go
```diff
@@ -405,33 +404,12 @@ func (re *Engine) Stop() {
 	re.cancel()
 	re.bookkeepingRunner.Stop()
 	re.recoveryRunner.Stop()
+	re.metadataRefreshRunner.Stop()
+	re.healthCheckQueueRunner.Stop()
 	re.wg.Wait()
```
dweitzman, line 409: Is this wait group still needed? I think all the wait group tasks might be gone now

rafael: We still need it for the queue workers. Once we perform that refactor, I think we can fully kill it

PR #575, opened 2026-02-03
feat(notice): show notices while running commands

### LINKED ISSUE #573: Missing NOTICE messages
Commands succeeds without emitting the NOTICE messages that standard PostgreSQL produces, e.g. when merging inherited column definitions.

Steps to Reproduce

```
CREATE TABLE person (
    name        text,
    age         int4,
    location    point
);

CREATE TABLE emp (
    salary      int4,
    manager     name
) INHERITS (person);

CREATE TABLE student (
    gpa         float8
) INHERITS (person);

CREATE TABLE stud_emp (
    percent     int4
) INHERITS (emp, student);
```

Expected Behavior (standard PostgreSQL)
```
NOTICE:  merging multiple inherited definitions of column "name"
NOTICE:  merging multiple inherited definitions of column "age"
NOTICE:  merging multiple inherited definitions of column "location"
CREATE TABLE
```

Actual Behavior (multigres)
```
CREATE TABLE
```

The NOTICE messages are silently dropped/ignored.

haritabh17:
This PR adds support for forwarding PostgreSQL NOTICE messages to clients during query execution.                                                
                                                                                                                                                   
  Previously, informational messages like NOTICE: merging multiple inherited definitions of column "name" were silently dropped when queries passed
   through the multigateway/multipooler stack. This caused a compatibility gap with standard PostgreSQL behavior, particularly noticeable when     
  running DDL commands that produce diagnostic output (e.g., table inheritance, implicit index creation).                                          
                                                                                                                                                   
  The fix extends the query protocol to carry notice messages alongside query results. Both simple and extended query protocols now propagate these
   messages back to the client, matching native PostgreSQL behavior.                                                                               
                                                                                                                                                   
  Resolves #573

### DIFF
--- a/go/common/sqltypes/sqltypes.go
+++ b/go/common/sqltypes/sqltypes.go
@@ -34,6 +34,24 @@ type Row struct {
 	Values []Value
 }
 
+// Notice represents a PostgreSQL notice response (non-fatal messages).
+type Notice struct {
+	Severity         string
+	Code             string
+	Message          string
+	Detail           string
+	Hint             string
+	Position         int32
+	InternalPosition int32
+	InternalQuery    string
+	Where            string
+	Schema           string
+	Table            string
+	Column           string
+	DataType         string
+	Constraint       string
+}
+
 // Result represents a query result with nullable values.
 type Result struct {
 	// Fields describes the columns in the result set.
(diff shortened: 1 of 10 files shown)


go/common/sqltypes/sqltypes.go
```diff
@@ -34,6 +34,24 @@ type Row struct {
 	Values []Value
 }
 
+// Notice represents a PostgreSQL notice response (non-fatal messages).
+type Notice struct {
```
GuptaManan100, line 38: Perfect, this was the only thing I wanted to ensure, that we don't use the proto type directly 👍

PR #578, opened 2026-02-03
fix(etcdtopo): Add graceful check for Close

rytswd:
This adds extra etcdtopo check. The current operator design relies on this logic, and transient connection failure would trigger this `Close()` method multiple times, causing it to panic.

### DIFF
--- a/go/common/topoclient/etcdtopo/store.go
+++ b/go/common/topoclient/etcdtopo/store.go
@@ -34,6 +34,7 @@ package etcdtopo
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"sync"
 	"time"
 
 	"github.com/spf13/pflag"
@@ -79,7 +80,9 @@ type etcdtopo struct {
 	// root is the root path for this client.
 	root string
 
-	running chan struct{}
+	running    chan struct{}
+	closeOnce  sync.Once
+	closeError error
 }
 
 func init() {
@@ -97,12 +100,17 @@ func registerEtcdTopoFlags(fs *pflag.FlagSet) {
 // It will nil out the global and cells fields, so any attempt to
 // re-use this server will panic.
 func (s *etcdtopo) Close() error {
-	close(s.running)
-	if err := s.cli.Close(); err != nil {
-		return err
+	if s == nil {
+		return nil
 	}
-	s.cli = nil
-	return nil
+	s.closeOnce.Do(func() {
+		close(s.running)
+		if s.cli != nil {
+			s.closeError = s.cli.Close()
+			s.cli = nil
+		}
+	})
+	return s.closeError
 }
 
 func newTLSConfig(certPath, keyPath, caPath string) (*tls.Config, error) {


go/common/topoclient/etcdtopo/store.go
```diff
@@ -97,12 +100,17 @@ func registerEtcdTopoFlags(fs *pflag.FlagSet) {
 // It will nil out the global and cells fields, so any attempt to
 // re-use this server will panic.
 func (s *etcdtopo) Close() error {
-	close(s.running)
-	if err := s.cli.Close(); err != nil {
-		return err
+	if s == nil {
```
dweitzman, line 103: What's the motivation for this `nil` check?

rytswd: The original panic was not about channel closed but nil reference on s, where it looked like the toposerver connection was lost unexpectedly, but operator kept using the same object reference. Maybe the sync.Once setup is sufficient in this case, but I wasn't clear how s can be nil so made this defensive logic

dweitzman: Interesting! I'm looking through all the places .Close() is called on topo and it seems like they all do an explicit `nil` check immediately beforehand.

The only thing suspicious I've spotted so far is that `store.go` doesn't hold any mutex while checking that `globalTopo` is non-nil before calling Close(). Is that where the panic() happened?

rytswd: This is the stack trace

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xfa5133]

goroutine 12778 [running]:
github.com/multigres/multigres/go/common/topoclient/etcdtopo.(*etcdtopo).Close(0x0)
    github.com/multigres/multigres@v0.0.0-20260202195144-f0c99746c6bc/go/common/topoclient/etcdtopo/store.go:100 +0x13
created by github.com/multigres/multigres/go/common/topoclient.(*WrapperConn).retryConnection.(*WrapperConn).retryConnection.(*Retry).Attempts.func4.(*WrapperConn).retryConnection-range1.func5 in goroutine 12748
    github.com/multigres/multigres@v0.0.0-20260202195144-f0c99746c6bc/go/common/topoclient/wrapperconn.go:148 +0xcd
```

dweitzman: Aha, looks like a subtle bug in `retryConnection`?

```
		conn, err := c.newFunc()
		mustContinue := func() bool {
			// We have to do this entire operation within a lock:
			// Once we check the value of c.closed, it should not be allowed
			// to change until we also set the value of c.wrapped. Otherwise,
			// it will conflict with c.Close.
			c.mu.Lock()
			defer c.mu.Unlock()

			if c.closed {
				// If the wrapper was closed, we have to close this extra
				// connection and stop retrying.
				if conn != nil { // !!! BUG IS HERE, DID NOT CHECK ERR FIRST !!!
					// No need to hold the lock while closing the connection.
					go conn.Close()
				}
				return false
			}
			if err != nil {
				return true
			}
			c.wrapped = conn
			return false
		}()
		if !mustContinue {
			return
		}
```

It's not sufficient to check that `conn != nil` without also checking that `err == nil` since a struct that's coerced to an interface can become a typed nil and type nils are not equal to untyped nils.

rytswd: Ah thanks for looking into this! It sounds like the Close function doesn't need update but retry logic needs to be corrected to handle err path?

dweitzman: Yes, made a PR: https://github.com/multigres/multigres/pull/584

dweitzman: Merged https://github.com/multigres/multigres/pull/584

rytswd: Thanks @dweitzman ! I can see the operator no longer panics with the connection. Closing this PR!

PR #582, opened 2026-02-04
feat(docs): Adds decision log for HA

rafael:
# Desc
- Adds decision log for HA components.

### DIFF
--- a/docs/ha/decision-log/2026-01-31-separate-voting-term-from-primary-term.md
+++ b/docs/ha/decision-log/2026-01-31-separate-voting-term-from-primary-term.md
@@ -55,20 +55,6 @@ If a node has accepted term 6 but is still primary for term 5, it must report pr
 - Delete the code in Postgres Monitor that changes the type to primary and use this as authoritative determination
 - Publish the term number in the topo
 
-## Related Reasoning and Implications
-
-### Transaction Ordering
-
-Transaction ordering and propagation will eventually need term numbers on transactions. During discovery, nodes should be able to report the term of their transaction state so the system can prefer the most advanced term state regardless of the election term.
-
-### Propagation Atomicity
-
-When propagating from higher term to lower term, the receiver must atomically update both the transaction log and its transaction term marker.
-
-### Term Completion
-
-Term completion is not guaranteed to be explicitly recorded. Revocation can be implicit by recruiting sync replicas. Only durable records count: any decision not recorded with durability never happened.
-
 ## Risks and Mitigations
 
 **Risk:** Confusion about which term to trust in different phases.


docs/ha/decision-log/2026-01-31-separate-voting-term-from-primary-term.md
```diff
@@ -0,0 +1,82 @@
...
+- Set primary term during establishment (promotion), not during discovery or propagation. Establishment is when primary authority becomes canonical
+- This also prevents two primaries from claiming the same primary term
+- Delete the code in Postgres Monitor that changes the type to primary and use this as authoritative determination
+- Publish the term number in the topo
+
+## Related Reasoning and Implications
+
+### Transaction Ordering
+
+Transaction ordering and propagation will eventually need term numbers on transactions. During discovery, nodes should be able to report the term of their transaction state so the system can prefer the most advanced term state regardless of the election term.
```
deepthi: How will this work even in theory? Transactions are stored in Postgres and have only an LSN associated with them, not a term number.

rafael: This refers to our plans of having the term number in postgres. When we get to two phase sync.

dweitzman: I feel like it ought to be possible to store that information outside the WAL even if it's impractical to change the WAL format for now, like maybe by writing a file to local disk that says, "transactions from LSN 1-5 are from term A and transactions with LSN >5 are from term B"

docs/ha/decision-log/2026-01-31-separate-voting-term-from-primary-term.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Propagation Atomicity
+
+When propagating from higher term to lower term, the receiver must atomically update both the transaction log and its transaction term marker.
+
+### Term Completion
+
+Term completion is not guaranteed to be explicitly recorded. Revocation can be implicit by recruiting sync replicas. Only durable records count: any decision not recorded with durability never happened.
```
dweitzman: Is this talking about a "voting term"? Might be helpful to be specific

docs/ha/decision-log/2026-01-31-separate-voting-term-from-primary-term.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Risks and Mitigations
+
+**Risk:** Confusion about which term to trust in different phases.
+
+**Mitigation:** Make "primary term is only set at establishment" a hard invariant and use it as the source of truth for primary determination.
```
dweitzman, line 62: I wonder if we the consensus docs or architecture should have an easy-to-find summary of moments during consensus and what they do or don't mean, like:

- "BeginTerm(5)" quorum revokes a voting term 4
- "Revoke(4)" quorum revokes the ability to commit new transactions to primary term 4, but individual nodes may still learn about more primary term 4 transactions through propagation
- Establishing a primary term 5 or higher finalizes where the end of primary term 4 transactions

etc.

docs/ha/decision-log/2026-01-31-separate-voting-term-from-primary-term.md
rafael: @dweitzman  - Agreed, I think that should be possible. As long as we can perform that operation in a transactional way between the write to the file and the transaction that starts the term in the DB. We can come back to this later.

For now, I will actually remove this part from this decision as it is not super relevant for the change here.

docs/ha/decision-log/2026-01-31-separate-voting-term-from-primary-term.md
rafael: Good idea. Added to my TODO

PR #583, opened 2026-02-04
feat(queryserving): preserve PostgreSQL error format through Multigres stack

### LINKED ISSUE #574: Error messages differ from PostgreSQL format
## Problem

Error messages from Multigres have a different format than native PostgreSQL, which breaks compatibility expectations.

## PostgreSQL Output

```
postgres=# SELECT bool 'test' AS error;
ERROR:  invalid input syntax for type boolean: "test"
LINE 1: SELECT bool 'test' AS error;
                    ^
```

## Multigres Output

```
postgres=> SELECT bool 'test' AS error;
ERROR:  query execution failed: query execution failed: stream receive error: rpc error: code = Unknown desc = query execution failed: ERROR: invalid input syntax for type boolean: "test" (SQLSTATE 22P02)
postgres=>
```

## Expected Behavior

Multigres error messages should match the PostgreSQL format exactly:
- Clean error message without internal RPC/streaming details
- Include the `LINE X:` indicator showing the problematic query line
- Include the `^` caret pointing to the error location

haritabh17:
Closes #574

## Summary

Preserves all 14 PostgreSQL diagnostic fields (Position, Schema, Table, Constraint, Hint, Detail, Where, etc.) as errors flow through the Multigres stack. Previously, clients lost this debugging context.

## Changes

- Unify error/notice handling with `PgDiagnostic` type implementing `error` interface
- Add `PgError` wrapper for system-level propagation
- Preserve diagnostics through gRPC via `RPCError.pg_diagnostic`
- Add `parseDiagnosticFields()` and `writePgDiagnosticResponse()` for unified parsing/writing

## Test plan

- Unit tests for `PgDiagnostic`, `PgError`, and gRPC round-trip
- E2E tests comparing error output between direct PostgreSQL and Multigres

### DIFF
--- a/go/common/pgprotocol/server/conn.go
+++ b/go/common/pgprotocol/server/conn.go
@@ -285,7 +285,7 @@ func (c *Conn) serve() error {
 	if err := c.handleStartup(); err != nil {
 		c.logger.Error("startup failed", "error", err)
 		// Try to send an error response before closing.
-		_ = c.writeErrorResponse("FATAL", "08P01", "connection startup failed", err.Error(), "")
+		_ = c.writeSimpleErrorWithDetail("FATAL", "08P01", "connection startup failed", err.Error(), "")
 		_ = c.flush()
 		return err
 	}
@@ -313,7 +313,7 @@ func (c *Conn) serve() error {
 		if err := c.handleMessage(msgType); err != nil {
 			c.logger.Error("error handling message", "type", string(msgType), "error", err)
 			// Send error response and continue (unless it's a fatal error).
-			_ = c.writeErrorResponse("ERROR", "XX000", "internal error", err.Error(), "")
+			_ = c.writeSimpleErrorWithDetail("ERROR", "XX000", "internal error", err.Error(), "")
 			_ = c.writeReadyForQuery()
 			_ = c.flush()
 			// For now, close connection on any error.
@@ -427,12 +427,9 @@ func (c *Conn) handleQuery() error {
 		return nil
 	})
 	if err != nil {
-		// Send error response with the actual error in the message for better visibility.
-		// lib/pq and other clients often only show the message field, not the detail field.
 		c.logger.Error("query execution failed", "query", queryStr, "error", err)
-		errMsg := fmt.Sprintf("query execution failed: %v", err)
-		if err := c.writeErrorResponse("ERROR", "42000", errMsg, "", ""); err != nil {
-			return err
+		if writeErr := c.writeError(err); writeErr != nil {
+			return writeErr
 		}
 	}
 
@@ -499,7 +496,7 @@ func (c *Conn) handleParse() error {
 	// Call the handler to validate and prepare the statement.
 	// The handler is responsible for storing any state it needs.
 	if err := c.handler.HandleParse(c.ctx, c, stmtName, queryStr, paramTypes); err != nil {
-		if writeErr := c.writeErrorResponse("ERROR", "42000", "parse failed", err.Error(), ""); writeErr != nil {
+		if writeErr := c.writeSimpleErrorWithDetail("ERROR", "42000", "parse failed", err.Error(), ""); writeErr != nil {
 			return writeErr
 		}
 		if writeErr := c.writeReadyForQuery(); writeErr != nil {
@@ -593,7 +590,7 @@ func (c *Conn) handleBind() error {
 
 	// Call the handler to create and bind the portal with parameters.
 	if err := c.handler.HandleBind(c.ctx, c, portalName, stmtName, params, paramFormats, resultFormats); err != nil {
-		if writeErr := c.writeErrorResponse("ERROR", "42000", "bind failed", err.Error(), ""); writeErr != nil {
+		if writeErr := c.writeSimpleErrorWithDetail("ERROR", "42000", "bind failed", err.Error(), ""); writeErr != nil {
 			return writeErr
 		}
 		if writeErr := c.writeReadyForQuery(); writeErr != nil {
@@ -685,7 +682,7 @@ func (c *Conn) handleExecute() error {
 		return nil
 	})
 	if err != nil {
-		if writeErr := c.writeErrorResponse("ERROR", "42000", "execution failed", err.Error(), ""); writeErr != nil {
+		if writeErr := c.writeError(err); writeErr != nil {
 			return writeErr
 		}
 		return c.flush()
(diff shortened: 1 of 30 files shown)


proto/query.proto
```diff
@@ -85,23 +87,34 @@ message Row {
...
+//
+// Reference: https://www.postgresql.org/docs/current/protocol-error-fields.html
+message PgDiagnostic {
+  // message_type is the PostgreSQL protocol message type byte.
+  // 'E' (0x45 = 69) for ErrorResponse, 'N' (0x4E = 78) for NoticeResponse.
+  // This determines whether the diagnostic is an error or notice.
+  int32 message_type = 15;
+
+  // severity is the severity level (e.g., "ERROR", "FATAL", "WARNING", "NOTICE", "INFO")
   string severity = 1;
```
GuptaManan100, line 114: Don't these 2 serve the same purpose?

go/common/mterrors/pgerror.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// NewPgErrorFromDiagnostic creates a new PgError from a PgDiagnostic.
+// Use this when reconstructing a PgError from gRPC responses.
+func NewPgErrorFromDiagnostic(diag *sqltypes.PgDiagnostic) *PgError {
+	return &PgError{
+		diag:  diag,
+		stack: callers(),
+	}
+}
```
GuptaManan100: They're the same function 😄

go/common/mterrors/pgerror.go
```diff
@@ -0,0 +1,120 @@
...
+
+// PgError represents a PostgreSQL error with full diagnostic information.
+// It wraps sqltypes.PgDiagnostic to preserve all 14 PostgreSQL error fields
+// as they pass through the Multigres system.
+//
+// PgError should NOT be wrapped with mterrors.Wrapf() - it should pass through
+// unchanged to preserve the original PostgreSQL error format for the client.
+type PgError struct {
+	diag  *sqltypes.PgDiagnostic
+	stack *stack
```
GuptaManan100: Not completely sure what value we're getting by adding this struct. We aren't using the stack for anything. We're just storing it. If that field is removed, then it is only keeping a singular field inside. Maybe remove this from the path altogether?

go/common/mterrors/pgerror.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// PgError represents a PostgreSQL error with full diagnostic information.
+// It wraps sqltypes.PgDiagnostic to preserve all 14 PostgreSQL error fields
+// as they pass through the Multigres system.
+//
+// PgError should NOT be wrapped with mterrors.Wrapf() - it should pass through
+// unchanged to preserve the original PostgreSQL error format for the client.
+type PgError struct {
+	diag  *sqltypes.PgDiagnostic
```
GuptaManan100: Maybe we just move `*sqltypes.PgDiagnostic` to mterrors package instead. 
I mean the only invariant we want is the querypb fields don't get used directly. It doesn't necessarily need to live in sqltypes. Having it in this package maybe better cause then we can remove this intermediate type. WDYT?

proto/query.proto
haritabh17: `message_type` is there to distinguish between the two different message types `ErrorResponse` and `NoticeResponse` and `severity` is the content inside the message

ref - https://www.postgresql.org/docs/current/protocol-error-fields.html

go/common/mterrors/pgerror.go
haritabh17: Makes sense! I've removed the PgError wrapper entirely.

go/common/mterrors/pgerror.go
haritabh17: Ahh, this whole file is removed now.

go/common/sqltypes/sqltypes.go
```diff
@@ -34,8 +109,13 @@ type Row struct {
 	Values []Value
 }
 
-// Notice represents a PostgreSQL notice response (non-fatal messages).
-type Notice struct {
+// PgDiagnostic represents a PostgreSQL diagnostic message (error or notice).
+// PostgreSQL uses the same wire format for both ErrorResponse ('E') and NoticeResponse ('N'),
+// differentiated by the MessageType field.
+type PgDiagnostic struct {
+	// MessageType is the PostgreSQL protocol message type byte.
```
GuptaManan100: What did you think of moving this to mterrors package? We could that way encapsulate everything error related there

go/common/pgprotocol/server/conn.go
```diff
@@ -427,12 +427,19 @@ func (c *Conn) handleQuery() error {
...
+		var diag *sqltypes.PgDiagnostic
+		if errors.As(err, &diag) {
+			if writeErr := c.writeErrorFromDiagnostic(diag); writeErr != nil {
+				return writeErr
+			}
+		} else {
+			// Non-PostgreSQL error: send with generic SQLSTATE and clean message.
+			if writeErr := c.writeSimpleError("XX000", err.Error()); writeErr != nil {
+				return writeErr
+			}
```
GuptaManan100: maybe instead of doing this here, we only have a single function to write an error, and that checks the error type, and makes decision accordingly.

go/multipooler/executor/executor.go
```diff
@@ -679,5 +680,23 @@ func int32ToInt16Slice(in []int32) []int16 {
...
+func wrapQueryError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// PostgreSQL errors (*sqltypes.PgDiagnostic) pass through unchanged
+	var diag *sqltypes.PgDiagnostic
+	if errors.As(err, &diag) {
+		return diag
+	}
```
GuptaManan100: I don't like this pattern either. I think it's better to have a Wrapf kind of a function in mterrors package, that allows wrapping all errors, and in the proposed function to print errors ☝️, we can get the RootCause. This should either be already available in mterrors or we can add that too. That would mean we don't need to worry about not wrapping postgres errors. And more importantly if we print errors in debug logs, then we get the full wrapped error with all the context, without paying the penalty of messing up the errors against postgres.

go/services/multigateway/poolergateway/grpc_query_service.go
```diff
@@ -186,6 +199,12 @@ func (g *grpcQueryService) PortalStreamExecute(
 	// Call the gRPC PortalStreamExecute
 	stream, err := g.client.PortalStreamExecute(ctx, req)
 	if err != nil {
+		// Convert gRPC error - if it's a PostgreSQL error, preserve it
+		grpcErr := mterrors.FromGRPC(err)
+		var pgDiag *sqltypes.PgDiagnostic
+		if errors.As(grpcErr, &pgDiag) {
+			return queryservice.ReservedState{}, grpcErr
+		}
```
GuptaManan100, line 216: If we do what I suggest ☝️, all of this goes away, and we can always just wrap the errors

go/common/sqltypes/sqltypes.go
haritabh17: Okay, this needs a bit of refactoring as currently this single struct handles both Errors and Notices.

go/common/pgprotocol/server/conn.go
haritabh17: Ack, moving the Diag struct to mterrors that will unify things

PR #584, opened 2026-02-04
fix(topoclient): prevent nil pointer panic in WrapperConn retry logic

dweitzman:
Check error before attempting to use conn in retryConnection to avoid panic from typed nil interfaces. When factory.Create() fails, it may return a typed nil (*etcdtopo)(nil) which becomes a non-nil interface value. Previously, the code only checked c.closed was not untyped `nil` before checking err, causing it to call Close() on an unusable typed nil connection.

Context: https://github.com/multigres/multigres/pull/578#discussion_r2764711182


rafael (approved):
LGTM

PR #586, opened 2026-02-05
feat(multipooler): Demote primary improvements

rafael:
# Desc
- This PR continues building on top the work that was introduced in https://github.com/multigres/multigres/pull/566 and https://github.com/multigres/multigres/pull/565.
- There are two main changes:
   - The demote stale primary action starts using the PrimaryTerm to discover stale primaries. The current implementation was not correct in edge cases where two primaries had the same Term (but different PrimaryTerm)
   - It removes a TODO from the BeginTerm code, that allow a pooler to vote, even though the revoke action failed.
- Outside this, the rest of the changes are small bug fixes that were discovered by flaky tests. I added comments in line to add context. 

# Tests
- Tests _seem_ less flaky in this PR.
- Added unit tests for the changes specific to this PR.

### DIFF
--- a/go/multipooler/manager/rpc_manager.go
+++ b/go/multipooler/manager/rpc_manager.go
@@ -1494,6 +1494,15 @@ func (pm *MultiPoolerManager) stopPostgresIfRunning(ctx context.Context) error {
 	// Stop postgres (no-op if already stopped)
 	stopReq := &pgctldpb.StopRequest{Mode: "fast"}
 	if _, err := pm.pgctldClient.Stop(ctx, stopReq); err != nil {
+		// Treat "already stopped" errors as success to make this truly idempotent.
+		// This handles race conditions where postgres was stopped between our check and stop call.
+		errMsg := err.Error()
+		if strings.Contains(errMsg, "not running") ||
+			strings.Contains(errMsg, "no child processes") ||
+			strings.Contains(errMsg, "no such process") {
+			pm.logger.InfoContext(ctx, "Postgres already stopped, continuing", "error", errMsg)
+			return nil
+		}
 		return mterrors.Wrap(err, "failed to stop postgres")
 	}
 


go/multipooler/grpcconsensusservice/service.go
```diff
@@ -47,7 +47,9 @@ func RegisterConsensusServices(senv *servenv.ServEnv, grpc *servenv.GrpcServer)
 func (s *consensusService) BeginTerm(ctx context.Context, req *consensusdata.BeginTermRequest) (*consensusdata.BeginTermResponse, error) {
 	resp, err := s.manager.BeginTerm(ctx, req)
 	if err != nil {
-		return nil, mterrors.ToGRPC(err)
+		// Return response even on error - the response may contain accepted=true
```
rafael, line 50: This was uncovered when I removed the extra code in the pooler that allows to vote when revoke fails

go/multipooler/manager/rpc_consensus.go
```diff
@@ -71,44 +71,6 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
 			"unknown BeginTerm action type: %v", req.Action)
 	}
 
-	// TEMPORARY: Check postgres health before accepting term with REVOKE action.
```
rafael, line 74: This is one of the main changes  from the PR description

go/multipooler/manager/rpc_manager.go
```diff
@@ -984,6 +984,28 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
 	}
 	defer pm.actionLock.Release(ctx)
 
+	// Check if already demoted by reading primary_term from disk.
```
rafael, line 987: Towards a more idempotent action. This was introducing some flakes in some tests where two multiorch race to fix a demote stale primary and then after one fixing it, another one attempted the fix again. 

This was ok from a functionality, but it broke the assertions in the test

go/multipooler/manager/rpc_manager.go
```diff
@@ -1036,23 +1058,28 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
...
-		return nil, mterrors.Wrap(err, "failed to update topology")
-	}
-
 	// Clear primary_term since this node is no longer primary for any term.
 	// Note: consensusState can't be nil here because validateTerm passed.
 	if err := pm.consensusState.SetPrimaryTerm(ctx, 0, false /* force */); err != nil {
 		return nil, mterrors.Wrap(err, "failed to clear primary term")
 	}
 
+	// Update consensus term to match the correct primary's term after successful demotion
```
rafael, line 1069: This makes sure consensus term is updated while replication is being fixed as part of demotion.

go/services/multiorch/recovery/actions/fix_replication.go
```diff
@@ -165,25 +165,6 @@ func (a *FixReplicationAction) fixNotReplicating(
 	}
 	consensusTerm := consensusResp.CurrentTerm
 
-	// Check if timelines have diverged, if so, rewind first
```
rafael, line 168: This is arguably the most controversial change here. I think there are still gremlins here that we need to iron out related to Postgres behavior that I still don't fully understand.

* Sometimes I was seeing that immediately after pg_rewind, the replica still reported the old timeline, so we were re-running rewind.
* This was okay from a functionality perspective, but it also introduced some flakiness in the tests.

I think short term, we can land this, but we should have a follow-up to revisit how we can know for sure if the replica diverged. Checking just the timeline info is not sufficient. Calling pg_rewind in dry mode is a heavy hammer because it requires stopping Postgres. These are the two approaches we have tried so far, but I'm not happy with either.

go/services/multiorch/recovery/analysis/generator.go
```diff
@@ -522,12 +529,70 @@ func (g *AnalysisGenerator) detectOtherPrimary(
 		}
 
 		if poolerType == clustermetadatapb.PoolerType_PRIMARY {
-			// Found another PRIMARY - one of them is stale!
-			analysis.OtherPrimaryInShard = pooler.MultiPooler.Id
+			// Extract consensus term and primary term
```
rafael, line 532: This is the other important change that is part of this PR

go/multipooler/manager/rpc_manager.go
```diff
@@ -984,6 +984,28 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
 	}
 	defer pm.actionLock.Release(ctx)
 
+	// Check if already demoted by reading primary_term from disk.
+	// If primary_term is 0, this node was already demoted (either by a previous call
+	// or by another multiorch instance). Return early to avoid redundant work.
+	// TODO (@rafael): This information should come from pooler state, instead of checking this
+	// invariant.
+	term, err := pm.consensusState.GetInconsistentTerm()
```
dweitzman: The action lock is held here, so it should be fine to call GetTerm() rather than GetInconsistentTerm()

go/multipooler/manager/rpc_manager.go
```diff
@@ -984,6 +984,28 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
...
+	if err != nil {
+		return nil, mterrors.Wrap(err, "failed to get current term")
+	}
+	if term.GetPrimaryTerm() == 0 {
```
dweitzman, line 998: To be clear, the invariant is that primary term must be set to a non-zero value anytime postgres might be in a primary state, so we would never promote before setting primary term or clear primary term prior to demoting?

go/services/multiorch/recovery/analysis/stale_primary_analyzer.go
```diff
@@ -71,39 +72,46 @@ func (a *StalePrimaryAnalyzer) Analyze(poolerAnalysis *store.ReplicationAnalysis
...
-	// If terms are equal, neither node demotes - this is an unusual state that
-	// should be resolved through normal failover or manual intervention.
+	// Skip if no other primaries detected
+	if len(poolerAnalysis.OtherPrimariesInShard) == 0 {
+		return nil, nil
+	}
 
-	if poolerAnalysis.ConsensusTerm == poolerAnalysis.OtherPrimaryTerm {
-		// Terms are equal - unusual state, skip
+	// Skip if no most advanced primary identified (tie in PrimaryTerm)
```
dweitzman: If there are two primaries within the same term for whatever reason, we wouldn't try to demote either one of them?

My intuition is that we should know which imposter primaries don't have quorum and demote those ones, but maybe that's not a "stale" primary and more of an rogue/imposter primary

go/services/multiorch/store/replication_analysis.go
```diff
@@ -68,9 +76,10 @@ type ReplicationAnalysis struct {
 	IsInPrimaryStandbyList bool // Whether this replica is in the primary's synchronous_standby_names
 
 	// Stale primary detection: populated for PRIMARY nodes only
-	OtherPrimaryInShard *clustermetadatapb.ID // ID of another PRIMARY in same shard (if detected)
-	OtherPrimaryTerm    int64                 // Consensus term of the other primary
-	ConsensusTerm       int64                 // This node's consensus term (from health check)
+	PrimaryTerm           int64          // This pooler's primary term (term when promoted)
+	OtherPrimariesInShard []*PrimaryInfo // All other primaries detected in the shard
+	MostAdvancedPrimary   *PrimaryInfo   // Primary with highest PrimaryTerm (rewind source)
```
dweitzman: Should this be called something more specific and literal like `HighestPrimaryTermPrimary`?

My worry with `MostAdvancedPrimary` is that it sounds like it could mean the primary with the highest LSN, which is not what's actually happening

go/services/multiorch/recovery/actions/fix_replication.go
dweitzman: Feels like a multipooler should theoretically may be able to figure out on its own that its replication is stuck and whether it should run rewind without multiorch's help

Seems like that wouldn't break consensus as long as a voting replica refuses to ACK a revoke request while it's still in the middle of a pg_rewind to sync itself with the outgoing primary

go/services/multiorch/recovery/analysis/generator_test.go
```diff
@@ -1099,3 +1100,229 @@ func TestPopulatePrimaryInfo_IsInPrimaryStandbyList(t *testing.T) {
...
+		require.NotNil(t, analysis.MostAdvancedPrimary)
+		assert.Equal(t, "primary-2", analysis.MostAdvancedPrimary.ID.Name)
+		assert.Equal(t, int64(6), analysis.MostAdvancedPrimary.PrimaryTerm)
+	})
+
+	t.Run("multiple other primaries detected", func(t *testing.T) {
+		store := setupMultiplePrimariesStore(t, []primaryConfig{
+			{id: "primary-1", primaryTerm: 5, consensusTerm: 9},
+			{id: "primary-2", primaryTerm: 4, consensusTerm: 10},
+			{id: "primary-3", primaryTerm: 6, consensusTerm: 11},
```
dweitzman: Maybe for this test case the highest primary term should not be the highest consensus term to test that the comparison is is on primary term rather than consensus term?

go/services/multiorch/recovery/analysis/stale_primary_analyzer_test.go
```diff
@@ -86,14 +117,14 @@ func TestStalePrimaryAnalyzer_Analyze(t *testing.T) {
...
-		assert.Contains(t, problem.Description, "stale-primary (term 5) is stale")
-		assert.Contains(t, problem.Description, "new-primary has term 6")
+		assert.Contains(t, problem.Description, "stale-primary (stale_primary_term 5) is stale")
+		assert.Contains(t, problem.Description, "new-primary (most_advanced_primary_term 6)")
 	})
 
-	t.Run("does not demote when terms are equal (prevents double demotion)", func(t *testing.T) {
-		// When terms are equal, neither node should demote to avoid both nodes
+	t.Run("does not demote when primary_terms are equal (prevents double demotion)", func(t *testing.T) {
+		// When primary_terms are equal, neither pooler should demote to avoid both poolers
```
dweitzman: Ah, this comment answers the earlier question about why not designate a random primary as rogue and demote it. Might be good to put this comment in the code and not just the test?

Seems like rather than having manual intervention, multiorch could also:
- start a new term
- reappoint a randomly-chosen primary is the primary for the new term, causing it to update it's primary_term
- now the rogue / imposter primaries have a stale term and can be safely demoted

go/multipooler/manager/rpc_manager.go
rafael: That's correct

go/services/multiorch/recovery/analysis/stale_primary_analyzer.go
rafael: I was thinking that if this were to happen, it implies that consensus invariants were broken. So no automated demotion in that situation. 

The second part matches my mental model. It would be more a rogue / imposter primary and that should not be possible

dweitzman: Reflecting on this more, I think I agree that we shouldn't rewind an imposter primary. But probably the right thing to do would be to force an imposter primary to shut down and preserve its data somewhere in case the pod/disk gets decommissioned?

With heartbeats and autovacuum and things like that it seems like it might be non-trivial to tell if split-brain writes are "real" data that requires human intervention vs bookkeeping data we can rewind and throw away to rejoin the cohort

go/services/multiorch/recovery/actions/fix_replication.go
rafael: Yeah, agree. I'll schedule something for next week to brainstorm next steps here.

PR #588, opened 2026-02-06
feat: add transaction handling with implicit-to-explicit transformation

GuptaManan100:
### Summary

- Add transaction support to multigateway: explicit (`BEGIN`/`COMMIT`), implicit (multi-statement batches), and deferred `BEGIN` execution
- Implement `ReserveStreamExecute` and `ConcludeTransaction` RPCs on multipooler for reserved connection management with multi-reason bitmask tracking
- Handle aborted transaction state, connection cleanup on disconnect, and timeout error propagation matching PostgreSQL semantics

### Details

**Implicit transaction transformation**: Multi-statement batches are wrapped in synthetic `BEGIN`/`COMMIT` so all statements use the same reserved backend connection. User `BEGIN` mid-batch is detected and the segment is marked explicit (preserving PostgreSQL's "adoption" semantics). Failures in implicit segments auto-rollback silently; failures in explicit segments transition to aborted state.

**Deferred BEGIN**: `BEGIN` sets `TxnStatusInBlock` without contacting the backend. The actual `BEGIN` is sent atomically with the first real query via `ReserveStreamExecute`, avoiding empty transactions and reducing round-trips.

**Reserved connection lifecycle**: ScatterConn uses 3-case routing (existing reserved conn / reserve+execute / pooled conn). Reservation reasons are tracked as a bitmask (`transaction | temp_table | portal | copy`), so concluding a transaction only releases the connection if no other reasons remain.

**Error handling**: Aborted transactions reject all queries unless the batch starts with `ROLLBACK` (matching PostgreSQL behavior for `ROLLBACK; SELECT 1;` batches). Client disconnect triggers `ReleaseReservedConnection` which rolls back transactions, aborts COPYs, and releases portals.


go/common/protoutil/reservation.go
```diff
@@ -0,0 +1,127 @@
...
+		parts = append(parts, "temp_table")
+	}
+	if HasPortalReason(reasons) {
+		parts = append(parts, "portal")
+	}
+	if len(parts) == 0 {
+		return "unknown"
+	}
+	return strings.Join(parts, "|")
+}
```
haritabh17, line 148: Consider adding a validation to ensure only known bits are set?

```
  func ValidateReasons(reasons uint32) error {
      validMask := ReasonTransaction | ReasonTempTable | ReasonPortal
      if reasons & ^validMask != 0 {
          return fmt.Errorf("invalid reservation reasons: %032b", reasons)
      }
      return nil
  }
```

go/multipooler/pools/reserved/reserved_conn.go
```diff
@@ -184,6 +184,41 @@ func (c *Conn) ReservedProps() *ReservationProperties {
...
+
+// RemoveReservationReason removes a reason from the reservation bitmask.
+// If all reasons are removed and no portals remain, clears reservedProps.
+// Returns true if all reservation reasons are gone (connection should be released).
+func (c *Conn) RemoveReservationReason(reason uint32) bool {
+	if c.reservedProps == nil {
+		return true
+	}
+	c.reservedProps.RemoveReason(reason)
+	if c.reservedProps.IsEmpty() && !c.reservedProps.HasPortals() {
```
haritabh17: Did not understand this fully, is there a scenario where Reasons == 0 but Portals map is non-empty?

go/services/multigateway/engine/transaction_primitive.go
```diff
@@ -0,0 +1,173 @@
...
+func (t *TransactionPrimitive) executeBegin(
+	ctx context.Context,
+	conn *server.Conn,
+	callback func(context.Context, *sqltypes.Result) error,
+) error {
+	// Set transaction state (deferred - no backend call yet)
+	conn.SetTxnStatus(protocol.TxnStatusInBlock)
+
+	// Return synthetic result to client
+	return callback(ctx, &sqltypes.Result{
```
haritabh17, line 101: Does the deferred BEGIN replay as plain BEGIN, dropping isolation level/read-only options??

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -284,6 +315,63 @@ func (sc *ScatterConn) Describe(
...
+		eo := &query.ExecuteOptions{
+			User:                 conn.User(),
+			SessionSettings:      state.GetSessionSettings(),
+			ReservedConnectionId: uint64(ss.ReservedConnectionId),
+		}
+
+		qs, err := sc.gateway.QueryServiceByID(ctx, ss.PoolerID, ss.Target)
+		if err != nil {
+			// Connection lost — clear stale state
+			state.ClearReservedConnection(ss.Target)
```
haritabh17: While iterating over ShardStates, ClearReservedConnection does swap-and-truncate on the same slice. Will it skip the swapped entry??

go/services/multigateway/handler/handler.go
```diff
@@ -211,10 +276,33 @@ func (h *MultiGatewayHandler) HandleClose(ctx context.Context, conn *server.Conn
...
-
-	// TODO: Handle transaction state
 	return nil
 }
 
+// ConnectionClosed is called when a client connection is closed.
+// It releases all reserved connections (rolling back transactions, aborting COPYs)
+// and removes prepared statement state.
+func (h *MultiGatewayHandler) ConnectionClosed(conn *server.Conn) {
+	// Release reserved connections if connection state exists.
```
haritabh17, line 288: ReleaseAll(conn.Context()) is called in ConnectionClosed, but by that point conn.Context() may already be cancelled. If ReleaseAll needs to send cleanup queries (like ROLLBACK) to the backend, could they fail??

go/services/multigateway/handler/handler.go
```diff
@@ -147,13 +198,27 @@ func (h *MultiGatewayHandler) HandleExecute(ctx context.Context, conn *server.Co
 	// Get the connection state.
 	state := h.getConnectionState(conn)
 
+	// Reject queries in aborted transaction state (except ROLLBACK via simple query).
+	// In the extended query protocol, ROLLBACK is typically sent via simple query,
+	// so we reject all Execute messages when aborted.
+	if conn.TxnStatus() == protocol.TxnStatusFailed {
+		return errAbortedTransaction
```
haritabh17: When transaction status is Failed, handleExecute rejects all Execute messages. But ROLLBACK is the one statement that should work in an aborted transaction — otherwise the client has no way to recover without disconnecting.

go/services/multigateway/engine/transaction_primitive.go
```diff
@@ -0,0 +1,173 @@
...
+
+	case ast.TRANS_STMT_COMMIT:
+		return t.executeCommit(ctx, exec, conn, state, callback)
+
+	case ast.TRANS_STMT_ROLLBACK:
+		return t.executeRollback(ctx, exec, conn, state, callback)
+
+	default:
+		// For other transaction statements (SAVEPOINT, etc.), pass through to backend
+		return exec.StreamExecute(ctx, conn, t.TableGroup, "", t.Query, state, callback)
```
haritabh17: The default case in the switch falls through to the regular (pooled) connection path? Does this mean that the SAVEPOINT would hit a random pooled connection instead?

go/services/multigateway/handler/transaction_helpers.go
```diff
@@ -0,0 +1,111 @@
...
+				conn.SetTxnStatus(protocol.TxnStatusFailed)
+			}
+			return err
+		}
+	}
+
+	// Auto-commit if we ended in an implicit transaction
+	if isImplicitTx {
+		if err := silentExecute(ast.NewCommitStmt()); err != nil {
+			return err
```
haritabh17, line 215: If the synthetic COMMIT to the backend fails after individual statements already succeeded from the client's perspective, the error propagates up — but does the client already think those statements committed, e.g. because we are sending SUCCESS message for each statement?

go/common/protoutil/reservation.go
GuptaManan100: Added the function, but not really sure where to use it?

go/services/multigateway/engine/transaction_primitive.go
GuptaManan100: Good point! The deferred BEGIN was always replaying as a plain `BEGIN`, dropping any isolation level or access mode options.
Fixed by:
1. Storing the original BEGIN query text (e.g. `"BEGIN ISOLATION LEVEL SERIALIZABLE"`) in `state.PendingBeginQuery` during `executeBegin`.
2. When `scatter_conn` creates the reserved connection (Case 2), it passes `PendingBeginQuery` via the new `begin_query` field on `ReservationOptions`.
3. The multipooler's `BeginWithQuery(ctx, query)` uses the original statement instead of hardcoded `"BEGIN"`.

go/multipooler/pools/reserved/reserved_conn.go
GuptaManan100: Purely defensive coding. There is no way that happens. We only remove the reason for Portals if we have no portals left. Would you prefer I remove it?

go/services/multigateway/scatterconn/scatter_conn.go
GuptaManan100: Good eye. The `range` caches the original slice header, so swap-and-truncate during iteration could cause elements to be skipped or processed twice.

Fixed by collecting updates into a slice during iteration and applying them after the loop completes. Also added `UpdateReservationReasons()` to `connection_state.go` for the non-clearing case.

Additionally, the loop now continues on errors instead of returning early - if one shard fails to conclude, we still conclude the others and return all errors joined together.  Probably not meaningful now, but still this is the right code.

go/services/multigateway/handler/handler.go
GuptaManan100: Looking at `listener.go`, `ConnectionClosed` is called *before* `conn.Close()`, which is where `cancel()` happens. So `conn.Context()` is still valid at this point. Added a `context.WithTimeout` wrapping `conn.Context()` (5s) so cleanup doesn't hang indefinitely, but no need for a detached context.

go/services/multigateway/handler/handler.go
GuptaManan100: You're right! PostgreSQL allows ROLLBACK even via the extended query protocol in an aborted transaction. Fixed by looking up the portal first, then checking if its underlying statement is a ROLLBACK before rejecting.

go/services/multigateway/engine/transaction_primitive.go
GuptaManan100: This actually works correctly. The default case calls `exec.StreamExecute()` which routes through `scatter_conn.StreamExecute()`. That method checks for an existing reserved connection (Case 1) or creates one if in a transaction (Case 2, which sends BEGIN + the query atomically). So a SAVEPOINT issued inside a transaction will always use/create a reserved connection, never a random pooled one.

go/services/multigateway/handler/transaction_helpers.go
GuptaManan100: This is a valid concern. Individual `CommandComplete` messages are sent to the client before the synthetic COMMIT, so if COMMIT fails those changes are rolled back even though the client saw success for individual statements. I don't have a way to fix this 🤯. Maybe we should discuss what we can do!

GuptaManan100: ALso we should see what Postgres does, it should be the same problem for them 🤔

GuptaManan100: Found what Postgres does, and emulated the behaviour. I'm warning you, it's functional, but it's not pretty! 😢

go/common/protoutil/reservation.go
haritabh17: Makes sense to call it in the gRPC handler when you first unpack ReservationOptions from the request — that's the trust boundary where invalid bitmasks should be rejected early rather than silently accepted, e.g. at the top of ReserveStreamExecute in grpcpoolerservice/service.go as this is the gRPC trust boundary where the request comes in from the wire.

go/multipooler/pools/reserved/reserved_conn.go
haritabh17: I'd lean towards removing it — the extra check may be confusing for readers trying to understand when it could be true. If you'd prefer to keep it as a safety net, maybe add a brief comment like `// defensive: should never happen` or make it as `Assert` so future readers don't spend time reasoning about it?

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -284,6 +321,97 @@ func (sc *ScatterConn) Describe(
...
+
+		qs, err := sc.gateway.QueryServiceByID(ctx, ss.PoolerID, ss.Target)
+		if err != nil {
+			// Connection lost — mark for clearing, continue to other shards
+			updates = append(updates, shardUpdate{target: ss.Target, clear: true})
+			errs = append(errs, fmt.Errorf("conclude transaction: pooler lookup failed for %s: %w", ss.Target, err))
+			continue
+		}
+
+		result, remainingReasons, err := qs.ConcludeTransaction(ctx, ss.Target, eo, conclusion)
```
haritabh17, line 392: Should there be a guard that returns an error (or at least logs a warning) when len(ShardStates) > 1 with a transaction reason? It could serve as a sentinel when we go with DT??

go/services/multigateway/handler/transaction_helpers.go
haritabh17: 😄 you mean intercepting the `CommandTag != ""` to hold off the last message before commit? 

Btw, if the last statement in a multi-statement batch is a COPY, the callback pattern might be different (CopyInResponse/CopyOutResponse instead of DataRow+CommandComplete). Guess that needs to be covered as well??

Just validated that copy command works in multi-statement batches:
```
postgres=# SELECT 1\; COPY copy_test FROM STDIN WITH (FORMAT csv)\; SELECT 2;
 ?column?
----------
        1
(1 row)

Enter data to be copied followed by a newline.
End with a backslash and a period on a line by itself, or an EOF signal.
>> 1, alice
>> 2, bob
>> \.
COPY 2
 ?column?
----------
        2
(1 row)

postgres=#
```

go/multipooler/executor/executor.go
```diff
@@ -690,5 +693,219 @@ func wrapQueryError(err error) error {
...
+			beginQuery = reservationOptions.BeginQuery
+		}
+		if err := reservedConn.BeginWithQuery(ctx, beginQuery); err != nil {
+			reservedConn.Release(reserved.ReleaseError)
+			return queryservice.ReservedState{}, err
+		}
+	}
+
+	// Execute the actual query
+	results, err := reservedConn.Query(ctx, sql)
```
haritabh17: The non-reserved path uses StreamExecute which actually streams rows as they arrive. For large result sets inside transactions, this buffers everything in memory. Should this use the streaming path instead, matching the method name?

go/services/multigateway/handler/transaction_helpers.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+			needsBegin = false
+			isImplicitTx = true
+		}
+
+		// User's BEGIN - skip execution (we already started), mark as explicit.
+		// Still send a result to the client so the response count matches their statements.
+		if ast.IsBeginStatement(stmt) {
+			isImplicitTx = false
+			if err := callback(ctx, &sqltypes.Result{CommandTag: "BEGIN"}); err != nil {
```
haritabh17, line 131: Consider the batch `SELECT 1; BEGIN ISOLATION LEVEL SERIALIZABLE; SELECT 2`. A synthetic BEGIN (plain, no isolation) is injected for the first statement. Then the user's BEGIN ISOLATION LEVEL SERIALIZABLE is detected and skipped — but the isolation level is never applied. The synthetic BEGIN already went to the backend as a plain BEGIN. Should the adoption path detect the user's isolation level and re-issue SET TRANSACTION ISOLATION LEVEL ... before continuing?

This is to handle the following error case:
```
postgres=# SELECT 1\; show transaction_isolation \; BEGIN ISOLATION LEVEL SERIALIZABLE \; SHOW transaction_isolation \; SELECT 2;
 ?column?
----------
        1
(1 row)

 transaction_isolation
-----------------------
 read committed
(1 row)

ERROR:  SET TRANSACTION ISOLATION LEVEL must be called before any query
postgres=#
```

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -284,6 +321,97 @@ func (sc *ScatterConn) Describe(
...
+		if u.clear {
+			state.ClearReservedConnection(u.target)
+		} else {
+			state.UpdateReservationReasons(u.target, u.remainingReasons)
+		}
+	}
+
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
```
haritabh17, line 429: The caller (TransactionPrimitive.executeCommit) sends this straight to the client as a PG ErrorResponse. PostgreSQL clients expect a single error with a SQLSTATE — a joined multi-line error might confuse ORMs or connection poolers. Would it be better to return the first error with a count, or wrap them into a structured multi-shard error type? For now, I would prefer sending single error.

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -482,14 +609,53 @@ func (sc *ScatterConn) CopyAbort(
...
+		if err := qs.ReleaseReservedConnection(ctx, ss.Target, eo); err != nil {
+			sc.logger.ErrorContext(ctx, "release: RPC failed",
+				"target", ss.Target,
+				"reserved_conn_id", ss.ReservedConnectionId,
+				"error", err)
+		}
+	}
+
+	state.ClearAllReservedConnections()
+	return nil
```
haritabh17: This function always returns nil — individual release errors are logged but never surfaced. The caller in ConnectionClosed checks the error and logs it, but it'll never see one. Either collect and return errors, or change the signature to not return an error to avoid misleading callers.

go/services/multigateway/handler/transaction_helpers.go
GuptaManan100: Good catch on COPY! We actually get a `CommandTag` in the CopyFinalize phase, so this code will still work! I'm finding other issues in the COPY integration with transactions. Fixing them now.

go/multipooler/executor/executor.go
GuptaManan100: You're right — `reservedConn.Query()` buffers the entire result set in memory, while the non-reserved path in `StreamExecute` uses streaming via `QueryStreaming`. For large result sets inside transactions, this could be a real issue.

The fix would be to use `reservedConn.QueryStreaming(ctx, sql, callback)` here instead of `Query()`, which would stream rows as they arrive. This is a straightforward change but I want to make sure the error handling (rollback on callback error) works correctly with the streaming path. I'll address this in a follow-up since it's an optimization that doesn't affect correctness for typical workloads.

go/common/protoutil/reservation.go
GuptaManan100: Good call! Added `protoutil.ValidateReasons()` at the top of `ReserveStreamExecute` in `grpcpoolerservice/service.go`. Returns `codes.InvalidArgument` if any unknown bits are set in the bitmask.

go/multipooler/pools/reserved/reserved_conn.go
GuptaManan100: Removed the `!c.reservedProps.HasPortals()` guard. The `IsEmpty()` check on the reasons bitmask is sufficient if the portal reason bit is set, `IsEmpty()` returns false, so we never clear props while portals exist.

go/services/multigateway/scatterconn/scatter_conn.go
GuptaManan100: Added a guard that counts shards with the transaction reason and logs a warning when > 1. This will serve as a sentinel before DT is implemented.

go/services/multigateway/handler/transaction_helpers.go
GuptaManan100: Great catch! Fixed with a two-case approach:

1. **Deferred BEGIN** (`PendingBeginQuery` still set): Replace it with the user's full `BEGIN ISOLATION LEVEL SERIALIZABLE` so the options are preserved when the reserved connection is created.
2. **Already-executed BEGIN** (`PendingBeginQuery` cleared, queries already ran): Return a `PgDiagnostic` error with SQLSTATE `25001` ("SET TRANSACTION ISOLATION LEVEL must be called before any query") — matching PostgreSQL's native behavior exactly.

This handles both `BEGIN ISOLATION LEVEL SERIALIZABLE; SELECT 1` (case 1, works correctly) and `SELECT 1; BEGIN ISOLATION LEVEL SERIALIZABLE; SELECT 2` (case 2, errors like PostgreSQL).

go/services/multigateway/scatterconn/scatter_conn.go
GuptaManan100: Agreed — changed to return only the first error. When multiple shards fail, the additional errors are logged with the full details but only the first error is surfaced to the client. This keeps the ErrorResponse clean for ORMs and connection poolers.

go/services/multigateway/scatterconn/scatter_conn.go
GuptaManan100: Fixed — errors are now collected during iteration and returned via \`errors.Join()\`. The caller in \`ConnectionClosed\` will now actually see release failures in its log.

go/multipooler/executor/executor.go
GuptaManan100: Fixed

go/services/multigateway/handler/transaction_helpers.go
GuptaManan100: Actually there's quite a few changes, maybe I'll do it as a follow up PR.

go/services/multigateway/handler/transaction_helpers.go
```diff
@@ -0,0 +1,224 @@
...
+					state.PendingBeginQuery = stmt.SqlString()
+				} else {
+					// Backend transaction already has queries executed.
+					// PostgreSQL rejects this with SQLSTATE 25001.
+					return &mterrors.PgDiagnostic{
+						MessageType: 'E',
+						Severity:    "ERROR",
+						Code:        "25001",
+						Message:     "SET TRANSACTION ISOLATION LEVEL must be called before any query",
+					}
```
haritabh17, line 128: When the user sends something like SELECT 1; BEGIN ISOLATION LEVEL SERIALIZABLE; ... and we return this error, is the implicit transaction from SELECT 1 is still open as there's no rollback before returning (the auto-rollback in the execErr block only handles regular statement failures). Should we silentExecute(ast.NewRollbackStmt()) before returning the 25001 error?

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -93,30 +97,56 @@ func (sc *ScatterConn) StreamExecute(
...
+	if conn.IsInTransaction() {
+		sc.logger.DebugContext(ctx, "creating reserved connection with BEGIN for transaction")
+
+		reservationOpts := protoutil.NewTransactionReservationOptions()
+		// Pass the original BEGIN query (e.g., "BEGIN ISOLATION LEVEL SERIALIZABLE")
+		// so the multipooler preserves transaction options instead of using plain "BEGIN".
+		if state.PendingBeginQuery != "" {
+			reservationOpts.BeginQuery = state.PendingBeginQuery
+			state.PendingBeginQuery = ""
+		}
```
haritabh17, line 134: state.PendingBeginQuery is cleared after the first shard consumes it. If a multi-shard transaction routes to a second shard, that shard gets a plain BEGIN instead of BEGIN ISOLATION LEVEL SERIALIZABLE. Should this be preserved for the transaction lifetime and only cleared on commit/rollback?

Dont think this needs to be addressed immediately, lets keep this noted.

go/services/multigateway/handler/transaction_helpers.go
GuptaManan100: Fixed!

go/services/multigateway/scatterconn/scatter_conn.go
GuptaManan100: Good catch. For the single-shard MVP this is fine. Longer term, I'm planning to restructure the ScatterConn interface - instead of taking separate tableGroup/shard strings, it should accept a Target that can span multiple shards. In that model, the transaction engine would hold the BEGIN query (with isolation level) as part of the transaction state, and each shard's ReserveStreamExecute would receive it when first joining the transaction. That way the isolation level is preserved across all shards naturally, rather than being a one-shot field consumed by the first shard. Added a NOTE comment to capture this for now.

go/services/multigateway/handler/transaction_helpers.go
GuptaManan100: Here is the follow up - #635

PR #591, opened 2026-02-06
feat(multiorch): fixes bug in discovery for consensus

rafael:
# Desc
- Fixes bug on discovery in the consensus implementation. We were selecting the candidate prior to revoking. 

# Tests
- Added unit tests.
- Integration tests did not cover this case (hard to repro at the moment). Existing tests continue to pass.

### DIFF
--- a/go/multipooler/manager/rpc_consensus_test.go
+++ b/go/multipooler/manager/rpc_consensus_test.go
@@ -135,6 +135,7 @@ func TestBeginTerm(t *testing.T) {
 		expectedAccepted                    bool
 		expectedTerm                        int64
 		expectedAcceptedTermFromCoordinator string
+		expectedWalPosition                 *consensusdatapb.WALPosition // nil means don't check
 		description                         string
 	}{
 		{
@@ -170,6 +171,7 @@ func TestBeginTerm(t *testing.T) {
 			expectedAccepted:                    true,
 			expectedTerm:                        10,
 			expectedAcceptedTermFromCoordinator: "candidate-B",
+			expectedWalPosition:                 &consensusdatapb.WALPosition{LastReceiveLsn: "0/2000000", LastReplayLsn: "0/2000000"},
 			description:                         "Acceptance should succeed when request term is newer than current term, even if already accepted leader in older term",
 		},
 		{
@@ -230,6 +232,7 @@ func TestBeginTerm(t *testing.T) {
 			expectedAccepted:                    true,
 			expectedTerm:                        5,
 			expectedAcceptedTermFromCoordinator: "candidate-A",
+			expectedWalPosition:                 &consensusdatapb.WALPosition{LastReceiveLsn: "0/3000000", LastReplayLsn: "0/3000000"},
 			description:                         "Acceptance should succeed when already accepted same candidate in same term (idempotent)",
 		},
 		{
@@ -287,6 +290,7 @@ func TestBeginTerm(t *testing.T) {
 			expectedAccepted:                    true,
 			expectedTerm:                        10,
 			expectedAcceptedTermFromCoordinator: "new-candidate",
+			expectedWalPosition:                 &consensusdatapb.WALPosition{LastReceiveLsn: "0/4000000", LastReplayLsn: "0/4000000"},
 			description:                         "Primary should accept term after successful demotion (idempotent case - already demoted)",
 		},
 		{
@@ -319,6 +323,7 @@ func TestBeginTerm(t *testing.T) {
 			expectedAccepted:                    true,
 			expectedTerm:                        10,
 			expectedAcceptedTermFromCoordinator: "new-candidate",
+			expectedWalPosition:                 &consensusdatapb.WALPosition{LastReceiveLsn: "0/5000000", LastReplayLsn: "0/5000000"},
 			description:                         "Standby accepts term with REVOKE action and pauses replication",
 		},
 		{
@@ -350,6 +355,7 @@ func TestBeginTerm(t *testing.T) {
 			expectedAccepted:                    true,
 			expectedTerm:                        10,
 			expectedAcceptedTermFromCoordinator: "new-candidate",
+			expectedWalPosition:                 &consensusdatapb.WALPosition{LastReceiveLsn: "0/6000000", LastReplayLsn: "0/6000000"},
 			description:                         "Standby should pause replication when accepting new term",
 		},
 		{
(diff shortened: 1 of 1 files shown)


go/services/multiorch/coordinator/leader_appointment.go
```diff
@@ -49,36 +48,50 @@ import (
 func (c *Coordinator) BeginTerm(ctx context.Context, shardID string, cohort []*multiorchdatapb.PoolerHealthState, quorumRule *clustermetadatapb.QuorumRule, proposedTerm int64) (*multiorchdatapb.PoolerHealthState, []*multiorchdatapb.PoolerHealthState, int64, error) {
 	c.logger.InfoContext(ctx, "Beginning term", "shard", shardID, "term", proposedTerm)
 
-	// Stage 1: Select Candidate - Choose based on WAL position
```
rafael, line 52: This is the main fix here

proto/consensusdata.proto
```diff
@@ -37,6 +37,10 @@ message WALPosition {
 
   // Timestamp when this position was recorded
   google.protobuf.Timestamp timestamp = 4;
+
+  // Reserved for future use: the term for which this node was promoted to primary.
+  // Set to 0 for now.
+  int64 primary_term = 5;
```
rafael: We are not using this yet, but we can use it when we get to implement the multiple failovers in a row

go/multipooler/manager/rpc_consensus.go
```diff
@@ -115,6 +115,36 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
 		PoolerId: pm.serviceID.GetName(),
 	}
 
+	// Populate WAL position for candidate selection
```
dweitzman: The WAL position shouldn't be determined until after doing the revoke action, I believe? At this point in the code the revoke (if requested) hasn't be applied, so the WAL position could continue to advance

go/services/multiorch/coordinator/leader_appointment.go
```diff
@@ -49,36 +48,50 @@ import (
...
+	if len(recruited) == 0 {
+		return nil, nil, 0, mterrors.New(mtrpcpb.Code_UNAVAILABLE,
+			"no poolers accepted the term")
+	}
 
-	// Stage 2: Recruit Nodes - Send BeginTerm RPC to all poolers in parallel
-	// Use REVOKE action to ensure old primary is demoted and standbys stop replicating
-	recruited, err := c.recruitNodes(ctx, cohort, proposedTerm, candidate, consensusdatapb.BeginTermAction_BEGIN_TERM_ACTION_REVOKE)
+	// Select Candidate - Choose from recruited nodes based on WAL position
+	// This eliminates the race condition
```
dweitzman: The comment "This eliminates the race condition" can probably be removed, or alternatively replaced with a comment explaining why the candidate is chosen during REVOKE instead of before?

go/multipooler/manager/rpc_consensus_test.go
```diff
@@ -396,7 +425,7 @@ func TestBeginTerm(t *testing.T) {
 			description:                         "NO_ACTION still respects term acceptance rules",
 		},
 		{
-			name:   "PostgresDownAcceptsTermButRevokeActionFails",
+			name:   "PostgresDownRejectsTerm_TemporaryBehavior",
```
dweitzman: This test name seems inconsistent with the test expectations?

go/services/multiorch/coordinator/leader_appointment_test.go
```diff
@@ -173,89 +196,311 @@ func TestSelectCandidate(t *testing.T) {
...
 
-		candidate, err := c.selectCandidate(ctx, cohort)
+		_, err := c.selectCandidate(ctx, recruited)
 		require.Error(t, err)
-		require.Nil(t, candidate)
-		require.Contains(t, err.Error(), "no healthy poolers available")
+		require.Contains(t, err.Error(), "no valid candidate found")
+	})
+
+	t.Run("selects primary with highest CurrentLsn among multiple primaries", func(t *testing.T) {
```
dweitzman, line 395: Is it worth adding a note here or in the relevant logic that we're trusting the primaries had sync replication configured correctly and therefore don't have diverging/split-brain timelines?

proto/consensusdata.proto
```diff
@@ -86,6 +90,9 @@ message BeginTermResponse {
   // before demotion. Standbys should replicate up to this point before switching
   // to a new primary. Empty if node was not a primary or didn't demote.
   string demote_lsn = 4;
+
+  // WAL position for candidate selection
+  WALPosition wal_position = 5;
```
dweitzman, line 91: Not a blocking comment, but something to think about for simplicity/API design

Feels like demote_lsn & wal_position here are a little bit competing with each other, and might only be relevant for certain actions?

I'm wondering if in the long term it might make sense for each action type to get its own response message, so the wal position here would be part of RevokeActionResponse if it's produced by the action of revoking

go/multipooler/manager/rpc_consensus.go
rafael: Yes! Good catch.

proto/consensusdata.proto
rafael: Agree. Another alternative for now is just to remove `demote_lsn` ? I think in this context, we can just use WalPosition always.

dweitzman: Sure, removing demote_lsn seems like a step in a better direction

go/services/multiorch/coordinator/leader_appointment.go
```diff
@@ -114,123 +125,120 @@ func (c *Coordinator) discoverMaxTerm(cohort []*multiorchdatapb.PoolerHealthStat
...
-				// For standbys, use receive position (includes unreplayed WAL)
-				status.walPosition = resp.WalPosition.LastReceiveLsn
-			}
-		}
-
-		statuses = append(statuses, status)
-	}
-
-	if len(statuses) == 0 {
+// selectCandidate chooses the best candidate from recruited poolers.
```
rafael, line 128: @sougou - is this clear and matching your mental model?

proto/consensusdata.proto
rafael: I created a ticket for this to follow up in a different PR.

go/multipooler/manager/rpc_consensus_test.go
dweitzman, line 1: Should any of these `Revoke` tests be inspecting the `WalPosition` value in the response?

go/services/multiorch/coordinator/leader_appointment.go
```diff
@@ -127,8 +125,48 @@ func (c *Coordinator) discoverMaxTerm(cohort []*multiorchdatapb.PoolerHealthStat
...
+//   - This provides logical ordering across conflicting timelines.
+//   - Limitation: Relies on clock synchronization (we accept theoretical
+//     clock skew for now).
+//   - TODO: Update selectCandidate to use timestamps
+//
+// Why *NOT* use PrimaryTerm?
+//   - We track PrimaryTerm at the pooler/coordinator level (which term the
+//     coordinator is operating in).
+//   - But we DON'T track which term each individual WAL transaction belongs to.
+//   - Without per-transaction term numbers embedded in WAL, we can't determine
```
dweitzman, line 162: This is a future problem, but it feels like we could have multipoolers save LSN ranges + primary term numbers to disk during propagation or reparenting so we could know what transaction each term belongs to deterministically without needing to wait for the WAL to be applied and run a live query

rafael: :nods: - Makes sense. We need to flush all these details for the two phase sync

PR #592, opened 2026-02-07
MUL-153 feat: forward client startup parameters to backend PostgreSQL

haritabh17:
Fixes startup parameters like PGDATESTYLE and PGOPTIONS being ignored by multigres.

Previously, when clients set parameters during connection startup (e.g., `PGDATESTYLE="Postgres, MDY"`), multigres would parse them but not forward to PostgreSQL. Clients always received hardcoded defaults like `DateStyle = "ISO, MDY"` regardless of what they requested.

Now startup params are properly forwarded. The implementation validates them during connection setup by establishing a backend connection before sending ReadyForQuery. Invalid params fail the connection immediately (matching PostgreSQL), and ParameterStatus messages reflect actual backend values. Connection pooling routes based on combined startup + session settings.

Example:
```bash
$ PGDATESTYLE="Postgres, MDY" psql -h localhost -p 15432
postgres=> show datestyle;
 DateStyle
-------------
 Postgres, MDY  ✅
```

### DIFF
--- a/go/common/pgprotocol/server/startup.go
+++ b/go/common/pgprotocol/server/startup.go
@@ -334,13 +334,6 @@ func (c *Conn) authenticateTrust() error {
 		return fmt.Errorf("failed to send BackendKeyData: %w", err)
 	}
 
-	// Validate startup parameters if the handler supports it.
-	if validator, ok := c.handler.(StartupValidator); ok {
-		if err := validator.ValidateStartup(c.ctx, c); err != nil {
-			return fmt.Errorf("startup parameter validation failed: %w", err)
-		}
-	}
-
 	// Send initial ParameterStatus messages.
 	if err := c.sendParameterStatuses(); err != nil {
 		return fmt.Errorf("failed to send ParameterStatus messages: %w", err)
@@ -426,13 +419,6 @@ func (c *Conn) authenticateSCRAM() error {
 	// Send BackendKeyData for query cancellation.
 	if err := c.sendBackendKeyData(); err != nil {
 		return fmt.Errorf("failed to send BackendKeyData: %w", err)
-	}
-
-	// Validate startup parameters if the handler supports it.
-	if validator, ok := c.handler.(StartupValidator); ok {
-		if err := validator.ValidateStartup(c.ctx, c); err != nil {
-			return fmt.Errorf("startup parameter validation failed: %w", err)
-		}
 	}
 
 	// Send initial ParameterStatus messages.
--- a/go/common/pgprotocol/server/handler.go
+++ b/go/common/pgprotocol/server/handler.go
@@ -96,10 +96,3 @@ type Handler interface {
 	// Called at the end of an extended query cycle to indicate transaction boundary.
 	HandleSync(ctx context.Context, conn *Conn) error
 }
-
-// StartupValidator is an optional interface that handlers can implement
-// to validate startup parameters during connection establishment.
-// If the handler does not implement this interface, validation is skipped.
-type StartupValidator interface {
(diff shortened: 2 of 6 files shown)


go/common/pgprotocol/server/conn.go
```diff
@@ -179,6 +179,22 @@ func (c *Conn) Database() string {
 	return c.database
 }
 
+// GetStartupParams returns the startup parameters sent by the client,
+// excluding 'user' and 'database' which are handled separately.
+func (c *Conn) GetStartupParams() map[string]string {
+	result := make(map[string]string, len(c.params))
+	for k, v := range c.params {
+		if k == "user" || k == "database" {
```
GuptaManan100, line 187: Are they required to be lowercased, thinking about the case where someone sends over `User` as the parameter name instead of `user`

go/common/pgprotocol/server/startup.go
```diff
@@ -240,6 +334,13 @@ func (c *Conn) authenticateTrust() error {
 		return fmt.Errorf("failed to send BackendKeyData: %w", err)
 	}
 
+	// Validate startup parameters if the handler supports it.
+	if validator, ok := c.handler.(StartupValidator); ok {
+		if err := validator.ValidateStartup(c.ctx, c); err != nil {
+			return fmt.Errorf("startup parameter validation failed: %w", err)
+		}
+	}
+
```
GuptaManan100: I'm actually thinking do we need any validation at all. What's the worst that can happen, the user gets an error when they run a query, right? I'm thinking that's better than running a query all the way to postgres on connection.

go/common/pgprotocol/server/conn.go
haritabh17: Yupp, they must be lowercase only.

```
haritabhgupta@Haritabhs-MacBook-Pro multigres % PGOPTIONS="-c work_mem=64MB" PGPORT=5433 PGHOST=localhost PGDATABASE=postgres PGUSER=haritabhgupta psql --User haritabhgupta
psql: unrecognized option `--User'
psql: hint: Try "psql --help" for more information.
haritabhgupta@Haritabhs-MacBook-Pro multigres % PGOPTIONS="-c work_mem=64MB" PGPORT=5433 PGHOST=localhost PGDATABASE=postgres PGUSER=haritabhgupta psql --USER haritabhgupta
psql: unrecognized option `--USER'
psql: hint: Try "psql --help" for more information.
haritabhgupta@Haritabhs-MacBook-Pro multigres %
```

go/common/pgprotocol/server/startup.go
haritabh17: Ack, updating this to not do any validation. We can add validation later as needed. Without validation, the user connection will become unusable as every query on that conn will hit invalid param error, so user will have to discard that conn to multigateway

go/services/multigateway/handler/connection_state.go
```diff
@@ -44,6 +44,10 @@ type MultiGatewayConnectionState struct {
 	// It keeps track of any reserved connections on each Shard currently open.
 	ShardStates []*ShardState
 
+	// StartupParams stores startup parameters from the client's StartupMessage.
+	// These are forwarded to the backend for every query.
+	StartupParams map[string]string
+
```
GuptaManan100, line 50: Just a comment here stating why this needs to be different from Settings would be helpful. (I can see the code and understand why, but it would be nice to have a comment for future reference)
Doesn't have to be in this PR, can be in the follow up one that you're intending to do.

GuptaManan100 (approved):
Rest LGTM! Feel free to merge 😍

PR #593, opened 2026-02-09
fix(reserved-pool): return pool slot when closing reserved connections

haritabh17:
## Problem

`reserved.Conn.Close()` was closing the TCP socket but not returning the pool slot to the underlying connection pool, causing permanent slot leaks.

Each leaked slot reduced the pool's effective capacity by 1. When all slots were leaked (typically 2 in initial capacity), new connection requests would block forever on `waitForConn()`, causing deadlock.

This manifested as a flaky test failure (1 in 111 runs) in `TestMultiGateway_CopyCommands`. The test only failed when running before the rebalancer increased capacity from 2→20 (10s window), and when COPY error paths leaked both slots via `Close()` calls.

## Root Cause

`Close()` only called `c.pooled.Close()` which closes the socket but doesn't call `pool.put()` to decrement the `active` counter or return the slot.

When `getNew()` tried to create a new connection, it saw `active >= capacity` and returned nil, forcing callers to block on `waitForConn()` indefinitely.

## Fix

Added `c.pooled.Taint()` before `c.pooled.Close()`:

```go
c.pooled.Taint()    // calls pool.put(nil) -> creates replacement connection
c.pooled.Close()    // closes the socket
```

`Taint()` calls `pool.put(nil)` which either:
- Creates a fresh replacement connection and hands it to a blocked waiter (unblocks immediately)
- Pushes the replacement to the idle stack for next caller
- On failure: calls `closedConn()` to decrement `active` (frees the slot)

This mirrors the cleanup pattern already used in `Pool.Close()` at `reserved_pool.go:293-294`.

## Testing

Added comprehensive unit tests:
- `TestCloseReturnsPoolSlot` - verifies Close() returns slots and prevents pool exhaustion
- `TestCloseRemovesFromActiveMap` - verifies Close() removes from reserved pool's active map
- `TestCloseIdempotency` - verifies Close() is idempotent and safe with Release()

All tests pass including the previously flaky `TestMultiGateway_CopyCommands`.

### DIFF
--- a/go/multipooler/pools/reserved/reserved_pool_test.go
+++ b/go/multipooler/pools/reserved/reserved_pool_test.go
@@ -516,3 +516,150 @@ func TestPool_NewConnAfterPoolRecreation(t *testing.T) {
 		conn.Release(ReleaseCommit)
 	}
 }
+
+func TestConn_Close_ReturnsPoolSlot(t *testing.T) {
+	// This test verifies that Close() properly returns the pool slot to the
+	// underlying connpool, preventing pool exhaustion. Before the fix, Close()
+	// only closed the socket without returning the slot, causing a permanent leak.
+	server := fakepgserver.New(t)
+	defer server.Close()
+	server.SetNeverFail(true)
+
+	// Create a pool with small capacity to easily exhaust it.
+	pool := NewPool(context.Background(), &PoolConfig{
+		InactivityTimeout: 5 * time.Second,
+		RegularPoolConfig: &regular.PoolConfig{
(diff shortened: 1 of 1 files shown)


GuptaManan100 (commented):
I actually don't think this is the right approach. Close is meant to be a low level operation that only closes the connection. That is the same in regular conn too. Instead from the executor, we should call Release. We should remove this Close function altogether. It's not being called from anywhere else. Let me fix it.

GuptaManan100:
@haritabh17 Look at https://github.com/multigres/multigres/pull/597 instead please

PR #596, opened 2026-02-09
feat(build): add docker build GitHub Action for multiadmin-web

rytswd:
This is to extend the Docker build setup for multiadmin-web, so that the image is publicly available using GHCR.


rytswd:
CC @rafael

sougou (approved):

PR #598, opened 2026-02-10
feat(multipooler): removes demote lsn

rafael:
# Desc
* Removes field that is redundant now

# Tests
* Updated relevant tests

### DIFF
--- a/proto/consensusdata.proto
+++ b/proto/consensusdata.proto
@@ -82,13 +82,8 @@ message BeginTermResponse {
   // ID of the responding pooler
   string pooler_id = 3;
 
-  // If this node was a primary and demoted itself, this contains the final LSN
-  // before demotion. Standbys should replicate up to this point before switching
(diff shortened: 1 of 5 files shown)


proto/consensusdata.proto
```diff
@@ -82,13 +82,8 @@ message BeginTermResponse {
...
   string pooler_id = 3;
 
-  // If this node was a primary and demoted itself, this contains the final LSN
-  // before demotion. Standbys should replicate up to this point before switching
-  // to a new primary. Empty if node was not a primary or didn't demote.
-  string demote_lsn = 4;
-
   // WAL position for candidate selection
-  WALPosition wal_position = 5;
+  WALPosition wal_position = 4;
```
dweitzman, line 86: The field number reuse here seems harmless since we're still prerelease

rafael: Right, I've been avoiding adding reserved fields for deprecation, since we can still change without breaking anything

PR #601, opened 2026-02-11
fix(connpool): tie connection lifetime to pool context instead of caller context

GuptaManan100:
### Summary                                                                                                                                    

- Connections created by a pool now have their lifetime context derived from the pool's lifecycle context (poolCtx) instead of the caller's request context (ctx). Previously, when a caller's context was cancelled (e.g., gRPC deadline), the connection's context was also cancelled, even though the connection goes back to the pool for reuse.
- The Connector type and client.Connect now accept a separate poolCtx parameter, cleanly separating the dial/startup context from the connection lifetime context.

### DIFF
--- a/go/multipooler/pools/connpool/pool_test.go
+++ b/go/multipooler/pools/connpool/pool_test.go
@@ -195,6 +195,47 @@ func TestPoolClose(t *testing.T) {
 	conn2.Recycle()
 }
 
+func TestPoolCallerContextCancelDoesNotCloseConnection(t *testing.T) {
+	// Verify that cancelling the caller's context (used for Get) does not
+	// close the underlying connection. The connection's lifetime should be
+	// tied to the pool's context, not the caller's.
+	var connPoolCtx context.Context
+	pool := NewPool[*mockConnection](context.Background(), &Config{
+		Name:         "test",
+		Capacity:     1,
+		MaxIdleCount: 1,
+	})
+	pool.Open(func(ctx context.Context, poolCtx context.Context) (*mockConnection, error) {
+		connPoolCtx = poolCtx
+		return newMockConnection(), nil
+	}, nil)
+	defer pool.Close()
+
+	// Get a connection using a cancellable context.
+	callerCtx, callerCancel := context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


go/multipooler/pools/connpool/pool_test.go
```diff
@@ -65,7 +65,7 @@ func newTestPool(capacity int64) *Pool[*mockConnection] {
 		Capacity:     capacity,
 		MaxIdleCount: capacity,
 	})
-	pool.Open(func(ctx context.Context) (*mockConnection, error) {
```
rafael, line 68: Could we add a test, validating that cancelling the context of a client, doesn't close the underlying connection in the pool? Is that somewhere? I did not see it

go/test/endtoend/multipooler/pgprotocol_client_test.go
```diff
@@ -40,7 +40,7 @@ func TestPgProtocolClientConnection(t *testing.T) {
 	t.Run("connect_and_close", func(t *testing.T) {
 		ctx := utils.WithTimeout(t, 10*time.Second)
 
-		conn, err := client.Connect(ctx, &client.Config{
+		conn, err := client.Connect(ctx, ctx, &client.Config{
```
rafael, line 43: not really necessary for this PR, but we've been using t.Context  / utils.WithTimeout , to get contexts in tests.

rafael (approved):
LGTM

go/multipooler/pools/connpool/pool_test.go
GuptaManan100: Done!

go/test/endtoend/multipooler/pgprotocol_client_test.go
GuptaManan100: Not sure, what I should change. The test is using `utils.WithTimeout`

haritabh17 (approved):

go/test/endtoend/multipooler/pgprotocol_client_test.go
rafael: Oh my bad. I misread how the context was created. This looks good

PR #604, opened 2026-02-11
feat(multipooler):  cohort durability when adding new replicas 

rafael:
# Desc

Extends the `leadership_history` table to audit replication configuration changes (add/remove/replace standbys) alongside existing promotion events. This will make sure that a replica doesn't get added to the standby list, unless the information has already been persisted in the history. 

In a follow up PR, we will start returning this data in the BeginTerm response and this will guarantee that we will discover the latest list of replicas in the cluster. 

## Test
- Added unit tests to cover this new codepath. If we fail to write to the table, ConfigureSyncReplication will now fail.

### DIFF
--- a/go/multipooler/manager/rpc_manager.go
+++ b/go/multipooler/manager/rpc_manager.go
@@ -393,27 +393,30 @@ func (pm *MultiPoolerManager) configureSynchronousReplicationLocked(ctx context.
 		)
 	}
 
-	// Insert history before making changes
-	if pm.consensusState != nil {
-		term, err := pm.consensusState.GetTerm(ctx)
-		if err == nil && term != nil {
-			// Convert standby IDs to application names for history
-			standbyNames := make([]string, len(standbyIDs))
-			for i, id := range standbyIDs {
-				standbyNames[i] = generateApplicationName(id)
-			}
-			if err := pm.insertHistoryRecord(ctx,
-				term.GetPrimaryTerm(),
-				"replication_config",
-				"", "", "", // leaderID, coordinatorID, walPosition
-				"configure",
-				"ConfigureSynchronousReplication called",
-				standbyNames,
-				nil, // acceptedMembers
-				force); err != nil {
-				return mterrors.Wrap(err, "failed to record replication config history")
-			}
-		}
+	// Insert history before applying GUCs.
+	// Rationale: we want to ensure that a new cohort is advertised
+	// before this primary can accept ACKs from it.
+	// This is for safe replica joining of the cluster.
+	// It will ensure multiorch can discover the new cohort during a failure.
+	term, err := pm.consensusState.GetTerm(ctx)
+	if err != nil {
+		return mterrors.Wrap(err, "failed to get consensus term")
+	}
+	// Convert standby IDs to application names for history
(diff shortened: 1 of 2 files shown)


go/multipooler/executor/mock/mock_querier.go
```diff
@@ -170,6 +170,13 @@ func (m *QueryService) Query(ctx context.Context, queryStr string) (*sqltypes.Re
 	return matched.result, nil
 }
 
+// QueryMultiStatement implements executor.InternalQueryService.
+// For the mock, this delegates to Query (result is discarded).
+func (m *QueryService) QueryMultiStatement(ctx context.Context, queryStr string) error {
```
rafael, line 175: I did not end up using this, but seems ok to keep this method. It could become handy in the future.

go/test/endtoend/multipooler/rpc_manager_replication_api_test.go
```diff
@@ -1463,6 +1463,7 @@ func TestConfigureSynchronousReplication(t *testing.T) {
 			NumSync:           1,
 			StandbyIds:        []*clustermetadatapb.ID{makeMultipoolerID("test-cell", "test-standby")},
 			ReloadConfig:      true,
+			Force:             true, // Required: fake standbys won't connect, so history insert would timeout
```
rafael, line 1466: these tests don't have a proper standby configured, so forcing so we can assert all the other behaviors.

go/multipooler/manager/rpc_manager.go
```diff
@@ -387,6 +387,29 @@ func (pm *MultiPoolerManager) configureSynchronousReplicationLocked(ctx context.
 		return err
 	}
 
+	// Insert history before making changes
+	if pm.consensusState != nil {
+		term, err := pm.consensusState.GetInconsistentTerm()
```
dweitzman: Any reason this needs to be an inconsistent term? Seems liek this method already has the action lock

go/multipooler/manager/rpc_manager.go
```diff
@@ -516,6 +531,47 @@ func (pm *MultiPoolerManager) UpdateSynchronousStandbyList(ctx context.Context,
 		return nil
 	}
 
+	// Map operation enum to string
+	var operationName string
```
dweitzman: Maybe this mapping should be extracted into a helper function?

go/multipooler/manager/rpc_manager.go
rafael: good catch. I'll update

go/multipooler/manager/rpc_manager.go
```diff
@@ -387,6 +387,35 @@ func (pm *MultiPoolerManager) configureSynchronousReplicationLocked(ctx context.
...
 	}
 
+	if pm.consensusState == nil {
+		return mterrors.New(mtrpcpb.Code_FAILED_PRECONDITION,
+			"consensus state must be available to configure synchronous replication",
+		)
+	}
+
+	// Insert history before making changes
+	if pm.consensusState != nil {
```
dweitzman: Might be worth a comment on why this inserts history and then updates the GUC, which you mentioned offline was for adding a new replica to an existing cohort

PR #611, opened 2026-02-12
feat(multiadmin): proxy queries to API in the nextjs app

rafael:
# Desc
- In kubernetes environments, the API url is not gonna be easily accessible by the frontend. Let's make sure that only the nextjs app talks to the mutliadmin API.

### DIFF
--- a/web/multiadmin/next-env.d.ts
+++ b/web/multiadmin/next-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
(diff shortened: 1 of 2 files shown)


web/multiadmin/tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
```
rafael, line 1: linter changes

web/multiadmin/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
```
rafael: nextjs updates this automatically for me. I did not make this update

Issue #567, opened 2026-02-02
isReplicaConnectedToPrimary should also check heartbeat


sougou:
I was testing a local kind cluster and noticed that a pod delete never results in a failover. Multiorch always waits for it to restart:

```
{"time":"2026-02-02T01:32:56.330290219Z","level":"WARN","msg":"primary pooler unreachable but postgres still running","shard_key":"postgres/default/0-inf","primary_pooler_id":"multipooler-zone1-0","action":"operator should restart pooler process"}
```

I believe this is because of how https://github.com/multigres/multigres/blob/574c4492d37d199374c24df4e06799656ba09857/go/services/multiorch/recovery/analysis/generator.go#L458-L490 is implemented. The function only checks if the connection was successfully established. So, this will always be true, even if a replica loses connection to the primary.

I believe it should also check if the replica is receiving heartbeats.

PR #615, opened 2026-02-13
feat(multiorch): use heartbeat in dead primary detection

rafael:
## Description
Closes a gap in the dead primary analyzer by strengthening the `isReplicaConnectedToPrimary` check with additional signals:
  - **WAL receiver streaming**: replica must have an active WAL receiver
    connection to the primary.
  - **Heartbeat health**: replica must have a fresh heartbeat from the primary,
    proving the primary was recently writing and replication is working.
  - **Lag guardrail**: a lagging replica is allowed to have a stale
    heartbeat without counting against the primary — that's a replication
    problem, not evidence of primary failure.

In my view, this brings to parity with what we had in vitess when it comes to detecting dead primaries. I also added a new analyzer that is relevant for multigres (RevokePrimary). 

## Notes
- I think I discovered a few issues in the conn pool that I need to review with @GuptaManan100.
- There is also something around pgbackrest that I need to review with @cuongdo 
- I left notes in the PR to make it easier to review.

## Tests

* Added unit tests for the new signals in `isReplicaConnectedToPrimary` (WAL receiver, heartbeat health, lag guardrail).

-----
* Fixes: MUL-74
* Fixes: https://github.com/multigres/multigres/issues/567

### DIFF
--- a/go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
+++ b/go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
@@ -23,8 +23,8 @@ import (
 	"github.com/multigres/multigres/go/services/multiorch/store"
 )
 
-// PrimaryIsDeadAnalyzer detects when a primary is unreachable, ALL replica poolers
-// are reachable, and none is connected to the primary. Strongest signal for failover.
+// PrimaryIsDeadAnalyzer detects when a primary is unreachable and all replica poolers
+// are reachable but none confirm the primary is alive. Strongest signal for failover.
 type PrimaryIsDeadAnalyzer struct {
 	factory *RecoveryActionFactory
 }
@@ -60,9 +60,11 @@ func (a *PrimaryIsDeadAnalyzer) Analyze(poolerAnalysis *store.ReplicationAnalysi
 		return nil, nil
 	}
 
-	// All replicas confirm primary is alive (streaming WAL + healthy heartbeat).
-	// Primary pooler may be down but Postgres is still running — don't failover.
-	if poolerAnalysis.AllReplicasConfirmPrimaryAlive {
+	// Any replica confirming primary is alive (streaming WAL + healthy heartbeat)
+	// is sufficient evidence the primary Postgres is still running — don't failover.
+	// We've already verified all replica poolers are reachable above, if we haven't
+	// found a connected replica then we know the primary has no connected replicas.
+	if poolerAnalysis.CountReplicasConfirmingPrimaryAliveInShard > 0 {
 		return nil, nil
 	}
 
--- a/go/services/multiorch/recovery/analysis/primary_is_dead_analyzer_test.go
+++ b/go/services/multiorch/recovery/analysis/primary_is_dead_analyzer_test.go
@@ -211,6 +211,28 @@ func TestPrimaryIsDeadAnalyzer_Analyze(t *testing.T) {
 		require.Nil(t, problem, "no replicas to failover to")
 	})
 
+	t.Run("does not trigger when any replica confirms primary is alive", func(t *testing.T) {
+		// If 1 of 2 replicas has a fresh heartbeat from the primary, that's enough evidence
+		// the primary is alive — don't failover due to a single replica's stale heartbeat.
+		primaryID := &clustermetadatapb.ID{Component: clustermetadatapb.ID_MULTIPOOLER, Cell: "zone1", Name: "primary1"}
+		analysis := &store.ReplicationAnalysis{
+			PoolerID:                            &clustermetadatapb.ID{Component: clustermetadatapb.ID_MULTIPOOLER, Cell: "zone1", Name: "replica1"},
+			ShardKey:                            commontypes.ShardKey{Database: "db", TableGroup: "tg", Shard: "0"},
+			IsPrimary:                           false,
+			IsInitialized:                       true,
+			PrimaryPoolerID:                     primaryID,
+			PrimaryReachable:                    false,
+			AllReplicasConfirmPrimaryAlive:      false, // NOT all — only 1 of 2
+			CountReplicaPoolersInShard:          2,
+			CountReachableReplicaPoolersInShard: 2,
+			CountReplicasConfirmingPrimaryAliveInShard: 1, // one replica sees the primary as alive
+		}
+
+		problem, err := analyzer.Analyze(analysis)
+		require.NoError(t, err)
+		require.Nil(t, problem, "should not trigger failover when any replica confirms primary is alive")
+	})
+
 	t.Run("does not trigger when no poolers reachable", func(t *testing.T) {
 		primaryID := &clustermetadatapb.ID{Component: clustermetadatapb.ID_MULTIPOOLER, Cell: "zone1", Name: "primary1"}
 		analysis := &store.ReplicationAnalysis{


go/common/constants/consensus.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+package constants
+
+import "time"
+
+// HeartbeatWriteInterval is the cadence at which the primary writes heartbeats
+// to the multigres.heartbeat table. Used by the multipooler heartbeat writer
+// and by multiorch to determine heartbeat staleness.
+const HeartbeatWriteInterval = 1 * time.Second
```
rafael, line 22: Brought this as a common constant as both consensus and multipooler writer would depend on this

go/common/mterrors/code.go
```diff
@@ -32,11 +32,15 @@ var (
 	// MT13002 Pooler Type Mismatch
 	MT13002 = errorWithoutState("MT13002", mtrpcpb.Code_FAILED_PRECONDITION, "pooler type mismatch: topology says %s but PostgreSQL is %s", "The pooler type in the topology does not match the actual PostgreSQL role. This indicates the pooler is in an inconsistent state and requires intervention.")
 
+	// MT13003 Heartbeat Stale
+	MT13003 = errorWithoutState("MT13003", mtrpcpb.Code_UNAVAILABLE, "heartbeat stale", "The heartbeat reader has not received a fresh heartbeat within the expected interval, indicating the primary may have stopped writing or replication is broken.")
```
rafael, line 36: Also, thought it might be sensible to make this a first class citizen error

go/multipooler/manager/pg_replication.go
```diff
@@ -310,6 +310,28 @@ func (pm *MultiPoolerManager) resetPrimaryConnInfo(ctx context.Context) error {
 	return nil
 }
 
+// disableRestoreCommand disables the restore_command to prevent the standby from
+// fetching WAL from the archive. After a REVOKE, the old primary may still archive
+// divergent WAL (e.g. locally-committed writes from cancelled sync rep waits).
+// Without this, archive-get could apply that divergent WAL before promotion.
+func (pm *MultiPoolerManager) disableRestoreCommand(ctx context.Context) error {
```
rafael, line 317: @cuongdo would love to sync about this with you.

After a REVOKE, I was still able to get WAL applied on a standby because it was being fetched from the archive via restore_command. That means we can accidentally apply divergent WAL (for example WAL archived by the old primary after REVOKE) before promotion.

As a stopgap I disabled restore_command during REVOKE and it fixes the issue, but I have not implemented the full lifecycle yet (i.e restore, the restore command)

go/multipooler/manager/rpc_consensus.go
```diff
@@ -171,10 +173,18 @@ func (pm *MultiPoolerManager) executeRevoke(ctx context.Context, term int64, res
 		}
 		response.WalPosition.CurrentLsn = demoteResp.LsnPosition
 		pm.logger.InfoContext(ctx, "Primary demoted", "lsn", demoteResp.LsnPosition, "term", term)
+		return mterrors.Errorf(mtrpcpb.Code_UNAVAILABLE, "primary demoted and postgres stopped (lsn=%s)", demoteResp.LsnPosition)
```
rafael, line 176: Why return an error on primary REVOKE?
                                                                                                                                                                      Not returning an error here was a bug (we just didn't have a test covering it). The scenario:            
1. Primary fails, multiorch begins recovery by revoking standbys.
2.  The old primary comes back online and receives the REVOKE RPC.
3. It shuts down postgres and returns success.
4. Because it returned success, multiorch includes it in the recruited nodes.
5. As the former primary, it likely has the most advanced LSN, making it the best failover candidate.
6.  Promotion fails because postgres is intentionally off.

Returning an error causes multiorch to exclude this node from the candidate set. 

I considered handling this on the multiorch side, but it's cleaner to reject at the source: a primary should never be a valid REVOKE target.

go/multipooler/pools/regular/regular_conn.go
```diff
@@ -342,15 +342,19 @@ func isConnectionError(err error) bool {
 
 // handleContextCancellation cancels the backend query if adminPool is available.
 // This is called when the context is cancelled while a query is in progress.
+// If cancellation fails (e.g. pool context expired), the connection is closed
```
rafael: This is one of the issues I need to discuss with @GuptaManan100. 

Without this change, I was consistently reproducing a problem that after revoking replicas, the primary can't close cleanly and it hangs waiting to cancel queries.

It looks like the ctx was already cancelled by the time it got to this place:
```
c.adminPool.CancelBackend(cancelCtx, c.ProcessID())
```

Unclear of the right fix here, but adding so we can discuss it.

go/services/multiorch/recovery/analysis/primary_is_dead_and_some_replicas_analyzer.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, nil
+	}
+
+	// If any reachable replica is still connected to the primary, the primary
+	// may be alive — don't failover with only partial visibility.
+	if poolerAnalysis.CountReplicasConfirmingPrimaryAliveInShard > 0 {
+		return nil, nil
+	}
+
+	return &types.Problem{
```
rafael, line 70: I was wondering if this should have a different timeout.  It could be possible that another orchestrator has a better view and can conclude with more certainty if the primary is dead.

go/test/endtoend/multiorch/dead_primary_recovery_test.go
```diff
@@ -240,6 +237,123 @@ func TestDeadPrimaryRecovery(t *testing.T) {
...
+		require.True(t, beginTermResp.Accepted, "Standby %s should accept BeginTerm with higher term", name)
+		t.Logf("Standby %s accepted BeginTerm (term=%d, revoked)", name, revokedTerm)
+	}
+
+	// Verify writes are blocked when sync standbys are disconnected.
+	// The context timeout cancels the sync rep wait, but PostgreSQL commits the data
+	// locally (known PG limitation: CancelRequest releases the sync rep wait, not the txn).
+	// After failover, pg_rewind will undo this locally-committed write.
+	//
+	// TODO(query-serving): This should go through the gateway instead of direct postgres.
```
rafael, line 303: This is the other issue I need to check with @GuptaManan100

go/multipooler/pools/regular/regular_conn.go
GuptaManan100: Fix for this is out in https://github.com/multigres/multigres/pull/636. Your fix is actually 💯 correct. I just added tests to verify we aren't reducing the pool capacity etc.

go/test/endtoend/multiorch/dead_primary_recovery_test.go
GuptaManan100: We need to work on query cancelation to handle this properly. That's an item in our tracking list, that I will pick up next 😄

go/multipooler/manager/rpc_consensus.go
dweitzman: I'm not totally sold on having multipooler pretend to reject the revoke. Philosophically it makes more sense to me to simultaneously accept the revoke and inform orch that postgres is intentionally stopped

That being said, this seems harmless for now

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
```diff
@@ -48,49 +46,33 @@ func (a *PrimaryIsDeadAnalyzer) Analyze(poolerAnalysis *store.ReplicationAnalysi
...
-	if !poolerAnalysis.PrimaryPoolerReachable && poolerAnalysis.ReplicasConnectedToPrimary {
-		// Primary pooler is down but Postgres is still running (replicas are connected).
-		// Do not trigger failover - operator should restart the pooler process.
-		a.factory.Logger().Warn("primary pooler unreachable but postgres still running",
-			"shard_key", poolerAnalysis.ShardKey.String(),
-			"primary_pooler_id", topoclient.MultiPoolerIDString(poolerAnalysis.PrimaryPoolerID),
-			"action", "operator should restart pooler process")
+	// All replicas confirm primary is alive (streaming WAL + healthy heartbeat).
+	// Primary pooler may be down but Postgres is still running — don't failover.
+	if poolerAnalysis.AllReplicasConfirmPrimaryAlive {
```
dweitzman: Should this be _any_ replica confirms the primary is alive rather than _all_?

go/services/multiorch/recovery/analysis/primary_revoked_analyzer.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Full visibility required: all replica poolers must be reachable
+	if poolerAnalysis.CountReachableReplicaPoolersInShard != poolerAnalysis.CountReplicaPoolersInShard {
+		return nil, nil
+	}
+
+	// Both conditions must hold to confirm a revoked primary:
+	// 1. This replica's consensus term exceeds the primary's promotion term,
+	//    proving the replica received a BeginTerm REVOKE with a higher term.
+	if poolerAnalysis.ConsensusTerm <= poolerAnalysis.PrimaryTerm {
```
dweitzman, line 86: I think we'll live to regret it if we assume a higher consensus term always implies revocation of a lower primary term. We've connected BeginTerm() with revoke for convenience, but that connection isn't theoretically necessary and personally I think it boxes us into a corner to interconnect those concepts when they don't need to be interconnected

This may be probably fine in practice, but my two cents is that I don't like it.
I'm not necessarily going to stand in the way of this as it is, but in the long term I don't think we should ever need to compare a voting term vs a primary term?... There's no theoretical reason why either one shouldn't be able to change freely without the other. Intertwining them feels like it'll lead to code that's more code complex, inflexible, and error-prone/brittle.

go/multipooler/manager/pg_replication.go
dweitzman: Seems like we probably need to disable restore during revoke, persist that status to disk in case multipooler crashes, and re-enable it after being assigned a new primary to replicate from?

go/services/multiorch/recovery/analysis/primary_revoked_analyzer.go
dweitzman: I'm going to soften my previous claim: during propagation, voting and primary term might be comparable?... Haven't thought through the details

However I still think the best evidence that a replica believes an old primary was revoked is direct evidence: if a replica can tell us at what logical time it last changed its view of what the primary is and that logical time is newer than when the potential stale primary was promoted, we know we need to revoke a stale primary

dweitzman:
I haven't looked at the tests yet. I don't see any blocking issues except the any vs all condition question

go/multipooler/manager/pg_replication.go
dweitzman: I guess the immediate question is, when/how does restore get re-enabled after this?

cuongdo: It should only be re-enabled by pgbackrest during a future restore (i.e. if a replica is unsalvageable). According to the [Postgres docs](https://www.postgresql.org/docs/18/warm-standby.html#STANDBY-SERVER-OPERATION), `restore_command` is only required during startup as a standby if the primary does not have the required WALs.

go/common/mterrors/code.go
mkindahl, line 36: You might want to write the expected interval here, it might help when debugging and reading logs.

go/multipooler/manager/pg_replication.go
```diff
@@ -310,6 +310,28 @@ func (pm *MultiPoolerManager) resetPrimaryConnInfo(ctx context.Context) error {
...
+	pm.logger.InfoContext(ctx, "Disabling restore_command")
+
+	execCtx, execCancel := context.WithTimeout(ctx, 500*time.Millisecond)
```
mkindahl, line 320: This is a minor thing, but why so short timeouts?

The command should execute quickly, and the timeout does not affect execution time. But if the system is heavily loaded and for some reason the timeout fires, we will abort the command unnecessarily and will then have to re-execute it or ask the admin to do the deed.

This is something that we want to avoid as much as possible, so it would make sense to set the timeout higher.

go/multipooler/manager/rpc_consensus.go
mkindahl: Maybe I am not understanding the situation, but if the primary comes back online during promotion of a standby, why is it wrong to select the old primary as the new primary? It should be in a good state for being a primary since it is the old primary, right?

dweitzman: I don't know if this is theoretically necessary, but it seems practical given how things work today

The emergency demote above completely shuts down postgres because there's no way to make postgres guaranteed read-only without first shutting it down

At that point we have two options:
- Restart as a standby with no primary configured and then restart w/ a pg_rewind once we learn about the primary
- Stay shut down and run pg_rewind + restart after learning about the primary

The first option seems overly-complex right now (and also exposes phantom transactions). We expect/hope to learn what the new primary is soon, so currently in the code the emergency-revoked primary just stays shut down waiting to be informed (hopefully soon) by multiorch who the new primary is to run pg_rewind and restart

It might not be wrong in theory for the old pooler to learn that it's been re-chosen as primary again if we have confidence restarting postgres is likely to succeed, succeed quickly, and we sort out any practical issues about how to do state changes if needed before postgres is running again.

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
dweitzman: Pushed a commit addressing this comment. @mkindahl, does my change make sense to you?

go/multipooler/manager/rpc_consensus.go
mkindahl: **This** is the main reason to why we do not want a restarting old primary be added as the new primary:

> ... exposes phantom transactions 

Since the transaction is committed before the secondaries acknowledges it, it means the server is in an "alternate future" that potentially not exists on the secondaries. It would be good to document this, or at least add it to the comment here.

The "proper" way to deal with this would be to "revert" all unacknowledged transaction on a restarting server and then incorporate it into the pool again. This would mean that the LSN will point to the last "committed and acknowledged" transaction and hence be in a correct state.

Not sure how simple this is to implement, so I agree that it is probably not a good idea to deal with this here.

go/services/multiorch/recovery/analysis/primary_and_replicas_dead_analyzer.go
```diff
@@ -0,0 +1,90 @@
...
+	if poolerAnalysis.CountReachableReplicaPoolersInShard != 0 {
+		return nil, nil
+	}
+
+	return &types.Problem{
+		Code:      types.ProblemPrimaryAndReplicasDead,
+		CheckName: "PrimaryAndReplicasDead",
+		PoolerID:  poolerAnalysis.PoolerID,
+		ShardKey:  poolerAnalysis.ShardKey,
+		Description: fmt.Sprintf("Primary for shard %s is unreachable and no replica poolers are reachable (%d total); likely multiorch network partition — no failover candidates available",
```
mkindahl, line 83: Nit: The number of pooler will always be zero since there is a check above. Do you need the printout?

go/services/multiorch/recovery/analysis/primary_revoked_analyzer.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, nil
+	}
+
+	// 2. No replicas are still replicating from the primary.
```
mkindahl, line 90: Hmmm.... I would argue that it if replicas are still replicating from a primary that is in an earlier term, this should be considered a fault. If secondaries have been moved to a later term this is because somewhere a decision was made that the primary was considered dead (it could be a false positive, but it is nevertheless declared dead) and the secondary should then not replicate from that primary since that could lead to a split-brain situation.

go/services/multiorch/recovery/analysis/primary_revoked_analyzer.go
```diff
@@ -0,0 +1,107 @@
...
+	if poolerAnalysis.CountReplicasConfirmingPrimaryAliveInShard > 0 {
+		return nil, nil
```
mkindahl, line 92: This could be reached with condition `poolerAnalysis.ConsensusTerm > poolerAnalysis.PrimaryTerm && poolerAnalysis.CountReplicasConfirmingPrimaryAliveInShard > 0`, which would mean that the secondary is in a later term than the primary and it is still replicating from it. There might be another leader elected then that other secondaries replicate from and hence you would have a split-brain situation.

go/services/multiorch/recovery/types/types.go
```diff
@@ -37,9 +37,11 @@ const (
...
 
 	// Primary problems (catastrophic - block everything else).
-	ProblemPrimaryIsDead      ProblemCode = "PrimaryIsDead"
-	ProblemPrimaryDiskStalled ProblemCode = "PrimaryDiskStalled"
-	ProblemStalePrimary       ProblemCode = "StalePrimary"
+	ProblemPrimaryIsDead                ProblemCode = "PrimaryIsDead"
+	ProblemPrimaryIsDeadAndSomeReplicas ProblemCode = "PrimaryIsDeadAndSomeReplicas"
+	ProblemPrimaryDiskStalled           ProblemCode = "PrimaryDiskStalled"
+	ProblemStalePrimary                 ProblemCode = "StalePrimary"
+	ProblemPrimaryRevoked               ProblemCode = "PrimaryRevoked"
```
mkindahl, line 44: You might want to be consistent and always use "Primary" as prefix.

go/services/multiorch/recovery/analysis/primary_revoked_analyzer.go
dweitzman: It's not necessarily a problem if voting term is higher than primary term.

Graceful shutdown of postgres at primary term 5 involves writing a checkpoint to WAL, which needs acknowledgement from primary term 5's replicas. If starting voting term 6 implies atomically revoking primary term 5, we'd block the possibility of a graceful switchover

I commented this elsewhere in the PR although Rafa isn't around to respond: I think we should try as much as possible to avoid comparing primary terms with voting/orch terms to avoid intertwining them in ways that may not be necessary or may cause bugs.

go/services/multiorch/recovery/analysis/primary_revoked_analyzer.go
dweitzman: Revoking a primary term technically only requires revoking enough replicas to break quorum. It doesn't require revoking _all_ replicas

If pooler A at primary term 5 requires ACK from both pooler B and pooler C, we could revoke primary term 5 by only reaching pooler B even though pooler A still believes it's a primary and pooler C is still set up to replicate from it

go/multipooler/manager/rpc_consensus.go
dweitzman: It's impossible for us to fully avoid making phantom transactions visible externally without changing postgres itself, since for example replicas can apply WAL before the primary. That both means that a "consistent" read from primary might show fewer transactions than a replica and also that if primary crashes we might throw away transactions during recovery that were previously visible on a replica.

Minimizing phantom reads seems like more of a best-effort aspiration today than a hard design constraint, assuming we use off-the-shelf Postgres.

> The "proper" way to deal with this would be to "revert" all unacknowledged transaction on a restarting server and then incorporate it into the pool again. This would mean that the LSN will point to the last "committed and acknowledged" transaction and hence be in a correct state.
>
> Not sure how simple this is to implement, so I agree that it is probably not a good idea to deal with this here.

Exactly. Long-term, we want to make contributions to postgres to implement two-phase commit for the WAL so we can distinguish proposed transactions (not yet externally visible) from transactions that are known to have achieved consensus.

Short-term, we'll probably have to accept phantom reads in some cases as a known limitation of current versions of Postgres

go/services/multiorch/recovery/analysis/primary_and_replicas_dead_analyzer.go
dweitzman: The check above is for _reachable_ (by this orch instance) replicas, but the number printed is total replicas, so printing the number here doesn't feel entirely unreasonable to me

go/multipooler/manager/rpc_consensus.go
deepthi: @mkindahl re
> The "proper" way to deal with this would be to "revert" all unacknowledged transaction on a restarting server and then incorporate it into the pool again. 

Have you looked at the rewind functionality that we have already implemented? Is there still a gap?

dweitzman: I think the question here is more about whether it's safe to restart postgres as a hot standby _before_ running pg_rewind if the primary knows its primary term is revoked but doesn't yet know the identity of a newer primary it should replicate from

My claim is that the reasons not to restart postgres prior to learning the identity of the new primary and rewinding are:
- it's more code and complexity without a lot of practical benefit
- it can expose phantom reads

One downside to not restarting is what Rafa is fixing here with this change: if we leave postgres stopped, orch can't change its mind as safely or quickly and decide to re-promote the just-stale-revoked primary

mkindahl: > [@mkindahl](https://github.com/mkindahl) re
> 
> > The "proper" way to deal with this would be to "revert" all unacknowledged transaction on a restarting server and then incorporate it into the pool again.
> 
> Have you looked at the rewind functionality that we have already implemented? Is there still a gap?

The rewind works as intended, but there are gaps that I think we need to deal with at cluster-level.

Here is the scenario:

1. Transaction T1 is started and committed. It is in the primary WAL.
2. T1 is sent over to the secondary and the primary is awaiting ack.
3. Secondary applies T1 and send an ack
4. Before primary receives the ack, it crashes.
5. It means the client never get the ack and we know have a transaction in the WAL of the primary and secondary that is applied but not acknowledged.
6. Running a rewind on the primary here will not remove the unacknowledged transaction since it is present and applied on the replica.
7. Client thinks that the transaction was aborted (I think, I am not sure if there are means to deal with this).

mkindahl: > I think the question here is more about whether it's safe to restart postgres as a hot standby _before_ running pg_rewind if the primary knows its primary term is revoked but doesn't yet know the identity of a newer primary it should replicate from
>
> My claim is that the reasons not to restart postgres prior to learning the identity of the new primary and rewinding are:
> 
> * it's more code and complexity without a lot of practical benefit
> * it can expose phantom reads

I agree that we should make sure that a failed primary stays failed until we decide it can be brought up again, so using an error to make sure it is not included in the candidate list seems fine with me. Even better is if we can make sure that the server does not automatically restart: then it cannot be in the candidate list either since it is down.

> One downside to not restarting is what Rafa is fixing here with this change: if we leave postgres stopped, orch can't change its mind as safely or quickly and decide to re-promote the just-stale-revoked primary

I do not think we should re-promote a primary, at least not until we understand the situation better. A crashed primary should trigger a new term and the crashed primary belongs to an old term and has old information.

PR #619, opened 2026-02-16
fix(multipooler): recover from PostgreSQL crash/restart without downtime

GuptaManan100:
## Description

When PostgreSQL crashes and is auto-restarted by the monitor, the multipooler's existing connections hold stale socket FDs that point to the old (dead) process. All queries through these connections fail until the connections are manually recycled. This PR fixes that with a two-pronged approach:

**Prong 1: Proactive reconnection on startup.** When `startPostgres` detects that PostgreSQL was restarted, it now calls `reopenConnections` to replace all stale socket FDs across every pool (regular + admin), then waits for the database to be ready before resuming. This handles the bulk replacement upfront.

**Prong 2: Per-query retry resilience.** Individual query methods (`QueryWithRetry`, `QueryStreamingWithRetry`, `QueryArgsWithRetry`) now detect connection errors and automatically reconnect + retry (up to 3 attempts). This handles any edge cases where a connection is still stale (e.g., single connection drops even if postgres doesn't restart), making the query path self-healing. Session state (SET parameters) is re-applied after reconnect; prepared statements are cleared since they don't survive a session reset.

### Key changes

- **`mterrors.IsConnectionError`**: New centralized connection error classifier that checks Go I/O errors (EOF, ECONNRESET, EPIPE) and PostgreSQL SQLSTATE codes (Class 08 connection exceptions, 57P01/57P02/57P03 shutdown codes). Replaces the old string-matching `isConnectionError` in regular_conn.
- **`client.Conn.Reconnect`**: New method that closes the broken socket, dials a fresh connection, and re-runs the PostgreSQL startup handshake - all in-place on the same `*Conn` object.
- **`regular.Conn.Reconnect`**: Wraps the client-level reconnect with session state management (re-applies settings, clears prepared statements).
- **Retry wrappers**: Generic `retryOnConnectionError[T]` drives `QueryWithRetry`, `QueryStreamingWithRetry`, and `QueryArgsWithRetry`. Admin conn gets its own `queryWithRetry`.
- **`manager.startPostgres`**: Now calls `reopenConnections` + `waitForDatabaseConnection` after a restart.
- **Executor**: All stateless pool queries switched to the `*WithRetry` variants.

### DIFF
--- a/go/common/fakepgserver/server.go
+++ b/go/common/fakepgserver/server.go
@@ -211,6 +211,15 @@ func (s *Server) Close() {
 	}
 }
 
+// CloseListener closes only the TCP listener, preventing new connections
+// while keeping existing connections alive. Use this for testing scenarios
+// where the initial connection should work but reconnect attempts should fail.
+func (s *Server) CloseListener() {
+	if err := s.listener.CloseListener(); err != nil {
+		s.t.Logf("fakepgserver: close listener error: %v", err)
+	}
+}
+
 // OrderMatters sets the orderMatters flag.
 func (s *Server) OrderMatters() {
 	s.orderMatters.Store(true)
--- a/go/common/pgprotocol/server/listener.go
+++ b/go/common/pgprotocol/server/listener.go
@@ -210,6 +210,15 @@ func (l *Listener) handleConnection(conn *Conn) {
 	conn.logger.Info("connection closed")
 }
 
+// CloseListener closes only the TCP listener and stops accepting new connections.
+// Existing connections are not affected. This is useful for testing reconnection
+// failure scenarios where initial connections should keep working but new dial
+// attempts should fail.
+func (l *Listener) CloseListener() error {
+	l.cancel()
+	return l.listener.Close()
+}
+
 // Close closes the listener and waits for all connections to finish.
 func (l *Listener) Close() error {
 	l.cancel()
--- a/go/multipooler/pools/admin/admin_test.go
+++ b/go/multipooler/pools/admin/admin_test.go
@@ -471,6 +471,39 @@ func TestQueryWithRetry_ClosesConnAfterMaxAttempts(t *testing.T) {
 	server.VerifyAllExecutedOrFail()
 }
 
+func TestQueryWithRetry_ReconnectFails(t *testing.T) {
+	server := fakepgserver.New(t)
+	defer server.Close()
+
+	server.OrderMatters()
+
+	// First attempt: connection error.
+	server.AddExpectedExecuteFetch(fakepgserver.ExpectedExecuteFetch{
+		Query: "SELECT 1",
+		Error: &mterrors.PgDiagnostic{
+			MessageType: 'E',
+			Severity:    "FATAL",
+			Code:        "57P01",
+			Message:     "terminating connection due to administrator command",
+		},
+	})
+
+	conn := newTestDirectConn(t, server)
+
+	// Close the listener so reconnect's dial will fail.
+	// Existing connections are not affected.
+	server.CloseListener()
+
+	_, err := conn.queryWithRetry(context.Background(), "SELECT 1")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "reconnect dial failed")
+
+	// Connection should be closed after reconnect failure.
+	assert.True(t, conn.IsClosed())
+
+	server.VerifyAllExecutedOrFail()
+}
+
 // --- execBackendFunc retry tests (via TerminateBackend/CancelBackend) ---
 
 func TestTerminateBackend_ReconnectsOnConnectionError(t *testing.T) {
--- a/go/multipooler/pools/regular/regular_conn_test.go
+++ b/go/multipooler/pools/regular/regular_conn_test.go
@@ -145,6 +145,34 @@ func TestQueryWithRetry_ClosesConnAfterMaxAttempts(t *testing.T) {
 	server.VerifyAllExecutedOrFail()
 }
 
+func TestQueryWithRetry_ReconnectFails(t *testing.T) {
+	server := fakepgserver.New(t)
+	defer server.Close()
+
+	server.OrderMatters()
+
+	// First attempt: connection error.
+	server.AddExpectedExecuteFetch(fakepgserver.ExpectedExecuteFetch{
+		Query: "SELECT 1",
+		Error: connErrFATAL(),
+	})
+
+	conn := newTestDirectConn(t, server)
+
+	// Close the listener so reconnect's dial will fail.
+	// Existing connections are not affected.
+	server.CloseListener()
+
+	_, err := conn.QueryWithRetry(context.Background(), "SELECT 1")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "reconnect dial failed")
+
+	// Connection should be closed after reconnect failure.
+	assert.True(t, conn.IsClosed())
+
+	server.VerifyAllExecutedOrFail()
+}
+
 func TestQueryWithRetry_StopsOnContextCancellation(t *testing.T) {
 	server := fakepgserver.New(t)
 	defer server.Close()


go/multipooler/pools/regular/regular_conn.go
```diff
@@ -311,35 +315,120 @@ func (c *Conn) RawConn() *client.Conn {
...
+	})
+}
+
+// QueryStreamingWithRetry executes a streaming query with automatic retry on
+// connection error. The callback must not have produced observable side effects
+// before the retry; if streaming has already started sending results, the error
+// is returned without retry.
+func (c *Conn) QueryStreamingWithRetry(ctx context.Context, sql string, callback func(context.Context, *sqltypes.Result) error) error {
+	_, err := retryOnConnectionError(c, ctx, func() (struct{}, error) {
+		return struct{}{}, c.conn.QueryStreaming(ctx, sql, callback)
```
haritabh17: The comment says "callback must not have produced observable side effects before the retry" — but what actually prevents that? If the connection dies mid-stream after the callback has already been called with some rows, retryOnConnectionError will reconnect and replay the entire query. Won't the callback see duplicate rows? Should there be a check like "if callback was already invoked, don't retry"?

go/common/pgprotocol/client/conn.go
```diff
@@ -164,6 +143,69 @@ func Connect(ctx context.Context, poolCtx context.Context, config *Config) (*Con
...
+// The caller is responsible for re-applying any session state (settings,
+// prepared statements) after a successful reconnect.
+func (c *Conn) Reconnect(ctx context.Context) error {
+	// Close the raw socket. Best-effort since the connection may already be broken.
+	// We intentionally don't call c.Close() because that cancels the lifetime
+	// context (derived from poolCtx) which we want to keep alive.
+	_ = c.conn.Close()
+
+	// Reset the closed flag so the connection is usable again.
+	c.closed.Store(false)
```
haritabh17, line 164: c.closed.Store(false) is called after c.conn.Close(), but between those two calls another goroutine could observe closed=false on the old (dead) connection and try to use it. Is there a mutex or other coordination that prevents concurrent access during reconnect, or are we relying on the caller to serialize?

go/common/mterrors/mterrors.go
```diff
@@ -461,3 +463,59 @@ func TruncateError(oldErr error, max int) error {
...
+		// Specific Class 57 (Operator Intervention) shutdown codes.
+		// Note: we intentionally exclude 57014 (query_canceled) and the
+		// generic 57000 since those don't indicate a lost connection.
+		switch diag.Code {
+		case "57P01", // admin_shutdown
+			"57P02", // crash_shutdown
+			"57P03": // cannot_connect_now
+			return true
+		}
+		return false
```
haritabh17: If an error wraps both a PgDiagnostic (non-connection, like 42601) AND an io.EOF, the function finds the PgDiagnostic first via errors.As, returns false, and never checks the I/O errors. Is that the intended precedence? Could a connection-level EOF get masked by an application-level PgDiagnostic in a wrapped chain?

go/multipooler/manager/manager.go
```diff
@@ -1643,6 +1643,21 @@ func (pm *MultiPoolerManager) startPostgres(ctx context.Context) error {
 	}
 
 	pm.logger.InfoContext(ctx, "MonitorPostgres: PostgreSQL started successfully")
+
+	// Reopen connections after postgres restart to replace stale socket FDs.
+	// Only when connection pool is initialized (the manager may not have
+	// connection infrastructure in unit tests with minimal setup).
+	if pm.connPoolMgr != nil {
+		if err := pm.reopenConnections(ctx); err != nil {
```
haritabh17, line 1651: As individual queries already retry via QueryWithRetry, is the eager reconnect redundant, or is it specifically for connections that aren't actively running queries?

go/multipooler/pools/admin/admin_conn.go
```diff
@@ -110,6 +115,33 @@ func (c *Conn) GetRolPassword(ctx context.Context, username string) (string, err
...
+		case !mterrors.IsConnectionError(err):
+			return nil, err
+		case attempt == maxQueryAttempts:
+			c.conn.Close()
+			return nil, err
+		}
+		if ctx.Err() != nil {
+			return nil, err
+		}
+		if reconnectErr := c.conn.Reconnect(ctx); reconnectErr != nil {
```
haritabh17, line 153: The regular conn's Reconnect re-applies session settings after reconnecting, but this calls c.conn.Reconnect(ctx) directly on the raw client.Conn — which doesn't re-apply anything. Does the admin connection ever have session state (like search_path or role) that needs restoring, or is it always a bare superuser connection?

go/multipooler/pools/admin/admin_conn.go
```diff
@@ -141,12 +173,13 @@ type queryResult struct {
...
 // If the context is cancelled while the query is running, the connection is closed
-// to abort the operation.
+// to abort the operation. The underlying query uses queryWithRetry for automatic
+// reconnection on stale connections.
 func (c *Conn) execBackendFunc(ctx context.Context, sql, operation string, processID uint32) (bool, error) {
 	// Run query in goroutine so we can respect context cancellation.
 	ch := make(chan queryResult, 1)
 	go func() {
-		results, err := c.conn.Query(ctx, sql)
+		results, err := c.queryWithRetry(ctx, sql)
```
haritabh17, line 198: If the context is cancelled during a retry's reconnect attempt, queryWithRetry checks ctx.Err() and bails — but execBackendFunc's select also fires and closes the connection. Is there a race where both paths try to close/reconnect simultaneously?

go/multipooler/pools/regular/regular_conn.go
```diff
@@ -311,35 +315,120 @@ func (c *Conn) RawConn() *client.Conn {
...
+func retryOnConnectionError[T any](c *Conn, ctx context.Context, op func() (T, error)) (T, error) {
+	for attempt := 1; attempt <= maxQueryAttempts; attempt++ {
+		val, err := execOnce(c, ctx, op)
+		switch {
+		case err == nil:
+			return val, nil
+		case !mterrors.IsConnectionError(err):
+			var zero T
+			return zero, err
+		case attempt == maxQueryAttempts:
```
haritabh17, line 441: If PG is restarting, the retry loop dials immediately after a connection error. With 3 attempts and no delay, all retries could fail before PG finishes starting up. Would a small backoff (even 100ms) between attempts improve the success rate for the crash recovery scenario?


Same for the admin_conn case

go/multipooler/pools/regular/regular_conn.go
GuptaManan100: Good catch. You're right, there was nothing actually preventing duplicate rows if the callback had already been invoked. Fixed: I now track whether the callback was invoked, and if so, the retry loop receives a sentinel error (`errStreamingAlreadyStarted`) that's not a connection error, so `retryOnConnectionError` stops and returns the original failure. This way, if the connection breaks mid-stream after rows have been delivered, the caller sees a clean error instead of duplicate data.

go/common/pgprotocol/client/conn.go
GuptaManan100: We rely on the pool's single-owner model: a connection is checked out to exactly one goroutine at a time, and `Reconnect` is only called from that goroutine's retry loop. There's no concurrent access to worry about here.

go/common/mterrors/mterrors.go
GuptaManan100: Good point. In practice it shouldn't happen because PgDiagnostic comes from successfully parsing an ErrorResponse (so the I/O was fine), but you're right that a wrapped chain could theoretically contain both. Fixed.

go/multipooler/manager/manager.go
GuptaManan100: Yes, it's specifically for idle connections that aren't running queries. After a PG restart, all pooled connections hold dead socket FDs. The eager `reopenConnections` bulk-replaces them upfront so the first queries after restart succeed immediately instead of each hitting a stale connection, reconnecting, and retrying. The per-query retry is the safety net for edge cases (single connection drops, etc).

GuptaManan100: Optimisation only. The e2e test works without this change. I've verified that too 😄

go/multipooler/pools/admin/admin_conn.go
GuptaManan100: Correct. Admin connections are always bare superuser connections with no session state. `Settings()` returns nil, `ApplySettings` panics, and `ResetSettings` is a no-op. The raw `client.Conn.Reconnect` is intentional here since there's nothing to re-apply.

go/multipooler/pools/admin/admin_conn.go
GuptaManan100: The race you describe is technically possible but benign: `execBackendFunc` closes the connection on context cancellation as its way of aborting the hung query, and the connection is not returned to the pool afterward. It's effectively discarded. If `queryWithRetry` is mid-reconnect when the select fires and closes the socket, the reconnect's dial or startup will fail, `queryWithRetry` returns an error, and that error is sent to the channel which is never read (buffered channel, size 1). The connection ends up closed either way. This pattern (goroutine + select for cancellation) is pre-existing and works because the connection is treated as disposable once the context fires.

go/multipooler/pools/regular/regular_conn.go
GuptaManan100: Good suggestion. Added a 100ms backoff (with context-cancellation awareness) between retry attempts in both `retryOnConnectionError` and admin's `queryWithRetry`. This gives PG time to accept connections after a restart without adding meaningful latency to the normal case. I can't believe I added retries without a timeout 💀

go/multipooler/pools/regular/regular_conn.go
```diff
@@ -311,35 +325,139 @@ func (c *Conn) RawConn() *client.Conn {
...
+	wrappedCallback := func(ctx context.Context, result *sqltypes.Result) error {
+		callbackInvoked = true
+		return callback(ctx, result)
+	}
+	_, err := retryOnConnectionError(c, ctx, func() (struct{}, error) {
+		if callbackInvoked {
+			// Callback was already called in a previous attempt — retrying
+			// would replay the query and send duplicate rows. Return the
+			// original connection error so the caller sees a clean failure.
+			return struct{}{}, errStreamingAlreadyStarted
```
haritabh17, line 405: When the callback was already invoked and streaming fails mid-stream, the caller gets errStreamingAlreadyStarted but the original connection error (e.g. 57P01) is lost. Would it make sense to wrap both, so the caller can still see what actually broke the connection?

go/multipooler/pools/admin/admin_conn.go
```diff
@@ -110,6 +119,40 @@ func (c *Conn) GetRolPassword(ctx context.Context, username string) (string, err
...
+		case err == nil:
+			return results, nil
+		case !mterrors.IsConnectionError(err):
+			return nil, err
+		case attempt == maxQueryAttempts:
+			c.conn.Close()
+			return nil, err
+		}
+		if ctx.Err() != nil {
+			return nil, err
```
haritabh17: The regular conn's retryOnConnectionError returns context.Cause(ctx) when the context is cancelled between retries, but the admin version returns the original connection error (err). Should this also return context.Cause(ctx) for consistency?

go/multipooler/pools/regular/regular_conn.go
```diff
@@ -311,35 +325,139 @@ func (c *Conn) RawConn() *client.Conn {
...
+			return zero, err
+		}
+		if ctx.Err() != nil {
+			var zero T
+			return zero, context.Cause(ctx)
+		}
+		// Brief backoff before reconnecting to give PostgreSQL time to
+		// finish starting up if the error is due to a restart.
+		select {
+		case <-time.After(retryBackoff):
```
haritabh17: nit: time.After creates a timer that leaks if ctx.Done() wins the select. At 100ms it's negligible, but time.NewTimer + defer timer.Stop() would be more idiomatic? (Totally optional)

Both here and in admin_pool

go/multipooler/pools/regular/regular_conn.go
haritabh17: Nice!
Would it be worth adding a test where the connection drops after the callback has been invoked with some rows? That would exercise the errStreamingAlreadyStarted path and verify the duplicate-row protection works as intended. The current streaming tests only cover failures before any rows are delivered.

go/multipooler/pools/admin/admin_conn.go
haritabh17: Makes sense — since this assumption is implicit and could silently break if someone adds session state to admin connections in the future, would it be worth adding a defensive check here? Something like:



```
if s := c.conn.ConnectionState().GetSettings(); s != nil && !s.IsEmpty() {
    // log warning or re-apply
}
```

go/multipooler/pools/admin/admin_conn.go
GuptaManan100: Good catch, fixed. The early `ctx.Err()` check in admin's `queryWithRetry` now returns `context.Cause(ctx)` instead of the connection error, matching the regular conn's `retryOnConnectionError`.

go/multipooler/pools/regular/regular_conn.go
GuptaManan100: Fixed in both `retryOnConnectionError` and admin's `queryWithRetry`. Replaced `time.After` with `time.NewTimer` + explicit `Stop()` in the `ctx.Done()` branch.

go/multipooler/pools/regular/regular_conn.go
GuptaManan100: Good point. Fixed with a post-swap approach: inside the retry loop, the sentinel `errStreamingAlreadyStarted` stops the retry (it's not a connection error, so `retryOnConnectionError` returns immediately). Then *after* the loop, `QueryStreamingWithRetry` replaces the sentinel with `mterrors.Wrapf(streamErr, "streaming already started, cannot retry")`. So callers get the real PostgreSQL error in the `Unwrap()` chain and can inspect it via `errors.As`. The sentinel is purely an internal mechanism, never exposed.

go/multipooler/pools/admin/admin_conn.go
GuptaManan100: The assumption is already enforced by the type system: `ApplySettings` panics, `Settings()` always returns nil, and `ResetSettings` is a no-op. If someone adds session state to admin connections in the future, the panicking `ApplySettings` would be a much louder signal than a runtime log warning. Added a comment at the Reconnect call site to make the assumption explicit.

go/multipooler/pools/regular/regular_conn.go
GuptaManan100: Great idea. Extended the fakepgserver with an `AfterCallbackError` field on `ExpectedExecuteFetch`, when set alongside `QueryResult`, the handler delivers the result via callback first, then returns the error. This simulates mid-stream failures where rows are delivered before the connection dies.

Added `TestQueryStreamingWithRetry_MidStreamFailureNoRetry` which:
1. Configures the server to deliver rows then fail with FATAL 57P01
2. Verifies the callback is invoked exactly once (no duplicate rows from retry)
3. Verifies the returned error wraps the original `PgDiagnostic` (callers can `errors.As` it)

go/common/mterrors/mterrors.go
```diff
@@ -461,3 +463,62 @@ func TruncateError(oldErr error, max int) error {
...
+	}
+
+	// Check PostgreSQL SQLSTATE codes for server-initiated disconnection.
+	// When PostgreSQL shuts down or crashes, it sends a FATAL error with
+	// a specific SQLSTATE before closing the connection.
+	var diag *PgDiagnostic
+	if errors.As(err, &diag) {
+		// Class 08: Connection Exception (all codes in this class).
+		if diag.IsClass("08") {
```
haritabh17: Instead of checking for specific error code, should we just check based on severity? Maybe just FATAL & PANIC?

go/multipooler/pools/admin/admin_conn.go
```diff
@@ -110,6 +119,45 @@ func (c *Conn) GetRolPassword(ctx context.Context, username string) (string, err
...
+		}
+		if ctx.Err() != nil {
+			return nil, context.Cause(ctx)
+		}
+		// Brief backoff before reconnecting to give PostgreSQL time to
+		// finish starting up if the error is due to a restart.
+		backoffTimer := time.NewTimer(retryBackoff)
+		select {
+		case <-backoffTimer.C:
+		case <-ctx.Done():
```
haritabh17, line 146: Have we considered using the existing go/tools/retry package here? It already implements exponential backoff with full jitter and context cancellation

go/multipooler/pools/regular/regular_conn.go
```diff
@@ -311,35 +325,149 @@ func (c *Conn) RawConn() *client.Conn {
...
+		if ctx.Err() != nil {
+			var zero T
+			return zero, context.Cause(ctx)
+		}
+		// Brief backoff before reconnecting to give PostgreSQL time to
+		// finish starting up if the error is due to a restart.
+		backoffTimer := time.NewTimer(retryBackoff)
+		select {
+		case <-backoffTimer.C:
+		case <-ctx.Done():
```
haritabh17, line 455: Have we considered using the existing go/tools/retry package here? It already implements exponential backoff with full jitter and context cancellation

go/test/endtoend/queryserving/postgres_crash_recovery_test.go
```diff
@@ -0,0 +1,121 @@
...
+	_ "github.com/lib/pq" // PostgreSQL driver
+	"github.com/stretchr/testify/require"
+
+	"github.com/multigres/multigres/go/test/endtoend/shardsetup"
+	"github.com/multigres/multigres/go/test/utils"
+
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
+)
+
```
haritabh17, line 33: Nice coverage! One case that seems missing: what happens when the first query fails with a connection error, but then Reconnect itself fails?

haritabh17 (approved):
Overall looks good ✌️

go/multipooler/pools/admin/admin_conn.go
GuptaManan100: Considered it, but the retry package is designed for indefinite retry loops (watch reconnection, topology polling, etc.), it doesn't have a max-attempts concept. Here we need exactly 3 attempts with error classification (connection vs non-connection), reconnection between attempts, and connection lifecycle management (close on exhaustion). Using the retry package would mean tracking attempts manually and breaking out of the range loop, which adds wiring without simplifying anything.

The 100ms delay is also intentionally fixed, we want a brief pause for PG startup, not growing exponential delays across 3 fast attempts. The real complexity here is the error classification and reconnect/close logic, which the retry package doesn't help with.

go/multipooler/pools/regular/regular_conn.go
GuptaManan100: Same reasoning as above.

go/test/endtoend/queryserving/postgres_crash_recovery_test.go
GuptaManan100: Good catch. Added `TestQueryWithRetry_ReconnectFails` for both regular and admin connections. The test uses a new `CloseListener()` helper on fakepgserver that closes just the TCP listener (so new dials fail) while keeping the existing connection alive. Flow:

1. First query returns FATAL 57P01 (connection error)
2. Retry loop waits 100ms, then calls Reconnect
3. Reconnect's `dial()` fails with "connection refused" (listener closed)
4. `retryOnConnectionError` closes the connection and returns the reconnect error

Verifies the connection is closed and the reconnect error is surfaced.

haritabh17 (approved):

PR #621, opened 2026-02-16
feat(pgregresstest): PostgreSQL regression test infrastructure and nightly CI

haritabh17:
Adds infrastructure to build and run official PostgreSQL 17.6 regression tests against multigres, producing a compatibility report with per-test pass/fail tracking. Failed tests include expandable diffs for easy debugging. The report is written to the GitHub Actions Job Summary and uploaded as an artifact.

A nightly CI workflow runs these tests at 3 AM UTC (with manual dispatch), building PostgreSQL from source, spinning up a multigres cluster, and executing the full `installcheck` suite (`parallel_schedule`).

To run only specific tests, set `PGREGRESS_TESTS="boolean char"`. When unset, the full suite runs.

## Test plan
- Trigger the workflow manually and verify the regression test report is generated correctly
- Confirm the report appears in the Job Summary and as a downloadable artifact
- Verify selective test execution via `PGREGRESS_TESTS` env var

### DIFF
--- a/.github/workflows/test-pgregress.yml
+++ b/.github/workflows/test-pgregress.yml
@@ -30,7 +30,6 @@ jobs:
       github.event_name != 'pull_request' ||
       contains(github.event.pull_request.labels.*.name, 'Run PgRegress Tests')
     runs-on: ubuntu-24.04
-
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:


.github/workflows/test-pgregress.yml
```diff
@@ -15,10 +15,9 @@
 name: PostgreSQL Regression Tests
 
 on:
-  push:
-    branches: [main]
-  pull_request:
-    branches: [main]
```
GuptaManan100, line 21: We should run this on PRs too, just skip them if a specific label isn't applied. We can create that label and call it `Run PgRegress Tests` or something

docs/README.md
```diff
@@ -11,3 +11,10 @@ to use Multigres for your applications, please refer to the
 - **[Working with us](./teamwork.md)**
 - **[Github workflow](./workflow.md)**
 - **[Contributing](./contributing.md)**
+
+## PostgreSQL Compatibility
+
+multigres runs the official PostgreSQL regression test suite to track compatibility.
+
+- **Results:** See the [latest workflow run](https://github.com/haritabh17/multigres-clone/actions/workflows/test-pgregress.yml) for the detailed compatibility report in the Job Summary.
```
GuptaManan100: Link shouldn't be to a fork of the repo

.github/workflows/test-pgregress.yml
haritabh17: Yup, will do that

docs/README.md
haritabh17: Ahh, missed to update this. Fixing now

GuptaManan100:
I've created the label required for running the tests 😄

PR #622, opened 2026-02-16
refactor(multigateway): simplify SET/RESET to local-only state updates

haritabh17:
## Summary

Simplifies SET/RESET command handling in multigateway by making both operations local-only — neither is routed to PostgreSQL.

**Previous approach:** SET commands were routed to PG for validation (with optimistic local state update and rollback on error). RESET was already local-only. This involved save/restore logic, rollback on PG rejection, and a Route dependency.

**New approach:** Both SET and RESET update local `SessionSettings` and return a synthetic `CommandComplete` immediately. The pool propagates these settings to backend connections on subsequent queries via `ApplySettings`.

**Trade-off:** Invalid SET values (bad variable names or values) are accepted at SET time without error. The error surfaces on the next query when the pool tries to apply the setting. The client must RESET the bad variable to recover. This deviation from PostgreSQL behaviour is documented in code comments and in `docs/query_serving/session_settings.md`.

### DIFF
--- a/docs/query_serving/session_settings.md
+++ b/docs/query_serving/session_settings.md
@@ -1,0 +1,53 @@
+# Session Settings (SET/RESET)
+
+## Overview
+
+Multigateway handles `SET` and `RESET` commands locally without forwarding them to PostgreSQL. Session variables are tracked in the gateway's connection state and propagated to backend connections via the pool's `ApplySettings` mechanism on each subsequent query.
+
+Inside transactions, the connection is reserved and bypasses the pool's normal `ApplySettings` path. To ensure settings changes (e.g., `SET search_path`) take effect on the reserved backend, the multipooler applies settings diffs to reserved connections before each query execution.
+
+## How It Works
+
+1. **SET variable = value**: The variable and value are stored in `SessionSettings`. A synthetic `CommandComplete (SET)` is returned to the client immediately.
+2. **RESET variable**: The variable is removed from `SessionSettings`. A synthetic `CommandComplete (RESET)` is returned.
+3. **RESET ALL**: All variables in `SessionSettings` are cleared. A synthetic `CommandComplete (RESET)` is returned.
+4. **SET LOCAL**: Passed through to PostgreSQL (transaction-scoped, not tracked by the gateway).
+
+On the next query, the pool merges `SessionSettings` with `StartupParams` (startup params from the client's initial connection). `SessionSettings` entries take precedence. The merged settings are applied to the backend connection before executing the query.
+
+When a variable is `RESET`, its entry is removed from `SessionSettings`, and the `StartupParams` value (if any) becomes visible again through the merge.
+
+## Behaviour Deviations from PostgreSQL
+
+### SET does not validate parameters
+
+`SET` commands are **not validated** against PostgreSQL. A client can `SET` an invalid variable name or an invalid value for a valid variable without receiving an immediate error.
+
+The error will surface on the **next query** when the connection pool attempts to apply the invalid setting to a backend connection. The client will receive repeated errors on every query until they `RESET` the problematic variable.
+
+**Example:**
+
+```sql
+SET nonexistent_variable = 'value';  -- Succeeds (no error)
+SELECT 1;                             -- Fails: unrecognized configuration parameter
+RESET nonexistent_variable;           -- Succeeds
+SELECT 1;                             -- Succeeds again
+```
+
+This trade-off was chosen to keep the SET/RESET code path simple. It may be revisited if stricter validation is needed.
+
+### SET inside a rolled-back transaction persists
+
+In PostgreSQL, `SET` (without `LOCAL`) inside a transaction that is later rolled back still reverts the variable to its pre-transaction value. Multigateway does **not** track transaction boundaries for session settings, so a `SET` inside a rolled-back transaction will persist in the gateway's tracked state.
+
+**Example:**
+
+```sql
+SET work_mem = '256MB';
+BEGIN;
+SET work_mem = '512MB';
+ROLLBACK;
+SHOW work_mem;  -- PostgreSQL: '256MB', Multigateway: '512MB'
+```
+
+**TODO:** Fix this by snapshotting session settings at `BEGIN` and restoring on `ROLLBACK`.
--- a/go/multipooler/connpoolmanager/manager.go
+++ b/go/multipooler/connpoolmanager/manager.go
@@ -376,6 +376,22 @@ func (m *Manager) GetReservedConn(connID int64, user string) (*reserved.Conn, bo
 	return pool.GetReservedConn(connID)
 }
 
+// ApplySettingsToConn ensures the connection's settings match the given session
+// settings. ApplySettings handles the diff internally: it resets removed
+// variables via individual RESET commands (safe inside transactions, unlike
+// RESET ALL) and applies desired variables via SET SESSION.
+func (m *Manager) ApplySettingsToConn(ctx context.Context, conn *regular.Conn, settings map[string]string) error {
+	desired := m.settingsCache.GetOrCreate(settings)
+	current := conn.Settings()
+
+	// Pointer equality — same *Settings means same settings (via cache interning)
+	if desired == current {
+		return nil
+	}
+
+	return conn.ApplySettings(ctx, desired)
+}
+
 // --- Stats ---
 
 // Stats returns statistics for all pools.
--- a/go/multipooler/executor/executor.go
+++ b/go/multipooler/executor/executor.go
@@ -79,6 +79,15 @@ func (e *Executor) ExecuteQuery(ctx context.Context, target *query.Target, sql s
 			return nil, fmt.Errorf("reserved connection %d not found for user %s", options.ReservedConnectionId, user)
 		}
 
+		// Apply settings if they changed (e.g., SET inside a transaction).
+		// Reserved connections bypass the pool's normal ApplySettings mechanism,
+		// so we must explicitly apply settings changes here.
+		if options.SessionSettings != nil {
+			if err := e.poolManager.ApplySettingsToConn(ctx, reservedConn.Conn(), options.SessionSettings); err != nil {
+				return nil, fmt.Errorf("failed to apply settings to reserved connection: %w", err)
+			}
+		}
+
 		results, err := reservedConn.Query(ctx, sql)
 		if err != nil {
 			return nil, wrapQueryError(err)
@@ -145,6 +154,15 @@ func (e *Executor) StreamExecute(
 		reservedConn, _ := e.poolManager.GetReservedConn(int64(options.ReservedConnectionId), user)
 		if reservedConn == nil {
 			return fmt.Errorf("reserved connection %d not found for user %s", options.ReservedConnectionId, user)
+		}
+
+		// Apply settings if they changed (e.g., SET inside a transaction).
+		// Reserved connections bypass the pool's normal ApplySettings mechanism,
+		// so we must explicitly apply settings changes here.
+		if options.SessionSettings != nil {
+			if err := e.poolManager.ApplySettingsToConn(ctx, reservedConn.Conn(), options.SessionSettings); err != nil {
+				return fmt.Errorf("failed to apply settings to reserved connection: %w", err)
+			}
 		}
 
 		if err := reservedConn.QueryStreaming(ctx, sql, callback); err != nil {
--- a/go/multipooler/pools/connpool/interfaces.go
+++ b/go/multipooler/pools/connpool/interfaces.go
@@ -35,12 +35,13 @@ type Connection interface {
 	// Close closes the connection and releases associated resources.
 	Close() error
 
-	// ApplySettings applies the given settings to the connection by executing
-	// the necessary SQL commands (e.g., SET commands).
-	// Returns an error if the settings cannot be applied.
-	ApplySettings(ctx context.Context, settings *connstate.Settings) error
+	// ApplySettings transitions the connection to the desired settings state.
+	// It diffs current tracked settings against desired: executes individual
+	// RESET commands for removed variables, then SET SESSION commands for all
+	// desired variables. Updates tracked state to desired.
+	ApplySettings(ctx context.Context, desired *connstate.Settings) error
 
-	// ResetSettings resets the connection to a clean state with no settings.
-	// This typically involves running RESET ALL or equivalent SQL.
-	ResetSettings(ctx context.Context) error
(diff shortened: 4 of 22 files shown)


go/services/multigateway/engine/session_state_route.go
```diff
@@ -0,0 +1,210 @@
...
+//     SessionSettings (which falls back to StartupParams) drives the pool to
+//     apply the correct SET commands to the connection
+//
+// Why RESET is not routed: PostgreSQL's RESET restores the server-level default,
+// which may differ from the client's startup parameter value. For example, the
+// Go driver sends extra_float_digits=2 at startup, but RESET extra_float_digits
+// sets PG to 1 (the server default). This creates a state mismatch: the pool
+// tracks the connection as having the startup value, but PG has the server default.
+// By not routing RESET and letting the pool handle it via ApplySettings, the
+// actual PG-side state always matches the pool's tracked state.
```
GuptaManan100: 👍 Also why route, its not like reset can fail, so we don't need it for validation.

go/services/multigateway/engine/session_state_route.go
```diff
@@ -0,0 +1,210 @@
...
+type SessionStateRoute struct {
+	// Route is the underlying route for SET commands (nil for RESET-only).
+	Route *Route
+
+	// VariableStmt is the SET/RESET statement from the AST.
+	VariableStmt *ast.VariableSetStmt
+
+	// SQL is the original SQL string.
+	SQL string
```
GuptaManan100: We're using `Query` field name in the other engines. Let's use that for consistency.

go/services/multigateway/engine/session_state_route.go
```diff
@@ -0,0 +1,210 @@
...
+	var err error
+	defer func() {
+		if err != nil {
+			if prevExists {
+				state.SetSessionVariable(s.VariableStmt.Name, prevValue)
+			} else {
+				state.ResetSessionVariable(s.VariableStmt.Name)
+			}
+		}
+	}()
```
GuptaManan100: Why don't we only update the state after the execution succeeds? Then we don't need to have any of this code of setting, and then resetting it.

go/services/multigateway/planner/planner.go
```diff
@@ -48,7 +48,7 @@ func NewPlanner(defaultTableGroup string, logger *slog.Logger) *Planner {
 // with switch on NodeTag for extensibility.
 //
 // Supported statement types:
-// - VariableSetStmt: SET/RESET commands → Sequence[Route, ApplySessionState]
+// - VariableSetStmt: SET/RESET commands → SessionStateRoute
```
GuptaManan100: Any reason to go away from the Sequence way of doing things?

haritabh17: While working on RESET, found that during SET (when using Sequence based implementation) the pool acquired a connection before local state was updated. So the connection was returned to the pool as "clean" even though PG had the dirty GUC value. Next query could get that connection without the setting applied.

Another potential way to fix this would be to use one reserved connection for the `Sequence` and have ApplySessionState update the pooler's setting map as well, that implementation looked like more complex for SET / RESET operation, hence inclining towards moving away from the Sequence based implementation.

go/services/multigateway/engine/session_state_route.go
haritabh17: Yupp, RESET is not routed, it simply does `state.ResetSessionVariable`

go/services/multigateway/engine/session_state_route.go
haritabh17: Ack, updated

go/services/multigateway/engine/session_state_route.go
haritabh17: The state must be updated before routing because the pool reads SessionSettings when acquiring the connection inside Route.StreamExecute. If we update after, the pool won't see the new setting during acquisition, and the connection gets tracked with stale state — this is the exact race condition that the old Sequence[Route, ApplySessionState] approach had. The optimistic update + rollback ensures the pool always sees correct settings at acquisition time.

go/multipooler/connpoolmanager/manager.go
```diff
@@ -376,6 +376,25 @@ func (m *Manager) GetReservedConn(connID int64, user string) (*reserved.Conn, bo
...
+func (m *Manager) ApplySettingsToReservedConn(ctx context.Context, conn *reserved.Conn, settings map[string]string) error {
+	desired := m.settingsCache.GetOrCreate(settings)
+	current := conn.Conn().Settings()
+
+	// Pointer equality — same *Settings means same settings (via cache interning)
+	if desired == current {
+		return nil
+	}
+
+	return conn.Conn().ApplySettings(ctx, desired)
```
GuptaManan100: You do need to reset regardless. Maybe not reset all, but then you need to reset each individual setting from the previous setting applied

go/services/multigateway/engine/session_state_route.go
```diff
@@ -0,0 +1,174 @@
...
+// will surface on the next query when the pool tries to apply the setting to a
+// backend connection. The client must RESET the bad variable to recover. This
+// trade-off was chosen intentionally to keep the SET/RESET path simple. It may
+// be revisited in the future if stricter validation is needed.
+//
+// For RESET/RESET ALL:
+// The variable is removed from SessionSettings. On the next query, the merged
+// settings (SessionSettings overlaid on StartupParams) will fall back to the
+// startup parameter value, and the pool will apply the correct SET commands.
+type SessionStateRoute struct {
```
GuptaManan100: Isn't ApplySessionState a better name instead of SessionStateRoute given its not a route anymore 😆

go/services/multigateway/handler/connection_state.go
```diff
@@ -225,16 +225,14 @@ func (m *MultiGatewayConnectionState) SetSessionVariable(name, value string) {
...
-	}
+	delete(m.SessionSettings, name)
 }
 
 // ResetAllSessionVariables clears all session variables (from RESET ALL command).
 func (m *MultiGatewayConnectionState) ResetAllSessionVariables() {
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	m.SessionSettings = make(map[string]string)
+	m.SessionSettings = nil
```
GuptaManan100, line 242: If we do then, then getting a session variable will panic cause the map will be nil

go/services/multigateway/engine/session_state_route.go
haritabh17: Ack, updating 😄

go/multipooler/connpoolmanager/manager.go
haritabh17: Yes, right. Adding an interface to selectively clear the RESET-ed variables to use here.

haritabh17: Nvm, while working on it realised better interface is to have only two functions -- one to apply the passed in final state (ApplySettings) - this would do RESET + SET and another to clear everything, ResetAll.

go/services/multigateway/handler/connection_state.go
haritabh17: Safe to keep nil, so lets keep it this way

go/multipooler/pools/connpool/interfaces.go
```diff
@@ -35,12 +35,13 @@ type Connection interface {
...
-	ResetSettings(ctx context.Context) error
+	// ApplySettings transitions the connection to the desired settings state.
+	// It diffs current tracked settings against desired: executes individual
+	// RESET commands for removed variables, then SET SESSION commands for all
+	// desired variables. Updates tracked state to desired.
+	ApplySettings(ctx context.Context, desired *connstate.Settings) error
+
+	// ResetAllSettings resets the connection to a clean state with no settings.
+	// This executes RESET ALL to clear all session variables at once.
+	ResetAllSettings(ctx context.Context) error
```
GuptaManan100, line 46: Perfect change. keeps the interface clean 👍

go/multipooler/connpoolmanager/manager.go
```diff
@@ -376,6 +376,22 @@ func (m *Manager) GetReservedConn(connID int64, user string) (*reserved.Conn, bo
 	return pool.GetReservedConn(connID)
 }
 
+// ApplySettingsToReservedConn ensures the reserved connection's settings match
+// the given session settings. ApplySettings handles the diff internally:
+// it resets removed variables via individual RESET commands (safe inside
+// transactions, unlike RESET ALL) and applies desired variables via SET SESSION.
+func (m *Manager) ApplySettingsToReservedConn(ctx context.Context, conn *reserved.Conn, settings map[string]string) error {
```
GuptaManan100: This function doesn't have to be particularly for reservedConn. Like it can just take the underlying conn, and then it can be used for any connection type. Probably not a big difference maker since there is no use case right now other than reserved conns. Cause regular conns should be gotten from the pool, and not need apply settings later, and admin conns don't support settings at all.

go/services/multigateway/engine/apply_session_state.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Supabase, Inc.
+// Copyright 2026 Supabase, Inc.
```
GuptaManan100: no need to update the year for an existing file

docs/query_serving/session_settings.md
```diff
@@ -0,0 +1,52 @@
+# Session Settings (SET/RESET)
+
+## Overview
+
+Multigateway handles `SET` and `RESET` commands locally without forwarding them to PostgreSQL. Session variables are tracked in the gateway's connection state and propagated to backend connections via the pool's `ApplySettings` mechanism on each subsequent query.
+
+Inside transactions, the connection is reserved and bypasses the pool's normal `ApplySettings` path. To ensure settings changes (e.g., `SET statement_timeout`) take effect on the reserved backend, the multipooler applies settings diffs to reserved connections before each query execution.
```
GuptaManan100: We should document that currently session settings don't always work correctly, like if someone sets them in a transaction, and then rollsback the transaction. Their connection to multigateway will still persist the settings change, which is incorrect.

go/multipooler/executor/executor.go
```diff
@@ -147,6 +147,15 @@ func (e *Executor) StreamExecute(
...
 
+		// Apply settings if they changed (e.g., SET inside a transaction).
+		// Reserved connections bypass the pool's normal ApplySettings mechanism,
+		// so we must explicitly apply settings changes here.
+		if options.SessionSettings != nil {
+			if err := e.poolManager.ApplySettingsToReservedConn(ctx, reservedConn, options.SessionSettings); err != nil {
+				return fmt.Errorf("failed to apply settings to reserved connection: %w", err)
+			}
+		}
+
```
GuptaManan100, line 167: Do we need a similar change for `ExecuteQuery` too. I know multigateway doesn't call it. Its mostly for internal use, but let's keep it correct too regardless.

go/multipooler/pools/regular/regular_conn.go
```diff
@@ -100,32 +101,60 @@ func (c *Conn) Close() error {
...
-	if settings == nil || settings.IsEmpty() {
+// ApplySettings transitions the connection to the desired settings state.
+// It diffs current tracked settings against desired: executes individual RESET
+// commands for removed variables, then SET SESSION commands for all desired
+// variables. This is safe inside transactions (individual RESETs don't destroy
+// SET LOCAL settings, unlike RESET ALL).
+func (c *Conn) ApplySettings(ctx context.Context, desired *connstate.Settings) error {
+	if desired == nil || desired.IsEmpty() {
 		return nil
 	}
```
GuptaManan100, line 118: If desired is nil or empty, then we should reset the previous settings 😄. It should only be a no-op if desired matches the current settings

go/services/multigateway/planner/variable_set_stmt_test.go
```diff
@@ -0,0 +1,123 @@
...
+		Kind: ast.VAR_SET_VALUE,
+		Name: "work_mem",
+		Args: &ast.NodeList{Items: []ast.Node{&ast.A_Const{Val: &ast.String{SVal: "256MB"}}}},
+	}
+
+	plan, err := p.planVariableSetStmt("SET work_mem = '256MB'", stmt, testConn.Conn)
+	require.NoError(t, err)
+	require.NotNil(t, plan)
+
+	// The primitive should be a ApplySessionState
```
GuptaManan100: nit: `an` ApplySessionState.

go/multipooler/pools/regular/regular_conn.go
```diff
@@ -100,32 +101,60 @@ func (c *Conn) Close() error {
...
-// This executes SET commands for each variable in the settings.
-func (c *Conn) ApplySettings(ctx context.Context, settings *connstate.Settings) error {
-	if settings == nil || settings.IsEmpty() {
+// ApplySettings transitions the connection to the desired settings state.
+// It diffs current tracked settings against desired: executes individual RESET
+// commands for removed variables, then SET SESSION commands for all desired
+// variables. This is safe inside transactions (individual RESETs don't destroy
+// SET LOCAL settings, unlike RESET ALL).
+func (c *Conn) ApplySettings(ctx context.Context, desired *connstate.Settings) error {
+	if desired == nil || desired.IsEmpty() {
```
GuptaManan100, line 113: Let's also add tests for this function. Specifically - SET overwriting an existing variable. We currently have no test for SET x = 'a'; SET x = 'b' to verify the second value takes precedence.

go/multipooler/executor/executor.go
haritabh17: Agree, better to add it. Adding.

go/multipooler/pools/regular/regular_conn.go
haritabh17: Good catch 💯  ! updating!

go/services/multigateway/engine/apply_session_state.go
haritabh17: Updated.

docs/query_serving/session_settings.md
haritabh17: Updated

go/services/multigateway/planner/variable_set_stmt_test.go
haritabh17: 🙂 updated

go/multipooler/pools/regular/regular_conn.go
haritabh17: Ack, added

go/multipooler/connpoolmanager/manager.go
haritabh17: Ack, updated

PR #623, opened 2026-02-17
fix(multigateway): readiness probe checks pooler discovery + retry on empty cells

haritabh17:
## Summary

During cluster startup, multigateway can report ready before discovering any poolers. This happens because `/ready` only checks topology registration, not pooler discovery. Additionally, `GlobalPoolerDiscovery` uses `GetCellNames()` once and then blocks on `<-ctx.Done()` forever — so new cells added later (e.g., new region or shard) are never discovered without a gateway restart.

This PR makes two changes. First, `/ready` now also requires at least one discovered pooler before reporting healthy. Second, cell discovery is replaced with a watch-based approach using `WatchRecursive` on the `cells/` directory, mirroring how `CellPoolerDiscovery` already watches for poolers. New cells are detected immediately via watch events, cell removal stops the corresponding watcher, and watch reconnection is handled by the existing retry loop with exponential backoff.

## Test plan

Tests cover cell addition at runtime, late pooler addition through the global discovery path, cell removal stopping the watcher, and `extractCellFromPath` parsing. Uses `lastCellRefresh` for deterministic synchronization instead of sleeps. All existing tests pass. Also verified manually against a local cluster with fresh etcd.


GuptaManan100:
There is a PR from @dweitzman #516 that is changing the discovery logic quite significantly. If we wait for that, then maybe this isn't a problem. In any case, maybe this change should be made after that one lands?

go/services/multigateway/discovery.go
```diff
@@ -488,6 +488,12 @@ func (gd *GlobalPoolerDiscovery) Start() {
 
 				gd.logger.Info("Discovered cells", "cells", cells)
 
+				// No cells found yet — retry with backoff instead of blocking forever
+				if len(cells) == 0 {
+					gd.logger.Warn("No cells discovered, will retry")
+					return
+				}
+
```
GuptaManan100: Maybe instead of this, we should setup a watch on the cell? Cause there is a chance a new cell comes up after the first 3, and that cell might get the primary, so we need the gateway to find cells that show up even after the first initialisation is done.

GuptaManan100 (commented):
Does 👇 make sense?

go/services/multigateway/discovery.go
haritabh17: Done — implemented using WatchRecursive on the cells/ directory. New cells are now detected immediately via watch events, and cell removal stops the corresponding watcher. Same pattern as CellPoolerDiscovery.

PR #630, opened 2026-02-18
test(pgctld): fix orphan detection test bug

dweitzman:
The orphan detection test had a bug that made it slow (~15s) and not a genuine test case. The test deleted the testdata directory expecting the watchdog to detect this and stop PostgreSQL. It turns out the orphan detector can't actually stop postgres in that case because `pg_ctl stop` doesn't work if there's no postgres directory with a postmaster.pid file.

Postgres does still crash eventually when it tries to use new files, but that has nothing to do with the orphan detector.

The test fix is to do orphan detection first, then remove the temporary directory.

Additional test improvements:
- Replace all time.Sleep() calls with require.Eventually() for deterministic waiting
- Add helper functions: waitForPostgresReady(), waitForPostgresSocket(), waitForPostgresStopped()

### DIFF
--- a/go/test/endtoend/pgctld/pgctld_test.go
+++ b/go/test/endtoend/pgctld/pgctld_test.go
@@ -58,6 +58,47 @@ func setupTestEnv(cmd *exec.Cmd) {
 	}
 }
 
+// waitForPostgresReady waits for PostgreSQL to be ready to accept TCP connections on the given port.
+func waitForPostgresReady(t *testing.T, port int, timeout time.Duration) {
+	t.Helper()
+	require.Eventually(t, func() bool {
+		conn, err := net.DialTimeout("tcp", fmt.Sprintf("localhost:%d", port), 500*time.Millisecond)
+		if err == nil {
+			conn.Close()
+			return true
+		}
+		return false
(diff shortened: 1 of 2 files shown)


go/test/endtoend/pgctld/pgctld_test.go
```diff
@@ -1167,23 +1207,15 @@ func TestOrphanDetectionWithRealPostgreSQL(t *testing.T) {
...
 	require.NoError(t, pgProcess.Signal(syscall.Signal(0)))
 
-	// Kill the pgctld server subprocess abruptly
+	// Kill the pgctld server subprocess abruptly to trigger watchdog via parent death detection
 	require.NoError(t, serverCmd.Process.Kill())
 	_, _ = serverCmd.Process.Wait()
 
-	// TODO(dweitzman): Start a process using sleep command and use that PID for orphan detection
+	// Wait for watchdog to detect parent death and stop PostgreSQL
+	waitForPostgresStopped(t, pgPort, 15*time.Second)
```
dweitzman, line 1215: Moving this earlier in the method is the actual fix for the test

PR #633, opened 2026-02-18
feat(executil): add process execution wrapper with graceful termination

dweitzman:
Reviving this from an old experimental branch because graceful termination makes tracing data much more useful in endtoend tests vs the KILL signals we're often sending now

Introduces executil package to replace direct use of exec.Command/CommandContext
with a safer, more production-ready API that prioritizes graceful termination,
trace propagation, and explicit environment handling.

Go's standard library exec package has doesn't handle some things we care about. By establishing executil as the standard subprocess library and enforcing its use via ruleguard, we support:

- Graceful termination (SIGTERM → SIGKILL) reduces telemetry data loss
- Automatic trace context propagation
- Explicit environment variable handling (AddEnv/SetEnv)

### DIFF
--- a/go/common/topoclient/etcdtopo/test_helpers.go
+++ b/go/common/topoclient/etcdtopo/test_helpers.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log/slog"
 	"net"
 	"net/http"
 	"os"
@@ -30,7 +29,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/multigres/multigres/go/test/utils"
-	"github.com/multigres/multigres/go/tools/ctxutil"
 	"github.com/multigres/multigres/go/tools/executil"
 	"github.com/multigres/multigres/go/tools/retry"
 )
@@ -152,8 +150,9 @@ func StartEtcdWithOptions(t *testing.T, opts EtcdOptions) (string, *executil.Cmd
 	peerAddr := fmt.Sprintf("http://localhost:%v", peerPort)
 	initialCluster := fmt.Sprintf("%v=%v", name, peerAddr)
 
-	// Wrap etcd with run_in_test.sh to ensure cleanup if test process dies
-	cmd := executil.Command(t.Context(), "run_in_test.sh", "etcd",
+	// Wrap etcd with run_in_test.sh for orphan protection. Stops gracefully when
+	// the test context is cancelled so run_in_test.sh can terminate etcd cleanly.
+	cmd := utils.CommandWithOrphanProtection(t.Context(), "etcd",
 		"-name", name,
 		"-advertise-client-urls", clientAddr,
 		"-initial-advertise-peer-urls", peerAddr,
@@ -175,20 +174,6 @@ func StartEtcdWithOptions(t *testing.T, opts EtcdOptions) (string, *executil.Cmd
 	defer cancel()
 	err = WaitForReady(ctx, clientAddr)
 	require.NoError(t, err, "etcd failed to become ready")
-
-	t.Cleanup(func() {
-		// Force kill if still running
-		// Use ctxutil.Detach since we're in cleanup after the test context is done
-		stopCtx, cancel := context.WithTimeout(ctxutil.Detach(ctx), 5*time.Second)
-		stopErr, _ := cmd.Stop(stopCtx)
-		cancel()
-		if stopErr != nil {
-			slog.Error("executil.KillProcess() failed killing etcd", "error", stopErr)
-		}
-
-		// Additional cleanup: try to release the ports
-		time.Sleep(50 * time.Millisecond)
-	})
 
 	return clientAddr, cmd
 }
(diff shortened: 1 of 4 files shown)


go/tools/executil/command.go
```diff
@@ -0,0 +1,407 @@
...
+// Command creates a new Cmd with graceful termination support.
+//
+// If the parent context is cancelled, the process receives SIGTERM and is given
+// DefaultGracePeriod to exit before SIGKILL. For explicit termination with a
+// different grace period, call Terminate(ctx) where ctx's timeout controls the wait.
+//
+// By default, the command inherits the parent process environment.
+// Use AddEnv() to add variables, or SetEnv() to replace the entire environment.
+func Command(ctx context.Context, name string, args ...string) *Cmd {
+	return CommandWithGracePeriod(ctx, DefaultGracePeriod, name, args...)
```
dweitzman, line 130: One idea here is that when the context runs out, it triggers a TERM instead of a KILL and there's a grace period before following up with a kill

go/tools/executil/command.go
```diff
@@ -0,0 +1,407 @@
...
+//
+// This is the recommended way to stop a process - always try graceful termination first.
+//
+// Example:
+//
+//	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+//	defer cancel()
+//	exitErr, stopped := cmd.Stop(ctx)
+//	// Tries SIGTERM for up to 10s, then SIGKILL with 100ms timeout
+func (c *Cmd) Stop(ctx context.Context) (error, bool) {
```
dweitzman, line 359: Another idea here is that if you explicitly call `Stop()` on a process, you can give a custom context for how long to wait before killing instead of relying on the per-configured default grace period

The default grace period is only needed when context runs out

go/tools/executil/terminate.go
```diff
@@ -0,0 +1,212 @@
...
+//
+// Uses exponential backoff starting at 1ms, doubling each time up to 100ms.
+// This responds quickly for fast exits while reducing CPU usage for slow exits.
+func waitForProcessExit(ctx context.Context, process *os.Process) bool {
+	// Check immediately before starting the timer
+	if err := process.Signal(syscall.Signal(0)); err != nil {
+		return true
+	}
+
+	delay, maxDelay := time.Millisecond, 100*time.Millisecond
```
cuongdo, line 196: Can this backoff/retry logic be replaced with `retry`?

go/tools/ruleguard/rules.go
```diff
@@ -73,3 +73,25 @@ func requireGrpcCommonNewClient(m dsl.Matcher) {
 				!m.File().PkgPath.Matches(`/grpccommon$|/test/|/testutil/|testutil$`)).
 		Report("use grpccommon.NewClient() instead of grpc.NewClient() to ensure telemetry instrumentation")
 }
+
+// disallowDirectExecCommandContext enforces use of executil.Command() for
+// graceful termination support, proper environment variable handling, and
+// trace propagation.
+func disallowDirectExecCommandContext(m dsl.Matcher) {
```
cuongdo, line 80: nice safeguard

go/tools/executil/terminate.go
dweitzman: Potentially, but my thinking here was that it's not necessary to have backoff here because this is an inexpensive operation that doesn't need the network or disk or anything like that.

go/tools/ruleguard/rules.go
dweitzman: I've become a big fan of ruleguard!

go/cmd/pgctld/testutil/temp_dirs.go
```diff
@@ -185,9 +181,9 @@ func cleanupMockProcesses(t *testing.T, tempDir string) {
 				pidStr := strings.TrimSpace(lines[0])
 				if pid, parseErr := strconv.Atoi(pidStr); parseErr == nil {
 					// Try to kill the process (ignore errors since process might already be dead)
-					if process, findErr := os.FindProcess(pid); findErr == nil {
-						_ = process.Kill()
-					}
+					killCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
cuongdo, line 184: can you use t.Context() here?

go/test/endtoend/localprovisioner/cluster_test.go
```diff
@@ -130,15 +130,13 @@ func killProcessByPID(pid int) error {
 		return fmt.Errorf("invalid PID: %d", pid)
 	}
 
-	process, err := os.FindProcess(pid)
-	if err != nil {
-		return fmt.Errorf("failed to find process %d: %w", pid, err)
-	}
+	// Use SIGKILL to forcefully terminate with a short timeout
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
cuongdo, line 134: can `t.Context()` be used here?

go/cmd/pgctld/testutil/temp_dirs.go
dweitzman: Not in this case, because this runs within a `t.Cleanup()` call where `t.Context()` is already expired

go/test/endtoend/localprovisioner/cluster_test.go
dweitzman: This is also called in at least one places during `t.Cleanup()`, so using `t.Context()` without changing that caller would potentially be breaking

PR #635, opened 2026-02-19
feat: support COPY FROM STDIN inside transactions

GuptaManan100:
### Summary
- Adds support for `COPY FROM STDIN` operations within explicit transactions (BEGIN/COMMIT/ROLLBACK) by reusing the transaction's reserved connection for COPY instead of creating a new one
- Use the multi-reason reservation tracking on pooler connections - COPY now adds/removes its own reservation reason alongside the transaction reason, so connections stay reserved until both COPY and transaction complete
- Handles deferred BEGIN: when a transaction has no reserved connection yet (lazy BEGIN), creates one via a lightweight `SELECT 1` before initiating COPY
- Supports COPY embedded in multi-statement simple query batches (implicit transactions), including correct CommandTag deferral for COPY as the last statement


go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -472,6 +472,33 @@ func (sc *ScatterConn) CopyInitiate(
...
+		// then reuse it for COPY.
+		reservationOpts := protoutil.NewTransactionReservationOptions()
+		if state.PendingBeginQuery != "" {
+			reservationOpts.BeginQuery = state.PendingBeginQuery
+			state.PendingBeginQuery = ""
+		}
+		reservedState, err := sc.gateway.ReserveStreamExecute(
+			ctx, target, "SELECT 1", execOptions, reservationOpts,
+			func(context.Context, *sqltypes.Result) error { return nil }, // discard SELECT 1 result
+		)
```
haritabh17: The SELECT 1 works but it's a workaround for CopyReady not accepting ReservationOptions — its signature only takes ExecuteOptions, so there's no way to pass a BeginQuery through. The clean path would be adding ReservationOptions to CopyReady and calling BeginWithQuery before InitiateCopyFromStdin inside the executor, but that touches the full interface chain (queryservice → gRPC proto → executor). Worth a TODO comment here noting this?

go/multipooler/executor/executor.go
```diff
@@ -655,12 +662,15 @@ func (e *Executor) CopyAbort(
 
 	e.logger.DebugContext(ctx, "COPY FAIL completed")
 
-	// If write or read failed, connection might be in bad state - close it
+	// If write or read failed, connection is in bad state — release regardless.
 	if writeFailed || readErr != nil {
```
haritabh17, line 680: This releases with ReleaseError unconditionally — but if we're inside a transaction, doesn't this kill the transaction's reserved connection too? Should this check RemoveReservationReason first and only do ReleaseError if the connection is truly unrecoverable?

go/test/endtoend/queryserving/copy_test.go
```diff
@@ -486,6 +486,202 @@ func TestMultiGateway_CopyCommands(t *testing.T) {
...
+			pgx.CopyFromRows([][]any{{2, "Bob"}, {3, "Charlie"}}),
+		)
+		require.NoError(t, err, "COPY inside transaction should succeed")
+		assert.Equal(t, int64(2), rowsAffected)
+
+		// COMMIT
+		_, err = conn.Exec(ctx, "COMMIT")
+		require.NoError(t, err)
+
+		// Verify all 3 rows persist
```
haritabh17, line 536: Nice coverage. Would be worth adding a case for BEGIN → COPY → SELECT (verify copied data is visible within txn) → COMMIT — confirms the connection stays reserved and usable after COPY completes inside a transaction.

go/multipooler/executor/executor.go
GuptaManan100: Yep, fixed now!

go/test/endtoend/queryserving/copy_test.go
GuptaManan100: Added

go/services/multigateway/scatterconn/scatter_conn.go
GuptaManan100: Instead of TODO comment, I fixed it properly ❤️

go/multipooler/executor/executor.go
```diff
@@ -645,22 +667,25 @@ func (e *Executor) CopyAbort(
...
-		e.logger.DebugContext(ctx, "error reading response after CopyFail (expected)", "error", readErr)
+		e.logger.ErrorContext(ctx, "failed to read response after CopyFail", "error", readErr)
 	}
 
 	e.logger.DebugContext(ctx, "COPY FAIL completed")
 
-	// If write or read failed, connection might be in bad state - close it
 	if writeFailed || readErr != nil {
+		// Connection is in a bad protocol state — release it.
 		reservedConn.Release(reserved.ReleaseError)
```
haritabh17, line 682: Minor: when ReleaseError fires inside a transaction, the executor destroys the connection but the gateway's ShardState still holds the stale ReservedConnectionId. The next query will fail cleanly with "reserved connection not found" — so no correctness issue, but the user won't know the transaction is dead until they try another statement. Worth returning an error from CopyAbort in this case so the caller can surface a more helpful message like "transaction aborted due to COPY failure"?

GuptaManan100: I'm gonna fix it in a follow up, because I think it needs a proper fix and not just for this RPC.

PR #637, opened 2026-02-19
fix(parser): remove duplicate %type declarations from grammar

mkindahl:
35 duplicate token declarations across 16 lines caused goyacc (homebrew/non-golang.org version) to error out. Remove the later duplicates, keeping the first declaration of each token.

Signed-off-by: Mats Kindahl <mats.kindahl@supabase.io>

### DIFF
--- a/go/common/parser/postgres.go
+++ b/go/common/parser/postgres.go
@@ -1216,7 +1216,7 @@ const (
 	yyInitialStackSize = 16
 )
 
-//line postgres.y:15615
+//line postgres.y:15601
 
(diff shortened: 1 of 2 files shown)


mkindahl:
Note that there are still other errors in the grammar causing errors with the non-golang version. In particular, there are `%nonassoc` clauses for comparison operators, but the grammar supports expressions like `a < b < c`, which are not correct SQL, but this is captured at semantic level in PostgreSQL and we want to stay as close to the original grammar file as possible for maintenance reasons.

PR #639, opened 2026-02-19
fix(multipooler): reopen manager instead of just connections after DemoteStalePrimary

GuptaManan100:
## Summary

- Fix cancelled pool context after `DemoteStalePrimary` by calling `pm.Open()` instead of `pm.reopenConnections()`
- Guard `reopenConnections` to be a no-op when the manager is closed

## Problem

When `DemoteStalePrimary` runs on a stale primary, the flow is:

1. `stopPostgresIfRunning()` calls `pm.Close()`, which cancels `pm.ctx`
2. pg_rewind runs
3. `restartPostgresAsStandby()` calls `pm.reopenConnections()`, which creates new connection pools using the **already-cancelled** `pm.ctx`

All connections created after this point inherit the cancelled context. When the heartbeat writer later runs and a write times out, `handleContextCancellation` tries to cancel the backend query using a context derived from `c.conn.Context()` - but that context is born-dead because it descends from the cancelled `pm.ctx`. This produces:

```
{"level":"ERROR","msg":"CONTEXT CANCELLED!"}
{"level":"ERROR","msg":"handleContextCancellation started"}
{"level":"ERROR","msg":"handleContextCancellation finished","err":"connection pool context already expired","ok":false}
{"level":"ERROR","msg":"Failed to write heartbeat","error":"failed to write heartbeat: context deadline exceeded"}
```

## Fix

- **`restartPostgresAsStandby`**: Replace `pm.reopenConnections()` with `pm.Open()`. This recreates `pm.ctx` with a fresh context, reopens connection pools, and restarts the monitor and heartbeat tracker.
- **`reopenConnections`**: Add an `!isOpen` guard so it's a no-op when the manager is closed, preventing misuse where callers accidentally reopen connections with a cancelled context.

### DIFF
--- a/go/multipooler/manager/manager.go
+++ b/go/multipooler/manager/manager.go
@@ -417,6 +417,10 @@ func (pm *MultiPoolerManager) reopenConnections(_ context.Context) error {
 	pm.mu.Lock()
 	defer pm.mu.Unlock()
 
+	if !pm.isOpen {
+		return nil
+	}
+
 	pm.closeConnectionsLocked()
 	return pm.openConnectionsLocked()
 }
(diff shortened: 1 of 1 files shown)


GuptaManan100:
Closing this PR - the bug described here doesn't occur anymore.
                                                                                                                         
The PR description says `stopPostgresIfRunning()` calls `pm.Close()`, which cancels `pm.ctx`. But PR #542 already changed `stopPostgresIfRunning` from `pm.Close()` to `pm.closeConnectionsLocked()`, which does not cancel pm.ctx and does not set `isOpen = false`. So `reopenConnections()` already works correctly with a valid context.

PR #642, opened 2026-02-19
feat(pgctld): integrate crash recovery into PgRewind RPC

dweitzman:
Crash recovery is only needed for pg_rewind dry-run operations (where dry-run cannot modify the database to complete recovery). Integrating it into PgRewind simplifies the API.

### DIFF
--- a/go/test/endtoend/pgctld/pgctld_test.go
+++ b/go/test/endtoend/pgctld/pgctld_test.go
@@ -15,6 +15,7 @@
 package pgctld
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -83,65 +61,42 @@ func extractClusterState(output string) string {
...
-		return fmt.Errorf("failed to create stdin pipe: %w", err)
+		return fmt.Errorf("failed to open /dev/null: %w", err)
 	}
+	defer devNull.Close()
 
-	// Start the command
-	if err := cmd.Start(); err != nil {
-		return fmt.Errorf("failed to start postgres --single: %w", err)
-	}
+	cmd.Stdin = devNull
```
deepthi, line 81: Does this solve the "waiting for input" problem? 💯

dweitzman: Yes. This idea is directly borrowed from pg_rewind itself, as is the use of template1 as a db name: https://github.com/postgres/postgres/blob/6e466e1e839e2d40b30af9cbda107af867d259f6/src/bin/pg_rewind/pg_rewind.c#L1197

PR #647, opened 2026-02-20
refactor(multipooler/manager): improve lifecycle management and fix rewind bug

dweitzman:
This commit refactors the MultiPoolerManager lifecycle methods to improve ergonomics and prevent bugs where the manager could stay closed after errors during maintenance operations.

Renamed Close() to Shutdown() to gently push people toward using Pause() for any temporary pauses, where the erognomics of Pause() encourage resuming afterwards.

There's no actual change in behavior here, since errors were never returned anyway. This philosophical shift treats Open() as a state transition ("connections are now allowed") rather than a guarantee that connections exist.

Introduced Pause() which returns a resume function. This pattern guarantees that the manager is always reopened, even on error paths:

```
    resume := pm.Pause()
    defer resume()  // Safety net: always reopens
    // ... maintenance work ...
    resume()  // Explicit resume at the right time
```

The resume function is idempotent (safe to call multiple times), allowing both explicit calls and defer cleanup.

Fixes the bug and adds a regression test.

### DIFF
--- a/.github/workflows/test-pgregress.yml
+++ b/.github/workflows/test-pgregress.yml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: PostgreSQL Regression Tests
+name: PostgreSQL Compatibility Tests
 
 on:
   schedule:
@@ -25,7 +25,7 @@ permissions: {}
 
 jobs:
   test-pgregress:
-    name: Run PostgreSQL regression tests
+    name: Run PostgreSQL compatibility tests
     if: >-
       github.event_name != 'pull_request' ||
       contains(github.event.pull_request.labels.*.name, 'Run PgRegress Tests')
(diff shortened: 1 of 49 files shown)


go/multipooler/manager/manager.go
```diff
@@ -284,49 +284,85 @@ func (pm *MultiPoolerManager) execArgs(ctx context.Context, sql string, args ...
...
+//	// ... perform maintenance ...
+//	resume()  // Explicit resume at the right time
+//	// defer will call resume() again safely
+//
+// Note: Pause() cancels the manager's context, but Open() creates a fresh one.
+func (pm *MultiPoolerManager) Pause() (resume func()) {
+	if !pm.closeLocked("paused") {
+		// Already closed, return no-op resume
+		return func() {}
+	}
```
mkindahl, line 333: It is correct that you cannot resume a closed manager, but if the caller makes a mistake and calls this on a closed manager, it will go unnoticed since something sane is returned. The calling code could just save away the resume() function and then call it later, with no effect, when the assumption is that the manager will resume.
I think you need some stronger indication that the Pause() function is used in an incorrect situation and is probably a bug in the code. Since this is usually an error in the code, it might be sufficient to just print a log message. It should not happen, but if it does, it would be good to capture this error.

dweitzman: Hmm, I wonder if the right thing to do here is to make pause() log an error if the manager is already paused and resume unconditionally resume. The manager should always be running between actions, so if Pause() did nothing that seems unexpected. I'm going to update the code to do that

dweitzman: Done. Let me know if you disagree, otherwise I'll merge with this update

mkindahl: Yes, that looks good. It is just to be able to notice that something was wrong.

PR #650, opened 2026-02-23
feat(pgctld): enforce PGDATA ownership and 0700 before start

Verolop:
this PR enforces Postgres' `PGDATA` ownership and `0700` mode before every `pg_ctl start`.

In Kubernetes, certain lifecycle paths can alter directory mode bits (e.g. introduce setgid `2700`), which can cause `pg_ctl start` to fail even though the data is valid.

## Changes

A new helper, `ensurePGDATAPermissions`, is invoked from `StartPostgreSQLWithResult` immediately before `pg_ctl start`.

It:

- Verifies `PGDATA` exists and is a directory  
- Verifies owner UID matches `os.Geteuid()`  
- Normalises mode to `0700` if needed (clears setgid)  
- Fails fast on ownership mismatch; does not attempt to `chown` 
- Doesn't recurse or modify parent directories.

## Testing

Unit tests covering:

- No-op when already `0700`  
- Normalisation of `2700`, `0750`, and `0777` to `0700`  
- Error when path does not exist or when it's not a directory

### DIFF
--- a/go/common/pgprotocol/client/conn.go
+++ b/go/common/pgprotocol/client/conn.go
@@ -446,6 +446,53 @@ func (c *Conn) ReadCopyDoneResponse(ctx context.Context) (string, uint64, error)
 	}
 }
 
+// ReadCopyFailResponse reads the expected ErrorResponse + ReadyForQuery sequence
+// after sending CopyFail. Unlike ReadCopyDoneResponse, this treats ErrorResponse
+// as the expected (normal) response and continues reading until ReadyForQuery,
+// leaving the connection in a clean protocol state.
+func (c *Conn) ReadCopyFailResponse(ctx context.Context) error {
+	c.bufmu.Lock()
+	defer c.bufmu.Unlock()
+
+	gotError := false
+
+	for {
+		msgType, err := c.readMessageType()
+		if err != nil {
+			return fmt.Errorf("failed to read message type: %w", err)
+		}
+
+		length, err := c.readMessageLength()
+		if err != nil {
+			return fmt.Errorf("failed to read message length: %w", err)
+		}
+
+		body, err := c.readMessageBody(length)
+		if err != nil {
+			return fmt.Errorf("failed to read message body: %w", err)
+		}
+
+		switch msgType {
+		case protocol.MsgErrorResponse:
+			// Expected after CopyFail — consume it and continue to ReadyForQuery.
+			_ = body
+			gotError = true
+
+		case protocol.MsgNoticeResponse:
+			continue
+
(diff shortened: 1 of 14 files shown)


go/cmd/pgctld/command/pgdata_permissions_test.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2025 Supabase, Inc.
```
cuongdo: this should be 2026

go/cmd/pgctld/command/start.go
```diff
@@ -213,6 +218,47 @@ func StartPostgreSQLWithConfig(logger *slog.Logger, config *pgctld.PostgresCtlCo
...
+		return fmt.Errorf("PGDATA %s owned by UID %d, expected %d, refusing to start (ownership mismatch is a configuration error)",
+			dataDir, stat.Uid, currentUID)
+	}
+
+	if info.Mode().Perm() == 0o700 && (info.Mode()&os.ModeSetgid) == 0 {
+		return nil
+	}
+
+	oldMode := fmt.Sprintf("%04o", stat.Mode&0o7777)
+	if err := os.Chmod(dataDir, 0o700); err != nil {
```
cuongdo, line 248: pgbackrest server requires direct access to PGDATA. will this still allow that?

go/cmd/pgctld/command/start.go
```diff
@@ -213,6 +218,47 @@ func StartPostgreSQLWithConfig(logger *slog.Logger, config *pgctld.PostgresCtlCo
...
+		return fmt.Errorf("failed to chmod PGDATA %s to 0700: %w", dataDir, err)
+	}
+
+	logger.Info("Normalized PGDATA permissions",
```
cuongdo: we already have a lot of logging. maybe change this to debug log level?

go/cmd/pgctld/command/start.go
Verolop: yes, I saw `pgbackrest` runs as the same postgres user in the same container (in Dockerfile.pgctld), so 0700 gives it full access. This is also what initdb sets by default and what `pgbackrest` expects. The ownership check adds a safety net, i.e. if the UIDs ever diverged, we'd catch it here before either postgres or pgbackrest hit failures at runtime

go/cmd/pgctld/command/start.go
Verolop: updated

cuongdo:
Merged. The only CI error is caused by a Coveralls issue.

PR #657, opened 2026-02-24
refactor(mterrors): consolidate error handling with MT error codes and proper SQLSTATE

GuptaManan100:
## Summary

- Introduces `MTError` type that produces `*PgDiagnostic` errors with the MT code (e.g. `MT10001`) as the SQLSTATE, making multigres-generated errors directly identifiable by clients
- Adds `NewPgError()` helper for spoofing native PostgreSQL errors with real SQLSTATE codes (e.g. `28P01`, `08P01`)
- Removes `writeSimpleError`/`writeSimpleErrorWithDetail` - all error paths now go through `writeError` with proper `PgDiagnostic` types
- Removes unused `State` type and `NewError`/`NewErrorf`/`ErrState` from the mterrors package
- Migrates scattered `fmt.Errorf` and hardcoded `PgDiagnostic` literals across handler, engine, planner, and server packages to use the centralized error constructors

### DIFF
--- a/go/cmd/pgctld/command/pgdata_permissions_test.go
+++ b/go/cmd/pgctld/command/pgdata_permissions_test.go
@@ -1,0 +1,99 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package command
+
+import (
+	"log/slog"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEnsurePGDATAPermissions(t *testing.T) {
+	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelDebug}))
+
+	t.Run("0700 is a no-op", func(t *testing.T) {
+		dir := t.TempDir()
+		require.NoError(t, os.Chmod(dir, 0o700))
+
+		err := ensurePGDATAPermissions(logger, dir)
+		require.NoError(t, err)
+
+		info, err := os.Stat(dir)
+		require.NoError(t, err)
+		assert.Equal(t, os.FileMode(0o700), info.Mode().Perm())
+	})
+
+	t.Run("setgid 2700 becomes 0700", func(t *testing.T) {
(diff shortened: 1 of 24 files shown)


go/services/multigateway/handler/handler.go
```diff
@@ -76,12 +76,7 @@ func (h *MultiGatewayHandler) Consolidator() *preparedstatement.Consolidator {
 
 // errAbortedTransaction is the error returned when queries are executed in an aborted transaction.
 // PostgreSQL returns SQLSTATE 25P02 (in_failed_sql_transaction) for this condition.
-var errAbortedTransaction = &mterrors.PgDiagnostic{
-	MessageType: 'E',
-	Severity:    "ERROR",
-	Code:        "25P02",
-	Message:     "current transaction is aborted, commands ignored until end of transaction block",
-}
+var errAbortedTransaction = mterrors.MT10001.New()
```
haritabh17: This changes the SQLSTATE sent to clients from 25P02 (the standard PostgreSQL code for in_failed_sql_transaction) to MT10001. Clients like psycopg, pgx, and JDBC often check SQLSTATE to detect aborted transactions and trigger retry logic — e.g. err.pgcode == '25P02'. With MT10001, that stops matching. More broadly, MTError.New() puts 7-character MT IDs into the 5-character SQLSTATE field, which isn't spec-compliant. Should MT errors carry both an MT code (for internal identification) and a proper SQLSTATE (for the wire)?

Similar for all other error codes.

go/common/pgprotocol/server/conn.go
```diff
@@ -321,7 +322,7 @@ func (c *Conn) serve() error {
 	if err := c.handleStartup(); err != nil {
 		c.logger.Error("startup failed", "error", err)
 		// Try to send an error response before closing.
-		_ = c.writeSimpleErrorWithDetail("FATAL", "08P01", "connection startup failed", err.Error(), "")
+		_ = c.writeError(mterrors.MT14001.New(err.Error()))
```
haritabh17: The old code put the original error in the Detail field with "connection startup failed" as the main Message. Now everything is concatenated into Message as "connection startup failed: <err>" with no Detail. Should we preserve the error message fields??

go/services/multigateway/handler/handler.go
GuptaManan100: Good catches on both points. Fixed in the latest commits:

- Errors with PG equivalents now use real SQLSTATE codes. errAbortedTransaction uses 25P02 (via NewPgError) and the transaction isolation error uses 25001. These are no longer MT errors, client drivers will see standard codes and retry logic works as expected.
- MT codes are now 5 characters. Renamed from 7-char (e.g. MT13004) to 5-char (e.g. MTD04) using category encoding: 13→D, 14→E. MT codes are reserved for multigres-specific conditions that have no PostgreSQL equivalent.

go/common/pgprotocol/server/conn.go
GuptaManan100: Fixed. Added NewWithDetail() method on MTError, wrapper errors (parse failed, bind failed, etc.) now put the short description in Message and the underlying error in Detail, restoring the old writeSimpleErrorWithDetail behavior. MTError.New() also populates Detail from the error's Description field.

PR #659, opened 2026-02-24
feat(multigateway): add server-side TLS for PostgreSQL client connections

haritabh17:
Multigateway currently declines all SSL requests from clients (sends `N`). This means all PostgreSQL wire protocol traffic — queries, results, and SCRAM authentication data — flows in plaintext between client and proxy.

This PR adds server-side TLS so clients can establish encrypted connections to multigateway, following the PostgreSQL wire protocol specification. Scope is client → multigateway only; client-side TLS (multipooler → PG) and SCRAM channel binding are follow-up work.

## Changes

The startup negotiation in `handleSSLRequest()` / `handleGSSENCRequest()` previously only accepted `ProtocolVersionNumber` after declining encryption. Per the PostgreSQL protocol, the client may try the other encryption method before sending StartupMessage — this was rejected. A shared `readAndDispatchStartup()` with `sslDone`/`gssDone` flags now handles the full fallback ordering, matching PostgreSQL's `ProcessStartupPacket(port, ssl_done, gss_done)` pattern.

When `tlsConfig` is set, `handleSSLRequest()` sends `S`, checks for buffer-stuffing (CVE-2021-23222 — detects MITM-injected data between SSLRequest and TLS handshake), performs the TLS handshake via `tls.Server`, resets the buffered reader to wrap the TLS connection, then reads the StartupMessage over TLS.

Multigateway gets two new flags: `--pg-tls-cert-file` and `--pg-tls-key-file`. When both are set, it builds a `*tls.Config` with `MinVersion: TLS 1.2` and `NextProtos: ["postgresql"]` for PG 17 ALPN forward compatibility, and passes it through `ListenerConfig` → `Listener` → `Conn`. Flag validation is symmetric — providing either flag without the other returns a clear error at startup, preventing silent plaintext misconfiguration. The validation logic is extracted into `buildPGTLSConfig()` with unit tests covering all combinations.

The e2e test infrastructure gets a `WithMultigatewayTLS()` setup option that generates ephemeral CA + server certificates (exported `GenerateCA`/`GenerateCert` from `provisioner/local/tls.go`) and passes the cert/key paths to the multigateway process.

Unit tests cover all negotiation paths (SSL/GSS fallback ordering, duplicate rejection, TLS handshake success/failure, buffer-stuffing prevention, TLS 1.1 rejection). E2E tests cover all six `sslmode` values (disable, allow, prefer, require, verify-ca, verify-full), negative cases (missing/wrong root cert, hostname mismatch), and SCRAM authentication over TLS.

A TODO is added in `conn.serve()` for a follow-up startup phase timeout (equivalent to PostgreSQL's `authentication_timeout`), which would protect against clients that stall during SSL handshake, startup packet reading, or SCRAM exchange.

### DIFF
--- a/go/common/pgprotocol/server/conn.go
+++ b/go/common/pgprotocol/server/conn.go
@@ -341,7 +341,14 @@ func (c *Conn) serve() error {
 	if err := c.handleStartup(); err != nil {
 		c.logger.Error("startup failed", "error", err)
 		// Try to send an error response before closing.
-		_ = c.writeError(mterrors.MTE01.NewWithDetail(err.Error()))
+		// If the error is already a PgDiagnostic (e.g., duplicate SSLRequest
+		// with native SQLSTATE), send it directly. Otherwise, wrap with MTE01.
+		var diag *mterrors.PgDiagnostic
+		if errors.As(err, &diag) {
+			_ = c.writePgDiagnosticResponse(protocol.MsgErrorResponse, diag)
+		} else {
+			_ = c.writeError(mterrors.MTE01.NewWithDetail(err.Error()))
+		}
 		_ = c.flush()
 		return err
 	}
--- a/go/common/pgprotocol/server/startup.go
+++ b/go/common/pgprotocol/server/startup.go
@@ -60,13 +60,13 @@ func (c *Conn) readAndDispatchStartup() error {
 	switch protocolCode {
 	case protocol.SSLRequestCode:
 		if c.sslDone {
-			return errors.New("duplicate SSLRequest: SSL negotiation already completed")
+			return mterrors.NewPgError("FATAL", "0A000", "duplicate SSLRequest: SSL negotiation already completed", "")
 		}
 		return c.handleSSLRequest()
 
 	case protocol.GSSENCRequestCode:
 		if c.gssDone {
-			return errors.New("duplicate GSSENCRequest: GSSAPI encryption negotiation already completed")
+			return mterrors.NewPgError("FATAL", "0A000", "duplicate GSSENCRequest: GSSAPI encryption negotiation already completed", "")
 		}
 		return c.handleGSSENCRequest()
 


go/common/pgprotocol/server/conn.go
```diff
@@ -69,6 +70,19 @@ type Conn struct {
...
+	// When nil, SSLRequest is declined with 'N'.
+	tlsConfig *tls.Config
+
+	// sslDone indicates that an SSLRequest has already been handled
+	// (accepted or declined) for this connection. Prevents double negotiation.
+	sslDone bool
+
+	// gssDone indicates that a GSSENCRequest has already been handled
+	// for this connection. Prevents double negotiation.
+	gssDone bool
```
GuptaManan100, line 85: These fields are only being used to throw errors. If the client is well behaved, then it shouldn't matter. Also does postgres return an error in case multiple rounds of negotiation happen?

GuptaManan100, line 85: These fields are only being used to throw errors. If the client is well behaved, then it shouldn't matter. Also does postgres return an error in case multiple rounds of negotiation happen?

GuptaManan100: I checked in postgres, its the same.

go/common/pgprotocol/server/startup.go
```diff
@@ -31,118 +32,132 @@ type StartupMessage struct {
...
 		return fmt.Errorf("failed to read protocol code: %w", err)
 	}
 
-	// Handle special protocol codes.
 	switch protocolCode {
 	case protocol.SSLRequestCode:
-		// Client is requesting SSL. We don't support SSL yet, so decline.
+		if c.sslDone {
+			return errors.New("duplicate SSLRequest: SSL negotiation already completed")
+		}
```
GuptaManan100, line 64: We should mirror the postgres error, I don't think it would be this. Possibly use the mterrors package

go/common/pgprotocol/server/startup.go
```diff
@@ -31,118 +32,132 @@ type StartupMessage struct {
...
 		return c.handleSSLRequest()
 
 	case protocol.GSSENCRequestCode:
-		// Client is requesting GSSAPI encryption. We don't support it, so decline.
+		if c.gssDone {
+			return errors.New("duplicate GSSENCRequest: GSSAPI encryption negotiation already completed")
+		}
```
GuptaManan100, line 70: Same

go/common/pgprotocol/server/startup.go
haritabh17: Updating

go/common/pgprotocol/server/startup.go
haritabh17: Updating

PR #660, opened 2026-02-25
fix(multipooler): return authoritative reserved state from all operations

GuptaManan100:
## Summary

- Make the multipooler the single source of truth for reserved connection state by returning `ReservedState` (connection ID, pooler ID, remaining reasons) from `StreamExecute`, `CopyFinalize`, `CopyAbort`, and `ConcludeTransaction`
- Replace the gateway's independent reason-tracking (`RemoveReservationReason`) with a unified `updateShardState` helper that applies the multipooler's authoritative state
- Add `remaining_reasons` to `StreamExecuteResponse`, `PortalStreamExecuteResponse`, `CopyBidiExecuteResponse`, and `ConcludeTransactionResponse` protos; change `StreamExecute` RPC to return `StreamExecuteResponse` instead of raw `QueryResultPayload`
- Handle edge cases: ROLLBACK on a destroyed connection succeeds gracefully, destroyed connections during transactions set `TxnStatusFailed`, and `CopyBidiExecute` no longer dereferences a nil request after stream errors

## Problem

The gateway maintained its own bookkeeping of reservation reasons independently from the multipooler. This could drift out of sync, for example, if the multipooler destroyed a connection the gateway might still think it was reserved, or vice versa. The `StreamExecute` RPC returned raw `QueryResultPayload` with no way to communicate reserved connection state changes back to the gateway.

## Solution

Every multipooler operation that touches a reserved connection now returns the authoritative `ReservedState` (connection ID + pooler ID + remaining reasons bitmask). A zero connection ID means the connection was released or destroyed. The gateway's `ScatterConn.updateShardState` helper replaces all per-operation reason manipulation with a single function that either clears shard state (connection gone) or updates the reasons bitmask (connection still alive). This eliminates the possibility of state drift between the two services.

### DIFF
--- a/go/multipooler/executor/executor.go
+++ b/go/multipooler/executor/executor.go
@@ -95,7 +95,7 @@ func (e *Executor) ExecuteQuery(ctx context.Context, target *query.Target, sql s
 		// so we must explicitly apply settings changes here.
 		if options.SessionSettings != nil {
 			if err := e.poolManager.ApplySettingsToConn(ctx, reservedConn.Conn(), options.SessionSettings); err != nil {
-				return nil, fmt.Errorf("failed to apply settings to reserved connection: %w", err)
+				return nil, e.buildReservedState(reservedConn), fmt.Errorf("failed to apply settings to reserved connection: %w", err)
 			}
 		}
 
@@ -175,7 +175,7 @@ func (e *Executor) StreamExecute(
 		// so we must explicitly apply settings changes here.
 		if options.SessionSettings != nil {
 			if err := e.poolManager.ApplySettingsToConn(ctx, reservedConn.Conn(), options.SessionSettings); err != nil {
-				return fmt.Errorf("failed to apply settings to reserved connection: %w", err)
+				return e.buildReservedState(reservedConn), fmt.Errorf("failed to apply settings to reserved connection: %w", err)
 			}
 		}
 


go/services/multigateway/scatterconn/scatter_conn_test.go
```diff
@@ -249,3 +267,313 @@ func TestScatterConn_Case3_StreamExecuteError(t *testing.T) {
...
+		func(_ context.Context, _ *sqltypes.Result) error { return nil })
+
+	require.Error(t, err)
+	// Shard state must be cleared
+	require.Nil(t, state.GetMatchingShardState(target))
+	// Transaction status must be set to Failed
+	require.Equal(t, protocol.TxnStatusFailed, conn.TxnStatus())
+}
+
+func TestScatterConn_ConcludeTransaction_RollbackOnDestroyedConn(t *testing.T) {
```
haritabh17, line 340: There's no corresponding test for COMMIT on a destroyed connection, which should have the opposite behavior — the error must propagate since the client needs to know their COMMIT didn't happen. Worth adding?

go/multipooler/grpcpoolerservice/service.go
```diff
@@ -101,13 +118,16 @@ func (s *poolerService) ExecuteQuery(ctx context.Context, req *multipoolerpb.Exe
...
 	if err != nil {
 		// Convert errors to gRPC format, preserving PostgreSQL error details
 		return nil, mterrors.ToGRPC(err)
 	}
 	return &multipoolerpb.ExecuteQueryResponse{
-		Result: res.ToProto(),
+		Result:               res.ToProto(),
+		ReservedConnectionId: reservedState.ReservedConnectionId,
+		PoolerId:             reservedState.PoolerID,
+		RemainingReasons:     reservedState.ReservationReasons,
```
haritabh17: Worth extracting into helper function? `ReservedStateToProto` and similarly `ReservedStateFromProto` ??

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -58,6 +59,30 @@ func NewScatterConn(gateway poolergateway.Gateway, logger *slog.Logger) *Scatter
...
 
+// updateShardState replaces independent bookkeeping with the authoritative reservation
+// state from the multipooler. If the reserved connection ID is zero, the connection was
+// destroyed or released — clear the shard state. Otherwise, update the reservation reasons.
+//
+// When a reserved connection is destroyed while in a transaction, the transaction is marked
+// as failed (TxnStatusFailed) so subsequent queries are rejected until ROLLBACK. This is
+// defense-in-depth — handler.go also sets TxnStatusFailed on query errors, but setting it
+// here ensures coverage for all code paths (COPY, portal, etc.).
+func (sc *ScatterConn) updateShardState(
```
haritabh17: updateShardState both updates shard state AND sets TxnStatusFailed on the connection. These are two distinct concerns — shard bookkeeping vs client protocol state. Renaming to something like `applyReservedState` that hints at broader side effects ?? Can even break into 2 separate funcs, but thats probably not needed atm.

GuptaManan100: Done!

go/multipooler/grpcpoolerservice/service.go
GuptaManan100: Done

go/services/multigateway/scatterconn/scatter_conn_test.go
GuptaManan100: Done!

PR #661, opened 2026-02-25
feat(pgctld): add dedicated internal PostgreSQL user with cert auth

mkindahl:
feat(pgctld): add dedicated internal PostgreSQL user with socket auth

Add a dedicated multigres superuser role used by pgctld for internal
PostgreSQL operations. The role is created idempotently after PostgreSQL
starts (not during init), avoiding a fragile temporary start/stop cycle
and port conflicts during initialization.

- Add --pg-ctld-user flag (default: multigres) to pgctld; pass it through gRPC and CLI paths so every Start creates the role if absent
- Move user creation from InitDataDirWithResult to StartPostgreSQLWithResult: ensureInternalUser runs a psql DO block after pg_isready succeeds; uses DO block for idempotency since CREATE ROLE has no IF NOT EXISTS in PG17/18; wrappers (StartPostgreSQLWithConfig, RestartPostgreSQL) pass "" to skip
- Simplify InitDataDirWithResult signature: drop pgCertsDir and pgCtldUser params; update server.go call sites
- Update integration tests: replace password-based auth subtests with a socket_authentication subtest (retaining database existence, postgres and multigres role privilege checks); fix TestEndToEndWithRealPostgreSQL to pass a free port to init (avoids port 5432 conflict); fix inverted condition in TestOrphanDetectionWithRealPostgreSQL
- Update unit tests: TestPgCtldServiceInitDataDir now uses mock binaries (init no longer needs a real PG socket, testing.Short skip removed); remove mock createuser (psql mock covers ensureInternalUser)

### DIFF
--- a/go/cmd/pgctld/command/server.go
+++ b/go/cmd/pgctld/command/server.go
@@ -668,9 +668,9 @@ func (s *PgCtldService) PgRewind(ctx context.Context, req *pb.PgRewindRequest) (
 		}
 	}
 
-	// Construct source server connection string (without password - will use PGPASSWORD env var)
+	// Construct source server connection string
 	// Include application_name if provided (used for replication identification)
-	sourceServer := fmt.Sprintf("host=%s port=%d user=postgres dbname=postgres",
+	sourceServer := fmt.Sprintf("host=%s port=%d user=multigres dbname=postgres",
 		req.GetSourceHost(), req.GetSourcePort())
 	if req.GetApplicationName() != "" {
 		sourceServer = fmt.Sprintf("%s application_name=%s", sourceServer, req.GetApplicationName())
--- a/config/pgbackrest/pgbackrest_server_template.conf
+++ b/config/pgbackrest/pgbackrest_server_template.conf
@@ -13,5 +13,5 @@ tls-server-auth=pgbackrest=*
 pg1-socket-path={{.Pg1SocketPath}}
 pg1-port={{.Pg1Port}}
 pg1-path={{.Pg1Path}}
-pg1-user=postgres
+pg1-user=multigres
 pg1-database=postgres
--- a/config/pgbackrest/pgbackrest_template.conf
+++ b/config/pgbackrest/pgbackrest_template.conf
@@ -26,7 +26,7 @@ process-max=8
 pg1-socket-path={{.Pg1SocketPath}}
 pg1-port={{.Pg1Port}}
 pg1-path={{.Pg1Path}}
-pg1-user=postgres
+pg1-user={{.Pg1User}}
 pg1-database=postgres
 
 # pg2 configuration is not included in this template because it varies by backup.
--- a/demo/local/failover-test.go
+++ b/demo/local/failover-test.go
@@ -65,6 +65,7 @@ type PoolerInfo struct {
 	Cell      string
 	ServiceID string
 	PoolerDir string
+	PgUser    string
 	PgPort    int
 }
 
@@ -389,6 +390,7 @@ func getPoolerInfo(cell, serviceID string, config *Config) *PoolerInfo {
 		ServiceID: serviceID,
 		PoolerDir: cellConfig.Multipooler.PoolerDir,
 		PgPort:    cellConfig.Pgctld.PgPort,
+		PgUser:    cellConfig.Pgctld.PgUser,
 	}
 }
 
@@ -715,7 +717,7 @@ func printReplicationStatus(ctx context.Context, config *Config) {
 
 func runSQLQuery(poolerInfo *PoolerInfo, query string) string {
 	socketPath := filepath.Join(poolerInfo.PoolerDir, "pg_sockets")
-	connStr := fmt.Sprintf("host=%s port=%d user=postgres database=postgres sslmode=disable", socketPath, poolerInfo.PgPort)
+	connStr := fmt.Sprintf("host=%s port=%d user=%s database=postgres sslmode=disable", socketPath, poolerInfo.PgPort, poolerInfo.PgUser)
 
 	db, err := sql.Open("postgres", connStr)
 	if err != nil {
--- a/go/multipooler/manager/pg_config.go
+++ b/go/multipooler/manager/pg_config.go
@@ -148,7 +148,7 @@ func parseSynchronousStandbyNames(value string) (*SyncStandbyConfig, error) {
 }
 
 // parseAndRedactPrimaryConnInfo parses a PostgreSQL primary_conninfo connection string into structured fields
-// Example input: "host=localhost port=5432 user=postgres application_name=cell_name"
+// Example input: "host=localhost port=5432 user=multigres application_name=cell_name"
 // Returns a PrimaryConnInfo message with parsed fields, or an error if parsing fails
 // Note: Passwords are redacted in the raw field for security
 func parseAndRedactPrimaryConnInfo(connInfoStr string) (*multipoolermanagerdata.PrimaryConnInfo, error) {
--- a/go/multipooler/manager/pg_config_test.go
+++ b/go/multipooler/manager/pg_config_test.go
@@ -15,11 +15,13 @@
 package manager
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/multigres/multigres/go/common/constants"
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	multipoolermanagerdata "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 )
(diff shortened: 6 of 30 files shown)


go/provisioner/local/local.go
```diff
@@ -805,7 +782,7 @@ func (p *localProvisioner) provisionMultipooler(ctx context.Context, req *provis
 		"--pooler-dir", poolerDir,
 		"--pg-port", strconv.Itoa(pgPort),
 		"--hostname", "localhost",
-		"--connpool-admin-password", "postgres", // Password created in initializePgctldDirectories
+		"--connpool-admin-password", "postgres",
```
Verolop, line 785: since pgctld is dropping the password, is that the plan for multipooler as well? The reason I ask is the Operator creates a password Secret that both pgctld and multipooler consume. If multipooler is also moving off passwords, we can design to remove it entirely in one shot vs. refactoring it as multipooler-only

mkindahl: Yes, I think we can use a single (local) user for both multi-pooler and pgctld. Right now these changes are not in place, but I have a draft PR for that as well. It is probably good to put into a separate PR since it is a little more involved than just using the new user. There are a bunch of environment variables and options for multi-pooler that also need to change.

go/cmd/pgctld/command/init.go
```diff
@@ -189,9 +160,50 @@ func initializeDataDir(logger *slog.Logger, poolerDir string, pgUser string) err
...
+// superuser role (if it doesn't already exist), then stops PostgreSQL.
+// The role is created with SUPERUSER LOGIN so pgctld can use it for all internal operations.
+// It authenticates locally via peer auth and remotely via certificate auth.
+func createInternalUser(logger *slog.Logger, config *pgctld.PostgresCtlConfig, pgCtldUser string) error {
+	logger.Info("Starting PostgreSQL temporarily to create internal user", "user", pgCtldUser)
+
+	// Start PostgreSQL temporarily
+	if err := startPostgreSQLWithConfig(logger, config); err != nil {
+		return fmt.Errorf("failed to start PostgreSQL for user creation: %w", err)
+	}
```
mkindahl: It might be better to defer this to after the server is started and create the role as first action before marking the server as ready.

mkindahl: Done.

go/cmd/pgctld/command/start.go
```diff
@@ -203,9 +183,33 @@ func StartPostgreSQLWithResult(logger *slog.Logger, config *pgctld.PostgresCtlCo
...
-// StartPostgreSQLWithConfig starts PostgreSQL with the given configuration
+// ensureInternalUser creates the pgCtldUser superuser role if it does not already exist.
+// It runs against a live PostgreSQL server via the Unix socket (trust auth).
+// Uses a DO block for idempotency since CREATE ROLE has no IF NOT EXISTS in PG17/18.
+func ensureInternalUser(logger *slog.Logger, config *pgctld.PostgresCtlConfig, pgCtldUser string) error {
+	socketDir := pgctld.PostgresSocketDir(config.PoolerDir)
+	sql := fmt.Sprintf(
+		"DO $$ BEGIN IF NOT EXISTS (SELECT FROM pg_roles WHERE rolname = '%s') THEN CREATE ROLE %s WITH SUPERUSER LOGIN; END IF; END $$;",
+		pgCtldUser, pgCtldUser,
+	)
```
haritabh17, line 194: Should this use PG's format('%I', ...) for safe identifier quoting?
```
DO $$ BEGIN
  IF NOT EXISTS (SELECT FROM pg_roles WHERE rolname = 'multigres') THEN
    EXECUTE format('CREATE ROLE %I WITH SUPERUSER LOGIN', 'multigres');
  END IF;
END $$;
```

Here its not that crucial hence minor.

go/cmd/pgctld/command/server.go
```diff
@@ -668,12 +668,6 @@ func (s *PgCtldService) PgRewind(ctx context.Context, req *pb.PgRewindRequest) (
...
 
-	// Resolve password using existing function
-	password, err := resolvePassword(s.poolerDir)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve password: %w", err)
-	}
-
 	// Construct source server connection string (without password - will use PGPASSWORD env var)
 	// Include application_name if provided (used for replication identification)
 	sourceServer := fmt.Sprintf("host=%s port=%d user=postgres dbname=postgres",
```
haritabh17: sourceServer still uses user=postgres with no password and no client cert params. The new pg_hba only has cert auth for the multigres user and scram-sha-256 for everyone else — but no password was ever set for postgres. How does cross-node pg_rewind authenticate now? Does this need to switch to user=multigres with sslcert/sslkey/sslrootcert in the connection string?

go/cmd/pgctld/command/start.go
```diff
@@ -280,6 +284,31 @@ func startPostgreSQLWithConfig(logger *slog.Logger, config *pgctld.PostgresCtlCo
...
+	// SSL parameters are passed as -c flags rather than written to postgresql.conf to keep
+	// the config file portable across environments (e.g. backups restored to a different host).
+	if config.PgCertsDir != "" {
+		caCert := filepath.Join(config.PgCertsDir, "ca.crt")
+		serverCert := filepath.Join(config.PgCertsDir, "server.crt")
+		serverKey := filepath.Join(config.PgCertsDir, "server.key")
+		_, caCertErr := os.Stat(caCert)
+		_, serverCertErr := os.Stat(serverCert)
+		_, serverKeyErr := os.Stat(serverKey)
+		switch {
```
haritabh17, line 297: If PgCertsDir is set but a cert file is missing, PG starts without SSL. But the pg_hba_template has hostssl lines for the multigres role — those will reject all connections to that role since SSL isn't enabled. Should missing certs be a hard error when PgCertsDir is configured?

go/cmd/pgctld/command/init.go
```diff
@@ -129,57 +125,13 @@ func initializeDataDir(logger *slog.Logger, poolerDir string, pgUser string) err
...
-	}
-
 	// Build initdb command
 	// It's generally a good idea to enable page data checksums. Furthermore,
 	// pgBackRest will validate checksums for the Postgres cluster it's backing up.
 	// However, pgBackRest merely logs checksum validation errors but does not fail
 	// the backup.
 	args := []string{"-D", dataDir, "--data-checksums", "--auth-local=trust", "--auth-host=scram-sha-256", "-U", pgUser}
 
-	// If password is provided, create a temporary password file for initdb
```
haritabh17, line 157: With password infrastructure removed, the postgres user has no password. TCP connections via scram-sha-256 will always fail. Is this intentional — are TCP connections as postgres no longer expected? If so, could switch to --auth-host=reject or --auth-host=cert to make the intent explicit.

go/cmd/pgctld/command/root.go
```diff
@@ -77,6 +79,12 @@ func GetRootCommand() (*cobra.Command, *PgCtlCommand) {
 			FlagName: "pooler-dir",
 			Dynamic:  false,
 		}),
+		pgCtldUser: viperutil.Configure(reg, "pg-ctld-user", viperutil.Options[string]{
+			Default:  constants.DefaultMultigresUser,
+			FlagName: "pg-ctld-user",
+			EnvVars:  []string{"MULTIGRES_USER"},
```
haritabh17, line 85: The provisioner reads MULTIGRES_USER via os.Getenv(), while pgctld uses viperutil with EnvVars. Worth using a shared constant for the env var name.

config/postgres/pg_hba_template.conf
```diff
@@ -23,6 +23,11 @@ local   replication     all                                     trust
 host    replication     all             127.0.0.1/32            trust
 host    replication     all             ::1/128                 trust
 
+# Remote connections: certificate authentication for Multigres internal user
+# Used by pg_rewind --source-server (normal DB connection, not replication protocol)
+hostssl all             multigres       0.0.0.0/0               cert clientcert=verify-full
+hostssl all             multigres       ::0/0                   cert clientcert=verify-full
```
haritabh17, line 29: The hostssl lines hardcode multigres instead of using a template variable like {{.PgCtldUser}}. Intentional?

go/cmd/pgctld/command/server.go
```diff
@@ -668,12 +668,6 @@ func (s *PgCtldService) PgRewind(ctx context.Context, req *pb.PgRewindRequest) (
 		}
 	}
 
-	// Resolve password using existing function
-	password, err := resolvePassword(s.poolerDir)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve password: %w", err)
-	}
-
 	// Construct source server connection string (without password - will use PGPASSWORD env var)
```
haritabh17: Comments still say "without password - will use PGPASSWORD env var" and "passing password separately" but password was fully removed from PgRewindWithResult. Should be cleaned up.

go/cmd/pgctld/command/server.go
mkindahl: Cross-node rewind authenticate using servers. Two new certificates are added for that purpose: server.crt (and key) plus multigres.crt (plus key), and yes, it should switch to use the configured multigres user, and also use certificates and SSL. Will fix.

go/cmd/pgctld/command/start.go
mkindahl: I'm not actually sure we should create the user at all: it is error prone and a potential security problem. It would require special powers to create the user and this is better done either when bootstrapping the image or when building the image. Going to look if I can remove this function entirely.

go/cmd/pgctld/command/start.go
mkindahl: Yes, I think we should have all certificates in place and now start without them.

go/cmd/pgctld/command/root.go
mkindahl: Added variable to `constants` with the environment variable name.

config/postgres/pg_hba_template.conf
mkindahl: Nope. Fixed (used `{{.User}}`).

go/cmd/pgctld/command/server.go
mkindahl: Removed them (all I think).

mkindahl:
Closing this issue in favor of #677

PR #662, opened 2026-02-25
feat(multigateway): add statement_timeout at the gateway layer

GuptaManan100:
## Summary

- Add `--statement-timeout` flag (default 30s) to multigateway for server-wide query timeout enforcement
- Implement `SET/SHOW/RESET statement_timeout` as a gateway-managed variable — intercepted locally without forwarding to PostgreSQL, preserving pooled connection hygiene
- Enforce timeouts via context deadlines that propagate through gRPC to the multipooler, which cancels the PostgreSQL backend query on expiry (SQLSTATE `57014`)
- Support startup parameter overrides (e.g., `options='--statement-timeout=5s'` in the connection string)

## Description

`statement_timeout` is the first "gateway-managed variable" — a session variable the gateway intercepts and does not forward to PostgreSQL. This avoids polluting pooled backend connections with per-client timeouts and allows enforcement via context deadlines rather than relying on PostgreSQL's internal timer.

**Timeout priority** (highest wins): per-query SQL directive (future) > `SET statement_timeout` > `--statement-timeout` flag.

The implementation introduces a `GatewayManagedVariable[T]` generic type that holds a default (from flag/startup params) and an optional session override. The planner intercepts SET/SHOW/RESET for gateway-managed variables and creates local-only primitives (`GatewaySessionState`, `GatewayShowVariable`) instead of routing to PostgreSQL. `RESET ALL` correctly resets gateway-managed variables via `ApplySessionState`.

For the extended query protocol, `PlanPortal` checks if a portal's statement needs local handling (gateway-managed variables, RESET ALL) and delegates to `Plan` for the actual planning. All other statements continue through `PortalStreamExecute` with bound parameters as before.

Two bugs were fixed during development:
- `Route.StreamExecute` was passing `conn.Context()` instead of the `ctx` parameter, discarding timeout deadlines set by `executeWithTimeout`.
- `PortalStreamExecute` skipped the planner entirely, so gateway-managed variables sent via the extended protocol (e.g., pgx) were forwarded to PostgreSQL instead of being intercepted.

### DIFF
--- a/go/services/multigateway/init.go
+++ b/go/services/multigateway/init.go
@@ -20,6 +20,8 @@ package multigateway
 
 import (
 	"context"
+	"crypto/tls"
+	"errors"
 	"fmt"
 	"time"
 
@@ -27,6 +29,7 @@ import (
 	"github.com/spf13/pflag"
 
 	"github.com/multigres/multigres/go/common/constants"
+	"github.com/multigres/multigres/go/common/pgprotocol/protocol"
 	"github.com/multigres/multigres/go/common/pgprotocol/server"
 	"github.com/multigres/multigres/go/common/servenv"
 	"github.com/multigres/multigres/go/common/servenv/toporeg"
@@ -48,6 +51,10 @@ type MultiGateway struct {
 	pgPort viperutil.Value[int]
 	// pgBindAddress is the address to bind the PostgreSQL listener to
 	pgBindAddress viperutil.Value[string]
+	// pgTLSCertFile is the path to the TLS certificate file for PostgreSQL SSL connections.
+	pgTLSCertFile viperutil.Value[string]
+	// pgTLSKeyFile is the path to the TLS private key file for PostgreSQL SSL connections.
+	pgTLSKeyFile viperutil.Value[string]
 	// poolerDiscovery handles discovery of multipoolers across all cells
 	poolerDiscovery *GlobalPoolerDiscovery
 	// poolerGateway manages connections to poolers
@@ -106,6 +113,18 @@ func NewMultiGateway() *MultiGateway {
 			Dynamic:  false,
 			EnvVars:  []string{"MT_STATEMENT_TIMEOUT"},
 		}),
+		pgTLSCertFile: viperutil.Configure(reg, "pg-tls-cert-file", viperutil.Options[string]{
+			Default:  "",
+			FlagName: "pg-tls-cert-file",
+			Dynamic:  false,
+			EnvVars:  []string{"MT_PG_TLS_CERT_FILE"},
+		}),
+		pgTLSKeyFile: viperutil.Configure(reg, "pg-tls-key-file", viperutil.Options[string]{
+			Default:  "",
+			FlagName: "pg-tls-key-file",
+			Dynamic:  false,
+			EnvVars:  []string{"MT_PG_TLS_KEY_FILE"},
+		}),
 		grpcServer: servenv.NewGrpcServer(reg),
 		senv:       servenv.NewServEnv(reg),
 		topoConfig: topoclient.NewTopoConfig(reg),
@@ -139,12 +158,16 @@ func (mg *MultiGateway) RegisterFlags(fs *pflag.FlagSet) {
 	fs.Int("pg-port", mg.pgPort.Default(), "PostgreSQL protocol listen port")
 	fs.String("pg-bind-address", mg.pgBindAddress.Default(), "address to bind the PostgreSQL listener to")
 	fs.Duration("statement-timeout", mg.statementTimeout.Default(), "Default statement execution timeout. 0 disables.")
+	fs.String("pg-tls-cert-file", mg.pgTLSCertFile.Default(), "path to TLS certificate file for PostgreSQL SSL connections")
+	fs.String("pg-tls-key-file", mg.pgTLSKeyFile.Default(), "path to TLS private key file for PostgreSQL SSL connections")
 	viperutil.BindFlags(fs,
 		mg.cell,
 		mg.serviceID,
 		mg.pgPort,
 		mg.pgBindAddress,
 		mg.statementTimeout,
+		mg.pgTLSCertFile,
+		mg.pgTLSKeyFile,
 	)
 	mg.senv.RegisterFlags(fs)
 	mg.grpcServer.RegisterFlags(fs)
@@ -204,6 +227,17 @@ func (mg *MultiGateway) Init() error {
 	// Create hash provider for SCRAM authentication using the pooler gateway
 	hashProvider := auth.NewPoolerHashProvider(&poolerSystemDiscovererAdapter{pg: mg.poolerGateway})
 
+	// Build TLS config if cert and key files are provided.
+	certFile := mg.pgTLSCertFile.Get()
+	keyFile := mg.pgTLSKeyFile.Get()
+	pgTLSConfig, err := buildPGTLSConfig(certFile, keyFile)
+	if err != nil {
+		return err
+	}
+	if pgTLSConfig != nil {
+		logger.Info("TLS configured for PostgreSQL listener", "cert_file", certFile, "key_file", keyFile)
+	}
+
 	// Create and start PostgreSQL protocol listener
 	mg.pgHandler = handler.NewMultiGatewayHandler(mg.executor, logger, mg.statementTimeout.Get())
 	pgAddr := fmt.Sprintf("%s:%d", mg.pgBindAddress.Get(), mg.pgPort.Get())
@@ -211,6 +245,7 @@ func (mg *MultiGateway) Init() error {
 		Address:      pgAddr,
 		Handler:      mg.pgHandler,
 		HashProvider: hashProvider,
+		TLSConfig:    pgTLSConfig,
 		Logger:       logger,
 	})
 	if err != nil {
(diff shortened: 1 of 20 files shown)


go/services/multigateway/handler/handler.go
```diff
@@ -213,7 +266,11 @@ func (h *MultiGatewayHandler) HandleExecute(ctx context.Context, conn *server.Co
 		}
 	}
 
-	err := h.executor.PortalStreamExecute(ctx, conn, state, portalInfo, maxRows, callback)
+	// Use the original query string for directive parsing (extended protocol preserves comments).
+	ast := portalInfo.PreparedStatementInfo.AstStmt()
```
haritabh17: Shadows the ast package import used a few lines above (ast.IsRollbackStatement). Works because the package ref is before the shadowing, but could cause subtle bugs if code is reordered. Rename to astStmt?

go/services/multigateway/planner/variable_set_stmt.go
```diff
@@ -74,6 +94,64 @@ func (p *Planner) planVariableSetStmt(
...
+	sql string,
+	stmt *ast.VariableSetStmt,
+	value string,
+) (engine.Primitive, error) {
+	name := strings.ToLower(stmt.Name)
+
+	switch stmt.Kind {
+	case ast.VAR_SET_VALUE:
+		switch name {
+		case "statement_timeout":
```
haritabh17, line 123: isGatewayManagedVariable and planGatewayManagedVariable are two separate switches that must stay in sync. Adding a variable to one but not the other causes a runtime "unrecognized parameter" error. Would a registry pattern (map[string]handler) be worth it here to make this impossible?

go/services/multigateway/engine/gateway_session_state.go
```diff
@@ -0,0 +1,110 @@
...
+			state.ResetStatementTimeout()
+		} else {
+			state.SetStatementTimeout(g.statementTimeout)
+		}
+	default:
+		return &mterrors.PgDiagnostic{
+			MessageType: 'E',
+			Severity:    "ERROR",
+			Code:        "42704", // undefined_object
+			Message:     fmt.Sprintf("unrecognized configuration parameter %q", g.variable),
```
haritabh17: Could this be a shared helper like mterrors.NewUnrecognizedParameter(name)? Could be used in the next file gateway_show_variable.go as well

go/services/multigateway/init.go
```diff
@@ -97,6 +100,12 @@ func NewMultiGateway() *MultiGateway {
 			Dynamic:  false,
 			EnvVars:  []string{"MT_PG_BIND_ADDRESS"},
 		}),
+		statementTimeout: viperutil.Configure(reg, "statement-timeout", viperutil.Options[time.Duration]{
+			Default:  30 * time.Second,
```
haritabh17, line 111: minor: default statement_timeout is 30s, but PG's default is 0 (disabled). Should the default be 0 to match PG, with 30s as a recommended production setting in the docs?

go/services/multigateway/handler/statement_timeout.go
```diff
@@ -0,0 +1,95 @@
...
+func outOfRangeParamError(paramName, value string) *mterrors.PgDiagnostic {
+	return mterrors.NewPgError("ERROR", mterrors.PgSSInvalidParameterValue,
+		fmt.Sprintf("%s is outside the valid range for parameter %q (0 .. 2147483647)", value, paramName), "")
+}
+
+// formatDurationAsMs formats a time.Duration as a PostgreSQL-compatible milliseconds string.
+// This matches PostgreSQL's display format for statement_timeout.
+func formatDurationAsMs(d time.Duration) string {
+	return strconv.FormatInt(d.Milliseconds(), 10)
+}
```
haritabh17: The doc comment says "matches PostgreSQL's display format" but PG's SHOW statement_timeout returns human-friendly units (5s, 1min, 500ms) via GUC_UNIT_MS conversion, not raw milliseconds. This returns "5000" where PG returns "5s". Is the raw format intentional, or should it match PG's unit conversion?

go/services/multigateway/handler/statement_timeout.go
```diff
@@ -0,0 +1,95 @@
...
+	// Try parsing as plain integer (milliseconds) first — this is the common PG case.
+	if ms, err := strconv.ParseInt(value, 10, 64); err == nil {
+		if ms < 0 {
+			return 0, outOfRangeParamError(paramName, value)
+		}
+		return time.Duration(ms) * time.Millisecond, nil
+	}
+
+	// Try Go duration format (e.g., "30s", "200ms", "1m").
+	d, err := time.ParseDuration(value)
```
haritabh17, line 66: minor: time.ParseDuration accepts formats PG doesn't — 1h30m, 1.5s, 300µs all work here but would fail on vanilla PG. Could this cause portability issues if someone develops against multigres then deploys to PG?

go/services/multigateway/engine/gateway_session_state.go
```diff
@@ -0,0 +1,106 @@
...
+// Values are parsed at plan time and stored in typed fields so execution is a
+// simple assignment with no parsing overhead per query.
+type GatewaySessionState struct {
+	sql      string // Original SQL for debugging
+	variable string // Variable name (e.g., "statement_timeout")
+	isReset  bool   // true for RESET, false for SET
+
+	// Typed fields for each gateway-managed variable.
+	// Only the field matching `variable` is used.
+	statementTimeout time.Duration
```
haritabh17, line 40: When more gateway-managed variables are added (lock_timeout, idle_in_transaction_session_timeout, etc.), each one needs a new typed field, a new constructor, and a new switch case. Would a registry or map-based pattern be worth considering to avoid that per-variable boilerplate?

We can add it later though, not immediately needed, can be added when we add those next vars.

go/services/multigateway/planner/variable_set_stmt.go
```diff
@@ -74,6 +94,56 @@ func (p *Planner) planVariableSetStmt(
...
+			return engine.NewStatementTimeoutSet(sql, d), nil
+		default:
+			return nil, mterrors.NewPgError("ERROR", mterrors.PgSSUndefinedObject,
+				fmt.Sprintf("unrecognized configuration parameter %q", name), "")
+		}
+
+	case ast.VAR_RESET, ast.VAR_SET_DEFAULT, ast.VAR_RESET_ALL:
+		// RESET, SET ... TO DEFAULT, and RESET ALL all revert to the flag default.
+		p.logger.Debug("planning RESET gateway-managed variable", "variable", name)
+		return engine.NewGatewaySessionStateReset(sql, name), nil
```
haritabh17, line 142: RESET ALL has stmt.Name = "" which won't pass isGatewayManagedVariable on line 47, so this VAR_RESET_ALL case is unreachable through the current control flow. Also means RESET ALL doesn't actually reset statement_timeout (goes through the old path → ResetAllSessionVariables which only clears SessionSettings). Dead code or routing issue?

go/services/multigateway/engine/gateway_session_state.go
```diff
@@ -0,0 +1,106 @@
...
+	switch g.variable {
+	case "statement_timeout":
+		if g.isReset {
+			state.ResetStatementTimeout()
+		} else {
+			state.SetStatementTimeout(g.statementTimeout)
+		}
+	default:
+		return mterrors.NewPgError("ERROR", mterrors.PgSSUndefinedObject,
+			fmt.Sprintf("unrecognized configuration parameter %q", g.variable), "")
```
haritabh17: The "unrecognized configuration parameter" error here is already caught by the planner in planGatewayManagedVariable. This default branch in StreamExecute should be unreachable — worth removing or replacing with an Assert to make that explicit?

go/services/multigateway/init.go
GuptaManan100: I think 30s default is fine. We ran with that in Vitess too. Having no timeout can be a way to completely block all traffic if you have 500 odd queries blocked on postgres

go/services/multigateway/handler/handler.go
GuptaManan100: Good catch! renamed to `astStmt` to avoid shadowing the `ast` package import.

go/services/multigateway/planner/variable_set_stmt.go
GuptaManan100: Agree the dual-switch is a maintenance risk. For now with just one variable I think it's fine, when we add the next gateway-managed variable (lock_timeout, idle_in_transaction_session_timeout), I'll refactor to a registry pattern that makes it impossible to add one without the other.

go/services/multigateway/engine/gateway_session_state.go
GuptaManan100: Added `mterrors.NewUnrecognizedParameter(name)` and updated all three call sites (gateway_session_state.go, gateway_show_variable.go, variable_set_stmt.go).

go/services/multigateway/handler/statement_timeout.go
GuptaManan100: Fair point. In practice the risk is low, anyone using compound durations like `1h30m` would likely be setting this via the gateway flag (where Go format is natural) rather than SQL `SET`. But I agree it could be a portability surprise. For now a superset I think is fine.

go/services/multigateway/engine/gateway_session_state.go
GuptaManan100: Agreed, will add when we add the next variable. One variable doesn't justify the abstraction yet.

go/services/multigateway/planner/variable_set_stmt.go
GuptaManan100: Great catch! removed `VAR_RESET_ALL` from `planGatewayManagedVariable` and added a comment explaining the routing: RESET ALL flows through Route + ApplySessionState (which calls both `ResetAllSessionVariables()` and `ResetStatementTimeout()`), so gateway-managed variables are correctly reset without needing a separate code path here.

go/services/multigateway/engine/gateway_session_state.go
GuptaManan100: Replaced with a panic that clearly documents the invariant: the planner validates the variable name before creating these primitives, so reaching the default means a code bug (new variable added to `isGatewayManagedVariable` but not to the switch).

go/services/multigateway/handler/statement_timeout.go
GuptaManan100: Fixed!

PR #631, opened 2026-02-18
feat(multipooler): support streaming health checks

dweitzman:
## Overview

This PR implements the health streaming infrastructure between multipooler and multigateway, establishing the foundation for health-aware load balancing.

**Important**: This PR implements the infrastructure for health monitoring but does **not yet integrate** health state into routing decisions. That integration is planned for follow-up work.

### Server-Side (Multipooler)

Implements health state management and streaming:

- **healthStreamer** manages health state with typed update methods (`UpdateServingStatus`, `UpdatePoolerType`, `UpdatePrimaryObservation`)
- **StreamPoolerHealth RPC** in gRPC service streams health updates to clients
- **Atomic updates**: state changes and broadcasts are synchronized
- **Periodic heartbeats**: 30-second broadcasts keep connections alive and detect staleness
- **Non-blocking sends**: slow clients get disconnected rather than blocking all broadcasts
- **HealthProvider interface** for integration with pooler server

### Client-Side (Multigateway)

Implements health stream subscription and monitoring:

- **PoolerConnection** subscribes to health streams automatically
- **Automatic reconnection** with exponential backoff on stream failures
- **Staleness detection** using server-provided timeout with client-side enforcement
- **Health state tracking**: serving status, pooler type, primary observation
- **Callback mechanism** (`onHealthUpdate`) designed for LoadBalancer integration (currently unused)

## What This PR Does NOT Do

- ❌ LoadBalancer **does not** filter unhealthy poolers from routing
- ❌ LoadBalancer **does not** use health state in `selectConnection()`
- ❌ The `onHealthUpdate` callback is passed as `nil` (not wired up)
- ❌ Health information is **collected but not consulted** for routing decisions

## Future Work

To integrate health into routing:
- Wire up `onHealthUpdate` callback in LoadBalancer.addPooler()
- Add health filtering in LoadBalancer.GetConnection() candidate selection
- Define health criteria (e.g., must be SERVING status, recent update within staleness threshold)

PR #665, opened 2026-02-25
feat(multipooler): streaming health check from pooler to gateway

haritabh17:
## Summary

Implements streaming health checks between multipooler and multigateway, enabling health-aware load balancing (MUL-60).

**Proto**
Adds a `StreamPoolerHealth` server-streaming RPC. Each response carries the pooler's serving status, its view of the current primary (`PrimaryObservation` with term), and a recommended staleness timeout so the client knows when to consider the stream dead.

**Server side (multipooler)**
A new `healthStreamer` in the manager owns all health state and fans out updates to subscribers. Typed mutators (`UpdateServingStatus`, `UpdatePoolerType`, `UpdatePrimaryObservation`) atomically update state and broadcast. A 30s heartbeat keeps streams alive when nothing changes. Clients that fall behind (full buffer) are disconnected so they reconnect with fresh state.

**Client side (multigateway)**
`PoolerConnection` maintains a long-lived health stream per pooler with exponential-backoff reconnection and a staleness timer. On each response it snapshots the health into an immutable `PoolerHealth` value that the `LoadBalancer` can read lock-free.

**Load balancing**
`GetConnection` now uses live health for routing:
- *Primary*: consults every pooler in the shard for its `PrimaryObservation` and picks the one with the highest term — handles split-brain during failover. Excluded (recently-failed) poolers are still consulted for their observations.
- *Replica*: tiered preference — local serving > remote serving > local not-serving > any.

`UpdatePoolerInfo` is now race-free (`atomic.Pointer`) so topology updates and health reads can happen concurrently.

**Other**
- Propagates a `shutdownCtx` through multigateway so health-stream goroutines stop promptly on shutdown.
- Fixes pgbackrest config artifact in grpcmanagerservice tests (`t.TempDir()`).

Based on work from PR #631 and `dw-primary-reconcile`, rebased onto current main.

### DIFF
--- a/go/services/multigateway/poolergateway/load_balancer.go
+++ b/go/services/multigateway/poolergateway/load_balancer.go
@@ -91,7 +91,30 @@ func (lb *LoadBalancer) AddPooler(pooler *clustermetadatapb.MultiPooler) error {
 
 	// Check if already exists - update info instead of skipping
 	if conn, exists := lb.connections[poolerID]; exists {
+		oldType := conn.Type()
 		conn.UpdatePoolerInfo(pooler)
+		newType := conn.Type()
+
+		// Invalidate seed if type changed away from PRIMARY
+		if oldType == clustermetadatapb.PoolerType_PRIMARY && newType != clustermetadatapb.PoolerType_PRIMARY {
+			key := shardKey{tableGroup: pooler.GetTableGroup(), shard: pooler.GetShard()}
+			if cached, ok := lb.cachedPrimaries[key]; ok && cached.conn == conn && cached.term == 0 {
+				delete(lb.cachedPrimaries, key)
+				lb.logger.Debug("invalidated seeded primary cache on type change",
+					"pooler_id", poolerID, "old_type", oldType, "new_type", newType)
+			}
+		}
+
+		// Seed if type changed to PRIMARY
+		if oldType != clustermetadatapb.PoolerType_PRIMARY && newType == clustermetadatapb.PoolerType_PRIMARY {
+			key := shardKey{tableGroup: pooler.GetTableGroup(), shard: pooler.GetShard()}
+			if existing, ok := lb.cachedPrimaries[key]; !ok || existing.term == 0 {
+				lb.cachedPrimaries[key] = &cachedPrimary{conn: conn, term: 0}
+				lb.logger.Debug("seeded primary cache on type change",
+					"pooler_id", poolerID, "old_type", oldType, "new_type", newType)
+			}
+		}
+
 		return nil
 	}
 
@@ -101,6 +124,18 @@ func (lb *LoadBalancer) AddPooler(pooler *clustermetadatapb.MultiPooler) error {
 	}
 
 	lb.connections[poolerID] = conn
+
+	// Seed primary cache from discovery type (term 0 = unconfirmed).
+	// Health stream callbacks will overwrite with real term data.
+	if pooler.Type == clustermetadatapb.PoolerType_PRIMARY {
+		key := shardKey{tableGroup: pooler.GetTableGroup(), shard: pooler.GetShard()}
+		if existing, ok := lb.cachedPrimaries[key]; !ok || existing.term == 0 {
+			lb.cachedPrimaries[key] = &cachedPrimary{conn: conn, term: 0}
+			lb.logger.Debug("seeded primary cache from discovery",
+				"pooler_id", poolerID, "tablegroup", key.tableGroup, "shard", key.shard)
+		}
+	}
+
 	lb.logger.Debug("added pooler connection",
 		"pooler_id", poolerID,
 		"type", pooler.Type.String(),
--- a/go/services/multigateway/poolergateway/load_balancer_test.go
+++ b/go/services/multigateway/poolergateway/load_balancer_test.go
@@ -364,7 +364,7 @@ func TestLoadBalancer_PrimaryCaching(t *testing.T) {
 		assert.Equal(t, poolerID(primary1), conn.ID(), "Should trust replica's observation with highest term")
 	})
 
-	t.Run("no observations returns UNAVAILABLE", func(t *testing.T) {
+	t.Run("no observations uses discovery seed", func(t *testing.T) {
 		logger := slog.Default()
 		lb := NewLoadBalancer(context.Background(), "zone1", logger)
 
@@ -374,18 +374,19 @@ func TestLoadBalancer_PrimaryCaching(t *testing.T) {
 		require.NoError(t, lb.AddPooler(primary))
 		require.NoError(t, lb.AddPooler(replica))
 
-		// No health updates — cache is empty, no type-based fallback
+		// No health updates — but discovery seed (term 0) provides a primary
 		target := &query.Target{
 			TableGroup: constants.DefaultTableGroup,
 			Shard:      "0",
 			PoolerType: clustermetadatapb.PoolerType_PRIMARY,
 		}
-		_, err := lb.GetConnection(target)
-		require.Error(t, err)
-		assert.Contains(t, err.Error(), "no primary cached")
+		conn, err := lb.GetConnection(target)
+		require.NoError(t, err)
+		assert.Equal(t, poolerID(primary), conn.ID(),
+			"Should return discovery-seeded primary before health stream connects")
 	})
 
-	t.Run("unwatched primary not cached", func(t *testing.T) {
+	t.Run("unwatched primary keeps discovery seed", func(t *testing.T) {
 		logger := slog.Default()
 		lb := NewLoadBalancer(context.Background(), "zone1", logger)
 
@@ -409,17 +410,17 @@ func TestLoadBalancer_PrimaryCaching(t *testing.T) {
 				PrimaryTerm: 100,
 			})
 
-		// Cache should NOT be updated (unknown-primary not in connections).
-		// No type-based fallback — returns UNAVAILABLE.
+		// Cache NOT updated (unknown-primary not in connections).
+		// Discovery seed (term 0) for primary1 remains intact.
 		target := &query.Target{
 			TableGroup: constants.DefaultTableGroup,
 			Shard:      "0",
 			PoolerType: clustermetadatapb.PoolerType_PRIMARY,
 		}
-		_, err := lb.GetConnection(target)
-		require.Error(t, err)
-		assert.Contains(t, err.Error(), "no primary cached",
-			"Should return UNAVAILABLE when observed primary not in connections")
+		conn, err := lb.GetConnection(target)
+		require.NoError(t, err)
+		assert.Equal(t, poolerID(primary1), conn.ID(),
+			"Discovery seed should survive when observed primary is not in connections")
 	})
 
 	t.Run("cache invalidated on pooler removal", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


proto/multipoolerservice.proto
```diff
@@ -383,4 +399,42 @@ message ReleaseReservedConnectionRequest {
...
+
+// PrimaryObservation represents a pooler's view of who the primary is.
+message PrimaryObservation {
+  // primary_id is the ID of the pooler this node believes is the primary.
+  // May be this pooler's own ID if it believes itself to be primary.
+  clustermetadata.ID primary_id = 1;
+
+  // term is the consensus term at which this observation was made.
+  // Higher terms indicate more recent observations.
+  int64 term = 2;
```
dweitzman: I think this should probably be a "primary_term", not consensus/coordinator term. At the beginning of a new coordinator term we may not have revoked the old primary term yet (in the case of graceful demote) and we may not have selected a new primary yet (since we don't select one until after revoking the previous primary term)

The primary never changes within a primary_term, though.

This name + comment change from term to primary_term would touch other files too (structs, etc.)

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -32,6 +29,13 @@ import (
 	"github.com/multigres/multigres/go/pb/query"
 )
 
+// GetConnectionOptions specifies options for connection selection.
+type GetConnectionOptions struct {
+	// ExcludePoolers is a list of pooler IDs to exclude from selection.
+	// This is used for retry logic to avoid retrying on a pooler that just failed.
+	ExcludePoolers []string
```
dweitzman: I think query retries are out of scope for this PR, so maybe the exclude poolers logic isn't needed yet? My inclination would be to wait until we need it to add the code

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -42,45 +46,46 @@ type LoadBalancer struct {
 	// logger for debugging
 	logger *slog.Logger
 
+	// ctx is the service-lifetime context for child goroutines (health streams)
+	ctx context.Context
+
 	// mu protects the connections map
-	mu sync.Mutex
+	mu sync.RWMutex
```
dweitzman: I think our opinion is generally to prefer Mutex over RWMutex for scenarios where locks are acquired frequently for short intervals, since RWMutex has significant overhead

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -96,67 +101,101 @@ func (lb *LoadBalancer) AddPooler(pooler *clustermetadatapb.MultiPooler) error {
...
-	return lb.selectConnection(candidates, target), nil
+	return lb.selectReplicaConnection(candidates), nil
+}
+
+// GetConnectionContext returns a PoolerConnection matching the target specification.
+// If no suitable connection is immediately available, it waits until one becomes
+// available or the context is cancelled.
+//
+// This is useful during failover when waiting for a new primary to be elected.
+func (lb *LoadBalancer) GetConnectionContext(ctx context.Context, target *query.Target, opts *GetConnectionOptions) (*PoolerConnection, error) {
```
dweitzman: Do we need this now? I think it's for query buffering which isn't implemented yet.

go/services/multigateway/poolergateway/pooler_connection.go
```diff
@@ -14,63 +14,164 @@
...
+	healthMu sync.Mutex
+
+	// health contains the current health state from the health stream.
+	// Updated atomically as a unit when new health responses arrive.
+	health *PoolerHealth
+
+	// timedOut indicates if the health stream has timed out.
+	// Accessed atomically because there's a race between the timeout
+	// goroutine and the stream processing. (Vitess pattern)
+	timedOut atomic.Bool
```
dweitzman: Maybe this should be called healthTimedOut for clarity?

go/services/multigateway/init.go
```diff
@@ -208,13 +212,17 @@ func (mg *MultiGateway) Init() error {
 	mg.serverStatus.LocalCell = mg.cell.Get()
 	mg.serverStatus.ServiceID = mg.serviceID.Get()
 
+	// Create a service-lifetime context cancelled on shutdown.
+	//nolint:gocritic // service entry point: this is the top-level context for the multigateway process
+	mg.shutdownCtx, mg.shutdownCancel = context.WithCancel(context.Background())
```
dweitzman: Instead of using context.Background(), `Init()` could take a context as a parameter that receives the cobra command's `.Context()` from go/cmd/multipooler

proto/multipoolerservice.proto
haritabh17: Good call — renamed to `primary_term` in the proto and all Go references. Updated the comment to clarify it tracks primary elections, not coordinator terms.

go/services/multigateway/poolergateway/load_balancer.go
haritabh17: Agreed, removed `GetConnectionOptions`, `ExcludePoolers`, and all the exclude logic from `GetConnection` and `selectPrimaryByTerm`. Will add back when we implement query retries.

go/services/multigateway/poolergateway/load_balancer.go
haritabh17: Done, switched to `sync.Mutex`.

go/services/multigateway/poolergateway/load_balancer.go
haritabh17: Removed. Will add back when we implement query buffering.

go/services/multigateway/poolergateway/pooler_connection.go
haritabh17: Renamed to `healthTimedOut`.

go/services/multigateway/init.go
haritabh17: Done — `Init()` now takes a `ctx` parameter, and `cmd.Context()` is passed from the cobra command. Also updated the `logger.Info` calls in `Init()` to use `logger.InfoContext(ctx, ...)` to satisfy sloglint.

proto/multipoolerservice.proto
```diff
@@ -383,4 +399,43 @@ message ReleaseReservedConnectionRequest {
...
+
+// PrimaryObservation represents a pooler's view of who the primary is.
+message PrimaryObservation {
+  // primary_id is the ID of the pooler this node believes is the primary.
+  // May be this pooler's own ID if it believes itself to be primary.
+  clustermetadata.ID primary_id = 1;
+
+  // primary_term is the primary term at which this observation was made.
+  // The primary never changes within a primary_term. Higher values indicate
+  // more recent primary elections.
```
dweitzman: Super nitpick: "primary election" -> "primary appointment"

haritabh17: Done.

go/services/multipooler/manager/health_provider_test.go
```diff
@@ -0,0 +1,313 @@
...
+		assert.Equal(t, clustermetadatapb.PoolerServingStatus_SERVING, received.ServingStatus)
+	case <-time.After(100 * time.Millisecond):
+		t.Fatal("ch1 did not receive broadcast")
+	}
+
+	select {
+	case received := <-ch2:
+		assert.Equal(t, "tg1", received.Target.TableGroup)
+		assert.Equal(t, clustermetadatapb.PoolerServingStatus_SERVING, received.ServingStatus)
+	case <-time.After(100 * time.Millisecond):
```
dweitzman: Nit: feels like this time.After() should ideally be set up at the same time as the previous one instead of doing them both in sequence? Not super important, though

go/services/multipooler/manager/health_provider_test.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// Channel should be closed due to buffer overflow
+	assert.Equal(t, 0, hs.clientCount(), "client should be removed after buffer overflow")
+
+	// Channel should be closed (reading should return zero value immediately)
+	select {
+	case _, ok := <-ch:
+		if ok {
+			// Drain any remaining buffered items
+			for range ch {
```
dweitzman, line 118: Maybe for safety this should check if t.Context() is cancelled and if so stop iterating? Just to avoid an infinite loop here if for whatever reason the test was failing

dweitzman (commented):
The non-test code LGTM. Haven't looked at all the test code, but I'll defer to @GuptaManan100 to approve from a query serving perspective

go/services/multigateway/poolergateway/pooler_connection_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 Supabase, Inc.
+// Copyright 2025 Supabase, Inc.
```
GuptaManan100, line 1: What? 😆

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -50,37 +50,35 @@ type LoadBalancer struct {
...
 		return nil
 	}
 
-	// Create new connection
-	conn, err := NewPoolerConnection(pooler, lb.logger)
+	// Create new connection.
+	// Pass nil for onHealthUpdate - the LoadBalancer currently uses Health()
+	// to check serving state on demand rather than maintaining a separate list.
+	// TODO: Consider adding a callback to proactively update routing when health changes.
+	conn, err := NewPoolerConnection(lb.ctx, pooler, lb.logger, nil)
```
GuptaManan100: This is the only place for NewPoolerConnection is called in production code, and it's passing a nil value for onHealthUpdate. I'm thinking would it make sense for the load balancer to instead own this. Like we have a method to register and deregister health updates. It would be rare for the function to be different for two different poolers. That way Load balancer can take over creating/removing them for pooler connections. WDYT?

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -139,7 +128,37 @@ func (lb *LoadBalancer) GetConnection(target *query.Target) (*PoolerConnection,
 	lb.mu.Lock()
 	defer lb.mu.Unlock()
 
-	// Collect candidates matching the target
+	targetType := target.PoolerType
+	if targetType == clustermetadatapb.PoolerType_UNKNOWN {
+		targetType = clustermetadatapb.PoolerType_PRIMARY
+	}
```
GuptaManan100: nit: We needed this before, I don't think we have any UNKNOWN routing's left. If there are, we should fix them to add PRIMARY instead.

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -139,7 +128,37 @@ func (lb *LoadBalancer) GetConnection(target *query.Target) (*PoolerConnection,
...
+			}
+		}
+
+		if len(shardPoolers) == 0 {
+			return nil, mterrors.Errorf(mtrpcpb.Code_UNAVAILABLE,
+				"no pooler found for target: tablegroup=%s, shard=%s, type=%s",
+				target.TableGroup, target.Shard, target.PoolerType.String())
+		}
+
+		conn := lb.selectPrimaryByTerm(shardPoolers)
```
GuptaManan100: Doing all of this computation on every iteration of us calling GetConnection seems wasteful, especially since we'll do this on every Query, so this could eat up a lot of CPU. I think it would be great if we did this computation on `onHealthUpdate` for a pooler connection! That way we can cache the primary term and primary for each shard on the load balancer. On every pooler update, we can update the information if we get a higher term! That would only trigger updates on health updates, which would be far rarer, and not consume as much CPU.

go/services/multigateway/poolergateway/pooler_connection.go
```diff
@@ -14,63 +14,164 @@
...
+const (
+	// defaultHealthRetryDelay is the initial delay before retrying a failed health stream.
+	// Matches Vitess DefaultHealthCheckRetryDelay.
+	defaultHealthRetryDelay = 5 * time.Second
+
+	// defaultHealthCheckTimeout is the timeout for detecting a stale health stream.
+	// If no message is received within this duration, the connection is marked unhealthy.
+	// Matches Vitess DefaultHealthCheckTimeout.
+	defaultHealthCheckTimeout = 1 * time.Minute
+)
```
GuptaManan100: Let's move them to `constants` package, and drop the Vitess references, our code should be self-explanatory.

go/services/multigateway/poolergateway/pooler_connection_test.go
haritabh17: 😬 updated.

go/services/multigateway/poolergateway/load_balancer.go
haritabh17: Ack. Removed.

go/services/multigateway/poolergateway/pooler_connection.go
haritabh17: Ack, done

go/services/multipooler/manager/health_provider_test.go
haritabh17: Ack, updating.

go/services/multipooler/manager/health_provider_test.go
haritabh17: Ack, done

go/services/multigateway/poolergateway/load_balancer.go
haritabh17: Done — AddPooler now wires lb.onPoolerHealthUpdate as the callback. LoadBalancer is the only creator of PoolerConnections, so it owns the callback directly.

go/services/multigateway/poolergateway/load_balancer.go
haritabh17: Done — Added cachedPrimaries map[shardKey]*cachedPrimary to LoadBalancer. The cache is updated by onPoolerHealthUpdate using term-based reconciliation and invalidated on pooler removal. Falls back to type-based lookup before any health updates arrive.

haritabh17: Done — Added cachedPrimaries map[shardKey]*cachedPrimary to LoadBalancer. The cache is updated by onPoolerHealthUpdate using term-based reconciliation and invalidated on pooler removal. Falls back to type-based lookup before any health updates arrive.

PR #667, opened 2026-02-27
feat(multigateway): implement cross-gateway query cancellation

### LINKED ISSUE #536: RFC: Cross-Gateway Query Cancellation
## Summary

This RFC proposes some solutions for handling connection visibility and query cancellation in a multi-multigateway deployment, addressing the limitations observed in Vitess.

## Background

In PostgreSQL's cancel protocol, when a client connects, the server returns a `BackendKeyData` message containing:
- **PID**: Process ID of the backend
- **Secret**: A random 32-bit cancellation key

When a user sends a cancel request (e.g., Ctrl+C in psql), the client opens a *new* connection and sends a `CancelRequest` with the pid/secret pair. The server uses these to locate and cancel the appropriate query.

### The Multi-MultiGateway Problem

```
                    ┌─────────────┐
                    │ Load        │
                    │ Balancer    │
                    └──────┬──────┘
                           │
          ┌────────────────┼────────────────┐
          │                │                │
          ▼                ▼                ▼
    ┌──────────┐     ┌──────────┐     ┌──────────┐
    │ Gateway  │     │ Gateway  │     │ Gateway  │
    │    A     │     │    B     │     │    C     │
    └──────────┘     └──────────┘     └──────────┘
```

1. Client connects through load balancer → lands on Gateway A → receives `(pid, secret)`
2. Client runs long query
3. Client hits Ctrl+C → psql opens *new* connection with `(pid, secret)`
4. Cancel request goes through load balancer → could land on Gateway B or C
5. Gateway B/C has no knowledge of the pid from Gateway A → **cancellation fails**

### Lessons from Vitess

Vitess exhibits a similar architectural gap:

| Command | Operates On | ID Type |
|---------|-------------|---------|
| `SHOW PROCESSLIST` | MySQL backend | MySQL backend IDs |
| `CONNECTION_ID()` | VTGate | VTGate ID |
| `KILL QUERY <id>` | VTGate | Expects VTGate ID |

**Key limitations in Vitess:**
- No cross-vtgate routing for KILL commands
- No discovery mechanism to find which vtgate owns a connection
- Connection ID collisions across vtgate instances
- No global view of connections

### 1. Connection Visibility

**Current behavior:** Queries like `SELECT * FROM pg_stat_activity` pass through to PostgreSQL, showing backend connections rather than client-to-gateway connections.

**Desired behavior:** Expose gateway-level connection information, potentially aggregated across all gateways.

### 2. Globally Unique PIDs

**Current behavior:** Each gateway assigns PIDs independently, leading to collisions and inability to route cancel requests.

**Desired behavior:** PIDs should encode routing information so any gateway can determine which gateway owns a given connection.

### 3. Cross-Gateway Cancel Routing

**Current behavior:** Cancel requests only work if they happen to land on the same gateway as the original connection.

**Desired behavior:** Any gateway should be able to route a cancel request to the correct gateway.

### PID Encoding Scheme

Encode gateway identity into the PID returned to clients:

```
┌─────────────────────────────────────┐
│            32-bit PID               │
├──────────────┬──────────────────────┤
│ Gateway ID   │ Local Connection ID  │
│  (N bits)    │     (32-N bits)      │
└──────────────┴──────────────────────┘
```

**Static bit allocation**
```
Gateway ID:     10 bits  → supports up to 1024 gateways
Connection ID:  22 bits  → supports ~4M concurrent connections per gateway
```

### Gateway-to-Gateway Communication

When a gateway receives a `CancelRequest`:

```
1. Gateway B receives CancelRequest(pid=0x00500042, secret=0xABCD1234)

2. Extract gateway ID from pid:
   gateway_id = pid >> 22  → 0x005 → Gateway A

3. If gateway_id == self:
   - Look up local connection, verify secret, cancel query

4. If gateway_id != self:
   - Forward cancel request to Gateway A via internal RPC
   
   ┌────────────┐    CancelRPC(pid, secret)    ┌────────────┐
   │ Gateway B  │ ───────────────────────────► │ Gateway A  │
   └────────────┘                              └────────────┘
```

## References

- [PostgreSQL Cancel Request Protocol](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-CANCELING-REQUESTS)

GuptaManan100:
## Summary

- Implements the RFC #536. Closes #536.

- Add cross-gateway query cancellation so that a PostgreSQL `CancelRequest` landing on any gateway is routed to the gateway that owns the connection
- Encode a unique gateway PID prefix into the upper 12 bits of each connection's process ID, with the lower 20 bits for the local connection ID
- Introduce a `MultiGatewayService` gRPC service for gateway-to-gateway cancel forwarding
- Add per-connection cancel context (`BeginQueryCancel`/`EndQueryCancel`) to allow in-flight queries to be interrupted

## Description

PostgreSQL clients cancel queries by opening a new TCP connection and sending a `CancelRequest` with the process ID and secret key they received at startup. In a multi-gateway deployment, this cancel connection may land on a different gateway than the one running the query.

To solve this, each gateway is assigned a unique PID prefix (12 bits) from topology on startup. Connection IDs combine this prefix with a local counter (20 bits) via the new `cancelkey` package, so any gateway can decode the prefix and route the cancel to the right peer.

The `CancelManager` handles the routing: if the prefix matches the local gateway it cancels directly via the `Listener`'s connection map; otherwise it looks up the owning gateway in topology and forwards the request over the new `MultiGatewayService.CancelQuery` gRPC endpoint. The `Listener` and `Conn` types gain cancel registration, secret-key verification, and context cancellation support so that an in-flight `HandleQuery` or `HandleExecute` is interrupted with a standard `57014 query_canceled` error.


go/services/multigateway/cancel.go
```diff
@@ -0,0 +1,243 @@
...
+			"process_id", processID,
+			"error", err,
+		)
+	}
+}
+
+// CancelQuery implements MultiGatewayServiceServer.
+// This is called by other gateways forwarding cancel requests via gRPC.
+func (cm *CancelManager) CancelQuery(ctx context.Context, req *multigatewayservicepb.CancelQueryRequest) (*multigatewayservicepb.CancelQueryResponse, error) {
+	cm.localCancelFn(req.ProcessId, req.SecretKey)
```
haritabh17, line 139: Called without verifying that the decoded PID prefix matches `cm.ownPrefix`. The map lookup in `CancelLocalConnection` implicitly handles this (wrong-prefix PIDs won't match any registered connection), so this isn't a correctness bug. However, if a stale prefix cache on the sender routes a cancel here by mistake, the lookup silently fails (`returns false`) with no log, making misdirected cancels invisible.

A short guard would make these visible for debugging:

```go
prefix, _ := pid.DecodePID(req.ProcessId)
if prefix != cm.ownPrefix {
    cm.logger.WarnContext(ctx, "received cancel for wrong prefix", "expected", cm.ownPrefix, "got", prefix)
}
```

go/common/pgprotocol/pid/pid.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Package pid provides encoding and decoding of PostgreSQL process IDs (PIDs)
+// for cross-gateway query cancellation. The 32-bit PID is split into a gateway prefix
+// and a local connection ID, allowing any gateway to route cancel requests to the
+// correct origin gateway.
+package pid
+
+const (
+	// PrefixBits is the number of bits used for the gateway PID prefix.
+	PrefixBits = 12
```
haritabh17: With `PrefixBits = 12`, the prefix occupies bits 20–31 of the PID. Prefixes ≥ 2048 set bit 31, making the PID appear negative when interpreted as PostgreSQL's signed `Int32`.

The cancel protocol itself is unaffected (raw 4 bytes round-trip correctly), but:

- If multigres ever exposes `pg_stat_activity` or similar views, monitoring tools (pgAdmin, Datadog, etc.) will see negative PIDs and likely break or confuse operators
- Client libraries that log the backend PID (`pgx`, `lib/pq`) will display negative values
- PgBouncer logs backend PIDs as signed int

Two options: (a) limit prefix to 11 bits (max 2047, still plenty) to keep bit 31 clear, or (b) document the decision and accept negative display PIDs. I'd lean toward (a) since it costs almost nothing.

go/common/pgprotocol/server/listener.go
```diff
@@ -248,6 +278,78 @@ func (l *Listener) Close() error {
 	return err
 }
 
+// assignConnectionID returns the next unused PID for a new connection.
+// It encodes the gateway prefix into the upper bits and skips PIDs that
+// are already in use or have a zero local ID (PID 0 is reserved in PostgreSQL).
+func (l *Listener) assignConnectionID() (uint32, bool) {
```
haritabh17, line 284: Acquires/releases `connsMu` on every iteration. In the common case (first attempt succeeds) this is one lock/unlock. But near `MaxLocalConnID` under high churn, this spins with repeated lock/unlock.

Consider holding the lock for the entire scan since the critical section is just a map lookup. `nextLocalID`'s CAS could also simplify to a plain increment under the lock?

go/services/multigateway/cancel.go
GuptaManan100: Fixed!

go/common/pgprotocol/pid/pid.go
GuptaManan100: Fixed

go/common/pgprotocol/server/listener.go
GuptaManan100: Fixed

PR #677, opened 2026-03-01
refactor(test): introduce semantic DSN helpers and add env var overrides

mkindahl:
Introduce semantic DSN helpers: GetTestUserDSN for test-client connections via multigateway, GetPostgresDSN for pgctld-level direct PostgreSQL connections. Simplify notice test helpers.

Add environment variables `POSTGRES_USER` and `POSTGRES_PASSWORD`. RunTestMain reads `POSTGRES_PASSWORD` (falling back to TestPostgresPassword) and sets `PGPASSWORD` for postgres binary invocations. `POSTGRES_USER` is set if not already present so `pgctld` subprocesses inherit it.

The `POSTGRES_USER` environment variable is wired to the `pgctld` flag `--pg-user` and the CLI flag takes precedence when both the flag and the environment variable is set.

### DIFF
--- a/go/test/endtoend/README.md
+++ b/go/test/endtoend/README.md
@@ -1,0 +1,197 @@
+# End-to-End Tests
+
+Integration tests that start real multigres components (multigateway, multipooler, pgctld,
+multiorch) and PostgreSQL, then exercise them end-to-end.
+
+## Prerequisites
+
+### 1. Build the binaries
+
+```bash
+make build
+```
+
+This must be run from the repository root before running integration tests. It compiles all service
+binaries into `bin/`.
+
+### 2. PostgreSQL binaries on PATH
+
+The tests look for `initdb`, `postgres`, `pg_ctl`, and `pg_isready`
+via `exec.LookPath`. If any are missing the tests skip the
+real-PostgreSQL paths and run in a degraded mode.
+
+On macOS with Homebrew, PostgreSQL is typically **not** on the default
+PATH. Prepend it explicitly:
+
+```bash
+# PostgreSQL 18 via Homebrew (adjust version as needed)
+export PATH="/opt/homebrew/Cellar/postgresql@18/18.2/bin:$PATH"
+
+# Verify
+which initdb   # should print the Homebrew path
+which postgres
+```
+
+On Linux, installing `postgresql` via the system package manager (apt,
+yum, etc.) usually puts the binaries on PATH automatically.
+
+## Running the Tests
+
+### Run all packages
+
+```bash
+go test -skip TestPostgreSQLRegression ./go/test/endtoend/...
+```
+
+If you run into issues with failing tests because of allocating shared
+memory segments, you can add `-p=1` to the test. Go runs test packages
+in parallel by default. Each package spins up its own PostgreSQL cluster,
+which allocates a shared memory segment via `shmget`. macOS has a hard
+system limit of 32 shared memory IDs (`kern.sysv.shmmni: 32`). Running
+all packages concurrently exhausts this pool, causing PostgreSQL's `initdb`
+to fail with the misleading error:
+
+```text
+FATAL: could not create shared memory segment: No space left on device
+```
+
+The `-p=1` flag forces packages to run sequentially, keeping at most
+one cluster bootstrapping at a time. This is not needed on Linux,
+which has a much higher default limit.
+
+**Why `-skip TestPostgreSQLRegression`?** The `pgregresstest` package
+clones PostgreSQL source and builds it from scratch (~10+
+minutes). Skip it during normal development; see
+[`pgregresstest/README.md`](pgregresstest/README.md) for how to run it
+explicitly.
+
+### Run a specific package
+
+```bash
+go test ./go/test/endtoend/multipooler/...
+```
+
+Single packages don't need `-p=1` (only one cluster at a time).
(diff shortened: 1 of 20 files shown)


go/common/constants/postgres.go
```diff
@@ -22,6 +22,16 @@ const (
...
 
+	// DefaultMultigresUser is the PostgreSQL user that pgctld and multipooler use
+	// when connecting to PostgreSQL for internal operations and query serving.
+	DefaultMultigresUser = "postgres"
+
+	// DefaultTestUser is the PostgreSQL user that tests use when connecting to
+	// multigateway or multipooler as a regular client.
+	// Currently identical to DefaultMultigresUser but intended to diverge in the
+	// future when a dedicated test user is introduced.
+	DefaultTestUser = "postgres"
```
dweitzman: Wondering if this constant should either be in a *_test.go file or if tests should just use DefaultPostgresUser

My gut reaction to seeing `DefaultTestUser` in a non-test file is that it feels like polluting production files with test logic

go/test/endtoend/shardsetup/cluster.go
```diff
@@ -86,6 +88,31 @@ type ShardSetup struct {
 	BaselineGucs map[string]map[string]string
 }
 
+// GetTestUserDSN returns a DSN for connecting to multigateway as a regular
+// test client. Uses DefaultTestUser and TestPostgresPassword.
+func (s *ShardSetup) GetTestUserDSN(host string, port int, args ...string) string {
```
dweitzman: Just thinking aloud, this isn't necessary or a blocker: I wonder if we'll eventually want a reusable common library for assembling a DNS that can handle quoting if needed.

go/test/endtoend/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Running the Tests
+
+### Run all packages
+
+```bash
+go test -p=1 -skip TestPostgreSQLRegression -timeout 45m ./go/test/endtoend/...
+```
+
+**Why `-p=1`?** Go runs test packages in parallel by default. Each
```
dweitzman: Have you been hitting this limit in practice?

I haven't run into limits caused by test concurrency. Usually when I hit the limit it's because of some bug preventing orphan test processes from cleaning up postgres instances if the tests were killed without gracefully terminiating. `run_command_if_parent_dies.sh` is supposed to detect and if that happens and clean up the dead postgres instance, or if a test deletes the temp directory it used postgres will probably crash itself after about 10 seconds or whenever it realizes that it can't open files anymore.

go/test/endtoend/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+```bash
+TEST_PRINT_LOGS=1 go test -v -run TestFailover -timeout 10m ./go/test/endtoend/multiorch/...
+```
+
+## Test Packages
+
+| Package            | What it tests                                      | Typical runtime |
+| ------------------ | -------------------------------------------------- | --------------- |
+| `endtoend`         | etcd topology, service discovery                   | ~7s             |
```
dweitzman: Seems like these documented estimated runtimes could become stale/inaccurate quickly

go/test/endtoend/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+**Symptom**: `[build failed]` in the output before any test runs.
+
+**Fix**: Run `make build` manually to see the full compiler error.
+
+### Test timeout
+
+**Symptom**: `panic: test timed out after 10m0s`
+
+**Fix**: Increase the timeout. The full suite needs at least 45 minutes:
```
dweitzman: This seems inaccurate. The tests take under 5 minutes. If we're getting timeouts it could be due to slower tests, but it's probably just as possible that it's because of flaky or inefficient tests or failing/stuck tests that never realize they're stuck because they're waiting for a condition.

go/common/constants/postgres.go
mkindahl: I agree with that and have the same thoughts, but didn't see a pre-existing "constants" package for the tests. I added the test user to cluster.go where it is used. Seems like the simplest solution since this is the main usage. I kept the variable exported since it it is used elsewhere.

go/test/endtoend/README.md
mkindahl: Yes. My tests fail consistently when I run them on my machine and this fixed it. I can elaborate and say something along the lines of "if you run into problems, you can use -p=1". Sounds good?

go/test/endtoend/shardsetup/cluster.go
mkindahl: Might be useful if we get a more diverse set of tests, but I do not see a need for this right now.

go/test/endtoend/README.md
mkindahl: Likely. Removed them.

go/test/endtoend/README.md
mkindahl: I ran into this and thought it might be useful to document. Note that this is when I run the tests locally on my machine, not in CI. Do you want me to remove it?

mkindahl: Removed all timeouts.

go/test/endtoend/README.md
mkindahl: Reworded.

PR #681, opened 2026-03-03
feat(multipooler): add pgBackRest server health metrics

Verolop:
### Description

When the pgBackRest TLS server crashes and fails to recover, backups and restores silently break with no signal. This PR adds two OTel gauges, `pgbackrest_server_up` and `pgbackrest_restart_count`, so we can detect both outright failures and crash-loop conditions.

---

### Changes

The new `metrics.go` file defines a `Metrics` struct with two `Int64ObservableGauge` instruments backed by atomics (because the OTel callback and the monitor loop run on different goroutines).

In `manager.go`, we initialise `Metrics` during `NewMultiPoolerManagerWithTimeout` and call `UpdateFromPgBackRestStatus` after every status poll in `discoverPostgresState`. The error and nil-status paths also update the gauges to 0, so we never report stale data when pgctld is unreachable.

I made `UpdateFromPgBackRestStatus` nil-receiver safe so existing tests that construct a bare `MultiPoolerManager` without metrics keep working without changes at every call site.

---

### Notes for reviewers

`NewMetrics()` takes no parameters and `otel.Meter()` only lives in metrics.go. The ruleguard rules enforce both so that metrics stay self-contained and all instrument registrations for a package are easy to find in one file.

### DIFF
--- a/go/services/multipooler/manager/rpc_backup.go
+++ b/go/services/multipooler/manager/rpc_backup.go
@@ -64,16 +64,7 @@ func (pm *MultiPoolerManager) Backup(ctx context.Context, forcePrimary bool, bac
 	}
 	defer pm.actionLock.Release(ctx)
 
-	pm.metrics.IncBackupAttempts(ctx)
-
-	result, err := pm.backupLocked(ctx, forcePrimary, backupType, jobID, overrides)
-	if err != nil {
-		pm.metrics.IncBackupFailures(ctx)
-		return "", err
-	}
-
-	pm.metrics.IncBackupSuccesses(ctx)
-	return result, nil
+	return pm.backupLocked(ctx, forcePrimary, backupType, jobID, overrides)
 }
 
 // backupLocked performs a backup. Caller must hold the action lock.
@@ -81,6 +72,15 @@ func (pm *MultiPoolerManager) backupLocked(ctx context.Context, forcePrimary boo
 	if err := AssertActionLockHeld(ctx); err != nil {
 		return "", err
 	}
+
+	pm.metrics.IncBackupAttempts(ctx)
+	defer func() {
+		if retErr == nil {
+			pm.metrics.IncBackupSuccesses(ctx)
+		} else {
+			pm.metrics.IncBackupFailures(ctx)
+		}
+	}()
 
 	pm.logger.InfoContext(ctx, "Starting backup operation", "backup_type", pm.backupConfig.Type(), "job_type", backupType)
 


go/services/multipooler/manager/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+
+	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
+)
+
+const meterName = "github.com/multigres/multigres/go/services/multipooler/manager"
+
+// Metrics holds OTel gauge instruments for pgBackRest server health.
+// Atomic values are used because the OTel callback and the monitor loop run on different goroutines.
+type Metrics struct {
+	serverUp     metric.Int64ObservableGauge
```
cuongdo: This doesn't feel like an easy fit for an OTel gauge. For me, the biggest mismatch is if the service updating this gauge goes down, the gauge becomes stale. I think up/down makes more sense as an HTTPS endpoint.

go/services/multipooler/manager/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	restartCount metric.Int64ObservableGauge
```
cuongdo: These metrics would be a better fit in pgctld than multipooler because: 

1. pgctld manages pgBackRest directly and calls `setPgBackRestStatus` on every state change. Multipooler learns about it via gRPC polling, so the metrics have inherent lag and reset to 0 when pgctld is unreachable, which is misleading for restartCount.
2. These metrics describe pgBackRest server's health, which is entirely pgctld's domain.

Also, the OTel infra (`otel.Meter`) already works in pgctld after telemetry.Init runs.

go/services/multipooler/manager/metrics.go
cuongdo: Server uptime seems like a better fit for a gauge. If it flatlines, at least we know that it's stale.

go/services/multipooler/manager/metrics.go
cuongdo: Metrics that multipooler could easily update are:

- Backup attempts
- Backup successes & failures

go/services/multipooler/manager/rpc_backup.go
```diff
@@ -64,7 +64,16 @@ func (pm *MultiPoolerManager) Backup(ctx context.Context, forcePrimary bool, bac
 	}
 	defer pm.actionLock.Release(ctx)
 
-	return pm.backupLocked(ctx, forcePrimary, backupType, jobID, overrides)
+	pm.metrics.IncBackupAttempts(ctx)
+
+	result, err := pm.backupLocked(ctx, forcePrimary, backupType, jobID, overrides)
```
cuongdo: This works in all cases, except when `backupLocked` is called directly. This only happens here:

https://github.com/multigres/multigres/blob/5f20798efe8f98161b02ce6ba03fdd1f51cdf180/go/services/multipooler/manager/rpc_initialization.go#L159

I suggest moving these metric increments to `backupLocked`. You can use the [defer pattern](https://github.com/multigres/multigres/blob/main/go/services/multipooler/manager/rpc_backup.go#L109) in that function to avoid having to increment the failure counter on every error return.

go/services/multipooler/manager/metrics.go
Verolop: Discussed this private, documenting for context:

Kept server_up as an OTel gauge. Now that pgctld emits it directly, if pgctld dies the SDK stops reporting and the metric shows as absent() in Prometheus, so we won't have any staleness ambiguity. An HTTP endpoint would have the same "down vs. lost visibility" problem at the polling layer, plus need its own alerting pipeline.

Kept server_up as binary 0/1 because replacing it with uptime would complicate alert rules (rate()/deriv() vs == 0). Added pgbackrest_server_uptime_seconds as a sister metric for staleness detection. We can drop it if we determine it produces more noise vs. signal.

go/services/multipooler/manager/metrics.go
Verolop: Moved server_up and restart_count from multipooler into pgctld, wired directly into setPgBackRestStatus

PR #682, opened 2026-03-03
chore(github): add CODEOWNERS file with per-service ownership

mkindahl:
A proposal for using CODEOWNERS so that we get reviewers automatically assigned.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,46 @@
+# CODEOWNERS
+#
+# GitHub uses the last matching pattern for a file, so the catch-all rule
+# must come first and specific overrides must follow.
+#
+# Teams must have at least Read access to this repository.
+
+# Global catch-all — owns everything not matched by a more specific rule below
+*  @multigres/Maintainers
+
+# Gateway service
+go/services/multigateway/  @multigres/owners-query-serving
+
+# Orchestration service
+go/services/multiorch/  @multigres/owners-cluster-management
+
+# PostgreSQL lifecycle manager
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,43 @@
...
+# Teams must have at least Read access to this repository.
+
+# Global catch-all — owns everything not matched by a more specific rule below
+*  @multigres/Maintainers
+
+# Gateway service
+go/services/multigateway/  @multigres/owners-query-serving
+
+# Orchestration service
+go/services/multiorch/  @multigres/owners-ha
```
rafael: Thinking that cluster-management can own ha for now. So we keep it a bit more flat. Not sure if long term, having a specific team for HA makes sense.

.github/CODEOWNERS
```diff
@@ -0,0 +1,43 @@
...
+go/services/multiadmin/  @multigres/Maintainers
+
+# CI/CD and GitHub configuration
+.github/  @multigres/Maintainers
+
+# Proto definitions (API surface)
+proto/  @multigres/Maintainers
+
+# Documentation
+docs/  @multigres/Maintainers
```
rafael, line 46: we could add topoclient to cluster management as well

.github/CODEOWNERS
mkindahl: Done

.github/CODEOWNERS
mkindahl: done

PR #683, opened 2026-03-03
feat(pgctld): add POSTGRES_USER and POSTGRES_PASSWORD env var support

mkindahl:
Add `POSTGRES_USER` and `POSTGRES_PASSWORD` constants following the supabase/postgres convention. Wire `POSTGRES_USER` to the `--pg-user` flag and `POSTGRES_PASSWORD` to password handling in `pgctld` so both can be configured via environment without CLI flags (flags take precedence when both are set).

Replace the `resolvePassword()` mechanism (which checked `pgpassword.txt` and `PGPASSWORD`) with a single source: the `POSTGRES_PASSWORD` environment variable, wired through `PgCtlCommand.pgPassword` using viperutil. No CLI flag is exposed for the password.

Part of MUL-68

### DIFF
--- a/go/test/endtoend/shardsetup/testmain.go
+++ b/go/test/endtoend/shardsetup/testmain.go
@@ -107,7 +107,6 @@ import (
 
 const (
 	// TestPostgresPassword is the password used for the postgres user in tests.
-	// This is set via PGPASSWORD env var before pgctld initializes PostgreSQL.
 	TestPostgresPassword = "test_password_123"
 )
 
@@ -161,9 +160,6 @@ func RunTestMain(m *testing.M) int {
 
 	// Set orphan detection environment variable as baseline protection
 	os.Setenv("MULTIGRES_TEST_PARENT_PID", strconv.Itoa(os.Getpid()))
-
-	// Set PGPASSWORD to a known value so tests can authenticate
-	os.Setenv("PGPASSWORD", TestPostgresPassword)
 
 	// Initialize telemetry (no-op if OTEL environment variables aren't set)
 	tel := telemetry.NewTelemetry()
--- a/go/cmd/pgctld/command/init.go
+++ b/go/cmd/pgctld/command/init.go
@@ -19,7 +19,6 @@ import (
 	"log/slog"
 	"os"
 	"os/exec"
-	"path/filepath"
 
 	"github.com/multigres/multigres/go/services/pgctld"
 
@@ -58,8 +57,7 @@ start the PostgreSQL server. Configuration can be provided via config file,
 environment variables, or CLI flags. CLI flags take precedence over config
 file and environment variable settings.
 
-Password can be set via PGPASSWORD environment variable or by placing a
-password file at <pooler-dir>/pgpassword.txt.
+Password can be set via the POSTGRES_PASSWORD environment variable.
 
 Examples:
   # Initialize data directory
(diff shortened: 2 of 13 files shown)


go/test/endtoend/shardsetup/testmain.go
```diff
@@ -106,8 +106,8 @@ import (
 )
 
 const (
-	// TestPostgresPassword is the password used for the postgres user in tests.
-	// This is set via PGPASSWORD env var before pgctld initializes PostgreSQL.
+	// TestPostgresPassword is the default password used for the postgres user in tests.
+	// Can be overridden by setting the POSTGRES_PASSWORD environment variable.
```
dweitzman: Seems like PGPASSWORD is used for client connections (supported by tools like `psql`) and POSTGRES_PASSWORD is more intended for calling things like `initdb`?

Wouldn't this be a user-facing password for the former case of client connections (where PGPASSWORD is more traditional) rather than the initdb case?

mkindahl: Not entirely sure if I understand your question, but let me elaborate on the situation.

[`PGPASSWORD` is used when calling tools that use `libpq`](https://www.postgresql.org/docs/18/libpq-envars.html) (for example `psql`) and is the means to communicate the password when connecting to a Postgres server. So PGPASSWORD is in our usage set by the tool (e.g., `pgctld`) to pass password information to the binary that needs it (e.g., `psql`).

`initdb` does not use PGPASSWORD since it does not really use `libpq`. The user provided here is the [bootstrap superuser](https://www.postgresql.org/docs/18/glossary.html#GLOSSARY-BOOTSTRAP-SUPERUSER) that is the owner of the database cluster.

The user configured in `POSTGRES_USER` is the bootstrap superuser as well as the user we use to connect and work with the server. This is information for `pgctld` on what user and password to use both when bootstrapping the server and when performing administrative actions on the server. In this sense, it is input information for `pgctld`.

As a result I think the comment "This is set via PGPASSWORD env var before pgctld initializes PostgreSQL" is incorrect since `initdb` does not use PGPASSWORD. However, pgctld *do* read the PGPASSWORD variable to be able to run initdb, so it is true that the password in PGPASSWORD is used for this, but it is used by pgctld.

I think we should use the following two rules:
- POSTGRES_USER and POSTGRES_PASSWORD is used as input on what pgctld should use. This is the same as what other containers are using.
- PGPASSWORD (and the other libpq variables) should be used by pgctld when executing libpq tools and set before calling the tools. (In most cases we do not have to set environment variables since there are options that provide the same information, but PGPASSWORD is an exception here.)

dweitzman: My reason for commenting here is that I think `TestPostgresPassword` will still be used with PGPASSWORD, like in line 167 of this file, for example

I wasn't sure about the intent of removing the reference to `PGPASSWORD` in this comment while the code still sets it sometimes in `PGPASSWORD`. It seems like we might want to expand the command to mention both `PGPASSWORD` and `POSTGRES_PASSWORD` if this constant is used with both?

mkindahl: > My reason for commenting here is that I think `TestPostgresPassword` will still be used with PGPASSWORD, like in line 167 of this file, for example

Ah, good point.

> I wasn't sure about the intent of removing the reference to `PGPASSWORD` in this comment while the code still sets it sometimes in `PGPASSWORD`. It seems like we might want to expand the command to mention both `PGPASSWORD` and `POSTGRES_PASSWORD` if this constant is used with both?

Intent is that we should only set this for commands that use libpq and make sure that it is available to all pgctld commands.

I'll look into it. BBIAB.

mkindahl: Code updated so that the only source of password from the outside is the `POSTGRES_USER` and `POSTGRES_PASSWORD`.

Internally, variable `PGPASSWORD` is set for commands that use `libpq` and `initdb` (which does not use `libpq`) still use a password file to communicate the password to the utility (since we do not want to expose the password on the command line).

dweitzman (approved):
I don't know if I fully appreciate the motivations for this, but I like replacing literals with pre-declared constants and I'm fine with deferring to your judgement on some of the context around this

PR #684, opened 2026-03-03
feat(multigateway): add OTel instrumentation for query serving path

haritabh17:
The multigateway query path (handler → executor → planner → scatterconn → poolergateway) has zero observability today. Operators cannot see query latency, error rates, row counts, or per-shard behavior. The bottom of the stack (`pgprotocol/client/query.go`) already has `"{op} postgresql"` spans, but the handler and scatterconn layers above it are invisible.

This PR adds OpenTelemetry metrics, distributed tracing, and structured query logging at the handler and scatterconn layers.

## Changes

Error classification helpers in `handler/error_helpers.go` extract SQLSTATE codes from the error chain (unwrapping `*mterrors.PgDiagnostic`), classify error sources as backend/client/routing/internal, and pull operation names from AST statements. These are pure functions shared by both handler and scatterconn instrumentation.

Handler-level metrics in `handler/metrics.go` follow the same wrapper-type pattern as `recovery/metrics.go` — typed `Record()` methods with noop fallback on init failure. Three instruments are defined: `mg.gateway.query.duration` (histogram, seconds), `mg.gateway.query.errors` (counter), and `mg.gateway.query.rows_returned` (histogram). Attributes include `db.namespace`, `db.operation.name`, `db.query.protocol`, `error.type`, and status.

Span creation in `handler/tracing.go` uses `telemetry.Tracer()` to start `gateway.query` spans with `SpanKindServer` and OTel semantic conventions (`semconv.DBSystemPostgreSQL`, `semconv.DBOperationName`, `semconv.DBNamespace`). `db.query.text` is intentionally excluded for security, matching the existing `pgprotocol/client` pattern.

Structured query logging in `handler/querylog.go` uses `slog.LogAttrs()` for minimal allocation on the latency-sensitive path. Normal queries log at INFO, errors and slow queries (threshold configurable, default 1s) log at WARN. The slog-OTel bridge (already configured in `telemetry.go`) auto-injects trace_id/span_id.

`HandleQuery` and `HandleExecute` in `handler/handler.go` are instrumented with timing, span creation, row counting (via a `countingCallback` wrapper), and a shared `recordQueryCompletion()` helper that records all three metrics and emits the query log.

ScatterConn-level metrics in `scatterconn/metrics.go` define `mg.scatter.execute.duration` and `mg.scatter.execute.errors` with tablegroup/shard attributes. `StreamExecute` in `scatter_conn.go` gets `shard.execute` spans wrapping the three execution paths (reserved connection, new reservation with BEGIN, pooled connection), with error recording via `recordShardError`.

The resulting span hierarchy is `gateway.query` → `shard.execute` → `{op} postgresql`, giving end-to-end visibility from client request to backend query.

All changes are additive — no existing behavior is modified. Unit tests cover error classification, metrics initialization, and query log emission.

### DIFF
--- a/go/common/mterrors/error_helpers.go
+++ b/go/common/mterrors/error_helpers.go
@@ -12,28 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package handler
+package mterrors
 
 import (
 	"errors"
 	"strings"
-
-	"github.com/multigres/multigres/go/common/mterrors"
-	"github.com/multigres/multigres/go/common/parser/ast"
 )
 
-// ExtractSQLSTATE unwraps the error chain looking for a *mterrors.PgDiagnostic
+// ExtractSQLSTATE unwraps the error chain looking for a *PgDiagnostic
 // and returns its SQLSTATE Code. Returns "" for nil errors and "XX000"
 // (internal_error) for non-PgDiagnostic errors.
 func ExtractSQLSTATE(err error) string {
 	if err == nil {
 		return ""
 	}
-	var diag *mterrors.PgDiagnostic
+	var diag *PgDiagnostic
 	if errors.As(err, &diag) {
 		return diag.Code
 	}
-	return mterrors.PgSSInternalError // "XX000"
+	return PgSSInternalError // "XX000"
 }
 
 // ClassifyErrorSource categorises the origin of an error for metric attribution.
@@ -48,11 +45,11 @@ func ClassifyErrorSource(err error) string {
 	}
 
 	// Connection errors (I/O failures, Class 08, shutdown codes) → routing.
-	if mterrors.IsConnectionError(err) {
+	if IsConnectionError(err) {
 		return "routing"
 	}
 
-	var diag *mterrors.PgDiagnostic
+	var diag *PgDiagnostic
 	if errors.As(err, &diag) {
 		// MT-prefixed codes (e.g. "MTD01", "MTE01") → internal.
 		if strings.HasPrefix(diag.Code, "MT") {
@@ -63,13 +60,4 @@ func ClassifyErrorSource(err error) string {
 	}
 
 	return "client"
-}
-
-// ExtractOperationName returns the SQL statement type (e.g. "SELECT", "INSERT")
-// from an AST node. Returns "UNKNOWN" if stmt is nil.
-func ExtractOperationName(stmt ast.Stmt) string {
-	if stmt == nil {
-		return "UNKNOWN"
-	}
-	return stmt.StatementType()
 }
--- a/go/services/multigateway/handler/handler.go
+++ b/go/services/multigateway/handler/handler.go
@@ -21,6 +21,7 @@ import (
 	"log/slog"
 	"time"
 
+	"github.com/multigres/multigres/go/common/constants"
 	"github.com/multigres/multigres/go/common/mterrors"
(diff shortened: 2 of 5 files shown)


go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -735,6 +795,28 @@ func (sc *ScatterConn) ReleaseAllReservedConnections(
 	return errors.Join(errs...)
 }
 
+// endAction records shard-level metrics and span status for both success and
+// error outcomes. Designed to be called via defer with a pointer to the named
+// error return, following Vitess's startAction/endAction pattern.
```
GuptaManan100: Let's not have Vitess references in our code's documentation. It should be self-explanatory.

go/services/multigateway/handler/handler.go
```diff
@@ -52,22 +52,33 @@ type Executor interface {
 	ReleaseAll(ctx context.Context, conn *server.Conn, state *MultiGatewayConnectionState) error
 }
 
+// defaultSlowQueryThreshold is the duration after which a query is logged at WARN level.
+const defaultSlowQueryThreshold = 1 * time.Second
+
```
GuptaManan100: Maybe we should put this in the constants package. It might be something we use in the UI etc, and having it in a common package will make it easier

go/services/multigateway/handler/metrics.go
```diff
@@ -0,0 +1,161 @@
...
+// HandlerMetrics holds all OTel metrics for the multigateway handler.
+// Following the recovery/metrics.go pattern: wrapper types embedding OTel
+// instruments with typed Record() methods that hide attribute key names.
+type HandlerMetrics struct {
+	queryDuration QueryDuration
+	queryErrors   QueryErrors
+	rowsReturned  RowsReturned
+}
+
+// QueryDuration wraps a Float64Histogram for recording query durations.
```
GuptaManan100, line 45: Eventually we'll need to make this per-query based. But that will happen after we have query normalisation.

go/common/mterrors/error_helpers.go
```diff
@@ -0,0 +1,75 @@
...
+	"strings"
+
+	"github.com/multigres/multigres/go/common/mterrors"
+	"github.com/multigres/multigres/go/common/parser/ast"
+)
+
+// ExtractSQLSTATE unwraps the error chain looking for a *mterrors.PgDiagnostic
+// and returns its SQLSTATE Code. Returns "" for nil errors and "XX000"
+// (internal_error) for non-PgDiagnostic errors.
+func ExtractSQLSTATE(err error) string {
```
GuptaManan100, line 25: This entire file I think should live in the mterrors package, one for modularity, and the other because it might be reused.

haritabh17: Done

go/services/multigateway/handler/handler.go
haritabh17: Moved to constants.DefaultSlowQueryThreshold in postgres.go.

go/services/multigateway/handler/metrics.go
haritabh17: Agreed, will address after query normalisation lands.

go/services/multigateway/scatterconn/scatter_conn.go
haritabh17: Done.

PR #687, opened 2026-03-04
feat(multiorch/recovery): replace polling loop with etcd watcher

mkindahl:
Add PoolerWatcher, a two-tier etcd watcher that keeps the pooler health store up-to-date in real time. A global watcher monitors the cells/directory; per-cell watchers monitor each cell's poolers/ directory. Poolers are added to the store on discovery and removed immediately when a deletion event arrives from etcd.

Resolves MUL-49

### DIFF
--- a/go/services/multiorch/recovery/pooler_watcher.go
+++ b/go/services/multiorch/recovery/pooler_watcher.go
@@ -364,20 +364,10 @@ func (cw *cellPoolerWatcher) sync(ctx context.Context) error {
 // handlePoolerEvent processes a single watch event for a pooler file.
 func (cw *cellPoolerWatcher) handlePoolerEvent(wd *topoclient.WatchDataRecursive) {
 	if wd.Err != nil {
-		// A deletion arrives as a NoNode error on the specific path.
-		if errors.Is(wd.Err, &topoclient.TopoError{Code: topoclient.NoNode}) {
-			if strings.HasSuffix(wd.Path, "/"+topoclient.PoolerFile) {
-				poolerID := cw.extractPoolerID(wd.Path)
-				if poolerID == "" {
-					return
-				}
-				if cw.store.Delete(poolerID) {
-					cw.logger.Info("pooler removed from topology, deleted from store", "pooler_id", poolerID)
-				} else {
-					cw.logger.Debug("pooler removed from topology, not in store (filtered or unknown)", "pooler_id", poolerID)
-				}
-			}
-		} else {
+		// Deletions (NoNode) are intentionally ignored here: the pooler store
+		// entry is left in place so that ongoing health checks can continue until
+		// bookkeeping removes it after the configured unseen threshold.
+		if !errors.Is(wd.Err, &topoclient.TopoError{Code: topoclient.NoNode}) {
 			cw.logger.Warn("watch error on pooler path", "error", wd.Err, "path", wd.Path)
 		}
 		return
@@ -438,19 +428,4 @@ func (cw *cellPoolerWatcher) matchesAnyTarget(pooler *clustermetadatapb.MultiPoo
 		}
 	}
 	return false
-}
-
-// extractPoolerID extracts the pooler ID string from a watch path.
-// The path format is: "...poolers/{pooler_name}/Pooler"
-func (cw *cellPoolerWatcher) extractPoolerID(watchPath string) string {
-	_, after, found := strings.Cut(watchPath, topoclient.PoolersPath+"/")
-	if !found {
-		return ""
-	}
-	// Drop the trailing "/Pooler"
-	name := strings.TrimSuffix(after, "/"+topoclient.PoolerFile)
-	if name == after {
-		return "" // suffix was not present
-	}
-	return name
 }
--- a/go/services/multiorch/recovery/pooler_watcher_test.go
+++ b/go/services/multiorch/recovery/pooler_watcher_test.go
@@ -323,7 +323,9 @@ func TestPoolerWatcher_NewCellDiscovered(t *testing.T) {
 }
 
 // TestPoolerWatcher_PoolerDeletedFromTopology verifies that deleting a pooler from topology
-// immediately removes it from the store.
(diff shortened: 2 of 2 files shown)


go/services/multiorch/recovery/pooler_watcher.go
```diff
@@ -0,0 +1,454 @@
...
+// handlePoolerEvent processes a single watch event for a pooler file.
+func (cw *cellPoolerWatcher) handlePoolerEvent(wd *topoclient.WatchDataRecursive) {
+	if wd.Err != nil {
+		// A deletion arrives as a NoNode error on the specific path.
+		if errors.Is(wd.Err, &topoclient.TopoError{Code: topoclient.NoNode}) {
+			if strings.HasSuffix(wd.Path, "/"+topoclient.PoolerFile) {
+				poolerID := cw.extractPoolerID(wd.Path)
+				if poolerID != "" {
+					cw.logger.Debug("pooler deleted from topology", "pooler_id", poolerID)
+					// Do not remove from store here; bookkeeping handles stale entry removal.
```
haritabh17: Makes sense to mirror polling behavior to keep this PR's scope tight. But with watches we now get explicit delete events that polling never had — worth a follow-up to remove from store immediately here? Bookkeeping's 4-hour eviction window (forgetLongUnseenInstances) would still serve as a safety net for edge cases.

go/services/multiorch/recovery/pooler_watcher.go
```diff
@@ -0,0 +1,454 @@
...
+	topoStore topoclient.Store
+	targets   func() []config.WatchTarget // live accessor, same as Engine.shardWatchTargets
+	store     *store.PoolerHealthStore
+	queue     *Queue
+	logger    *slog.Logger
+
+	// Control
+	ctx       context.Context
+	cancel    context.CancelFunc
+	waitGroup sync.WaitGroup
```
dweitzman: I think we usually name this `wg` instead of `waitGroup`

go/services/multiorch/recovery/pooler_watcher.go
dweitzman, line 1: This file seems to have quite a bit of overlap with `go/services/multigateway/discovery.go`

Do you think it might be worth having a shared topo watcher helper somewhere under go/common/topoclient that both gateway and orch can use?

go/services/multiorch/recovery/pooler_watcher.go
dweitzman: For consensus, the deciding if/when to evict a pooler from the cohort feels like a potentially nuaned and interesting question that feels like it should be addressed separately from topology watching

The disappearance of a pooler from the topology seems like a signal, not a command. Finding the right balance between evicting cohort members too eagerly vs keeping them too long seems like something that we may have to tune based on experience.

go/services/multiorch/recovery/engine.go
```diff
@@ -258,7 +260,7 @@ type Engine struct {
 	cancel      context.CancelFunc
 
 	// WaitGroup to track goroutines for graceful shutdown
-	wg sync.WaitGroup
+	waitGroup sync.WaitGroup
```
dweitzman: Similar comment to elsewhere: I think we should be consistent about wg vs waitGroup, and I think `wg` is considered more typical Go naming

go/services/multiorch/recovery/pooler_watcher.go
mkindahl: Replaced `waitGroup` with `wg`

go/services/multiorch/recovery/engine.go
mkindahl: Added a separate commit to fix this. Will squash all into one before merging.

go/services/multiorch/recovery/pooler_watcher.go
mkindahl: I noted the existing watcher and was thinking along the same lines. I'll look into it.

mkindahl: Added it as a separate commit since it can make sense as a refactoring alone. If you want me to create a separate pull request for it, please tell me.

go/services/multiorch/recovery/pooler_watcher.go
mkindahl: > Makes sense to mirror polling behavior to keep this PR's scope tight. But with watches we now get explicit delete events that polling never had — worth a follow-up to remove from store immediately here?

AFAICT, it only happens if you explicitly delete a pooler from the topology, so this should normally not happen except if you for some reason do a manual delete. I think it is safe to delete the entry from the store since it does not seem to trigger any actions when removing it: it is just stale information.

I think the correct action is to log this for debugging purposes and then see if we get this error message in running systems. It indicates a lingering issue.

We might want to have some special tag on log messages with "anomalies" like this so that we can see if they happen. They should normally not happen and if they do, we need to pay attention to it and understand why they happen.

go/services/multiorch/recovery/pooler_watcher.go
dweitzman: I was thinking we could go even further with reusing logic between the two of them, making it typesafe instead of exposing direct etcd paths

Here's an idea that refactors on top of this code to create topoclient/pooler_cache.go which is essentially just a local cache of all pooler data with the ability to subscribe to updates, which both gateway and orch discovery are able to use: https://github.com/multigres/multigres/compare/main...dweitzman:multigres:prototype/typed-watch-helpers?expand=1. The lines of code nets out to roughly no increase or decrease (ignoring test files) but I think it makes both the gate and orch code simpler and maintainable. Thoughts?

go/services/multiorch/recovery/pooler_watcher.go
mkindahl: I added a separate commit that just remove the pooler id if a NoNode comes in.

go/services/multiorch/recovery/pooler_watcher.go
mkindahl: Seems like a useful refactoring. As agreed, created a new branch https://github.com/multigres/multigres/pull/710 for the refactoring effort and remove it from this pull request.

Issue #658, opened 2026-02-24
RFC: Failover Buffering in Multigateway


GuptaManan100:
## Problem

During PRIMARY failovers, queries fail immediately with `UNAVAILABLE` errors. The LoadBalancer is fail-fast: no PRIMARY pooler → error to the application. This causes a burst of visible errors during the brief window (typically 1-10s) between the old PRIMARY going away and the new one appearing in topology.

These errors are avoidable. If multigateway held in-flight requests briefly instead of failing them, applications would see zero downtime during planned failovers.

## Goals

1. **Zero application-visible errors during planned PRIMARY failovers**
2. **Per-shard buffering** — each shard buffers and drains independently
3. **Bounded resource usage** — global limit on total buffered requests
4. **Observable** — metrics for production visibility
5. **Safe defaults** — opt-in, conservative timeouts, no silent data loss

## Non-Goals (v1)

- REPLICA buffering (use another replica)
- In-transaction requests (reserved connections are pinned; the transaction is lost anyway)
- Resharding support
- Active health probing

### Where: PoolerGateway

Buffering wraps the existing `LoadBalancer.GetConnection()` call in `PoolerGateway`.

- Sits between ScatterConn and LoadBalancer — the natural retry point
- Has full target visibility (tablegroup, shard, pooler type)
- Doesn't interfere with reserved connections (`QueryServiceByID` bypasses the buffer)
- Zero overhead on the hot path when not buffering

**Not buffered:** reserved connection queries (transactions, COPY, portals), REPLICA queries.

### Error Classification

Errors fall into three categories. Only PRIMARY traffic is eligible for buffering.

**From LoadBalancer (`GetConnection` fails):**

| Error Code | Response |
|------------|----------|
| `Code_UNAVAILABLE` — no PRIMARY in topology | **Buffer** |
| Anything else | **Fail** |

**From multipooler (query execution fails):**

| Error Code | Response |
|------------|----------|
| `Code_CLUSTER_EVENT` — PRIMARY shutting down / being demoted | **Buffer** |
| `Code_READ_ONLY` — PRIMARY was demoted to replica, topology update pending | **Buffer** |
| `Code_UNAVAILABLE` — pooler starting up, transient issue | **Retry** |
| Connection errors (EOF, ECONNRESET) | **Retry** |
| Anything else (syntax error, constraint violation, etc.) | **Fail** |

[...]

PR #690, opened 2026-03-05
feat(multigateway): failover buffering for zero-downtime planned failovers

GuptaManan100:
Closes https://github.com/multigres/multigres/issues/658

## Summary

- Add failover buffering to multigateway that holds PRIMARY queries during planned failovers and retries them once a new primary is available, achieving zero application-visible errors
- Use streaming health checks (from #665) to detect the new primary and stop buffering — a direct signal from the pooler with no topology propagation delay
- Use MTF01 buffering errors from multipooler's state management (from #755) to trigger buffering during graceful drain and NOT_SERVING transitions

## Design

During a planned failover, the old primary transitions to NOT_SERVING and starts returning MTF01 errors. The gateway's `classifyError` detects MTF01 (or PostgreSQL 25006 for in-flight queries) and buffers the request in a global FIFO queue. Each shard has an independent state machine (IDLE → BUFFERING → DRAINING → IDLE).

When the new primary starts serving, the `LoadBalancer`'s streaming health check callback detects it (via `PrimaryObservation` term reconciliation) and calls `buffer.StopBuffering()`, which drains all buffered entries with configurable concurrency.

See `docs/query_serving/failover_buffering.md` for the full design.

## Key Changes

**Buffer package** (`go/services/multigateway/buffer/`):
- Global FIFO queue with weighted semaphore for capacity management
- Per-shard state machines for independent failover tracking
- Timeout thread for per-request window eviction
- Safety rails: buffer-full eviction (MTB01), window timeout (MTB02), max failover duration, timing guard, shutdown eviction (MTB03)

**Buffering integration** (`poolergateway/pooler_gateway.go`):
- `withBuffering()` wraps query execution with proactive buffering (wait before sending if shard is already failing over) and reactive buffering (buffer on MTF01/25006 errors)

**Health-stream-driven buffer stop** (`poolergateway/load_balancer.go`, `init.go`):
- `LoadBalancer.onPoolerHealthUpdate()` invokes `onPrimaryServing` callback when a new primary is detected via streaming health check
- Direct signal from the pooler — no etcd propagation delay

**Custom buffering errors** (`poolerserver/pooler.go`, `grpcpoolerservice/service.go`):
- `StartRequest()` returns MTF01 for both NOT_SERVING and shutting-down states
- gRPC service wraps `StartRequest()` errors with `mterrors.ToGRPC()` so the MTF01 code survives the gRPC round-trip
- Removed dead SERVING_RDONLY check from `Executor()`

## Test plan

- [x] Unit tests: buffer lifecycle, eviction, timeout, multi-shard independence (`buffer/buffer_test.go`)
- [x] Unit tests: error classification and retry loop (`poolergateway/pooler_gateway_test.go`)
- [x] Unit tests: StartRequest returns MTF01 (`poolerserver/pooler_test.go`)
- [x] Integration test: zero failed writes during planned failover (`queryserving/buffer_test.go`)

### DIFF
--- a/.claude/skills/fix-markdown-lint/SKILL.md
+++ b/.claude/skills/fix-markdown-lint/SKILL.md
@@ -1,0 +1,86 @@
+---
+name: "fix-markdown-lint"
+description: "Fix markdown linting and prettier formatting issues in markdown files"
+disable-model-invocation: true
+---
+
+# Fix Markdown Lint
+
+Fix markdown lint errors and prettier formatting issues so CI passes.
+
+## Usage
+
+```text
+/fix-markdown [file ...]
+```
+
+If no files are specified, auto-detect changed `.md` files:
+
+```bash
+git diff --name-only HEAD upstream/main -- '*.md'
+```
+
+## Instructions
+
+### 1. Identify errors
+
+Run markdownlint with the repo config to see all errors:
+
+```bash
+markdownlint-cli2 --config .github/linters/.markdownlint.json <files>
+```
+
+Read each file that has errors before fixing.
+
+### 2. Fix errors
+
+Apply fixes for each error type:
+
+**MD013 — Line length** (max 120 chars, code blocks and tables exempt):
+
+- Break long lines at sentence boundaries
+- Keep related phrases together when possible
+- Do not break URLs or inline code spans
+
+**MD040 — Fenced code block language**:
+
+- Add a language identifier after the opening triple backticks
+- Use `text` for plain text, CLI output, or pseudo-code diagrams
+- Use the actual language (`bash`, `go`, `json`, `yaml`, `sql`, `markdown`) when the content is code
+
+**MD060 — Table column style**:
+
+- Add spaces around pipes in separator rows: `| --- | --- |` not `|---|---|`
+- Prettier will handle full table alignment afterward
+
+**Other rules**:
+
+- Read the markdownlint rule documentation if you encounter an unfamiliar rule
+- Fix according to the rule's requirements
+
+### 3. Run prettier
+
+After fixing lint errors, run prettier to normalize formatting:
+
+```bash
+npx prettier --write <files>
+```
+
+Prettier will align tables, normalize spacing, and fix other formatting.
+
+### 4. Verify
+
+Re-run markdownlint to confirm all errors are resolved:
+
+```bash
+markdownlint-cli2 --config .github/linters/.markdownlint.json <files>
+```
+
+If errors remain, fix them and repeat until clean.
+
+## Config reference
+
+- Markdownlint config: `.github/linters/.markdownlint.json`
+  - Line length: 120 (code blocks and tables exempt)
+- Prettier: default config (no `.prettierrc`)
+- CI runs both checks via super-linter
--- a/docs/query_serving/prepared_statements_design.md
+++ b/docs/query_serving/prepared_statements_design.md
@@ -105,56 +105,85 @@ Each connection in the MultiPooler's pool must track:
 - Which prepared statements exist on that connection
 - The mapping between logical statement names and physical statement names
 
-## Prepared Statement Consolidator
+## Prepared Statement Consolidation
 
-Both MultiGateway and MultiPooler can use the same consolidation strategy to
-manage prepared statements efficiently.
+The gateway and pooler have different consolidation needs and use separate
+implementations.
 
-### Data Structure
+### Deduplication Key
+
+Both consolidators deduplicate by **(query text, parameter types)** — not
+query text alone. The same SQL with different type hints (e.g.,
+`SELECT $1` with `INT4` vs `TEXT`) produces different PostgreSQL prepared
+statements with different plans and type coercion, so they must be tracked
+separately.
+
+### Gateway Consolidator (`Consolidator`)
+
+The gateway consolidator maps `(connectionID, clientName) → canonical name`.
+It has real per-client connection IDs to namespace by.
 
 ```go
-type PreparedStatementConsolidator struct {
-    // Map from statement body to canonical prepared statement
-    Stmts map[string]*PreparedStatement
+type Consolidator struct {
+    // Map from (query, paramTypes) dedup key to canonical prepared statement
+    Stmts map[string]*PreparedStatementInfo
 
     // Map from connection ID and statement name to prepared statement reference
-    Incoming map[int]map[string]*PreparedStatement
-
-    // Map from prepared statement to its name on the underlying connection
-    Outgoing map[*PreparedStatement]string
+    Incoming map[uint32]map[string]*PreparedStatementInfo
 
     // Reference count: number of connections using each prepared statement
-    UsageCount map[*PreparedStatement]int
+    UsageCount map[*PreparedStatementInfo]int
 }
 ```
 
-### Algorithm
+**Algorithm** — when processing `PREPARE stmt1 AS body1` with `paramTypes`:
 
-When processing a PREPARE request like `PREPARE stmt1 AS body1`:
+1. **Check for existing statement**: Look up `dedupKey(body1, paramTypes)` in
+   `Stmts`
+2. **If exists**: increment usage count, store
+   `Incoming[connectionId]["stmt1"] = existingPS`
+3. **If not exists**: create a new canonical name (e.g., `stmt0`), store in
+   `Stmts`, initialize usage count, store incoming mapping
 
-1. **Check for existing statement**: Look up `body1` in the `Stmts` map
-2. **If exists**:
-   - Increment the usage count for the statement
-   - Store the mapping: `Incoming[connectionId]["stmt1"] = preparedStatement`
-   - Return the existing prepared statement
-3. **If not exists**:
-   - Create a new prepared statement on the underlying connection with a
-     unique name (e.g., `ppstmt1`)
-   - Store it in `Stmts[body1]`
-   - Store the outgoing name mapping:
-     `Outgoing[preparedStatement] = "ppstmt1"`
-   - Initialize usage count to 1
-   - Store the incoming mapping:
-     `Incoming[connectionId]["stmt1"] = preparedStatement`
+**Name translation**: clients use their own names (`stmt1`, `myquery`); the
+consolidator maps these to canonical names (`stmt0`, `stmt1`) shared across
+connections with the same query.
 
-### Name Translation
+### Pooler Consolidator (`PoolerConsolidator`)
 
-Since multiple clients may use different names for the same logical prepared
-statement:
+The pooler consolidator is intentionally simpler. It receives requests from
+multiple stateless gateway replicas, each of which independently assigns
+canonical names starting from `stmt0`. Since different gateways can assign
+the same name to different queries, the pooler **ignores incoming names
+entirely** and deduplicates purely by (query text, parameter types).
 
-- **Client-facing names**: Stored in the `Incoming` map (e.g., `stmt1`,
-  `myquery`)
-- **Backend names**: Stored in the `Outgoing` map (e.g., `ppstmt1`, `ppstmt2`)
+```go
+type PoolerConsolidator struct {
+    // Map from (query, paramTypes) dedup key to canonical name
+    Stmts map[string]string
+}
+```
 
-This allows clients to use their own naming conventions while sharing the
-underlying prepared statement.
+**Algorithm** — `CanonicalName(query, paramTypes) → name`:
+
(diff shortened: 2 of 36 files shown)


go/services/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -158,47 +235,71 @@ func (pg *PoolerGateway) PortalStreamExecute(
...
+func (pg *PoolerGateway) ReserveStreamExecute(
+	ctx context.Context,
+	target *query.Target,
+	sql string,
+	options *query.ExecuteOptions,
+	reservationOptions *multipoolerpb.ReservationOptions,
+	callback func(context.Context, *sqltypes.Result) error,
+) (*query.ReservedState, error) {
+	var state *query.ReservedState
+	err := pg.withBuffering(ctx, target, func(conn *PoolerConnection) error {
```
haritabh17, line 302: ReserveStreamExecute goes through withBuffering, but if the request is for an existing reserved connection (options.ReservedConnectionId > 0), the retry after failover would target a different pooler that doesn't have that connection. Should withBuffering skip buffering for requests on existing reserved connections?

go/common/mterrors/code.go
```diff
@@ -144,6 +144,24 @@ var (
...
+		Description: "The request was evicted because the failover buffer is at capacity. Retry the query.",
+	}
+
+	MTB02 = &MTError{
+		ID: "MTB02", Severity: "ERROR",
+		Format:      "failover buffer timeout",
+		Description: "The request was evicted because the failover did not complete within the buffer window. Retry the query.",
+	}
+
+	MTB03 = &MTError{
```
haritabh17, line 162: MTB01/MTB02/MTB03 use the MT error ID as the SQLSTATE (e.g., "MTB01"). This falls outside the standard PG SQLSTATE classes, which could confuse monitoring tools and client error-handling middleware that parse SQLSTATE to categorize errors. Worth using a standard code like 57P01 (admin_shutdown) or a custom code under an unassigned class prefix?

go/services/multigateway/buffer/buffer.go
```diff
@@ -0,0 +1,313 @@
...
+	sb, ok := b.buffers[key]
+	if !ok {
+		sb = newShardBuffer(b, key)
+		b.buffers[key] = sb
+	}
+	return sb
+}
+
+// enqueue adds a new entry to the global FIFO queue. If the buffer is full,
+// the oldest entry globally is evicted to make room.
```
haritabh17: Global eviction means during multi-shard failover, shard B's new entries can evict shard A's older entries even if shard A is closer to recovering. Fine for single-shard failover (the common case) — worth a comment documenting the tradeoff?

go/services/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -47,26 +50,131 @@ type Gateway interface {
...
+)
+
+// classifyError determines whether an error is eligible for failover buffering.
+// Only PRIMARY traffic with specific error codes is buffered.
+func classifyError(err error, target *query.Target) errorAction {
+	if target.PoolerType != clustermetadatapb.PoolerType_PRIMARY {
+		return actionFail
+	}
+	code := mterrors.Code(err)
+	switch code {
```
haritabh17: classifyError includes UNAVAILABLE and READ_ONLY because execution streams don't check queryServingState before executing — queries reach PG directly during demotion.

Would it make sense to add a serving state check at the top of each query handler (returning CLUSTER_EVENT when not serving)? Not the full state manager from the TODO — just the gate. That way classifyError only needs CLUSTER_EVENT, and we avoid false-triggering buffering on pooler restarts.

go/services/multigateway/buffer/shard_buffer.go
```diff
@@ -0,0 +1,261 @@
...
+	// Fast path: if draining or idle with recent failover, skip.
+	sb.mu.Lock()
+	switch sb.state {
+	case stateDraining:
+		// Already draining — the new PRIMARY is available. Signal the caller
+		// to retry immediately. A recursive retry loop is unlikely because
+		// the LoadBalancer listener is registered before the BufferListener,
+		// so the new PRIMARY is already in the LoadBalancer by the time we
+		// reach here. It is bounded by context timeout in any case.
+		sb.mu.Unlock()
```
haritabh17, line 109: If the topo watcher sees the new primary before the first failed request arrives (low QPS), OnPoolerChanged calls StopBuffering while the shard is still IDLE (no-op). Then the first failed request starts BUFFERING, but nobody calls StopBuffering again — stuck until MaxFailoverDuration. Would tracking a lastPrimaryChange timestamp from OnPoolerChanged and checking it in the IDLE→BUFFERING transition handle this?

haritabh17:
One thought on the overall approach — right now buffering is reactive (first failed query triggers it), which means at least one query sees an error before buffering kicks in. One way to make it proactive is with a two-phase demotion:

1. Before demoting, the pooler broadcasts "buffer now" to connected gateways via health stream
2. Gateways start buffering and ack back
3. Pooler gates new queries (returning CLUSTER_EVENT) and waits for in-flight queries to drain (with a timeout)
4. Once drained (or timeout), demotion proceeds as normal
5. New primary comes up → gateways stop buffering → retry

This gets us to zero query failures for normal OLTP workloads. The reactive path stays as a fallback for gateways that miss the signal.

It doesn't require implementing the full state manager. If not in this PR — then maybe worth considering as a follow-up since most of the pieces are already here. Thoughts?

GuptaManan100:
No not really, the first error triggers buffering, but that query gets buffered and retried too, so the app sees no errors.

go/services/multigateway/buffer/shard_buffer.go
GuptaManan100: Yes. True the best approach is to use the health stream instead of topo transitions, and I'll wait for #665 to land first, and then this will be more complete

go/services/multigateway/poolergateway/pooler_gateway.go
GuptaManan100: I'm going to add a buffering specific error code that should be sent for buffered queries.

go/services/multigateway/poolergateway/pooler_gateway.go
GuptaManan100: `ReserveStreamExecute` always creates a **new** reservation, it doesn't operate on an existing one. Methods that operate on existing reserved connections (`ConcludeTransaction`, `ReleaseReservedConnection`, `CopySendData`, `CopyFinalize`, `CopyAbort`) already bypass `withBuffering` and go directly to `GetConnection`, so they're routed to the specific pooler holding their reserved connection. No change needed here.

go/common/mterrors/code.go
GuptaManan100: Intentional design: all MT errors use custom SQLSTATE codes (MTD01–MTD08, MTE01, MTF01). This makes MT-specific errors immediately identifiable by clients. Using a standard PG code would lose that distinction and risk conflicting with real PostgreSQL errors. If monitoring tools need to handle these, they can filter on the `MT` prefix.

go/services/multigateway/buffer/buffer.go
GuptaManan100: Agreed, added a comment on `enqueue()` documenting that global FIFO eviction means shard B's entries can evict shard A's older entries during multi-shard failover, and that this is intentional to keep the implementation simple. For the common case (single-shard failover) there's no cross-shard interference.

go/services/multigateway/poolergateway/pooler_gateway.go
GuptaManan100: Done. PR #755 added `StartRequest()` which gates every gRPC handler before executing. We now return `MTF01` from `StartRequest()` for both `NOT_SERVING` and shutting-down states, and wrap it with `mterrors.ToGRPC()` so the code survives the gRPC round-trip. `classifyError` now only needs `MTF01` and `25006` (for in-flight queries that hit PostgreSQL directly during demotion). No more UNAVAILABLE/CLUSTER_EVENT in the buffering path.

go/services/multigateway/buffer/shard_buffer.go
GuptaManan100: Addressed! Replaced topo-based `BufferListener` with streaming health checks from PR #665. The `LoadBalancer.onPoolerHealthUpdate` callback fires when the health stream detects a new primary (via `PrimaryObservation` term reconciliation), then calls `buffer.StopBuffering()`. This is a direct signal from the pooler itself, no topo propagation delay or ordering issues.

go/services/multipooler/poolerserver/pooler.go
```diff
@@ -163,11 +156,11 @@ func (s *QueryPoolerServer) StartRequest(allowOnShutdown bool) error {
 	defer s.mu.Unlock()
 
 	if s.servingStatus != clustermetadatapb.PoolerServingStatus_SERVING && !s.shuttingDown {
-		return ErrNotServing
+		return mterrors.MTF01.New()
 	}
 
 	if s.shuttingDown && !allowOnShutdown {
-		return ErrShuttingDown
+		return mterrors.MTF01.New()
```
haritabh17, line 163: For GetAuthCredentials — which hardcodes a PRIMARY target — it could trigger unnecessary buffering if the only known primary is still initializing. Worth adding a separate error code for "not yet serving" ?

go/services/multigateway/buffer/buffer.go
```diff
@@ -0,0 +1,316 @@
...
+
+// drainEntriesForShard removes and returns all entries for the given shard
+// from the global queue.
+// Must NOT be called with b.mu held.
+func (b *Buffer) drainEntriesForShard(shardKey commontypes.ShardKey) []*entry {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	var drained []*entry
+	remaining := b.queue[:0]
```
haritabh17, line 306: Nit: b.queue[:0] reuses the backing array. After the loop, slots b.queue[len(remaining):cap(remaining)] still hold pointers to drained entries, preventing GC until the queue grows back or shuts down. Negligible at buffer size 1000, but for correctness: nil out the tail after the loop?

go/services/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -88,68 +190,37 @@ func (pg *PoolerGateway) QueryServiceByID(ctx context.Context, id *clustermetada
...
-	pg.logger.DebugContext(ctx, "selected pooler for target",
-		"tablegroup", target.TableGroup,
-		"shard", target.Shard,
-		"pooler_type", target.PoolerType.String(),
-		"pooler_id", conn.ID())
-
-	// Delegate to the pooler's QueryService
-	return conn.QueryService().StreamExecute(ctx, target, sql, options, callback)
+	var state *query.ReservedState
+	err := pg.withBuffering(ctx, target, func(conn *PoolerConnection) error {
```
haritabh17, line 212: Just noting that the callback-based methods have an implicit contract: the callback must be safe to invoke from scratch on retry. Worth a comment on withBuffering documenting this.

go/services/multipooler/poolerserver/pooler.go
GuptaManan100: Good catch on the edge case. Buffering during initial startup is actually the desired behavior here, the auth request gets held until the primary becomes SERVING, then retries successfully. This is better than returning an error to the client during startup.

The concern about conflating "not yet ready" with "failover in progress" is fair from a metrics perspective (it would record a spurious "failover" event). In practice this is a one-time occurrence at initial cluster startup, well before any real failovers happen, so it won't interfere with `MinTimeBetweenFailovers` guards in any realistic deployment.

A separate error code (e.g., `MTF02` for "not yet initialized") would add complexity for a case that resolves naturally and happens at most once per cluster lifecycle. I'd prefer to defer this unless we see real-world confusion in monitoring. Happy to add a comment on `GetAuthCredentials` noting this behavior if that helps.

go/services/multigateway/buffer/buffer.go
GuptaManan100: Fixed. Added a loop to nil out the stale tail pointers after partitioning.

go/services/multigateway/poolergateway/pooler_gateway.go
GuptaManan100: Fixed. Added an `IMPORTANT` paragraph to the `withBuffering` doc comment documenting the contract: `inner` must be safe to call from scratch on retry with a fresh `PoolerConnection`, and streaming callbacks must tolerate being invoked on a new connection without partial state carried over.

PR #691, opened 2026-03-05
fix(pgctld,provisioner): move locale handling from pgctld to  the provisioner

mkindahl:
`pgctld` was injecting `LC_ALL=C` into the environment of `initdb` and `pg_ctl` on macOS. Production code should not override the caller's locale: it can mask misconfigured deployments, change collation and encoding of the initialized cluster, and violates the principle of least surprise.

Remove the `LC_ALL` injection from `pgctld` and move it to the provisioner, which is responsible for providing a correct execution environment to the processes it spawns. `LC_ALL=C` is only injected when neither `LC_ALL` nor `LANG` is set in the caller's environment, so an existing locale is always preserved unchanged.

Fix tests that built Env slices manually without locale vars by adding inline darwin checks in all affected test sites.

### DIFF
--- a/go/cmd/pgctld/command/init.go
+++ b/go/cmd/pgctld/command/init.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"runtime"
 
 	"github.com/multigres/multigres/go/services/pgctld"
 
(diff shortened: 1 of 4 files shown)


go/cmd/pgctld/command/init.go
```diff
@@ -183,17 +182,6 @@ func initializeDataDir(logger *slog.Logger, poolerDir string, pgUser string) err
 
 	cmd := exec.Command("initdb", args...)
 
-	// On macOS, ensure locale environment variables are set for initdb.
-	// initdb requires valid locale settings to avoid "invalid locale settings" errors.
-	// This is specific to macOS where LC_ALL may not be set by default.
-	if runtime.GOOS == "darwin" {
```
cuongdo, line 189: This could cause the postmaster crashing issue when running a local cluster on macOS, which affects both development and anyone trying Multigres from source. Or is that not an issue?

mkindahl: Good point. I would assume that you set these correctly in whatever scripts you use to start your local cluster, but I will investigate and check if that is the case or if we need to add it.

cuongdo: The typical way I start my local cluster is:

```shell
multigres cluster init
multigres cluster start
```

mkindahl: Updated the patch so that it sets `LC_ALL` inside the provisioner if `LC_ALL` was not set. If you call `pgctld` directly without `LC_ALL` set you will still get an error.

Unfortunately the `initdb` error, which contains information about `LC_*` and `LANG`, is not shown, so it might be a good idea to fix that for `pgctld` so that a proper error is logged.

PR #693, opened 2026-03-06
fix(guc): use set_config() for GUC propagation to fix list-valued GUCs

haritabh17:
`ApplyQuery()` generates `SET SESSION name = 'value'` which wraps the entire value in single quotes. This breaks list-valued GUCs like `search_path`, `DateStyle`, and `lc_monetary`. For example, `SET SESSION search_path = 'temp_func_test, public'` makes PostgreSQL treat the whole string as a single schema name `"temp_func_test, public"` instead of two separate schemas. The SQL parser produces a single `A_Const` node, and `flatten_set_variable_args` with `GUC_LIST_QUOTE` calls `quote_identifier` on it — one identifier, not two. This causes `CREATE FUNCTION` and similar DDL to fail with "no schema has been selected to create in" when `search_path` has multiple schemas.

This PR switches `ApplyQuery()` from `SET SESSION` to `SELECT pg_catalog.set_config()`, which is the same approach pg_dump uses (see `pg_dump.c`). `set_config()` passes the flat string directly to `set_config_option()`, which internally splits comma-separated values for `GUC_LIST_INPUT` variables. Single quotes in names and values are escaped by doubling (`'` → `''`), matching PostgreSQL's standard escaping.

The planner also gets handling for three previously-unsupported `VariableSetStmt` kinds. `VAR_SET_DEFAULT` (`SET var TO DEFAULT`) is treated as `RESET`, matching PostgreSQL's `ExecSetVariableStmt` which falls through from `VAR_SET_DEFAULT` to `VAR_RESET`. `VAR_SET_MULTI` (`SET TRANSACTION ...`) and `VAR_SET_CURRENT` (`SET var FROM CURRENT`) are passed through to PostgreSQL since they need backend execution and don't need session tracking.

## Test plan

Unit tests cover `ApplyQuery()` with list GUCs, SQL injection escaping, empty settings, and single quotes in variable names. Planner tests cover `VAR_SET_DEFAULT`, `VAR_SET_MULTI`, and `VAR_SET_CURRENT`. All existing multipooler tests updated to match the new `set_config()` pattern. Full pgregress suite run locally — 5 new passes (`create_function_sql`, `btree_index`, `gist`, `select_implicit`, `spgist`).

### DIFF
--- a/go/services/multigateway/planner/variable_set_stmt.go
+++ b/go/services/multigateway/planner/variable_set_stmt.go
@@ -59,42 +59,28 @@ func (p *Planner) planVariableSetStmt(
 		return plan, nil
 	}
 
-	// Handle each SET kind:
-	// - VAR_SET_VALUE, VAR_RESET, VAR_RESET_ALL: tracked locally in SessionSettings
-	// - VAR_SET_DEFAULT: equivalent to RESET (PG passes NULL to set_config_option)
-	// - VAR_SET_MULTI: SET TRANSACTION / SET SESSION CHARACTERISTICS — pass through to PG
-	// - VAR_SET_CURRENT: SET FROM CURRENT — pass through to PG
+	// SET var TO DEFAULT is equivalent to RESET var in PostgreSQL
+	// (PG's ExecSetVariableStmt falls through from VAR_SET_DEFAULT to VAR_RESET).
+	// Normalize before the switch so it shares the same tracking path.
+	if stmt.Kind == ast.VAR_SET_DEFAULT {
+		p.logger.Debug("SET TO DEFAULT treated as RESET",
+			"variable", stmt.Name)
+		stmt = &ast.VariableSetStmt{
+			Kind: ast.VAR_RESET,
+			Name: stmt.Name,
+		}
+	}
+
 	switch stmt.Kind {
 	case ast.VAR_SET_VALUE, ast.VAR_RESET, ast.VAR_RESET_ALL:
 		// These are tracked locally
 
-	case ast.VAR_SET_DEFAULT:
-		// SET var TO DEFAULT is equivalent to RESET var in PostgreSQL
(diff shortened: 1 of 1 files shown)


go/services/multigateway/planner/variable_set_stmt.go
```diff
@@ -59,13 +59,45 @@ func (p *Planner) planVariableSetStmt(
...
+		// (PG's ExecSetVariableStmt falls through from VAR_SET_DEFAULT to VAR_RESET)
+		p.logger.Debug("SET TO DEFAULT treated as RESET",
+			"variable", stmt.Name)
+		resetStmt := &ast.VariableSetStmt{
+			Kind: ast.VAR_RESET,
+			Name: stmt.Name,
+		}
+		primitive := engine.NewApplySessionState(sql, resetStmt)
+		plan := engine.NewPlan(sql, primitive)
+		return plan, nil
```
GuptaManan100: Why don't we add `ast.VAR_SET_DEFAULT` to the previous switch itself

go/services/multigateway/planner/variable_set_stmt.go
```diff
@@ -59,13 +59,45 @@ func (p *Planner) planVariableSetStmt(
...
+		// commands that must be executed directly on the backend connection.
+		// They set multiple GUCs internally (transaction_isolation, etc.)
+		// and don't need to be tracked in SessionSettings.
+		p.logger.Debug("SET MULTI (SET TRANSACTION) passing through to PostgreSQL",
+			"variable", stmt.Name)
+		return p.planDefault(sql, conn)
+
+	case ast.VAR_SET_CURRENT:
+		// SET var FROM CURRENT — reads current PG value and sets it.
+		// Must be executed on the backend.
```
GuptaManan100: These 2 can also be clubbed together

PR #697, opened 2026-03-09
feat(multipooler): replace admin/internal users with POSTGRES_USER/PASSWORD

mkindahl:
Use the standard `POSTGRES_USER` and `POSTGRES_PASSWORD` environment variables for the PostgreSQL superuser credentials instead of the service-specific `CONNPOOL_ADMIN_USER`, `CONNPOOL_ADMIN_PASSWORD`, and `CONNPOOL_INTERNAL_USER`.

The `--connpool-admin-user` and `--connpool-admin-password` flags and their associated environment variables are kept for backwards compatibility, but are deprecated and should not be used.

The old internal user used trust auth with no password, which is incompatible with supabase postgres (requires SCRAM-SHA-256 on all connections). Internal queries (heartbeat, replication tracking) are now routed through the admin pool, which already supports password auth. The `internalUser` concept is removed entirely.

Admin credentials are renamed `pgUser`/`pgPassword` (viper keys `connpool.pg.user` / `connpool.pg.password`, flag `--connpool-pg-user`) to align with `pgctld` naming. Passwords are configured via env var only -- no CLI flag is registered.

Fixes MUL-205


mkindahl:
Just noting this down.

The integration tests fails with different tests each time and has to be run several times to pass. It seems to be related to GetFreePort() returning the same port in different processes causing race conditions. Going to investigate this separately.

Here is a successful run: https://github.com/multigres/multigres/actions/runs/22952456900/job/66753711621?pr=697

go/services/multipooler/connpoolmanager/config.go
```diff
@@ -151,21 +151,16 @@ func NewConfig(reg *viperutil.Registry) *Config {
...
-		// Admin credential settings
-		adminUser: viperutil.Configure(reg, "connpool.admin.user", viperutil.Options[string]{
-			Default:  "postgres",
-			FlagName: "connpool-admin-user",
-			EnvVars:  []string{"CONNPOOL_ADMIN_USER"},
-		}),
-		adminPassword: viperutil.Configure(reg, "connpool.admin.password", viperutil.Options[string]{
-			Default:  "",
-			FlagName: "connpool-admin-password",
-			EnvVars:  []string{"CONNPOOL_ADMIN_PASSWORD"},
```
mkindahl, line 163: We should keep these flags and variables until Multigres operator are not using them any more.

go/services/multipooler/connpoolmanager/config.go
```diff
@@ -240,10 +235,8 @@ func NewConfig(reg *viperutil.Registry) *Config {
 
 // RegisterFlags registers all connection pool flags with the given FlagSet.
 func (c *Config) RegisterFlags(fs *pflag.FlagSet) {
-	// Admin credential flags
-	fs.String("connpool-admin-user", c.adminUser.Default(), "Admin pool user (PostgreSQL superuser for control operations)")
-	fs.String("connpool-admin-password", c.adminPassword.Default(), "Admin pool password (can also be set via CONNPOOL_ADMIN_PASSWORD env var)")
```
mkindahl, line 245: Same as above. Keep them until operator has removed them.

go/services/multipooler/connpoolmanager/config.go
mkindahl: Moved to a [separate PR](https://github.com/multigres/multigres/pull/726), to be merged after Multigres operator has stopped using the variables.

go/services/multipooler/connpoolmanager/config.go
mkindahl: Also moved to a [separate pull request](https://github.com/multigres/multigres/pull/726).

go/services/multipooler/executor/internal_query.go
```diff
@@ -48,15 +45,15 @@ type InternalQueryService interface {
...
+// It executes a query using the admin connection pool and returns the first result.
 // Internal queries include SQL text in trace spans since they use system functions.
 func (e *Executor) Query(ctx context.Context, queryStr string) (*sqltypes.Result, error) {
 	// Enable SQL text in trace spans for internal queries (safe - no user data)
 	ctx = client.WithQueryTracing(ctx, client.QueryTracingConfig{
 		IncludeQueryText: true,
 	})
 
-	conn, err := e.poolManager.GetRegularConn(ctx, e.poolManager.InternalUser())
+	conn, err := e.poolManager.GetAdminConn(ctx)
```
GuptaManan100: I'm not sure about this change specifically. If we end up exhausting the Admin connections for queries, then we might not have any left to kill queries/connections

mkindahl: Hmm... good point. I missed that there was a cap on the number of administrative connections.

I was thinking about having a dedicated function for this, but instead just exposed the `PgUser` through the interface and restored the use of `GetRegularConn` instead. It is still the single PostgreSQL user for the internal connection, so this should be dealt with.

One issue that comes to mind is that since we are using a user with SUPERUSER privileges we can kill _any_ backend or cancel _any_ query even if we do not own it. If we make a mistake in the code, it would mean that we kill a backend for a different user than the owner. It might be an idea to configure the connection with the user for which we want to cancel the query or terminate the backend to avoid mistakes here. In practice it means picking a connection from the user's pool.

GuptaManan100: That has issues of its own, like how many connections do you keep for every user ear-marked for connection killing? If we keep 1 connection per user, then even that can lead to connections being wasted, for example if we have 20 users, then 20 connections aren't being used. And if we don't do something like this, then we could end up in a situation where we have exhausted max connections, and there is no way to cancel a query

go/services/multipooler/executor/internal_query.go
```diff
@@ -48,15 +45,15 @@ type InternalQueryService interface {
 var _ InternalQueryService = (*Executor)(nil)
 
 // Query implements InternalQueryService for simple internal queries.
-// It executes a query using the configured internal user and returns the first result.
+// It executes a query using the admin connection pool and returns the first result.
```
GuptaManan100: This comment needs to be updated

mkindahl: Fixed.

PR #702, opened 2026-03-09
fix(bootstrap): shard lock TTL

dweitzman:
## Summary

- **Short shard lock TTL**: The etcd shard lock acquired during bootstrap now has a 60-second TTL, so the lock auto-expires if orch crashes.

- **Initial backup captures full state**: Moved the initial pgBackRest backup to after durability policy creation and leadership history insertion. Standbys restoring from this backup now start with a fully-initialized schema rather than a partial one. This may not be critical, but it doesn't feel entirely unreasonable

- **`isInitialized()` doesn't cache its result without confirming schema**: If postgres is down we trust the marker file but re-check the schema on the next call.

- **Fix nil-interface panic in `InternalQueryService()`**: Returning a nil `*Executor` directly produced a non-nil interface wrapping a nil pointer, causing callers' `== nil` guards to pass but method calls to panic.

- **Test improvements**: Added `TestIsInitialized` unit tests for the new two-condition semantics; cleaned up `TestInitializeEmptyPrimary`; switched from `context.Background()` to `t.Context()` throughout.


go/cmd/multigres/command/pooler/status.go
```diff
@@ -41,7 +41,7 @@ Key fields returned:
   - pooler_type: Whether this is a PRIMARY or REPLICA pooler
   - postgres_running: Whether the PostgreSQL process is running
   - postgres_role: Actual database role (primary/standby/unknown)
-  - is_initialized: Whether the pooler has been fully initialized
+  - has_backup: Whether the pooler has been fully initialized
```
cuongdo: `has_backup` doesn't imply fully initialized to someone who hasn't seen our discussions on this PR. maybe `fully_initialized`? or keep `is_initialized` but strengthen the definition?

go/common/topoclient/bootstrap_attempt.go
```diff
@@ -0,0 +1,110 @@
...
+// pathForBootstrapAttempt returns the topology path for the BootstrapAttempt record of a shard.
+func pathForBootstrapAttempt(shardKey types.ShardKey) string {
+	return path.Join(DatabasesPath, shardKey.Database, shardKey.TableGroup, shardKey.Shard, bootstrapAttemptFile)
+}
+
+// UpdateBootstrapAttempt reads the current BootstrapAttempt for the shard (nil if none),
+// passes it to fn, and writes the result back if fn returns a non-nil record.
+//
+// Returns (true, nil)  if fn returned a non-nil record and it was successfully written.
+// Returns (false, nil) if fn returned nil (chose not to write, e.g. rate-limited).
```
cuongdo: what type of rate limiting would apply to this method?

go/common/topoclient/bootstrap_attempt.go
```diff
@@ -0,0 +1,110 @@
...
+// Returns (false, err) if the topology read/write failed or fn returned an error.
+//
+// The shard lock must be held by the caller. Retries transparently on CAS conflicts.
+func (ts *store) UpdateBootstrapAttempt(
+	ctx context.Context,
+	shardKey types.ShardKey,
+	fn func(*clustermetadatapb.BootstrapAttempt) (*clustermetadatapb.BootstrapAttempt, error),
```
cuongdo: nit: maybe it's just me, but readability of this method if this name was more descriptive. e.g. `shouldUpdateFn`

go/common/topoclient/bootstrap_attempt.go
```diff
@@ -0,0 +1,110 @@
...
+) (bool, error) {
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+	filePath := pathForBootstrapAttempt(shardKey)
+
+	for {
```
cuongdo: Should this have a limit on the number of iterations, or is that up to `fn`?

go/common/topoclient/bootstrap_attempt_test.go
```diff
@@ -0,0 +1,241 @@
...
+				OrchId:    orchID,
+				StartedAt: timestamppb.New(startedAt),
+			}, nil
+		})
+	require.NoError(t, err)
+}
+
+// TestUpdateBootstrapAttempt_NoExistingRecord: fn receives nil when no record exists, write succeeds.
+func TestUpdateBootstrapAttempt_NoExistingRecord(t *testing.T) {
+	ctx := context.Background()
```
cuongdo: should this be `t.Context()` instead?

cuongdo: same question applies to other tests below

go/common/topoclient/bootstrap_attempt_test.go
```diff
@@ -0,0 +1,241 @@
...
+	ts, _ := memorytopo.NewServerAndFactory(ctx, "zone-1")
+	defer ts.Close()
+
+	shard0 := types.ShardKey{Database: "mydb", TableGroup: "mygroup", Shard: "0"}
+	shard1 := types.ShardKey{Database: "mydb", TableGroup: "mygroup", Shard: "1"}
+
+	write := func(shardKey types.ShardKey, orchID string) (bool, error) {
+		return ts.UpdateBootstrapAttempt(ctx, shardKey, func(_ *clustermetadatapb.BootstrapAttempt) (*clustermetadatapb.BootstrapAttempt, error) {
+			return &clustermetadatapb.BootstrapAttempt{
```
cuongdo: should this function verify that no prior BootstrapAttempt exists?

cuongdo:
Also, it would be greatly beneficial to have adversarial end-to-end tests for this code. For example, we could set up conditions that make dueling bootstraps guaranteed or at least very likely.

dweitzman:
> Also, it would be greatly beneficial to have adversarial end-to-end tests for this code. For example, we could set up conditions that make dueling bootstraps guaranteed or at least very likely.

The idempotency here is throwaway code, so I'm debating how much to invest in testing 

For a theoretically sound solution we'd need to update the provisioner so it can somehow indicate both the initial desired durability rules and what backup to begin the cohort with (or whether to run initdb to create a new shard)

Need to sync first with folks working on provisioner to negotiate exactly how the provisioner will communicate those things

go/common/topoclient/bootstrap_attempt.go
dweitzman: For this PR, `bootstrapCooldown = 30 * time.Second`. I'm thinking of all this as throwaway code, though. This method would be deleted once we've settled on a way for provisioner to safely communicate what poolers & backup are needed for the initial cohort

dweitzman:
> I am unsure what problem it is trying to solve. I have a feeling I am missing something here and would like to double check.
> 
> > We haven't been retrying to make an initial shard backup if the first attempt fails, which can get a shard stuck. This isn't the most sound theoretical solution, but it may be an adequate solution for now: a pooler can record in etcd if it's about to try to make a bootstrap backup, and other poolers will not try within 30 seconds assuming clocks are accurate across poolers
> 
> What is the situation where two orchestrators can both end up with trying to initialize an empty primary (I assume this is the case you're trying to handle).
> 
> From what I can tell, the sequence is roughly this (I hope this renders correctly):
> 
> ```mermaid
> sequenceDiagram
>     participant OA as Orch-A
>     participant etcd
>     participant MP as Multipooler
>     participant pgctld
>     participant PG as Postgres
>     participant PBR as pgBackRest
> 
>     note over OA,PBR: Orch-A acquires shard lock
> 
>     OA->>etcd: LockShard
>     etcd-->>OA: lock ok
> 
>     OA->>etcd: Write BootstrapAttempt(orchId=A, startedAt=T)
>     etcd-->>OA: ok
> 
>     note over OA,PBR: Orch-A dispatches InitializeEmptyPrimary
> 
>     OA->>MP: InitializeEmptyPrimary(term=1, policy=ANY_2)
>     MP->>MP: acquire actionLock
>     MP->>MP: pause postgres monitor
> 
>     note over MP,PG: Initialize data directory and start Postgres
> 
>     MP->>pgctld: InitDataDir
>     pgctld->>PG: initdb
>     PG-->>pgctld: ok
>     pgctld-->>MP: ok
> 
>     MP->>MP: write archive_mode to postgresql.auto.conf
> 
>     MP->>pgctld: Start
>     pgctld->>PG: pg_ctl start
>     PG-->>pgctld: running
>     pgctld-->>MP: ok
> 
>     MP->>PG: wait for connection
> 
>     note over MP,PG: Create multigres schema + tables and insert tablegroup + shard records
> 
>     MP->>PG: create multigres schema + tables
>     PG-->>MP: ok
>     MP->>PG: insert tablegroup + shard records
>     PG-->>MP: ok
> 
>     MP->>MP: write consensus term=1 to disk
> 
>     note over MP,PBR: Initialize pgBackRest stanza
> 
>     MP->>PBR: pgbackrest stanza-create (5s timeout)
>     PBR-->>MP: ok
> 
>     note over MP,etcd: Register as PRIMARY in topology
> 
>     MP->>etcd: RegisterMultiPooler(type=PRIMARY)
>     etcd-->>MP: ok
> 
>     MP->>MP: write primary_term=1 to disk
> 
>     note over MP,PG: Insert durability policy and leadership history
> 
>     MP->>PG: insert durability policy + leadership history
>     PG-->>MP: ok
> 
>     note over MP,PBR: Take initial full backup
> 
>     MP->>PBR: pgbackrest backup --type=full
>     PBR->>PG: connect + stream base backup
>     PG-->>PBR: data
>     PBR-->>MP: backup_id=...
> 
>     MP->>PBR: pgbackrest verify --set=backup_id
>     PBR-->>MP: ok
> 
>     note over MP,PBR: Mark bootstrap complete
> 
>     MP->>MP: write MULTIGRES_HAS_BACKUP marker
>     MP->>MP: release actionLock
>     MP->>MP: resume postgres monitor
> 
>     MP-->>OA: InitializeEmptyPrimaryResponse(success=true, backup_id=...)
> 
>     note over OA,etcd: Configure synchronous replication and release lock
> 
>     OA->>MP: ConfigureSynchronousReplication
>     MP-->>OA: ok
> 
>     OA->>etcd: ReleaseLock
> ```
> Note that in this case the lock protect the entire sequence of operations, so no other orchestrator can come in and try to bootstrap the server since they do not get the lock on the shard if they try.
> 
> Once the lock is released, be it because of a failure or a success, another orchestrator can come in and pick up the job and either complete it or notice that it's done.
> 
> Given this context, what will the "cool down period" solve? I have a feeling I am missing something here.
> 
> There is a problem of another kind here in that the RPC timeout is set to 60s (I have not found a different setting) and I would not trust that for the bootstrapping of the server. If the operation is aborted with a timeout, the server might still be bootstrapping and if an orchestrator comes in at this point, we have a problem.
> 
> (There is another problem here in that the lock attempt will block, so an orchestrator that want to bootstrap a server should probably do a try-lock rather than lock and wait since another orchestrator might have decided to pick up the job. This is however not relevant here.)

Ah, you're right! I didn't realize there was already an etcd lock on this. It accomplishes the same thing, more or less, I think we just need to set a shorter TTL (the default is 24 hours, which seems excessive for bootstrap)

So I think what's needed here is:
- Make the etcd lock shorter
- Keep the changes to allow bootstrap attempts if the repo exists but the stanza has no actual backups, but revert the changes for the 30-second grace period
- See if we have any agreement on whether the rename is necessary

I'll start working on the first two

dweitzman:
Simplified this substantially to mostly be about setting a shard lock TTL. PTAL

mkindahl:
> Ah, you're right! I didn't realize there was already an etcd lock on this. It accomplishes the same thing, more or less, I think we just need to set a shorter TTL (the default is 24 hours, which seems excessive for bootstrap)

I'm unsure what TTL is in this case. Normally having a TTL means evicting some cache information from a cache to trigger a reload, but what information here is it that is expiring? Are you actually referring to a timeout (a time after which an operation is considered failed if it does not progress)? And yes, 24 hours seems excessive for both a TTL and a timeout.

> So I think what's needed here is:
> 
> * Make the etcd lock shorter

What is it you want to accomplish by making the lock shorter?

> * Keep the changes to allow bootstrap attempts if the repo exists but the stanza has no actual backups, but revert the changes for the 30-second grace period

Seems reasonable to try to continue the bootstrap if it stops halfway through for some reason, yes.

> * See if we have any agreement on whether the rename is necessary

No opinion on names. I'll accept whatever you think is good.

go/services/multiorch/store/replication_analysis.go
```diff
@@ -45,7 +45,7 @@ type ReplicationAnalysis struct {
 	// valid status response
 	LastCheckValid   bool
 	IsStale          bool
-	IsInitialized    bool // Whether this pooler has been initialized
+	IsInitialized    bool // Whether this pooler is fully initialized and ready to join the cohort
```
cuongdo, line 48: This comment makes the intent of this field much clearer

cuongdo:
> I'm unsure what TTL is in this case. Normally having a TTL means evicting some cache information from a cache to trigger a reload, but what information here is it that is expiring? Are you actually referring to a timeout (a time after which an operation is considered failed if it does not progress)? And yes, 24 hours seems excessive for both a TTL and a timeout.

I'm fairly certain I know this answer to this, but I'd like to hear it from @dweitzman for confirmation.

dweitzman:
> > Ah, you're right! I didn't realize there was already an etcd lock on this. It accomplishes the same thing, more or less, I think we just need to set a shorter TTL (the default is 24 hours, which seems excessive for bootstrap)
> 
> I'm unsure what TTL is in this case. Normally having a TTL means evicting some cache information from a cache to trigger a reload, but what information here is it that is expiring? Are you actually referring to a timeout (a time after which an operation is considered failed if it does not progress)? And yes, 24 hours seems excessive for both a TTL and a timeout.

The TTL here is a little nuanced. I didn't realize that nuance so this is a good question, but I think this is what's actually happening: it's a TTL on how long after a loss of network connectivity between orch and etcd before etcd gives up and evicts that orch as the lock holder

If orch is still running and is still able to reach etcd, the lock can actually be held indefinitely. Our `lock.go` calls the etcd Lease's "KeepAlive" function: https://github.com/etcd-io/etcd/blob/4e814e204934c3c682d9e185db1dfb646d2510b3/client/v3/lease.go#L122

> 
> > So I think what's needed here is:
> > 
> > * Make the etcd lock shorter
> 
> What is it you want to accomplish by making the lock shorter?

This TTL only address the scenario where network connectivity between orch and etcd is interrupted and etcd doesn't have conclusive evidence as to whether orch crashed, finished its operation, or the network is just slow or unreliable. This should be rare in practice, because we only use the lock for bootstrap (which is uncommon to begin with) and we'd need some networking bad luck or a violently crashed process while that lock is held to end up in a state where it needs to be externally released 

> 
> > * Keep the changes to allow bootstrap attempts if the repo exists but the stanza has no actual backups, but revert the changes for the 30-second grace period
> 
> Seems reasonable to try to continue the bootstrap if it stops halfway through for some reason, yes.
> 
> > * See if we have any agreement on whether the rename is necessary
> 
> No opinion on names. I'll accept whatever you think is good.

mkindahl:
> > > Ah, you're right! I didn't realize there was already an etcd lock on this. It accomplishes the same thing, more or less, I think we just need to set a shorter TTL (the default is 24 hours, which seems excessive for bootstrap)
> > 
> > 
> > I'm unsure what TTL is in this case. Normally having a TTL means evicting some cache information from a cache to trigger a reload, but what information here is it that is expiring? Are you actually referring to a timeout (a time after which an operation is considered failed if it does not progress)? And yes, 24 hours seems excessive for both a TTL and a timeout.
> 
> The TTL here is a little nuanced. I didn't realize that nuance so this is a good question, but I think this is what's actually happening: it's a TTL on how long after a loss of network connectivity between orch and etcd before etcd gives up and evicts that orch as the lock holder

Thank you for the explanation, I think all is good, but let me expand on my reasoning.

Having a TTL on the locks is reasonable, provided all actors are aware of that and collaborate.
The lock can be seen as a contract or agreement that "I have the exclusive right to work with this server for this long".

If the lock is held for, say, 30 seconds and then expires, but the orch continue with the operation beyond the promised lock TTL, then another orch might "crash the party" and start other operations that conflict with the first operation. In other words, this strategy with a TTL only works if all actors agree to stop working when the lock/contract expires. From looking at `runLongCommand` it seems that the operations _are_ aborted (using SIGKILL) if the gRPC call expires, so all should be fine.

> If orch is still running and is still able to reach etcd, the lock can actually be held indefinitely. Our `lock.go` calls the etcd Lease's "KeepAlive" function: https://github.com/etcd-io/etcd/blob/4e814e204934c3c682d9e185db1dfb646d2510b3/client/v3/lease.go#L122

I think this is fine. If the backup needs to take some time, there is no way to make it move faster. Especially not by killing it and trying again. The problematic case is when there is a network partition and the actors need to act alone.

> > > So I think what's needed here is:
> > > 
> > > * Make the etcd lock shorter
> > 
> > 
> > What is it you want to accomplish by making the lock shorter?
> 
> This TTL only address the scenario where network connectivity between orch and etcd is interrupted and etcd doesn't have conclusive evidence as to whether orch crashed, finished its operation, or the network is just slow or unreliable. This should be rare in practice, because we only use the lock for bootstrap (which is uncommon to begin with) and we'd need some networking bad luck or a violently crashed process while that lock is held to end up in a state where it needs to be externally released

Yes, but what improves by taking a shorter lock if this event is rare in practice?

The lock time is just there as a precaution for the case where you lose connection, but operations will normally conclude earlier. Aborting an operation is fine if the operation _should_ be aborted and (essentially) never executed again and the system is fine with this. But if the operation _needs_ to be done (for example, we cannot skip bootstrapping a new server if we need one), lowering the lock time just risks creating more aborted operations. 

To illustrate the problem, suppose that the execution time of the operation has a mean of 25 seconds and a standard deviation of 5 seconds. Assuming a normal distribution, that means that the p99 for the operation is roughly 40 seconds. If the timeout is 60 seconds, the operation will on average finish in 25 seconds and 95% of the queries will finish in 35 seconds. Operations will get aborted with a probability of 10e-12 (tail of N(25,5) above 60), which is like 0.00000000001%.

Now, if the lock time is lowered, it will not affect the execution time, so p99 is still 40 seconds, but now the abort rate is something like 15% instead (tail of N(25,5) above 30). This is going to be a lot of aborts, but it will not improve the execution time of the operation since instead of letting the operation conclude, it is aborted mid-stream and have to be re-executed.

dweitzman:
> Yes, but what improves by taking a shorter lock if this event is rare in practice?

That's a fair question. I think there's an interesting theoretical discussion to be had on the tradeoffs between different wrong options.

Realistically, I think the real answer is that orch shouldn't be making the decision of whether or not to call `initdb` in the first place, it should be a provisioner decision. If multigres owns calling `initdb`, I think it only works if we lean heavily on consistent reads from `etcd` and have a contract that the provisioner may not start any multipooler unless etcd is known to be up-to-date (at least with regards to bootstrap instructions)

I'm open to dropping the TTL from this PR on the grounds of theoretical nebulousness (or making it a bigger number like 10 minutes) but I think some of the other changes are useful on their own

mkindahl:
> > Yes, but what improves by taking a shorter lock if this event is rare in practice?
> 
> That's a fair question. I think there's an interesting theoretical discussion to be had on the tradeoffs between different wrong options.
> 
> Realistically, I think the real answer is that orch shouldn't be making the decision of whether or not to call `initdb` in the first place, it should be a provisioner decision. If multigres owns calling `initdb`, I think it only works if we lean heavily on consistent reads from `etcd` and have a contract that the provisioner may not start any multipooler unless etcd is known to be up-to-date (at least with regards to bootstrap instructions)
> 
> I'm open to dropping the TTL from this PR on the grounds of theoretical nebulousness (or making it a bigger number like 10 minutes) but I think some of the other changes are useful on their own

I'm all for that. I am just wary of setting timeouts too low without a good reason. I have approved the PR and still think you should merge. We can discuss the timeout separately.

PR #706, opened 2026-03-10
feat(pgctld,multipooler): read database name from POSTGRES_DB env var

mkindahl:
Add support for the POSTGRES_DB environment variable (and --pg-database CLI flag) to pgctld and multipooler, mirroring the docker-library/postgres convention.

When POSTGRES_DB names a database other than the default "postgres", pgctld init now creates that database after initdb using a transient PostgreSQL server started on a private socket directory. The private socket prevents multipooler's background goroutines from connecting to the transient server and misreporting the pooler as ready before the multigres schema is created.

The transient server logic is encapsulated in a pgInstance struct (pg_instance.go) with methods for construction (newPgInstance), cleanup (stop), and command execution (psql).

Fixes MUL-247

### DIFF
--- a/go/test/endtoend/localprovisioner/cluster_test.go
+++ b/go/test/endtoend/localprovisioner/cluster_test.go
@@ -169,9 +169,15 @@ func cleanupTestProcesses(tempDir string) error {
 	return nil
 }
 
-// createTestConfigWithPorts creates a test configuration file with custom ports
-// The number of zones is determined by len(portConfig.Zones)
+// createTestConfigWithPorts creates a test configuration file with custom ports using "postgres"
+// as the database name. The number of zones is determined by len(portConfig.Zones).
 func createTestConfigWithPorts(tempDir string, portConfig *testPortConfig) (string, error) {
+	return createTestConfigWithDatabase(tempDir, portConfig, "postgres")
+}
+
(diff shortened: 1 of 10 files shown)


go/test/endtoend/localprovisioner/cluster_test.go
```diff
@@ -1256,43 +1256,38 @@ func TestClusterLifecycle(t *testing.T) {
...
 
-	t.Run("multipooler requires database flag", func(t *testing.T) {
-		// This test verifies that multipooler binary requires --database flag
-		// We'll test this by trying to run the provisioned multipooler directly
-		// without the --database flag and expecting it to fail
-
-		tempDir, err := os.MkdirTemp("/tmp", "mlt")
-		require.NoError(t, err)
-		defer os.RemoveAll(tempDir)
+	t.Run("multipooler database flag defaults to postgres", func(t *testing.T) {
```
cuongdo, line 1277: Also, it would be good to test that the local provisioner passes on the `--database` flag to the other services.

mkindahl: Good point, will check.

mkindahl: Added a test for that.

PR #711, opened 2026-03-10
fix(multipooler): prevent pool starvation from unbounded background connections

### LINKED ISSUE #694: Multipooler Status RPC Hangs Due to Connection Pool Starvation
## Finding Summary

The multipooler's `Status` gRPC RPC hangs indefinitely when the internal connection pool becomes permanently starved. This causes multiorch (which polls Status every 500ms with a 5-second timeout) to declare all poolers dead, which prevents the gateway from routing queries. The observer detects this as SQL probe timeouts on the gateway.

## Observed Context

This bug was observed on a `kind` cluster after a scale-down/up test that triggered the synchronous replication write stall (#695). The cluster was **not** idle — it had gone through manual scaling activity that caused a sync write stall on the primary. The pool starvation persisted permanently after the sync stall self-healed, suggesting the stall window caused irreversible damage to the connection pool's internal state.

### Timeline (2026-03-07)

| Time (UTC) | Event |
|------------|-------|
| `01:02:51` | `kubectl apply` creates MultigresCluster `m` |
| `01:02:58` | pool-0 PVC + pod created by operator |
| `01:03:03` | pool-0 multipooler process starts; postgres sidecar still initializing |
| `01:03:14` | pool-0 becomes Ready; pool-1 PVC + pod created |
| `01:03:21` | pool-0 postgres socket appears |
| `01:03:25` | pool-2 PVC + pod created — all 3 pods up within 27s |
| `01:03:26` | **First heartbeat failure on pool-0: `context deadline exceeded`** |
| `01:03:26–01:20:43` | 82 heartbeat write failures on pool-0 (intermittent) |
| `01:03:34` | All 3 pods running: pool-0=PRIMARY, pool-1=REPLICA, pool-2=REPLICA |
| `~01:04–01:16` | **pool-1 and pool-2 drained and deleted** (manual scaling test) — triggers sync replication write stall on primary |
| `01:16:12` | DrainCompleted for pool-1, pool-2; PVCs deleted |
| `01:20:05` | pool-1 and pool-2 PVCs + pods recreated (second incarnation) |
| `01:20:28` | pool-1 multipooler starts |
| `01:20:39` | pool-2 multipooler starts |
| `~01:20–01:25` | Sync stall self-heals as new standbys connect and satisfy `ANY 1` quorum |
| `~01:50` | pool-2 last seen by multiorch (inferred from "not seen for 7h51m" at 09:41) |
| `06:51:17+` | Earliest visible multiorch poll failure (earlier logs rotated out) |
| `09:43` | Confirmed: Status RPC still hangs on **all 3 poolers**, 8+ hours later |

The sync stall lasted ~4 minutes (`01:16:12` to `~01:20:25`). During this window, ALL synchronous writes on the primary blocked (including heartbeat INSERTs). The stall self-healed when the recreated standbys connected. **But the Status RPC never recovered** — it was still hanging 8+ hours later, indicating permanent pool damage.

The heartbeat failure at `01:03:26` (before any scaling) may also have been caused by `synchronous_standby_names` being configured before standbys had fully connected as streaming replicas.

### Evidence (from live kind cluster, 8h after deployment)

| Test | Result |
|------|--------|
| `grpc.health.v1.Health/Check` | Instant, returns `SERVING` |
| `MultiPoolerManager/State` | Instant, returns `{"state": "ready"}` |
| `MultiPoolerManager/Status` | **HANGS** for full timeout (10s), `DeadlineExceeded` |
| `psql` directly to postgres (socket and TCP) | Instant response |
| `pg_stat_activity` on pool-0 (primary) | **4 connections, all idle** (3 idle 8h+) |
| `pg_stat_activity` on pool-1 (replica) | **3 connections, all idle** (2 idle 7h+) |
| `pg_stat_activity` on pool-2 (replica) | **3 connections, all idle** (2 idle 7h55m+) |
| multipooler pool capacity (logged) | `regular_capacity=80, reserved_capacity=20` |
| multiorch logs | `"pooler poll failed"` on all 3 poolers, every 5s, `DeadlineExceeded` |
| observer `/api/status` | 2 error findings: SQL probes timeout on both gateways |

**Pattern:** RPCs that return cached state (State, Health) complete instantly. RPCs that query postgres (Status, PrimaryStatus, PrimaryPosition) all hang. No queries reach postgres during the hang. Postgres itself is fully responsive to direct `psql` connections.

### Pool starvation evidence

The pool has capacity 100 (80 regular + 20 reserved) but postgres has only 3–4 connections per pod, all idle. The pool's internal state is inconsistent with reality — it cannot create new connections despite having massive headroom. The most likely explanation is that the `active` atomic counter is inflated far beyond the number of real connections, which prevents `getNew()` from creating connections and forces all callers into `waitForConn()`.

**Note:** We did not capture goroutine stack dumps or `active` counter values during this incident. The counter leak theory is based on code analysis (see below), not direct runtime observation.

### What we know for certain

1. **The Status RPC hangs permanently** — observed, not theory
2. **The hang is in the connection pool** — RPCs that bypass the pool (Health, State) work; RPCs that use it (Status) hang
3. **Postgres is healthy** — direct `psql` works instantly, few idle connections
4. **The pool never self-heals** — stuck for 8+ hours with no recovery
5. **The incident followed a sync replication write stall** — scaling activity at ~01:04-01:16 triggered the stall described in #695

### Call chain (verified)

```
Status RPC (grpcmanagerservice/service.go)
  → MultiPoolerManager.Status(ctx) (manager/rpc_manager.go:281)
    → pm.isInitialized(ctx) / pm.getRole(ctx) / pm.isPrimary(ctx) / ...
      → pm.query(ctx, sql) (manager/manager.go:264)
        → pm.internalQueryService().Query(ctx, sql) (executor/internal_query.go:53)
          → poolManager.GetRegularConn(ctx, internalUser) (connpoolmanager/manager.go:325)
            → userPool.GetRegularConn(ctx) (connpoolmanager/user_pool.go:205)
              → regularPool.Get(ctx) (pools/regular/pool.go:83)
                → connpool.Pool.get(ctx) (pools/connpool/pool.go:654)
                  → pool.waitForConn(ctx, ...) ← BLOCKS HERE
```

### Theory: `active` counter leak via `pool.ctx` paths

The pool's `active` atomic counter tracks connection slots in use. When `active >= capacity`, `getNew()` refuses to create new connections. Three code paths create connections using `pool.ctx` (the pool's lifecycle context — **no deadline**), combined with `DialTimeout=0`. If `connNew(pool.ctx)` hangs in any of these paths, the `active` slot is permanently consumed.

**Path 1: `put(nil)` — broken connection replacement** (`pool.go:453`)

When a broken connection is recycled via `Recycle()` → `put(nil)`, the pool calls `connNew(pool.ctx)` (not `getNew` — no CAS increment) to create a replacement. The `active` slot was already counted from the original `getNew` call. If `connNew` hangs, the goroutine is stuck forever and the existing slot is never freed via `closedConn()`.

**Path 2: `satisfyWaitersOnCapacityIncrease()` — rebalancer** (`pool.go:874`)

When the rebalancer increases capacity, this calls `getNew(pool.ctx)` for each queued waiter. `getNew` does a CAS increment of `active` **before** calling `connNew`. This is the only path that can inflate `active` beyond the number of real connections. If `connNew` hangs, the CAS increment is never rolled back.

**Path 3: `closeIdleResources()` — expired connection replacement** (`pool.go:895`)

When idle connections expire, this calls `getNew(pool.ctx)` to replace them. Same CAS-before-connNew pattern as Path 2.

### Where `connNew` could hang

For Unix socket connections (the default in this deployment), `dial()` to a non-existent socket returns immediately with `ENOENT` — it does **not** hang. So the hang cannot be in `dial()` during the startup window when the postgres socket doesn't exist yet.

The potential hang points are:
1. **`startup()` handshake** — after `dial()` succeeds (socket exists, kernel accepts connection), the postgres protocol handshake could hang if postgres is overwhelmed or unresponsive. With `pool.ctx` (no deadline), this hangs forever.
2. **Sync write stall cascade** — during the sync stall, connections that attempt writes block and timeout. `ForceClose` + `Recycle` routes them through `put(nil)` → `connNew(pool.ctx)`. If postgres is slow to complete the handshake under load, these replacement attempts can hang.

### What this means for the trigger

The sync replication write stall (Bug 1) creates a window where:
1. Multiple connections timeout on blocked writes
2. All are recycled as broken → `put(nil)` → `connNew(pool.ctx)`
3. If any of these `connNew` calls hang (in `startup()`, not `dial()`), the slot is permanently consumed
4. After the sync stall ends, the `active` counter damage is permanent
5. The pool can never create new connections → permanent starvation

## Verified Code Vulnerabilities

These are real code issues confirmed by reading the source, regardless of whether they caused this specific incident:

### 1. Missing `DialTimeout` (core enabler)

`buildClientConfig()` at `connpoolmanager/manager.go:160` does not set `DialTimeout`. The `client.Config.DialTimeout` defaults to zero, meaning `net.Dialer.Timeout = 0`. Connection creation relies entirely on the caller's context for timeout, but the three `pool.ctx` paths pass an unbounded context.

```go
// manager.go:~162
func (m *Manager) buildClientConfig(user, password string) *client.Config {
    return &client.Config{
        SocketFile: m.connConfig.SocketFile,
        Host:       m.connConfig.Host,
        Port:       m.connConfig.Port,
        Database:   m.connConfig.Database,
        User:       user,
        Password:   password,
        // DialTimeout NOT SET — defaults to 0
    }
}
```

### 2. `Reconnect` leaves `IsClosed()=false` on dial failure (latent bug)

In `client/conn.go:157`, `Reconnect()` sets `c.closed.Store(false)` before calling `dial()`. If `dial` fails, `closed` remains `false`. The caller (`retryOnConnectionError`) compensates by calling `c.conn.Close()` on failure, so this doesn't directly cause the counter leak. However, any other code that checks `IsClosed()` would get a wrong answer.

```go
// conn.go:157
func (c *Conn) Reconnect(ctx context.Context) error {
    _ = c.conn.Close()
    c.closed.Store(false)  // set false BEFORE dial
    netConn, err := dial(ctx, c.config)
    if err != nil {
        return err  // closed remains false!
    }
    // ...
}
```

### 3. No pool-level connection health check

The pool returns connections from the idle stack without validating they're alive. A connection that died while idle is returned to the caller, which must handle the failure through `QueryWithRetry`'s reconnect logic.

### 4. `active` counter has three decrement paths (not one)

| Location | Function | Mechanism |
|----------|----------|-----------|
| Line 632 | `closedConn()` | `active.Add(-1)` |
| Line 548 | `closeOnOverCapacity()` | `active.CompareAndSwap(open, open-1)` |
| Line 564 | `closeOnIdleLimitReached()` | `active.CompareAndSwap(open, open-1)` |

## Fix Location

**Upstream multigres** — the operator is not at fault.

### Suggested Fixes (Upstream)

1. **Set `DialTimeout` in `buildClientConfig`** (`connpoolmanager/manager.go:160`). A reasonable default would be 2-3 seconds. This prevents `connNew → client.Connect → dial/startup` from blocking indefinitely. This single fix would prevent permanent starvation because all three `pool.ctx` paths would fail fast instead of hanging forever.

2. **Use bounded contexts for background connection creation.** The three `pool.ctx` call sites (`put(nil)`, `satisfyWaitersOnCapacityIncrease`, `closeIdleResources`) should wrap `pool.ctx` with `context.WithTimeout(pool.ctx, dialTimeout)` before calling `connNew`/`getNew`. Defense-in-depth.

3. **Fix `Reconnect` to reset `closed` on dial failure.** In `client/conn.go:164-168`, if `dial()` fails, call `c.closed.Store(true)` before returning.

4. **Consider a pool-level liveness check.** Before returning a connection from the idle stack, validate it's alive (e.g., zero-byte read with short deadline).

5. **Consider a stale-slot reaper.** A background goroutine that periodically audits `active` vs actual connections and reclaims phantom slots would prevent any counter leak from becoming permanent.

## Key Files

| File (relative to multigres repo) | Role |
|------|------|
| `go/services/multipooler/pools/connpool/pool.go:453` | `put(nil)` — broken conn replacement via `connNew(pool.ctx)` **(leak path 1)** |
| `go/services/multipooler/pools/connpool/pool.go:548` | `closeOnOverCapacity()` — `active` decrement path 2 |
| `go/services/multipooler/pools/connpool/pool.go:564` | `closeOnIdleLimitReached()` — `active` decrement path 3 |
| `go/services/multipooler/pools/connpool/pool.go:632` | `closedConn()` — `active` decrement path 1 |
| `go/services/multipooler/pools/connpool/pool.go:635` | `getNew()` — CAS increments `active`, calls `connNew(ctx)` |
| `go/services/multipooler/pools/connpool/pool.go:654` | `pool.get()` — enters `waitForConn` when `active >= capacity` |
| `go/services/multipooler/pools/connpool/pool.go:874` | `satisfyWaitersOnCapacityIncrease()` — `getNew(pool.ctx)` **(leak path 2)** |
| `go/services/multipooler/pools/connpool/pool.go:895` | `closeIdleResources()` — `getNew(pool.ctx)` **(leak path 3)** |
| `go/services/multipooler/pools/connpool/pooled.go:50` | `Recycle()` — routes to `put(nil)` when `IsClosed()=true` |
| `go/services/multipooler/pools/connpool/waitlist.go:48` | `waitForConn()` — blocking select on channel |
| `go/services/multipooler/connpoolmanager/manager.go:38` | `initialUserPoolCapacity=10` |
| `go/services/multipooler/connpoolmanager/manager.go:160` | `buildClientConfig()` — missing `DialTimeout` **(core config bug)** |
| `go/services/multipooler/connpoolmanager/config.go` | `globalCapacity=100`, `reservedRatio=0.2`, `minCapacityPerUser=10` |
| `go/services/multipooler/manager/rpc_manager.go:281` | Status RPC handler |
| `go/services/multipooler/manager/manager.go:264` | `pm.query()` — through internal pool |
| `go/services/multipooler/executor/internal_query.go:53` | `Executor.Query()` — acquires pool connection |
| `go/services/multipooler/pools/regular/regular_conn.go:465` | `retryOnConnectionError` — reconnect on failure |
| `go/services/multipooler/pools/regular/regular_conn.go:530` | `execOnce` — context cancellation + ForceClose |
| `go/common/pgprotocol/client/conn.go:121` | `client.Connect()` — dial + startup handshake |
| `go/common/pgprotocol/client/conn.go:157` | `Reconnect()` — sets `closed=false` before dial **(latent bug)** |
| `go/common/pgprotocol/client/conn.go:184` | `dial()` — uses `DialTimeout` (defaults to 0) |
| `go/services/multipooler/heartbeat/writer.go:91` | Heartbeat writer — default 1s timeout (configurable), uses same pool |
| `go/services/multiorch/recovery/healthcheck.go:133` | MultiOrch poll — 5s timeout on Status RPC |

## Why It Never Self-Heals

Once the pool's internal state is corrupted, it cannot recover:

1. **Hung goroutines never return.** Goroutines stuck in `connNew(pool.ctx)` with no timeout block until the pool's lifecycle context is cancelled (pod restart). They hold `active` slots permanently.

2. **Capacity increases don't help.** The rebalancer raises capacity from 10 to 100, but `satisfyWaitersOnCapacityIncrease()` creates connections using `pool.ctx` — if postgres is slow or the handshake hangs, these attempts also consume slots.

3. **Even after the trigger resolves, the damage is permanent.** Once `active >= capacity` (even though few real connections exist), all `get()` calls enter `waitForConn()`. No new connections can be created.

4. **No stale-slot reaper.** There is no background goroutine that audits `active` vs actual connections and reclaims leaked slots.

The cluster remained broken for 8+ hours until we investigated. It would have stayed broken indefinitely without a pod restart.

## Cascade Effect

```
Sync replication write stall (trigger — see #695)
  → Synchronous writes block, connections timeout
  → ForceClose + Recycle → put(nil) → connNew(pool.ctx) — may hang
  → active counter inflated, pool slots permanently consumed
  → Pool starved: all get() calls enter waitForConn()
  → Status RPC hangs (server-side goroutine blocks in waitForConn)
  → MultiOrch 5s timeout fires, logs "pooler poll failed"
  → MultiOrch marks all poolers as dead
  → Gateway has no healthy poolers to route to
  → All client queries hang or fail
```

## Observability Gap: All Components Report Healthy During Total Outage

During the 8+ hour outage, **every pod in the cluster reported Ready=true with zero restarts**. Kubernetes, the operator, and any external monitoring would see a perfectly healthy cluster. This is a critical observability failure.

### Current state of health reporting

| Component | Probes | Behavior during outage | Correct? |
|-----------|--------|----------------------|----------|
| **Multipooler** | Startup, liveness (`/live`), and readiness (`/ready`) probes on port 15200 (configured by operator on the sidecar init container) | All probes return 200. gRPC `Health/Check` → SERVING, `State` → "ready". All return cached state, never touch the connection pool. `ready=true, restarts=0` after 8h of total failure. | **Wrong.** The component cannot serve any SQL queries but all health endpoints say it's fine. |
| **MultiOrch** | HTTP `/live` + `/ready` on :15300 | Returns 200 OK while logging "pooler poll failed" on all 3 poolers every 5s | **Wrong.** It knows the entire data plane is dead but reports healthy. |
| **Gateway** | HTTP `/live` + `/ready` on :15100 | Returns 200 OK while every SQL query through it times out | **Wrong.** It has zero healthy routing targets but reports ready to receive traffic. |
| **Pool pod (Kubernetes)** | Probes configured on multipooler sidecar (see above), none on postgres container | Pod stays Ready=true indefinitely because multipooler probes pass | **Wrong.** Probes exist but don't detect the failure. |

### Why this matters

1. **No automatic recovery.** Without a failing liveness probe, Kubernetes never restarts the multipooler. A restart would clear the corrupted pool state and restore service.
2. **No traffic shedding.** Without a failing readiness probe, the gateway Service keeps the pool pod in its endpoints.
3. **Silent failure.** No alerts fire, no dashboards turn red. The cluster is fully broken but invisible to monitoring.

### Recommendations

1. **Fix multipooler `/ready` and `/live` endpoints to reflect connection pool health** (upstream fix). The `/ready` endpoint should exercise the connection pool — ideally a lightweight query like `SELECT 1` through the internal pool, or at minimum check whether the `Status` RPC can complete within a short timeout. A failing liveness probe would trigger a container restart by Kubernetes, which is the correct recovery action.

2. **MultiOrch `/ready` should reflect pooler quorum.** If multiorch cannot reach any poolers for its Status poll, its `/ready` endpoint should return non-200.

3. **Gateway `/ready` should reflect routing availability.** If the gateway has zero healthy pooler backends, it should go not-ready.

## Verification Steps

To confirm this diagnosis on a running cluster:

```bash
# 1. Verify Status hangs while State works
grpcurl -plaintext -d '{}' localhost:15270 multipoolermanager.MultiPoolerManager/State   # Should return instantly
grpcurl -plaintext -d '{}' localhost:15270 multipoolermanager.MultiPoolerManager/Status  # Will hang

# 2. Check pg_stat_activity during the hang
kubectl exec -it <pool-pod> -c postgres -- psql -U postgres -c "SELECT pid, state, query, application_name FROM pg_stat_activity WHERE datname IS NOT NULL"

# 3. Check multiorch logs for poll failures
kubectl logs -l app.kubernetes.io/component=multiorch --tail=200 | grep -E 'poll.*failed|dead|quorum'

# 4. Check multipooler logs for heartbeat errors
kubectl logs <pool-pod> -c multipooler --tail=200 | grep -E 'error|failed|heartbeat'

# 5. To get definitive proof of counter leak (requires adding debug logging upstream):
# - Log pool.active.Load() in getNew() and closedConn()
# - Or expose active counter via /debug/vars or prometheus metrics
```

GuptaManan100:
## Summary

- Add configurable `--connpool-dial-timeout` (default 5s) to bound `net.Dialer.Timeout` for all pool connections
- Add `ConnectTimeout` to `connpool.Config` that bounds background connection attempts (dial + startup) via a `connectionCtx` helper inside `connNew`/`connReopen`
- Fix `Reconnect()` to set `closed=true` when dial fails, so `IsClosed()` returns the correct state

## Problem

The multipooler's connection pool has several background code paths (`put` replacement, max-lifetime reopen, capacity-increase, idle-replacement) that create connections using `pool.ctx` — the pool's lifecycle context with no deadline. Combined with `DialTimeout=0` in the client config, if a dial hangs in any of these paths, the `active` slot is permanently consumed, leading to pool starvation that never self-heals. This causes the `Status` gRPC RPC to hang indefinitely (#694). Fixes #694 

## Solution

Two layers of defense:

1. **`DialTimeout` on the wire**: A new `--connpool-dial-timeout` flag (default 5s) is propagated to `client.Config.DialTimeout`, which sets `net.Dialer.Timeout` for all connections. This bounds the TCP/Unix dial itself.

2. **`ConnectTimeout` on the pool context**: `connpool.Config.ConnectTimeout` (set to `2 × dialTimeout` to cover dial + startup handshake) is applied inside `connNew` and `connReopen` via a `connectionCtx` helper. When a background connection attempt exceeds this timeout, the context is cancelled, the error propagates, and the active slot is released — preventing permanent starvation.

Additionally, `Reconnect()` now correctly sets `closed=true` when dial fails, ensuring `IsClosed()` reflects the actual connection state.


go/services/multipooler/pools/connpool/pool.go
haritabh17: Using the original unbounded `ctx` instead of `connCtx`. When `connReopen` is called from `put()`, `ctx` is `pool.ctx` which has no deadline. If `connect` succeeds but `ApplySettings` hangs (PG accepts connection but blocks on a SET), the active slot is consumed forever — same starvation this PR is fixing.

Should this be `ApplySettings(connCtx, settings)` to bound the entire reconnect (dial + startup + settings application)?

GuptaManan100: Yep 👍

PR #712, opened 2026-03-10
refactor(consensus): rename durability policy ANY_N to AT_LEAST_N

dweitzman:
In postgres sync replication settings, "ANY 2" means a primary applies a transaction after any 2 replica ACKs it, so the transaction would be on 3 nodes (primary + 2 replicas)

In multigres durability policies, a policy of "ANY 2" means a transaction is durable if any 2 nodes have it, so the transaction could be on just 2 nodes (primary + 1 replica)

This overloading of "ANY" is potentially confusing, so we can rename the multigres concept to "AT_LEAST_N" to avoid being too similar to the slightly different concept of "ANY N" in postgres sync replication

The old policy names are temporary still here for backwards compatibility, but we can remove them soon.

### DIFF
--- a/go/pb/clustermetadata/clustermetadata.pb.go
+++ b/go/pb/clustermetadata/clustermetadata.pb.go
@@ -1247,8 +1247,8 @@ type QuorumRule struct {
 	// quorum_type determines which quorum algorithm to use
 	QuorumType QuorumType `protobuf:"varint,1,opt,name=quorum_type,json=quorumType,proto3,enum=clustermetadata.QuorumType" json:"quorum_type,omitempty"`
 	// required_count: number of nodes/cells required
-	//   - For QUORUM_TYPE_AT_LEAST_N (or deprecated QUORUM_TYPE_ANY_N): number of nodes required from discovered cohort
-	//   - For QUORUM_TYPE_MULTI_CELL_AT_LEAST_N (or deprecated QUORUM_TYPE_MULTI_CELL_ANY_N): number of distinct cells required,
+	//   - For QUORUM_TYPE_AT_LEAST_N: number of nodes required from discovered cohort
+	//   - For QUORUM_TYPE_MULTI_CELL_AT_LEAST_N: number of distinct cells required,
 	//     with at least one node from each cell
 	RequiredCount int32 `protobuf:"varint,2,opt,name=required_count,json=requiredCount,proto3" json:"required_count,omitempty"`
 	// Human-readable description
--- a/go/test/endtoend/multiorch/bootstrap_test.go
+++ b/go/test/endtoend/multiorch/bootstrap_test.go
@@ -207,9 +207,9 @@ func TestBootstrapInitialization(t *testing.T) {
 		// Verify Description
 		description, ok := quorumRule["description"].(string)
 		require.True(t, ok, "description should be a string")
-		assert.Equal(t, "Any 2 nodes must acknowledge", description)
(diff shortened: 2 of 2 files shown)


proto/clustermetadata.proto
```diff
@@ -319,12 +319,18 @@ enum QuorumType {
...
   QUORUM_TYPE_UNKNOWN = 0;
 
-  // QUORUM_TYPE_ANY_N requires any N nodes from discovered cohort
+  // Deprecated: use QUORUM_TYPE_AT_LEAST_N
   QUORUM_TYPE_ANY_N = 1;
 
-  // QUORUM_TYPE_MULTI_CELL_ANY_N requires nodes from multiple cells (availability zones)
-  // with at least one node from each of the required cells
+  // Deprecated: use QUORUM_TYPE_MULTI_CELL_AT_LEAST_N
   QUORUM_TYPE_MULTI_CELL_ANY_N = 2;
```
mkindahl, line 326: I know the normal convention is to add new constants rather than renaming existing ones. This is because the protocol encodes these as integers, and for that reason it might happen that an existing deployment read this as a different constant, so it is for backwards compatibility.

However, we have not yet any deployments running, so it would be easier for us to just rename the enumeration constants until we have an actual release out? This will avoid keeping a bunch of "never used for real" constants in the enum and avoid cluttering the code.

proto/clustermetadata.proto
```diff
@@ -319,12 +319,18 @@ enum QuorumType {
...
+
+  // QUORUM_TYPE_AT_LEAST_N requires at least N nodes from the discovered cohort to acknowledge
+  QUORUM_TYPE_AT_LEAST_N = 3;
```
mkindahl, line 329: The traditional term used for database replication is ["n-safety"](https://www.semanticscholar.org/paper/The-dangers-of-replication-and-a-solution-Gray-Helland/14b2986687b0a6140bedcd12429cbb4702f5aea0) (or "n-safe"). (This is because we set `synchronous_commit` to `on`.)

Maybe we should focus on the properties of interest to Multigres and use "n-safe" (since this is what we care about). This would also allow us to distinguish it from "n-visibility", if we ever add that in the future.

proto/clustermetadata.proto
dweitzman: I plan to delete these as soon as the operator is updated, but it's in a different repo

proto/clustermetadata.proto
dweitzman: I don't have strong feelings about the naming here, except to avoid the "any" overloading

One thing I like about "at least" is that it feels simple and approach to people regardless of their background in distributed systems (or lack thereof). My intuition is that "n-safe" or "n-visible" will feel a bit more jargon-y/technical and if people try to search for them they won't get useful results.

mkindahl: I agree that we should not overload the "any" jargon and stick to our own definitions. I am fine with merging this in the current state and take this separately, but let me just clarify my thinking.

Using "n-safe" *is* esoteric in the way you mention it. I think "at least" is better in this sense (not esoteric).
However, then the question is what is it "at least" of? Is it "at least durable on N servers", or does it mean "at least sent to N server", or does it mean "at least visible on N servers"?

If we only use one of these concepts, it is not really a problem since you can just check the documentation, but what if you want to support other concepts so that the user can pick either "at least visible on 2 standbys" or "at least safe on 2 standbys"?

PR #713, opened 2026-03-11
feat(multigateway): add per-table query metrics

haritabh17:
Builds on the OTel instrumentation from #684 to add per-table visibility. During query planning, we now extract table names from the AST and emit a `mg.gateway.query.table_queries` counter so hot tables are easy to spot.

To plumb the table names from executor back to handler, this introduces an `ExecuteResult` struct (defined in the handler package to keep dependency direction clean). For multi-statement batches, table names are aggregated across all statements.

Also adds four Grafana dashboard panels: query duration, query errors, queries per table, and rows returned.

### DIFF
--- a/go/pb/multipoolermanagerdata/multipoolermanagerdata.pb.go
+++ b/go/pb/multipoolermanagerdata/multipoolermanagerdata.pb.go
@@ -1036,9 +1036,12 @@ type SynchronousReplicationConfiguration struct {
 	// Number of synchronous standbys
 	NumSync int32 `protobuf:"varint,3,opt,name=num_sync,json=numSync,proto3" json:"num_sync,omitempty"`
 	// List of standby IDs that can participate in synchronous replication
-	StandbyIds    []*clustermetadata.ID `protobuf:"bytes,4,rep,name=standby_ids,json=standbyIds,proto3" json:"standby_ids,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	StandbyIds []*clustermetadata.ID `protobuf:"bytes,4,rep,name=standby_ids,json=standbyIds,proto3" json:"standby_ids,omitempty"`
+	// Application names for standbys in standby_ids (format: {cell}_{name}).
+	// Parallel to standby_ids: standby_application_names[i] is the application name for standby_ids[i].
+	StandbyApplicationNames []string `protobuf:"bytes,5,rep,name=standby_application_names,json=standbyApplicationNames,proto3" json:"standby_application_names,omitempty"`
+	unknownFields           protoimpl.UnknownFields
+	sizeCache               protoimpl.SizeCache
 }
 
 func (x *SynchronousReplicationConfiguration) Reset() {
@@ -1095,6 +1098,13 @@ func (x *SynchronousReplicationConfiguration) GetNumSync() int32 {
 func (x *SynchronousReplicationConfiguration) GetStandbyIds() []*clustermetadata.ID {
 	if x != nil {
 		return x.StandbyIds
+	}
+	return nil
+}
+
+func (x *SynchronousReplicationConfiguration) GetStandbyApplicationNames() []string {
+	if x != nil {
+		return x.StandbyApplicationNames
 	}
 	return nil
 }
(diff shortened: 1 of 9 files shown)


go/services/multigateway/planner/tables.go
```diff
@@ -0,0 +1,71 @@
...
+	cteNames := make(map[string]struct{})
+	ast.Rewrite(stmt, func(cursor *ast.Cursor) bool {
+		cte, ok := cursor.Node().(*ast.CommonTableExpr)
+		if ok && cte.Ctename != "" {
+			cteNames[cte.Ctename] = struct{}{}
+		}
+		return true
+	}, nil)
+
+	// Second pass: collect real table names from RangeVar nodes.
```
GuptaManan100: Why do we need 2 passes? Can't we do  both in 1 pass?

go/services/multigateway/planner/tables.go
```diff
@@ -0,0 +1,71 @@
...
+package planner
+
+import "github.com/multigres/multigres/go/common/parser/ast"
+
+// ExtractTablesUsed walks the AST and returns deduplicated, schema-qualified
+// table names from all RangeVar nodes. CTE names are excluded since they are
+// virtual tables, not real ones. Returns nil for statements that don't
+// reference tables (SET, SHOW, BEGIN, etc.).
+func ExtractTablesUsed(stmt ast.Stmt) []string {
+	if stmt == nil {
```
GuptaManan100: This function seems better lived in the ast package, since it relies purely on ast structs.

go/services/multigateway/handler/transaction_helpers.go
```diff
@@ -65,19 +67,37 @@ func (h *MultiGatewayHandler) executeWithImplicitTransaction(
...
+		if result == nil {
+			return
+		}
+		for _, t := range result.TablesUsed {
+			if _, exists := seen[t]; !exists {
+				seen[t] = struct{}{}
+				allTables = append(allTables, t)
+			}
+		}
+	}
```
GuptaManan100: Why are we collecting all the tables for the entire multi statement? Shouldn't we keep the metrics per query?

go/services/multigateway/handler/metrics.go
```diff
@@ -108,6 +109,26 @@ func (m RowsReturned) Record(
...
+func (m TableQueries) Add(
+	ctx context.Context,
+	dbNamespace string,
+	tableName string,
+	operationName string,
+) {
+	m.Int64Counter.Add(ctx, 1,
+		metric.WithAttributes(
+			attribute.String("db.namespace", dbNamespace),
+			attribute.String("db.collection.name", tableName),
```
GuptaManan100, line 127: With table name as an attribute, this could create unbounded cardinality on the metric in multi-tenant environments or with dynamic table names (e.g., temp_<uuid>). Should we bound to some reasonable number of tables? Is it possible to do that in metrics?

go/services/multigateway/handler/handler.go
```diff
@@ -277,7 +290,7 @@ func (h *MultiGatewayHandler) HandleExecute(ctx context.Context, conn *server.Co
 		portalErr := mterrors.NewPgError("ERROR", mterrors.PgSSInvalidCursorName,
 			fmt.Sprintf("portal \"%s\" does not exist", portalName), "")
 		// Record before span creation since we don't have an operation name yet.
-		h.recordQueryCompletion(ctx, conn, "UNKNOWN", "extended", 0, 0, time.Since(queryStart), 0, portalErr)
+		h.recordQueryCompletion(ctx, conn, "UNKNOWN", "extended", 0, 0, time.Since(queryStart), 0, nil, portalErr)
```
GuptaManan100, line 294: We should collect the table names from the AST in the portal info, otherwise we won't get metrics from queries using the extended query protocol

demo/observability/grafana-dashboard.json
```diff
@@ -197,6 +197,183 @@
...
+      "id": 51,
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "Prometheus"
+          },
+          "editorMode": "code",
+          "exemplar": true,
+          "expr": "histogram_quantile(0.95, sum by (db_operation_name) (rate(mg_gateway_query_duration_seconds{job=\"multigateway\"}[$__rate_interval])))",
```
GuptaManan100, line 232: Did we test this? Claude seems to think its wrong with two problems:
1. Missing le label in sum by - histogram_quantile requires grouping by le (the bucket boundary label)
2. Missing _bucket suffix - Prometheus histograms expose _bucket suffixed metrics; the query should target mg_gateway_query_duration_seconds_bucket

I don't know whether its right, but let's test it to make sure it works

haritabh17: Tested end-to-end with the OTel collector + Prometheus stack — the histogram panels render correctly.

Image

go/services/multigateway/planner/tables.go
haritabh17: Ack, rewrote it as a single pass that collects both CTE names and table refs together, then filters out CTE references after the walk.

go/services/multigateway/planner/tables.go
haritabh17: Agreed, moved to ast.ExtractTablesUsed() in helper.go alongside the other statement helpers.

go/services/multigateway/handler/transaction_helpers.go
haritabh17: Ack, changed to emit per-statement with the correct db.operation.name.

go/services/multigateway/handler/handler.go
haritabh17: Good catch — PortalStreamExecute now extracts tables from the portal's AST even when there's no local plan, so extended protocol queries get table metrics too.

go/services/multigateway/handler/metrics.go
haritabh17: Thinking that this may not be the right layer to handle cardinality bounding — it's better done at the OTel collector level (e.g. filter/transform processors) where operators can configure what they need. Keeps the instrumentation simple.

go/common/parser/goyacc/goyacc.go
```diff
@@ -1,44 +1,42 @@
...
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-THE SOFTWARE.
-*/
+// Derived from Inferno's utils/iyacc/yacc.c
```
haritabh17, line 1: The pre-commit hook requires // style license comments for Go files, and this file came in with /* */ format during the merge. Converted to match the hook's expectation. No functional change.

PR #716, opened 2026-03-11
ci(workflows): upload test log artifacts for offline analysis

mkindahl:
Add an upload-artifact step to the integration, short, and race detection test workflows so that the raw JSONL test logs can be downloaded and analyzed offline after a CI run.


mkindahl:
> Thanks for doing this. Let's try this out. I'm not clear on whether we should also upload the results of pg-regress, but that can be a separate PR

Good point. I'll upload this right now since these logs are the post problematic ones.

PR #724, opened 2026-03-12
feat: use PGDATA env var as source of truth for PostgreSQL data directory

mkindahl:
- Add `PgDataDirEnvVar` and `ConsensusTermFile` constants; replace all hardcoded strings
- Add `pg_data_dir` field to MultiPooler proto for `multiadmin` backup path
- Require `PGDATA` to be set at `pgctld` and multipooler startup
- Multigres-specific marker files move to `$PGDATA/multigres/`
- Add `PGDATA` to k8s manifest and local provisioner subprocess env
- Fix e2e tests: inject `PGDATA` into subprocess environments and correct pg_data subdirectory paths
- Rename `ProcessInstance.DataDir` to `PoolerDir` in e2e test infrastructure

poolerDir is still required for three purposes unrelated to PostgreSQL data:
- Unix socket directory (poolerDir/pg_sockets)
- pgBackRest operational files (server config, client config, logs) under poolerDir/pgbackrest/

Fixes MUL-228

### DIFF
--- a/go/common/constants/postgres.go
+++ b/go/common/constants/postgres.go
@@ -34,12 +34,22 @@ const (
 	// PgDatabaseEnvVar is the environment variable for the PostgreSQL database name.
 	PgDatabaseEnvVar = "POSTGRES_DB"
 
+	// PgDataDirEnvVar is the environment variable for the PostgreSQL data directory.
+	PgDataDirEnvVar = "PGDATA"
+
 	// DefaultPostgresDatabase is the default database that always exists in PostgreSQL.
 	// This database is created during cluster initialization.
 	DefaultPostgresDatabase = "postgres"
 
 	// PostgresExecutable is the name of the PostgreSQL server binary.
 	PostgresExecutable = "postgres"
+
+	// MultigresMarkerDirectory is the name of the directory used by pgctld to
+	// mark a PostgreSQL data directory as managed by pgctld. This is also where
+	// all marker files are stored, such as the file indicating that the cluster
+	// is in the process of being initialized. This directory is created inside
+	// the PostgreSQL data directory.
+	MultigresMarkerDirectory = "multigres"
 
 	// DefaultSlowQueryThreshold is the duration after which a query is logged at WARN level.
 	DefaultSlowQueryThreshold = 1 * time.Second
--- a/go/services/multipooler/grpcconsensusservice/service_test.go
+++ b/go/services/multipooler/grpcconsensusservice/service_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"log/slog"
 	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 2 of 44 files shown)


go/common/constants/postgres.go
```diff
@@ -34,6 +34,9 @@ const (
 	// PgDatabaseEnvVar is the environment variable for the PostgreSQL database name.
 	PgDatabaseEnvVar = "POSTGRES_DB"
 
+	// PgDataDirEnvVar is the environment variable for the PostgreSQL data directory.
+	PgDataDirEnvVar = "PGDATA"
+
```
mkindahl, line 39: This is added as a constant, but I am not entirely sure if this is needed. After all, the name of PGDATA will not change that often, so there is less reason to keep this as a global constant.

go/services/multipooler/grpcconsensusservice/service_test.go
```diff
@@ -98,7 +98,9 @@ func TestConsensusService_BeginTerm(t *testing.T) {
 	pgDataDir := tmpDir + "/pg_data"
 	require.NoError(t, os.MkdirAll(pgDataDir, 0o755))
 	require.NoError(t, os.WriteFile(pgDataDir+"/PG_VERSION", []byte("16\n"), 0o644))
-	require.NoError(t, os.WriteFile(pgDataDir+"/MULTIGRES_INITIALIZED", []byte("initialized\n"), 0o644))
+	t.Setenv(constants.PgDataDirEnvVar, pgDataDir)
+	require.NoError(t, os.MkdirAll(pgDataDir+"/multigres", 0o755))
```
mkindahl: We should probably have a global constant for the marker directory though.

mkindahl: Done.

go/services/multipooler/manager/term_storage.go
```diff
@@ -21,31 +21,35 @@ import (
...
 	_, err := os.Stat(pgVersionFile)
 	return err == nil
 }
 
 // consensusTermPath returns the path to the consensus term file
-func consensusTermPath(poolerDir string) string {
-	dataDir := postgresDataDir(poolerDir)
-	return filepath.Join(dataDir, "consensus", "consensus_term.json")
+func consensusTermPath() string {
+	return filepath.Join(multigresDataDir(), "consensus_term.json")
```
dweitzman: This seems like an unintended change: `consensus_term.json` was previously at `poolerdir/consensus/consensus_data.json`. I think this code is changing it to be at `pgdatadir/multigres/consensus_term.json`?...

Consensus term is specific to the pooler rather than the shard, so it doesn't seem like it belongs in pgdata directory or in backups

I think the main misleading thing here is `multigresDataDir()`. In reality there are probably two multigres data directories: one under pgdata (that becomes part of a backups) and one _not_ under pgdata the survives multipooler restarts but not a change of pooler iD

mkindahl: Looking at the original code, the file was, AFAICT, placed in poolerDir/pg_data/consensus/consensus_data.json, so it would have been backed up.

Note that in the original code, we had

- dataDir := postgresDataDir(poolerDir) --> dataDir = poolerDir/pg_data
- filepath.Join(dataDir, "consensus", "consensus_term.json") --> poolerDir/pg_data/consensus/consensus_term.json

Which is why I kept the consensus file in the data directory that is backed up (PGDATA)

I am not sure why we should or should not back up the consensus term information, it depends a little on what we want to do with it after a restore. Backing it up means that we know what term the backup was taken for, which we might use to figure out what changes that we need to process if we restore the backup on a machine. If we have other means to figure out the current term and recover data from old terms (e.g., an LSN), this might not be needed.

Looking at the code, it seems like the consensus term is loaded when multipooler manager starts, and then promptly overwritten. I see no reason to take a backup of the file suggesting, as you say, that we should not place it in the PostgreSQL data directory.

I'll move it elsewhere under the poolerDir, similar to, e.g., pg_sockets.

mkindahl: Done, and update the PR description.

mkindahl: Seems like there are tests and code that assume that you include the consensus term file in the backup. I moved this to commit #749 so we can take it from there.

PR #730, opened 2026-03-12
refactor(multiorch): put PoolerHealthStore into PoolerStore

dweitzman:
The context here is that we may want to store per-pooler information other than health data. `recentPollCache` is one example of a map that exists directly on the engine. I'm thinking of introducing more similar behavior, like throttling how often we log that a pooler is unreachable, etc.

To me it feels natural to have PoolerStore be responsible for the full lifetime of knowing things about poolers rather than having some of that knowledge exist as separate maps directly on the engine.

It makes cleaning up old data easier, for example: if we forget that a pooler exists, it'll be easier to forget all the non-health data we may have saved for that pooler also.

### DIFF
--- a/go/services/multiorch/recovery/discovery_test.go
+++ b/go/services/multiorch/recovery/discovery_test.go
@@ -484,7 +484,7 @@ func TestPoolerWatcher_DirectDiscovery(t *testing.T) {
 	defer ts.Close()
 
 	cfg := config.NewTestConfig(config.WithCell("zone1"))
-	poolerStore := store.NewPoolerHealthStore()
+	poolerStore := store.NewPoolerStore(nil, slog.Default())
 	queue := NewQueue(slog.Default(), cfg)
 	watchTargets := []config.WatchTarget{{Database: "mydb", TableGroup: "tg1"}}
 	poolerWatcher := NewPoolerWatcher(ctx, ts, func() []config.WatchTarget { return watchTargets }, poolerStore, queue, slog.Default())
--- a/go/services/multiorch/recovery/pooler_watcher.go
+++ b/go/services/multiorch/recovery/pooler_watcher.go
@@ -44,7 +44,7 @@ import (
 type PoolerWatcher struct {
 	topoStore topoclient.Store
 	targets   func() []config.WatchTarget // live accessor, same as Engine.shardWatchTargets
-	store     *store.PoolerHealthStore
+	store     *store.PoolerStore
 	queue     *Queue
 	logger    *slog.Logger
 
@@ -64,7 +64,7 @@ func NewPoolerWatcher(
 	ctx context.Context,
 	topoStore topoclient.Store,
 	targets func() []config.WatchTarget,
-	poolerStore *store.PoolerHealthStore,
+	poolerStore *store.PoolerStore,
 	queue *Queue,
 	logger *slog.Logger,
 ) *PoolerWatcher {
@@ -248,7 +248,7 @@ type cellPoolerWatcher struct {
 	topoStore topoclient.Store
 	cell      string
 	targets   func() []config.WatchTarget
-	store     *store.PoolerHealthStore
+	store     *store.PoolerStore
 	queue     *Queue
 	logger    *slog.Logger
 
@@ -263,7 +263,7 @@ func newCellPoolerWatcher(
 	topoStore topoclient.Store,
 	cell string,
 	targets func() []config.WatchTarget,
-	poolerStore *store.PoolerHealthStore,
+	poolerStore *store.PoolerStore,
 	queue *Queue,
 	logger *slog.Logger,
 ) *cellPoolerWatcher {
--- a/go/services/multiorch/recovery/pooler_watcher_test.go
+++ b/go/services/multiorch/recovery/pooler_watcher_test.go
@@ -27,7 +27,6 @@ import (
 	"github.com/multigres/multigres/go/common/topoclient"
 	"github.com/multigres/multigres/go/common/topoclient/memorytopo"
 	"github.com/multigres/multigres/go/pb/clustermetadata"
-	multiorchdatapb "github.com/multigres/multigres/go/pb/multiorchdata"
 	"github.com/multigres/multigres/go/services/multiorch/config"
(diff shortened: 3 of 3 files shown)


go/services/multiorch/store/pooler_store.go
```diff
@@ -28,38 +28,64 @@ import (
...
-// NewPoolerStore creates a new PoolerStore wrapper.
-func NewPoolerStore(
-	store *ProtoStore[string, *multiorchdatapb.PoolerHealthState],
-	rpcClient rpcclient.MultiPoolerClient,
-	logger *slog.Logger,
-) *PoolerStore {
+// NewPoolerStore creates a new PoolerStore.
+// rpcClient and logger are used by FindHealthyPrimary; they may be nil in tests
+// that do not exercise that method.
+func NewPoolerStore(rpcClient rpcclient.MultiPoolerClient, logger *slog.Logger) *PoolerStore {
```
dweitzman, line 43: The essence of this refactoring is making PoolerStore the future home for all pooler-specific information. For example, in a follow-up PR I would move `recentPollCache` from being directly on the engine to being part of the PoolerStore

dweitzman: For context, here's what the move of `recentPollCache` will look like: https://github.com/multigres/multigres/commit/ca9db3076aded4bf477d8e54f82a8b806ea254b7

It'll also simplify the code because `recentPollCache` won't be trying to guess when to delete entries, the lifetime of recent poll observations for a pooler will exactly equal the lifetime of health records for that pooler

go/services/multiorch/recovery/actions/bootstrap_shard_test.go
```diff
@@ -81,10 +81,10 @@ func TestBootstrapShardAction_ExecuteNoCohort(t *testing.T) {
 	defer ts.Close()
 
 	logger := slog.Default()
-	poolerStore := store.NewProtoStore[string, *multiorchdatapb.PoolerHealthState]()
+	ps := store.NewPoolerStore(nil, nil)
 	coord := newTestCoordinator(ts, nil, logger)
 
-	action := NewBootstrapShardAction(nil, nil, poolerStore, ts, coord, logger)
+	action := NewBootstrapShardAction(nil, nil, ps.Health(), ts, coord, logger)
```
mkindahl, line 87: Is there a particular reason to why you pass in the PoolerHealthStore rather than the PoolerStore? If you want to collect all information about the pooler into the PoolerStore (which I think is a good idea), it would make sense to hide the internal health store entirely and just pass around the PoolerStore instance.

go/services/multiorch/recovery/actions/bootstrap_shard_test.go
```diff
@@ -144,15 +144,15 @@ func TestBootstrapShardAction_ConcurrentExecutionPrevented(t *testing.T) {
 	defer ts.Close()
 
 	logger := slog.Default()
-	poolerStore := store.NewProtoStore[string, *multiorchdatapb.PoolerHealthState]()
+	ps := store.NewPoolerStore(nil, nil)
```
mkindahl: You're creating a logger just before this line. I was expecting it to be passed in to the pooler store. I think this is benign for this test, but I am wondering if there is a good reason to not pass it in?

```suggestion
	ps := store.NewPoolerStore(nil, logger)
```

go/services/multiorch/recovery/actions/bootstrap_shard_test.go
```diff
@@ -232,7 +232,7 @@ func TestBootstrapShardAction_ConfiguresSyncReplication(t *testing.T) {
...
```
mkindahl: Might be a good idea to pass in the logger here unless there is a particular reason not to.

```suggestion
	ps := store.NewPoolerStore(nil, logger)
```

go/services/multiorch/recovery/actions/bootstrap_shard_test.go
```diff
@@ -414,7 +414,7 @@ func TestBootstrapShardAction_QuorumCheckFailsWithInsufficientPoolers(t *testing
...
```
mkindahl:
```suggestion
	ps := store.NewPoolerStore(nil, logger)
```

go/services/multiorch/recovery/actions/bootstrap_shard_test.go
```diff
@@ -490,7 +490,7 @@ func TestBootstrapShardAction_QuorumCheckPassesWithEnoughPoolers(t *testing.T) {
...
```
mkindahl:
```suggestion
	ps := store.NewPoolerStore(nil, logger)
```

go/services/multiorch/recovery/actions/bootstrap_shard_test.go
```diff
@@ -587,7 +587,7 @@ func TestBootstrapShardAction_FullBootstrapFlow(t *testing.T) {
...
```
mkindahl:
```suggestion
	ps := store.NewPoolerStore(nil, logger)
```

go/services/multiorch/recovery/actions/bootstrap_shard_test.go
```diff
@@ -675,7 +675,7 @@ func TestBootstrapShardAction_SkipsIfAlreadyInitialized(t *testing.T) {
...
```
mkindahl:
```suggestion
	ps := store.NewPoolerStore(nil, logger)
```

go/services/multiorch/recovery/actions/bootstrap_shard_test.go
dweitzman: The only real reason here was to minimize the diff in the PR: a lot of code is already taking something that look like a health store and if I keep `.Health()` and pass it to places that previously took the health-only poolerStore it makes this refactoring more minimal and the diff smaller

I'm fine with dropping the .Health(), but maybe it'd make sense to do in a follow-up PR to keep this one more focused on the real goal of the refactoring?

dweitzman:
I'll merge as-is and we can follow up with removing Health() in another PR, to keep this focused

PR #733, opened 2026-03-13
fix(multipooler): follow consensus rules when updating synchronous standby names

dweitzman:
Previously: WAL was always updated before GUC
Now: the more restrictive of the before/after policy is applied during the WAL write

This PR also removes the "REPLACE" command to update the sync standby list. It's easier to make the caller split up adds vs removes than to try to have the primary deal with both simultaneously


go/services/multipooler/manager/rpc_manager.go
```diff
@@ -560,33 +558,59 @@ func (pm *MultiPoolerManager) UpdateSynchronousStandbyList(ctx context.Context,
...
+	if operation == multipoolermanagerdatapb.StandbyUpdateOperation_STANDBY_UPDATE_OPERATION_REMOVE {
+		if err = applyGUC(); err != nil {
+			return err
+		}
+		if err = maybeReload(); err != nil {
+			return err
+		}
+		if err = insertHistoryRecord(); err != nil {
+			return mterrors.Wrap(err, "failed to record replication config history")
+		}
```
mkindahl, line 605: If the insert fails, do you need to undo the modifications of the GUC above? I don't think so since this is just a log of decisions already made, but wanted to check.

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -560,33 +558,59 @@ func (pm *MultiPoolerManager) UpdateSynchronousStandbyList(ctx context.Context,
...
+	} else {
+		if err = insertHistoryRecord(); err != nil {
+			return mterrors.Wrap(err, "failed to record replication config history")
+		}
+		if err = applyGUC(); err != nil {
```
mkindahl, line 610: If the applyGUC fails here, the record added to the WAL is not correct. Can this lead to problems?

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -560,33 +558,59 @@ func (pm *MultiPoolerManager) UpdateSynchronousStandbyList(ctx context.Context,
...
+	// but the WAL record has not been written yet. A long-term fix would track the term at
+	// which each GUC update was last applied and have the postgres monitor poll
+	// leadership_history periodically to re-apply any GUC updates it missed.
+	if operation == multipoolermanagerdatapb.StandbyUpdateOperation_STANDBY_UPDATE_OPERATION_REMOVE {
+		if err = applyGUC(); err != nil {
+			return err
+		}
+		if err = maybeReload(); err != nil {
+			return err
+		}
```
mkindahl, line 602: If pg_reload_conf() is not called for any reason, the insert below might fail if some server that should be removed is not responsive.

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -560,33 +558,59 @@ func (pm *MultiPoolerManager) UpdateSynchronousStandbyList(ctx context.Context,
...
+		if err = insertHistoryRecord(); err != nil {
```
mkindahl, line 603: The configuration is not immediately reloaded, so this statement may be executed with the old set of standby names, which could lead to problems if some server is not responsive.

mkindahl (commented):
I think it is good to remove the "replace" functionality. It is a lot easier to reason with the changes if there are only add and remove actions.

We have two potential problems here, but they were present the patch.

1. If `pg_reload_conf()` is not called, the changes to `synchronous_standby_names` will not be used for the insert below update of the GUC, which might cause it to never return if we have unresponsive servers in the list.
2. After the call to `pg_reload_conf()` the reconfigure of `synchronous_standby_names` is not instantaneous. The function just sends a `SIGHUP` and then it might take while for it to be reloaded, depending on the load of the system.

I think we need to:
1. Remove the `ReloadConfig` flag and just not use it. We should always call `pg_reload_conf()`. It seems like the flag is never used, not even in tests, so I think it is safe to just remove it.
2. After the call to `pg_reload_conf()` there needs to be a wait for it to actually be reloaded. This wait can be added to `applySynchronousStandbyNames` by using `pg_conf_reload_time()` and wait until it increases.

dweitzman:
Closing this old PR since I'll have a newer PR soon that accomplishes the same goals

PR #734, opened 2026-03-14
feat(multipooler): strengthen typing for replication IDs and application names

dweitzman:
- Change coordinator_id, cohort_members, and accepted_members in proto to use clustermetadata.ID instead of string for stronger protocol typing
- Introduce applicationName type as a validated wrapper around PostgreSQL application_name strings
- Validate application names at construction time: reject nil IDs, empty fields, underscores, and names exceeding PostgreSQL's 63-character limit (NAMEDATALEN - 1), returning an error rather than letting Postgres silently truncate the name

### DIFF
--- a/go/common/topoclient/database_test.go
+++ b/go/common/topoclient/database_test.go
@@ -20,13 +20,13 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	"google.golang.org/protobuf/proto"
 
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 
 	"github.com/multigres/multigres/go/common/topoclient"
 	"github.com/multigres/multigres/go/common/topoclient/memorytopo"
 	"github.com/multigres/multigres/go/test/utils"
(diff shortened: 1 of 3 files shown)


go/services/multipooler/manager/pg_replication.go
```diff
@@ -47,22 +47,76 @@ import (
...
+	// in the application_name format (cell_name). Allowing underscores would break parsing.
+	if strings.Contains(id.Cell, "_") {
+		return "", mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT,
+			"cell contains underscore: %q (underscores not allowed)", id.Cell)
+	}
+	if strings.Contains(id.Name, "_") {
+		return "", mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT,
+			"name contains underscore: %q (underscores not allowed)", id.Name)
+	}
+	name := fmt.Sprintf("%s_%s", id.Cell, id.Name)
```
dweitzman, line 82: This check is the new validation. Everything else in the PR is intended to be pure refactoring

PR #737, opened 2026-03-16
ci(coverage): fix flaky Coveralls parallel upload and check permissions

mkindahl:
The `dorny/test-reporter need to add report to the check runs, hence require  `checks: write` privileges, while `action/checkout` needs `contents:read` privileges to work reliably.

The job `coveralls-finish` used `if: success()`, so any upstream job failure  would skip it. As a result, Coverall would keep current partial runs uploaded, and on the next run these old runs would interfere with data from the new runs. To handle this, use `if: always()` to trigger the finish signal to always be sent.

However, the workflow permissons were `permissions: {}`, which prevented dorny/test-reporter from creating check runs (since this requires `checks:  write`), causing coverage jobs to fail even when tests passed. Fixed by adding `checks: write` and `contents: read` permissions to the workflow.

Also remove some whitespace to make "prettier" YAML file checker happy and permissions are on a per-job basis to make `zizmor` GitHub workflow linter happy.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -11,41 +11,34 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 name: Code Coverage
-
 on:
   push:
     branches: [main]
   pull_request:
     branches: [main]
-
 permissions: {}
-
 jobs:
   coverage-direct:
     name: Direct test coverage
     runs-on: ubuntu-24.04
-
+    permissions:
+      checks: write
+      contents: read
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
-
       - name: Set up Go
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -77,3 +77,6 @@ demo/k8s/data/
 # Generated PR description
 pr-description.md
 
+# pgBackRest working directory created during integration tests
+go/services/multipooler/grpcmanagerservice/pgbackrest/
```
dweitzman: I think the underly issue here is that the integration tests shouldn't be writing into the source directory. I believe it's happening due to some misconfiguration where the actual test temp directory isn't be passed to whatever writes the pgbackrest config and it's assuming the current working directory instead

cuongdo: This is bizarre. This directory doesn't exist in my local repo. I'll try to figure out how this is happening.

mkindahl: File was added accidentally because I added the entire directory

> I think the underly issue here is that the integration tests shouldn't be writing into the source directory. I believe it's happening due to some misconfiguration where the actual test temp directory isn't be passed to whatever writes the pgbackrest config and it's assuming the current working directory instead

I agree on this. We should use a "build" directory or similar for temporary or generated files. I'll remove this from the `.gitignore` and we'll deal with it separately.

mkindahl: Done.

PR #738, opened 2026-03-16
feat(multipooler): track active postgres action in status RPC

dweitzman:
Adds a PostgresAction enum (STARTING, RESTORING_FROM_BACKUP) and duration to the multipooler Status response.

Surfaces this in bootstrap status logging so end-to-end test failures stuck during shard setup show whether a node is actively restoring from backup (and for how long) vs not doing anything at all.

Motivation: https://github.com/multigres/multigres/actions/runs/23158744659?pr=730 showed a TestPoolerDownNoFailover flake where a replica stayed REPLICA/down for the full 60s bootstrap timeout with no indication of why. This change makes the SUMMARY line for each pooler show something like this, making it immediately clear if a restore-from-backup is still in progress:

```
main_test.go:25: checkBootstrapStatus: SUMMARY primary=pooler-1 initialized=1/3 latest_backup="20260316-151841F" [pooler-1: queryable, type=PRIMARY, sync_replication_configured (standbys=2/2) | pooler-2: not_queryable, action=POSTGRES_ACTION_RESTORING_FROM_BACKUP(0s) | pooler-3: not_queryable]
main_test.go:25: checkBootstrapStatus: SUMMARY primary=pooler-1 initialized=1/3 latest_backup="20260316-151841F" [pooler-1: queryable, type=PRIMARY, sync_replication_configured (standbys=2/2) | pooler-2: queryable, type=REPLICA, primary_conn_info_waiting, action=POSTGRES_ACTION_RESTORING_FROM_BACKUP(1s) | pooler-3: not_queryable, action=POSTGRES_ACTION_RESTORING_FROM_BACKUP(0s)]
main_test.go:25: checkBootstrapStatus: SUMMARY primary=pooler-1 initialized=1/3 latest_backup="20260316-151841F" [pooler-3: not_queryable, action=POSTGRES_ACTION_RESTORING_FROM_BACKUP(1s) | pooler-1: queryable, type=PRIMARY, sync_replication_configured (standbys=2/2) | pooler-2: queryable, type=REPLICA, primary_conn_info_waiting]
```

It's still technically possible that restore attempts are happening frequently but fail so quickly that status is always reported between restore attempts instead of while one is active. In production we'd catch that with metrics on restore attempt frequency, but for endtoend tests I'm not attempting to log anything that would catch that scenarion in this PR


go/services/multipooler/manager/manager.go
cuongdo, line 1704: should we clear the action here, since the restore from backup has completed by this point?

dweitzman: The action will auto-clear when the action lock is released

At the moment that means it'll release almost immediately after `takeRemedialAction()` returns, since we don't currently try to take multiple remedial actions in a single run of the postgres monitor loop

My feeling is this is probably fine and if/when we decide to run multiple remedial actions during a single run of the postgres monitor we could revisit whether an explicit clear is useful

PR #739, opened 2026-03-16
refactor(multiorch): remove Health() accessor, callers use PoolerStor…e directly

dweitzman:
All components that previously received *PoolerHealthStore via Health() now take *PoolerStore instead. PoolerHealthStore is made unexported as a pure implementation detail.

### DIFF
--- a/go/services/multiorch/store/pooler_health_state.go
+++ b/go/services/multiorch/store/pooler_health_state.go
@@ -54,7 +54,7 @@ func (s *poolerHealthStore) len() int {
 
 // range iterates over all poolers. Each value passed to the callback is a deep
 // clone safe to mutate. Iteration stops early if the callback returns false.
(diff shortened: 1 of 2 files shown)


go/services/multiorch/store/pooler_health_state.go
```diff
@@ -18,46 +18,43 @@ import (
...
-func (s *PoolerHealthStore) Len() int {
+// len returns the number of poolers in the store.
+func (s *poolerHealthStore) len() int {
 	return s.proto.Len()
 }
 
-// Range iterates over all poolers. Each value passed to the callback is a deep
+// range iterates over all poolers. Each value passed to the callback is a deep
 // clone safe to mutate. Iteration stops early if the callback returns false.
-func (s *PoolerHealthStore) Range(fn func(key string, value *multiorchdata.PoolerHealthState) bool) {
```
dweitzman, line 60: Unexported/lowercase "range" would conflict with the Go keyword, so it's slightly renamed

PR #740, opened 2026-03-17
feat(pgctld): support POSTGRES_INITDB_ARGS for extra initdb arguments

mkindahl:
Add support for passing extra arguments to `initdb` via the `--pg-initdb-args` flag or the `POSTGRES_INITDB_ARGS` environment variable. The flag takes precedence over the environment variable when both are set.

`POSTGRES_INITDB_ARGS` is the standard Docker/supabase convention for supplying `initdb` options such as ICU locale settings at cluster initialization time. Supporting it lets `pgctld` be a drop-in replacement in environments that already set this variable, without requiring any change to how those environments are configured.

The argument string is split on whitespace with strings.Fields before being forwarded to `initdb`, matching the shell-splitting semantics that Docker applies to the same variable.

The pull request also includes a refactoring: The four commonly-repeated parameters `pgPort`, `pgUser`, `pgDatabase`, and `pgPassword` have been grouped into a `PgCtldServiceConfig` struct, reducing argument-list length on `NewPgCtldService`, `InitDataDirWithResult`, and `initializeDataDir`. Within `PgCtldService`, the two config fields are now named `ctldConfig` (holding `PgCtldServiceConfig`) and `pgConfig` (holding `*pgctld.PostgresCtlConfig`) to make their roles unambiguous. A package-level `testServiceConfig` variable eliminates repeated struct literals across thirteen test call sites.

Fixes MUL-248

### DIFF
--- a/go/cmd/pgctld/command/init.go
+++ b/go/cmd/pgctld/command/init.go
@@ -61,12 +61,18 @@ file and environment variable settings.
 
 Password can be set via the POSTGRES_PASSWORD environment variable.
 
+Extra initdb arguments can be set via the POSTGRES_INITDB_ARGS environment variable,
+or overridden with the --pg-initdb-args flag.
+
 Examples:
   # Initialize data directory
   pgctld init --pooler-dir /var/lib/pooler-dir
 
   # Initialize with existing configuration
   pgctld init -d /var/lib/pooler-dir
+
+  # Initialize with ICU locale provider and specific locale en_US.UTF-8
+  pgctld init --pg-initdb-args "--locale=icu --icu-locale=en_US.UTF-8" -d /var/lib/pooler-dir
 
   # Initialize using config file settings
   pgctld init --config-file /etc/pgctld/config.yaml`,
--- a/config/README.md
+++ b/config/README.md
@@ -58,6 +58,31 @@ Cluster orchestration service for consensus and failover.
 **Configuration file**: `multiorch.yaml`
 **Environment prefix**: `MULTIORCH_`
 
+## PostgreSQL Environment Variables
+
+The following environment variables are recognized by `pgctld` (and to
+some extent, Multipooler) and follow the Docker `postgres` image
+convention.
+
+| Variable               | CLI flag               | Default    | Description                           |
+| ---------------------- | ---------------------- | ---------- | ------------------------------------- |
+| `POSTGRES_USER`        | `--pg-user` / `-U`     | `postgres` | PostgreSQL user name                  |
+| `POSTGRES_PASSWORD`    | _(none)_               | _(empty)_  | PostgreSQL password                   |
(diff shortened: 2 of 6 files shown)


go/cmd/pgctld/command/init.go
```diff
@@ -213,9 +226,14 @@ func initializeDataDir(logger *slog.Logger, poolerDir string, pgUser string, pgP
...
 		args = append(args, "--pwfile="+tempPwFile)
-		logger.Info("Setting password during initdb", "user", pgUser, "password_source", "POSTGRES_PASSWORD environment variable")
+		logger.Info("Setting password during initdb", "user", cfg.User, "password_source", "POSTGRES_PASSWORD environment variable")
 	} else {
-		logger.Warn("No password provided - skipping password setup", "user", pgUser, "warning", "PostgreSQL user will not have password authentication enabled")
+		logger.Warn("No password provided - skipping password setup", "user", cfg.User, "warning", "PostgreSQL user will not have password authentication enabled")
+	}
+
+	if cfg.InitdbArgs != "" {
+		args = append(args, strings.Fields(cfg.InitdbArgs)...)
```
dweitzman, line 234: Does this need to deal with quoted strings when splitting fields?

go/multipooler/grpcmanagerservice/pgbackrest/pgbackrest.conf
dweitzman, line 1: Some of these config files seem unrelated?... I think it can happen if the tests don't have the right variables set for what temp directory to operate in and end up writing config to the directory the test code is in instead

cuongdo: I am also confused about these config files being in this PR. We typically render `pgbackrest_template.conf`. `repo1-path` and `repo1-type` can be passed in as the command-line parameters `--repo1-path` and `--repo1-type`, respectively.

mkindahl: They were added by accident. Removed them.

go/cmd/pgctld/command/init.go
mkindahl: There is no need for dealing with quotes for the `initdb` flags that we are using. In general, the flags provided to `initdb` does not contain spaces (with the exception of `--pwfile`, which can contain spaces).

As a safety measure we can parse the string and deal with quotes, but since these flags are only passed through `POSTGRES_INITDB_ARGS` there should be no way to exploit this.

cuongdo: This seems correct for now. In the future, we may need to consider. For example, the user may want `--waldir="/tmp/WAL Archive"`.

We can punt on this for now though.

mkindahl: I added documentation to make the current situation clear.

PR #743, opened 2026-03-17
test(shardsetup): add EventuallyPoolerCondition for endtoend tests

dweitzman:
There are lot of places in endtoend tests that are waiting for pooler statuses to get into a certain state, and not all of those places print rich debug information to make it obvious on failure what incorrect state the poolers are in

This change tries to start unifying on a central helper for waiting on pooler status that prints a fair amount of detail about each pooler as time is passing but the goal cluster state hasn't been reached, so you can see at a glance if a test failed because postgres wasn't running, because it had the wrong replication settings, etc.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,3 @@
+[codespell]
+skip = */go/common/parser/*
+ignore-words-list = copys,usera,anull
--- a/.github/linters/.codespellrc
+++ b/.github/linters/.codespellrc
@@ -1,0 +1,1 @@
+../../.codespellrc
--- a/.github/super-linter.env
+++ b/.github/super-linter.env
@@ -1,5 +1,5 @@
 DEFAULT_BRANCH=main
-FILTER_REGEX_EXCLUDE=(external/.*)|(\.env$)|(demo/k8s/.*)
+FILTER_REGEX_EXCLUDE=(external/.*)|(\.env$)|(demo/k8s/.*)|(.*\.tsbuildinfo$)|(.*\.pb\.go$)|(.*\.svg$)|(go\.sum$)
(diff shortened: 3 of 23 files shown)


go/test/endtoend/multiorch/multiorch_helpers.go
```diff
@@ -49,111 +47,64 @@ func connectToPostgres(t *testing.T, socketDir string, port int) *sql.DB {
...
+		func(statuses []shardsetup.PoolerStatusResult) (any, bool, string) {
+			standbyCount := 0
+			syncConfigured := false
+			for _, r := range statuses {
+				if r.Err != nil || r.Status == nil {
+					continue
+				}
+				if r.Status.PoolerType == clustermetadatapb.PoolerType_PRIMARY {
+					if r.Status.PrimaryStatus != nil && r.Status.PrimaryStatus.SyncReplicationConfig != nil {
+						syncConfigured = len(r.Status.PrimaryStatus.SyncReplicationConfig.StandbyIds) >= expectedCount
```
cuongdo: can `>=` be `==`?

dweitzman: I've both simplified and hardened this code so it's looking for an exact number of replicas and also verifying that the primary and replicas are all in agreement in their GUC about what the required ACKs are

dweitzman: Made another change here that leaked into .proto files: I added `standby_application_names` to `SynchronousReplicationConfiguration` in addition to `standby_ids` so tests can easily verify that the application name for a replica in the primary's GUC matches the application name configured for that replica on itself

PR #745, opened 2026-03-17
fix(pgctld): remove stale standby.signal before crash recovery in PgRewind

dweitzman:
When a previous RewindToSource attempt failed during postgres startup (e.g. due to a timeline incompatibility), standby.signal was left in the data directory. On the next attempt, single-user crash recovery mode (postgres --single) would fail with "standby mode is not supported by single-user servers", preventing pg_rewind from recovering the dirty data dir and causing an infinite failure loop.

Remove standby.signal before crash recovery when postgres is not cleanly stopped. The caller re-creates it via Restart --as-standby after rewind.


dweitzman:
This is folded into https://github.com/multigres/multigres/pull/835 now

PR #749, opened 2026-03-18
refactor(multipooler): move consensus term file out of PGDATA

mkindahl:
The consensus term was previously stored inside PGDATA (`pg_data/consensus/consensus_term.json`). This meant a pgBackRest restore — which replaces PGDATA — would also reset the term to whatever value was in the backup, forcing the node to re-join consensus at a stale lower term.

The term file is now stored at `poolerDir/consensus_term.json`, directly alongside other multipooler state, outside the PostgreSQL data directory. The file is intentionally not included in backups — the term only needs to be monotonically increasing, so the pre-restore value is the correct one to keep.

Because the term file is no longer backed up, a restored node could have a term file on disk that is ahead of the restored PGDATA. restoreFromBackupLocked now deletes the term file (and resets in-memory state to term 0) immediately after pgBackRest completes and before PostgreSQL starts. Multiorch advances the term to the current cluster value on the first BeginTerm RPC after the node rejoins.

To allow a post-restore node (term 0) to accept the first `BeginTerm` RPC without special-casing, the guard in validateAndUpdateTerm that rejected all RPCs when the local term was 0 is removed. A positive request term is now accepted and initializes the local term regardless of whether term 0 came from an uninitialized state or a just-deleted term file.

Changes:
- Term file path simplified from pg_data/consensus/consensus_term.json to poolerDir/consensus_term.json (no subdirectory)
- New ConsensusTermFile constant in go/common/constants/postgres.go
- Term storage functions (consensusTermPath, getConsensusTerm, setConsensusTerm) converted from package-level functions taking poolerDir to methods on ConsensusState
- New ConsensusState.DeleteTermFile() method, idempotent if file absent
- restoreFromBackupLocked calls DeleteTermFile() after restore completes
- validateAndUpdateTerm: remove zero-term rejection guard
- Tests updated to use the new ConsensusState method API and to set the PGDATA env var where term_storage needs it
- New term_storage_test.go with unit tests for DeleteTermFile
- backup_test.go: assert term resets to 0 after restore

Fixes MUL-257

### DIFF
--- a/go/services/multipooler/manager/manager.go
+++ b/go/services/multipooler/manager/manager.go
@@ -765,14 +765,6 @@ func (pm *MultiPoolerManager) validateAndUpdateTerm(ctx context.Context, request
 		return fmt.Errorf("failed to get current term: %w", err)
 	}
 
-	// Check if consensus term has been initialized (term 0 means uninitialized)
-	if currentTerm == 0 {
-		pm.logger.ErrorContext(ctx, "Consensus term not initialized",
-			"service_id", pm.serviceID.String())
-		return mterrors.New(mtrpcpb.Code_FAILED_PRECONDITION,
-			"consensus term not initialized, must be set via BeginTerm (use force=true to bypass)")
-	}
-
 	// If request term == current term: ACCEPT (same term, execute)
 	// If request term < current term: REJECT (stale request)
 	// If request term > current term: UPDATE term and ACCEPT (new term discovered)
--- a/go/services/multipooler/manager/rpc_backup.go
+++ b/go/services/multipooler/manager/rpc_backup.go
@@ -322,31 +322,26 @@ func (pm *MultiPoolerManager) restoreFromBackupLocked(ctx context.Context, backu
 		return err
 	}
 
-	// Set consensus term after restore.
-	// If the cluster is at a higher term,
-	// validateAndUpdateTerm will automatically update our term when multiorch fixes replication.
-	var term int64 = 0
-	if err := telemetry.WithSpan(ctx, "restore/reload-consensus-term", func(ctx context.Context) error {
+	// Delete the consensus term file after restore.
+	//
+	// The term file lives outside PGDATA and is not included in the backup, so
+	// its on-disk value may be ahead of what the restored PGDATA participated in.
+	// Deleting it resets the node to term 0; multiorch will advance the term to
+	// the current cluster value on first contact via BeginTerm.
+	//
+	// TODO: Revisit this when we restore backups for other reasons than to
+	// bootstrap a new node, e.g. point-in-time recovery for an existing node.
+	if err := telemetry.WithSpan(ctx, "restore/reset-consensus-term", func(ctx context.Context) error {
 		if pm.consensusState != nil {
-			pm.logger.InfoContext(ctx, "Loading consensus term that was restored from backup")
-			pm.loadConsensusTermFromDisk()
-			term = pm.consensusState.term.TermNumber
-
-			// Clear primary_term from backup since this is a standby restore
-			// The backup may contain a non-zero primary_term if it was taken from a primary
-			if err := pm.consensusState.SetPrimaryTerm(ctx, 0, false /* force */); err != nil {
-				return mterrors.Wrap(err, "failed to clear primary_term after restore")
+			pm.logger.InfoContext(ctx, "Deleting consensus term file after restore; node will re-join consensus from term 0")
+			if err := pm.consensusState.DeleteTermFile(); err != nil {
+				return mterrors.Wrap(err, "failed to delete consensus term file after restore")
 			}
 		}
 		pm.healthStreamer.UpdatePrimaryObservation(nil)
-
 		return nil
 	}); err != nil {
 		return err
-	}
-
-	if term == 0 {
-		pm.logger.ErrorContext(ctx, "MonitorPostgres: term is uninitialized even after restore")
 	}
 
 	if err := telemetry.WithSpan(ctx, "restore/reopen-pooler", func(ctx context.Context) error {
--- a/go/services/multipooler/manager/term_storage.go
+++ b/go/services/multipooler/manager/term_storage.go
@@ -36,13 +36,13 @@ func multigresDataDir() string {
 }
 
 // consensusTermPath returns the path to the consensus term file
-func consensusTermPath(poolerDir string) string {
-	return filepath.Join(poolerDir, "pg_data", "consensus", "consensus_term.json")
+func (cs *ConsensusState) consensusTermPath() string {
+	return filepath.Join(cs.poolerDir, constants.ConsensusTermFile)
 }
 
 // getConsensusTerm retrieves the current consensus term information from disk
-func getConsensusTerm(poolerDir string) (*multipoolermanagerdatapb.ConsensusTerm, error) {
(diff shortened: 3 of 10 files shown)


go/test/endtoend/multipooler/backup_test.go
```diff
@@ -241,23 +241,23 @@ func TestBackup_CreateListAndRestore(t *testing.T) {
 						}
 						return statusResp.Status.PostgresRunning
 					}, 10*time.Second, 100*time.Millisecond, "PostgreSQL should be running after restore")
-					t.Logf("Term after restore: %d (expected: 1)", restoredTerm)
-					assert.Equal(t, int64(1), restoredTerm, "Term should be restored to backup's original value (1)")
+					t.Logf("Term after restore: %d (expected: %d)", restoredTerm, higherTerm)
+					assert.Equal(t, higherTerm, restoredTerm, "Term should be preserved after restore (consensus term file is outside PGDATA)")
```
mkindahl: This means that if you restore a backup the term you had before remains, which I am not sure is correct behavior.

So, if you take a backup at term 5, do some stuff to end up at term 10, then when you restore the backup the term will still say 10 even though the backup represents a backup at term 5.

mkindahl: Added a commit that will remove the consensus term file if you have restored a backup. This means that the term will be set to zero and a new term is expected to be fetched from the cluster.

go/services/multipooler/manager/rpc_backup.go
```diff
@@ -322,33 +322,24 @@ func (pm *MultiPoolerManager) restoreFromBackupLocked(ctx context.Context, backu
...
-			pm.logger.InfoContext(ctx, "Loading consensus term that was restored from backup")
-			pm.loadConsensusTermFromDisk()
-			term = pm.consensusState.term.TermNumber
-
-			// Clear primary_term from backup since this is a standby restore
-			// The backup may contain a non-zero primary_term if it was taken from a primary
-			if err := pm.consensusState.SetPrimaryTerm(ctx, 0, false /* force */); err != nil {
-				return mterrors.Wrap(err, "failed to clear primary_term after restore")
+			pm.logger.InfoContext(ctx, "Deleting consensus term file after restore; node will re-join consensus from term 0")
+			if err := pm.consensusState.DeleteTermFile(); err != nil {
```
dweitzman, line 337: Feels like this is worthy of a big caveat that the behavior here is not (currently) intended to be theoretically-sound. Feels like there should be a TODO to revisit this when start allowing restore for reasons other than provisioning new poolers in a pre-existing cluster

go/provisioner/local/local.go
```diff
@@ -2078,6 +2084,46 @@ func (p *localProvisioner) validateBinaryExists(binaryPath, serviceName string)
 	return nil
 }
 
+// validateAllPorts checks every configured port for conflicts before any service is started.
+// Running this first means users get a precise "port X is already in use" error rather than
+// a misleading failure that surfaces only after partially starting services.
+func (p *localProvisioner) validateAllPorts() error {
```
dweitzman: Seems like this would need to consider processes that were already successfully started in a previous local provisioner run?

The various provisionXYZ() methods generally call `findRunningXYZService()` and if one is already running for the current configuration (by looking at process IDs in state files & checking for those processes on the operating system) they return the existing port numbers without changing anything

go/services/multipooler/manager/term_storage.go
```diff
@@ -34,66 +30,3 @@ func postgresDataDir() string {
 func multigresDataDir() string {
 	return filepath.Join(postgresDataDir(), constants.MultigresMarkerDirectory)
 }
-
-// consensusTermPath returns the path to the consensus term file
```
dweitzman, line 38: There's not a lot left in this file if everything moves to `consensus_state.go`

Personally I think it's a nice for this file to focus on modifying files safely and `consensus_state.go` to focus on consensus algorithm issues, but I'm not strongly attached to that if you think otherwise

go/services/multipooler/manager/manager.go
```diff
@@ -765,14 +765,6 @@ func (pm *MultiPoolerManager) validateAndUpdateTerm(ctx context.Context, request
 		return fmt.Errorf("failed to get current term: %w", err)
 	}
 
-	// Check if consensus term has been initialized (term 0 means uninitialized)
```
dweitzman, line 768: Can we no longer distinguish between a pooler with bad disk (lost term data) vs a newly-initialized pooler (or could we not before)? Makes me a little nervous from a defense-in-depth perspective

I don't know if there's a particularly elegant alternative, though, short of implementing the sort of things I'm personally in favor of like letting observers see through multischema if they've become a cohort member and initialize their promise-state then, and then remove it when they leave the cohort. I'd also be a little nervous about the idea of doing a backup restore on a cohort member. I kinda think we should explicitly only allow restore when the consensus term is unset and block restore after joining any term.

Not requesting any changes here, but we should probably chat about these things at some point

go/test/endtoend/localprovisioner/cluster_test.go
```diff
@@ -868,6 +868,11 @@ func TestClusterLifecycle(t *testing.T) {
 	_, err := exec.LookPath("etcd")
 	require.NoError(t, err, "etcd binary must be available in PATH for cluster lifecycle tests")
 
+	// Require PostgreSQL binaries to be available (required for local provisioner)
+	if !utils.HasPostgreSQLBinaries() {
```
dweitzman: This makes me a little nervous about some kind of accident where we skip this test on CI or somewhere it matters due to not having postgres binaries in the path

Seems like endtoend tests should generally always run with the postgres binaries on their path, and if folks don't want to run this test the can do `go test -short ...`?...

mkindahl: This crept in while testing. Agree that we should always have binaries and not allow tests to be executed without binaries available. The user needs to add necessary tools to the path to run the tests.

Removed it.

go/services/multipooler/manager/term_storage.go
mkindahl: I have no strong opinion. The reason for moving the code is because they are now methods in ConsensusState rather than free functions and it felt prudent to move it to that file.

We can keep the methods here as well if we want to split the methods for ConsensusState into separate logical units rather than based on the structure the methods belong to.

I'll move them back and we can see how it looks.

go/services/multipooler/manager/rpc_backup.go
mkindahl: Added a TODO, but I think the critical information will be stored in the leadership history, including terms and other similar information.

Meaning, if we restore a backup, it needs to replay all changes from a known primary, including leadership history, so this file is not going to be needed anyway.

go/services/multipooler/manager/manager.go
mkindahl: If critical files are lost because of a corrupted disk, we will have problems yes. The same applies if the `standby.signal` file is lost, which would make a server read-write, so I do not think we are in a new position here.

go/provisioner/local/local.go
mkindahl: This was just an attempt at avoiding some flakes. I'll remove it for now and we can consider it when we implement a "port server" (we currently have race conditions in some tests because ports can be "stolen" when you restart a server inside a test.)

PR #750, opened 2026-03-18
feat(multigateway): support LISTEN/NOTIFY with shared listener

haritabh17:
## Summary

- Add LISTEN/NOTIFY support using a shared listener architecture — one dedicated PG connection per multipooler fans out notifications to all gateway clients via gRPC streaming
- Implement PubSubListener with channel refcounting, automatic reconnect, and serialized event loop
- Handle transaction semantics (pending LISTEN/UNLISTEN on COMMIT/ROLLBACK) and multi-statement batches
- Add design doc, proto definitions, and comprehensive e2e tests

## Description

Instead of pinning a backend connection per client session (which defeats pooling), a single PubSubListener in the multipooler maintains a dedicated PG connection that issues LISTEN/UNLISTEN based on aggregate demand. Channels are refcounted — first subscriber triggers LISTEN, last unsubscriber triggers UNLISTEN. Notifications are fanned out through a StreamNotifications gRPC stream to each gateway, where GRPCNotificationManager fans out locally to client sessions.

On the gateway side, the planner routes LISTEN/UNLISTEN to a ListenNotifyPrimitive that updates per-connection state, while NOTIFY goes to PostgreSQL as a regular query. LISTEN/UNLISTEN inside transactions are buffered as pending and applied on COMMIT or discarded on ROLLBACK. The handler's async pusher uses the connection-scoped context so notification delivery survives past individual query timeouts.

Notifications during a PubSubListener reconnect window may be lost — this is documented as a known limitation.

### DIFF
--- a/go/services/multigateway/planner/prepared_stmt_test.go
+++ b/go/services/multigateway/planner/prepared_stmt_test.go
@@ -114,7 +114,7 @@ type testSetup struct {
 func newTestSetup(t *testing.T) *testSetup {
 	t.Helper()
 	logger := slog.New(slog.NewTextHandler(bytes.NewBuffer(nil), nil))
-	p := NewPlanner("default", logger)
+	p := NewPlanner("default", logger, nil)
 	exec := &mockIExecute{}
 
 	// The primitive calls conn.Handler().HandleParse/HandleBind/HandleClose,


go/services/multipooler/pubsub/listener.go
```diff
@@ -0,0 +1,525 @@
...
+	done    chan struct{} // closed when request is processed
+}
+
+// Listener manages a single dedicated PG connection for LISTEN/NOTIFY,
+// with channel refcounting and fan-out to subscribers.
+//
+// Architecture: split read/write on a single TCP connection.
+// TCP is full-duplex; bufferedReader and bufferedWriter are independent bufio
+// instances. The reader goroutine exclusively reads; the event loop exclusively
+// writes. This avoids reconnecting to issue LISTEN/UNLISTEN commands, which
```
GuptaManan100: `This avoids reconnecting to issue LISTEN/UNLISTEN commands` seems grammatically incorrect. And also, I don't see how this prevents LISTEN/UNLISTEN reconnecting issue.

go/services/multipooler/pubsub/listener.go
```diff
@@ -0,0 +1,525 @@
...
+	result := make([]byte, 0, len(s))
+	for i := 0; i < len(s); i++ {
+		if s[i] == '"' {
+			result = append(result, '"', '"')
+		} else {
+			result = append(result, s[i])
+		}
+	}
+	return string(result)
+}
```
GuptaManan100: We have `ast.QuoteIdentifier` that we should use instead

go/services/multipooler/pubsub/listener.go
```diff
@@ -0,0 +1,525 @@
...
+func (l *Listener) UnsubscribeCh(channel string, notifCh chan *Notification) {
+	req := request{
+		typ:     reqUnsubscribe,
+		channel: channel,
+		subCh:   notifCh,
+		done:    make(chan struct{}),
+	}
+	l.requests <- req
+	<-req.done
+}
```
GuptaManan100: This is the same function as `Unsubscibe`

go/services/multipooler/pubsub/listener.go
```diff
@@ -0,0 +1,525 @@
...
+	req := request{
+		typ:     reqSubscribe,
+		channel: channel,
+		subCh:   ch,
+		done:    make(chan struct{}),
+	}
+	l.requests <- req
+	<-req.done
+	return ch
+}
```
GuptaManan100: This is unused

proto/multipoolerservice.proto
```diff
@@ -439,3 +444,24 @@ message PrimaryObservation {
...
+  repeated string channels = 2;
+}
+
+// StreamNotificationsResponse delivers a notification or warning from the shared listener.
+message StreamNotificationsResponse {
+  // notification contains the PG notification data.
+  query.PgNotification notification = 1;
+
+  // is_warning indicates this is a reconnect warning, not a real notification.
+  bool is_warning = 2;
```
GuptaManan100: Why do we need this field, can't we just use the string being non-empty to know we have a warning

go/services/multipooler/pubsub/listener.go
```diff
@@ -0,0 +1,525 @@
...
+	"github.com/multigres/multigres/go/common/pgprotocol/protocol"
+	"github.com/multigres/multigres/go/common/sqltypes"
+)
+
+// Notification wraps a PG notification.
+type Notification struct {
+	PID     int32
+	Channel string
+	Payload string
+}
```
GuptaManan100: We have this sturct defined 3 times. We should reuse the one in `sqltypes`

go/services/multigateway/handler/notification_manager.go
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+package handler
+
+// Notification represents an async notification from PostgreSQL.
+type Notification struct {
+	PID     int32
+	Channel string
+	Payload string
+}
```
GuptaManan100: Same we have this struct 3 times

go/services/multigateway/poolergateway/notification_adapter.go
```diff
@@ -0,0 +1,228 @@
...
+// the pooler's StreamNotifications gRPC. It manages per-channel streams
+// and fans out notifications to subscriber channels.
+//
+// NOTE: Currently opens one gRPC stream per unique PG channel. If the number
+// of distinct channels grows large, consider consolidating into a single
+// bidirectional stream per gateway-pooler pair with dynamic channel add/remove.
+// The StreamNotifications RPC already accepts repeated channels, so the
+// multipooler side supports multi-channel subscriptions on a single stream.
+// HTTP/2 multiplexes all streams over one TCP connection, so the overhead is
+// small in practice.
```
GuptaManan100: Not entirely sure about this. Should we benchmark it? There is also the overhead of additional go routines waiting on listening on the streams that we don't account for

go/services/multipooler/pubsub/listener.go
```diff
@@ -0,0 +1,525 @@
...
+
+	// requests is the serialized command channel for the event loop.
+	requests chan request
+
+	// cancel stops the background goroutines.
+	cancel context.CancelFunc
+
+	// wg tracks background goroutines.
+	wg sync.WaitGroup
+}
```
GuptaManan100, line 77: This doesn't account for state changes. We need to open and close the listener in the state manager when we transition to and from a PRIMARY

go/services/multipooler/pubsub/listener.go
```diff
@@ -0,0 +1,525 @@
...
+// Listener manages a single dedicated PG connection for LISTEN/NOTIFY,
+// with channel refcounting and fan-out to subscribers.
+//
+// Architecture: split read/write on a single TCP connection.
+// TCP is full-duplex; bufferedReader and bufferedWriter are independent bufio
+// instances. The reader goroutine exclusively reads; the event loop exclusively
+// writes. This avoids reconnecting to issue LISTEN/UNLISTEN commands, which
+// would cause notification loss for existing channels.
+type Listener struct {
+	config *client.Config
```
GuptaManan100: Not quite sold on keeping the client config to connect. Why don't we get the pool manager to give us a reserved connection instead?

go/services/multigateway/engine/listen_notify.go
```diff
@@ -0,0 +1,121 @@
...
+	ListenActionUnlisten
+	ListenActionUnlistenAll
+)
+
+// ListenNotifyPrimitive handles LISTEN/UNLISTEN commands by updating connection state.
+// The actual subscription to the pooler's PubSubListener is managed by the handler.
+type ListenNotifyPrimitive struct {
+	Action  ListenAction
+	Channel string
+	SQL     string
```
GuptaManan100: We should call this Query

go/common/pgprotocol/server/conn.go
```diff
@@ -988,3 +991,92 @@ func (c *Conn) handleSync() error {
...
+	ch     chan *NotificationPayload
+	cancel context.CancelFunc
+}
+
+// NotificationPayload carries notification data for async delivery.
+type NotificationPayload struct {
+	PID     int32
+	Channel string
+	Payload string
+}
```
GuptaManan100: Same as the notification struct in sqltypes package

go/common/pgprotocol/server/query.go
```diff
@@ -584,3 +584,38 @@ func writeString(w io.Writer, s string) error {
 	}
 	return nil
 }
+
+// WriteNotificationResponse writes a NotificationResponse ('A') message to the client.
+// Format: Byte1('A') + Int32(length) + Int32(pid) + String(channel) + String(payload)
+// This is used to deliver asynchronous notifications to clients that have issued LISTEN.
+func (c *Conn) WriteNotificationResponse(pid int32, channel, payload string) error {
```
GuptaManan100: We already have this in conn.go too

go/services/multigateway/handler/handler.go
```diff
@@ -194,6 +196,11 @@ func (h *MultiGatewayHandler) HandleQuery(ctx context.Context, conn *server.Conn
 				conn.SetTxnStatus(protocol.TxnStatusFailed)
 			}
 		}
+
+		// Post-execution: sync LISTEN/NOTIFY subscriptions.
+		if err == nil {
+			h.handleListenStateChanges(ctx, conn, st, asts[0])
```
GuptaManan100: Why don't we do this in the StreamExecute primitives of LISTEN and NOTIFY themselves?

go/services/multipooler/pubsub/listener.go
haritabh17: Fixed — updated the comment to clarify the split read/write design.

go/services/multipooler/pubsub/listener.go
haritabh17: Done

go/services/multipooler/pubsub/listener.go
haritabh17: Fixed

go/services/multipooler/pubsub/listener.go
haritabh17: Removed.

proto/multipoolerservice.proto
haritabh17: Agreed — removed the separate boolean field.

go/services/multipooler/pubsub/listener.go
haritabh17: Done — consolidated to use sqltypes.Notification everywhere.

go/services/multigateway/handler/notification_manager.go
haritabh17: Fixed — same as above, using sqltypes.Notification

go/services/multipooler/pubsub/listener.go
haritabh17: Addressed — added OnStateChange(poolerType, servingStatus) method. The listener starts only when poolerType == PRIMARY && servingStatus == SERVING, and stops on any other state.

go/services/multipooler/pubsub/listener.go
haritabh17: Done — switched from *client.Config to connpoolmanager.PoolManager.

go/services/multigateway/engine/listen_notify.go
haritabh17: Done — renamed the field to Query.

go/common/pgprotocol/server/conn.go
haritabh17: Fixed — server.Conn now uses sqltypes.Notification

go/common/pgprotocol/server/query.go
haritabh17: Removed — The functionality is now in conn.go's writeNotificationResponseMsg (unexported, used internally by FlushPendingNotifications and the async pusher).

go/services/multigateway/handler/handler.go
haritabh17: Ack - moved to primitives

go/services/multigateway/poolergateway/notification_adapter.go
haritabh17: Good point on the per-stream goroutine overhead. Added a NOTE comment documenting the consolidation path (single bidirectional stream per gateway-pooler pair) if this becomes an issue. Created MUL-276 to add OTel metrics for active stream count, notification drops, and reconnect gap duration so we have production visibility. We can benchmark the consolidation path if the numbers warrant it.

go/common/pgprotocol/server/query.go
```diff
@@ -584,3 +584,12 @@ func writeString(w io.Writer, s string) error {
...
 }
+
+// WriteNotice sends a NoticeResponse with the given severity and message to the client.
+func (c *Conn) WriteNotice(severity, message string) error {
+	diag := &mterrors.PgDiagnostic{
+		Severity: severity,
+		Message:  message,
+	}
+	return c.writeNoticeResponse(diag)
+}
```
GuptaManan100: This is unused

go/services/multigateway/engine/listen_notify.go
```diff
@@ -0,0 +1,132 @@
...
+		} else {
+			state.RemoveListenChannel(channel)
+			state.SubSync.SyncSubscriptions(conn.Context(), conn, state, nil, []string{channel}, false)
+		}
+	case ListenActionUnlistenAll:
+		if conn.IsInTransaction() {
+			state.AddPendingUnlistenAll()
+		} else {
+			state.ClearListenChannels()
+			state.SubSync.SyncSubscriptions(conn.Context(), conn, state, nil, nil, true)
```
GuptaManan100, line 96: Can't we call the SubSync functions internally in the corresponding listen functions. Like call `state.SubSync.SyncSubscriptions(conn.Context(), conn, state, nil, nil, true)` inside `state.ClearListenChannels()`, and so on?

go/common/pgprotocol/server/query.go
haritabh17: Ack, removed

go/services/multigateway/engine/listen_notify.go
haritabh17: We could move SubSync into AddListenChannel/ RemoveListenChannel/ ClearListenChannels for the autocommit cases, and also into CommitPendingListens for the transaction case. The atomicity contract wouldn't change — CommitPendingListens would still compute the net diff and call SyncSubscriptions once with the full batch.    
                            
The main reason I kept it external is separation of concerns: connection_state.go is currently a pure state container with no I/O or context/conn dependencies. Moving SubSync in would require adding those parameters to all the state methods and mixing I/O into the state layer. Keeping all SubSync calls in the engine primitives makes the "mutate state, then sync" pattern explicit and consistent across both autocommit and transaction paths.

PR #751, opened 2026-03-18
feat(multigateway): session pinning for temporary tables

haritabh17:
## Summary

- Add session pinning when `CREATE TEMP TABLE` is detected, routing all subsequent queries through a reserved connection to maintain session-local state
- Add `DiscardTempTables` RPC following the `ConcludeTransaction` pattern — gateway detects `DISCARD TEMP`/`DISCARD ALL` via AST, sends dedicated RPC to multipooler which removes `ReasonTempTable` and releases the connection if no other reasons remain
- Add extended query protocol (EQP) support for temp table pinning and unpinning
- Auto-detect inline transaction state from PG's `ReadyForQuery` status so deferred BEGIN/COMMIT works correctly on pinned connections
- Send `DISCARD TEMP` on release as defense-in-depth for client disconnect cleanup

## Description

Temporary tables in PostgreSQL are session-local, but multigres uses connection pooling, so queries within a single client session can land on different backends. This PR detects temp table creation via AST inspection and pins the session to a dedicated reserved connection for its lifetime.

**Detection** covers `CREATE TEMP TABLE`, `CREATE TABLE AS SELECT ... TEMPORARY`, `SELECT INTO TEMPORARY`, and `CREATE TEMPORARY VIEW` — all via AST node inspection (`RelPersistence == 't'`), with support for both simple and extended query protocols.

**DISCARD TEMP** follows the same semantics as `COMMIT` — a dedicated `DiscardTempTables` RPC removes `ReasonTempTable` from the reserved connection, releasing it back to the pool if no other reasons (e.g., active transaction) remain. No string matching; detection is fully AST-based via the planner.

**Transaction interop**: pinned sessions use deferred BEGIN (prepended inline to the first query), and the multipooler syncs transaction state from PG's wire protocol after each `StreamExecute`.

pgregress results improved from 122 to 134 passing tests (+12).

### DIFF
--- a/go/services/multigateway/engine/discard_temp_primitive.go
+++ b/go/services/multigateway/engine/discard_temp_primitive.go
@@ -61,13 +61,7 @@ func (d *DiscardTempPrimitive) StreamExecute(
 		// transaction (PG rejects it), so a pending BEGIN that was never
 		// sent to PG should be discarded.
 		state.PendingBeginQuery = ""
-
-		err := exec.DiscardTempTables(ctx, conn, state, callback)
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return exec.DiscardTempTables(ctx, conn, state, callback)
 	}
 
 	// No temp table reservation — return synthetic result.
--- a/go/services/multigateway/handler/connection_state.go
+++ b/go/services/multigateway/handler/connection_state.go
@@ -301,9 +301,7 @@ func (m *MultiGatewayConnectionState) GetSessionVariable(name string) (string, b
 }
 
 // HasTempTableReservation returns true if any shard state has a reserved
-// connection with the temp table reason set. This replaces the old
-// SessionPinned bool field with a derived check on the authoritative
-// reservation state from the multipooler.
+// connection with the temp table reason set.
 func (m *MultiGatewayConnectionState) HasTempTableReservation() bool {
 	m.mu.Lock()
 	defer m.mu.Unlock()


proto/multipoolerservice.proto
```diff
@@ -382,6 +387,35 @@ message ConcludeTransactionResponse {
...
+  // caller_id identifies the caller
+  mtrpc.CallerID caller_id = 2;
+
+  // options must include reserved_connection_id to identify the connection
+  query.ExecuteOptions options = 3;
+
+  // sql is the original SQL statement to execute (e.g., "DISCARD TEMP" or "DISCARD ALL").
+  // The multipooler sends this to PostgreSQL instead of hardcoding "DISCARD TEMP",
+  // so that DISCARD ALL correctly resets all session state on the backend.
+  string sql = 4;
```
GuptaManan100: this doesn't seem required anymore. The comment is also wrong and requires updating.

go/test/endtoend/shardsetup/temp_table_test.go
```diff
@@ -0,0 +1,977 @@
+// Copyright 2025 Supabase, Inc.
```
GuptaManan100: Year needs to be fixed

go/test/endtoend/shardsetup/temp_table_test.go
```diff
@@ -0,0 +1,977 @@
...
+	err := conn.QueryRowContext(t.Context(), "SELECT pg_backend_pid()").Scan(&pid)
+	require.NoError(t, err, "failed to get backend pid")
+	return pid
+}
+
+// requirePinned asserts that two consecutive backend PID queries return the
+// same value, proving the multigateway session is routed to a single reserved
+// connection (session pinned).
+func requirePinned(t *testing.T, conn *sql.Conn) int {
+	t.Helper()
```
GuptaManan100: I'm thinking should we create a utils package with some of these functions. I feel we are repeating too many times in tests.

go/services/multipooler/grpcpoolerservice/service.go
```diff
@@ -529,6 +529,36 @@ func (s *poolerService) ConcludeTransaction(ctx context.Context, req *multipoole
...
+	executor, err := s.pooler.Executor()
+	if err != nil {
+		return nil, errors.New("executor not initialized")
+	}
+
+	// Discard temp tables — pass original SQL so DISCARD ALL works correctly
+	sql := req.GetSql()
+	if sql == "" {
+		sql = "DISCARD TEMP" // backward compatibility
+	}
```
GuptaManan100: Not sure about this.

go/services/multigateway/handler/connection_state.go
```diff
@@ -59,6 +59,11 @@ type MultiGatewayConnectionState struct {
 	// connection so the multipooler can use the exact statement instead of plain "BEGIN".
 	PendingBeginQuery string
 
+	// SessionPinned indicates this connection requires session affinity.
+	// When true, all queries are routed through a reserved connection
+	// to maintain session-local state (e.g., temp tables).
+	SessionPinned bool
+
```
GuptaManan100: Why do we need this? Having a reserved connection in the shard state should be enough to know if we have a pinned session or not?

proto/multipoolerservice.proto
haritabh17: Done — removed the `sql` field entirely. The multipooler now hardcodes `"DISCARD TEMP"` since DISCARD ALL handling was scoped out of this PR.

go/test/endtoend/shardsetup/temp_table_test.go
haritabh17: Fixed — changed to 2026.

go/test/endtoend/shardsetup/temp_table_test.go
haritabh17: Done — moved `GetBackendPID` and `RequirePinned` to `go/test/utils/conn.go` for cross-package reuse. Moved `openMultigatewayConn` to `shardsetup_test.go` (next to `skipIfShort`) since it depends on the `ShardSetup` type.

go/services/multipooler/grpcpoolerservice/service.go
haritabh17: Removed along with the `sql` field in the proto. The fallback was only needed when DISCARD ALL was in scope.

go/services/multigateway/handler/connection_state.go
haritabh17: Good call — eliminated it entirely. Temp table creation is now detected in the planner and routed through a `TempTableRoute` primitive that calls `ReserveAndExecute` (new `IExecute` method). All other code derives the pinned state from `HasTempTableReservation()` which checks shard state for `ReasonTempTable`. No more duplicated state.

go/services/multigateway/engine/discard_temp_primitive.go
```diff
@@ -0,0 +1,95 @@
...
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	// No temp table reservation — just execute via regular path.
+	// DISCARD TEMP on an unreserved session is harmless.
+	return exec.StreamExecute(ctx, conn, d.TableGroup, constants.DefaultShard, d.Query, state, callback)
```
GuptaManan100: Why are we doing this? Even if its harmless its not required, right?

go/services/multigateway/engine/engine.go
```diff
@@ -58,6 +58,26 @@ type IExecute interface {
...
+	// consumed as part of the reservation.
+	ReserveAndExecute(
+		ctx context.Context,
+		conn *server.Conn,
+		tableGroup string,
+		shard string,
+		sql string,
+		state *handler.MultiGatewayConnectionState,
+		reasons uint32,
+		callback func(context.Context, *sqltypes.Result) error,
```
GuptaManan100: I'm not sure if we need this. We can still use the StreamExecute even for this purpose. We just need to add a case to that, like `conn.IsInTransaction()`, we need to add `conn.HasTemTable` or something

go/services/multigateway/engine/temp_table_route.go
```diff
@@ -0,0 +1,66 @@
...
+	exec IExecute,
+	conn *server.Conn,
+	state *handler.MultiGatewayConnectionState,
+	callback func(context.Context, *sqltypes.Result) error,
+) error {
+	reasons := protoutil.ReasonTempTable
+	if conn.IsInTransaction() {
+		reasons |= protoutil.ReasonTransaction
+	}
+	return exec.ReserveAndExecute(ctx, conn, t.TableGroup, t.Shard, t.Query, state, reasons, callback)
```
GuptaManan100: We can just call StreamExecute, and add a case for temp tables

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -532,6 +658,102 @@ func (sc *ScatterConn) ConcludeTransaction(
...
+		clear         bool
+		reservedState *querypb.ReservedState
+	}
+	var updates []shardUpdate
+	var errs []error
+	var callbackResult *sqltypes.Result
+
+	// Iterate over all shard states with reserved connections.
+	// Only discard on shards that have the temp table reason.
+	for _, ss := range state.ShardStates {
```
GuptaManan100, line 614: We should lock state before reading the ShardStates

go/services/multigateway/engine/discard_temp_primitive.go
haritabh17: Done — replaced with a synthetic "DISCARD" result

go/services/multigateway/engine/engine.go
haritabh17: Agreed — removed ReserveAndExecute entirely. TempTableRoute now sets a one-shot PendingTempTableReservation flag and delegates to StreamExecute. ScatterConn handles it alongside conn.IsInTransaction().

go/services/multigateway/engine/temp_table_route.go
haritabh17: Done — same fix as above. TempTableRoute.StreamExecute now just sets the flag and calls exec.StreamExecute.

go/services/multigateway/scatterconn/scatter_conn.go
haritabh17: ShardStates is only accessed from the connection's goroutine. Do we need locking? 🤔

GuptaManan100: Maybe not, you're right

go/services/multigateway/engine/discard_temp_primitive.go
```diff
@@ -0,0 +1,94 @@
...
+	if state.HasTempTableReservation() {
+		// Clear any deferred BEGIN — DISCARD TEMP cannot run inside a
+		// transaction (PG rejects it), so a pending BEGIN that was never
+		// sent to PG should be discarded.
+		state.PendingBeginQuery = ""
+
+		err := exec.DiscardTempTables(ctx, conn, state, callback)
+		if err != nil {
+			return err
+		}
```
GuptaManan100: nit: Can be simplified to `return exec.DiscardTempTables(ctx, conn, state, callback)`

go/services/multigateway/handler/connection_state.go
```diff
@@ -294,6 +300,21 @@ func (m *MultiGatewayConnectionState) GetSessionVariable(name string) (string, b
 	return value, exists
 }
 
+// HasTempTableReservation returns true if any shard state has a reserved
+// connection with the temp table reason set. This replaces the old
+// SessionPinned bool field with a derived check on the authoritative
```
GuptaManan100: We don't need to talk about SessionPinned here. It was never committed to main

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -159,44 +159,86 @@ func (sc *ScatterConn) StreamExecute(
...
+		// this reserved connection.
+		beganTransaction := false
+		if state.PendingBeginQuery != "" && protoutil.HasTempTableReason(ss.ReservedState.GetReservationReasons()) {
+			sc.logger.DebugContext(ctx, "prepending deferred BEGIN to reserved query",
+				"pending_begin", state.PendingBeginQuery)
+			sql = state.PendingBeginQuery + "; " + sql
+			state.PendingBeginQuery = ""
+			beganTransaction = true
+		}
+
```
GuptaManan100, line 173: Oooh, I see why you have to do this, but it's not a great solution, just for the reason that the connection pooler on Postgres won't see a transaction has started! So the reservation reasons will be all wrong

go/services/multigateway/engine/discard_temp_primitive.go
haritabh17: Ack done.

go/services/multigateway/handler/connection_state.go
haritabh17: Updated

go/services/multigateway/scatterconn/scatter_conn.go
haritabh17: Will refactor in follow up PR

PR #754, opened 2026-03-19
feat(multipooler): add request gating and graceful drain for state transitions

GuptaManan100:
## Summary

- Add `RequestsWaiter` to track in-flight requests and notify when count reaches zero
- Gate all gRPC query endpoints with `StartRequest`/`EndRequest` to reject requests when not serving
- Implement two-phase graceful shutdown in `OnStateChange`: block new queries, then drain in-flight requests with a configurable grace period
- Allow cleanup operations (COMMIT, ROLLBACK, release reserved connection) to proceed during shutdown via `allowOnShutdown`

## Description

When multipooler transitions to `NOT_SERVING`, it now performs a graceful drain instead of immediately cutting off queries. The `StartRequest` method checks serving state and returns `ErrNotServing` or `ErrShuttingDown` as appropriate. During shutdown, requests on existing reserved connections (identified by `reservedConnectionId > 0`) and transaction-concluding operations are still permitted, ensuring clients can cleanly finish in-progress transactions.

The `RequestsWaiter` primitive provides a channel-based notification when all tracked requests complete, enabling `select`-based drain with timeout support.

### DIFF
--- a/go/services/multipooler/poolerserver/pooler.go
+++ b/go/services/multipooler/poolerserver/pooler.go
@@ -111,22 +111,34 @@ func (s *QueryPoolerServer) OnStateChange(ctx context.Context, poolerType cluste
 	}
 
 	// NOT_SERVING transition: begin two-phase shutdown.
-	// Phase 1: set shuttingDown to block new non-shutdown requests.
+	// Phase 1: set shuttingDown to block new non-shutdown requests and capture
+	// the drain channel atomically. This prevents the TOCTOU race where a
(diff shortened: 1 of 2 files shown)


GuptaManan100:
Closing in favor of #755. Request-level tracking (StartRequest/EndRequest) has a correctness issue: the drain counter drops to zero between individual gRPC requests on the same reserved connection, so the drain can complete while a transaction is still in-flight (e.g., after the last query finishes but before COMMIT). Connection-level tracking in #755 avoids this because reserved connections stay "lent" for the entire transaction lifetime.

PR #755, opened 2026-03-19
feat(multipooler): graceful drain during NOT_SERVING state transitions

GuptaManan100:
## Summary

- Add connection tracking across all pool types (regular borrow/recycle, reserved reserve/release) to know when in-flight queries have completed
- Implement `WaitForDrain` on the pool manager that blocks until all lent connections are returned
- Add `StartRequest` admission gate to reject new queries during shutdown while allowing existing reserved connections to finish their transactions
- Wire `StartRequest` checks into all gRPC service endpoints (StreamExecute, ExecuteQuery, Describe, Copy, Reserve, etc.)

## Problem

When transitioning to `NOT_SERVING`, the pooler immediately stopped accepting requests but had no mechanism to wait for in-flight queries to complete. This could cause clients to see abrupt connection drops mid-transaction.

## Solution

Two-phase graceful drain in `OnStateChange(NOT_SERVING)`:
1. Set `shuttingDown=true` — new requests are rejected via `StartRequest`, but existing reserved connections (in-flight transactions) are still admitted
2. Wait up to `gracePeriod` (3s) for all lent connections to drain back to the pool
3. Complete the transition to `NOT_SERVING`

Connection tracking uses a simple atomic counter with a channel-based signal (`zeroCh`) so `WaitForDrain` can efficiently block without polling.

### DIFF
--- a/go/services/multipooler/connpoolmanager/manager.go
+++ b/go/services/multipooler/connpoolmanager/manager.go
@@ -469,6 +469,10 @@ func (m *Manager) lentAdd(n int64) {
 	m.drainMu.Lock()
 	defer m.drainMu.Unlock()
 
+	if m.lentCount+n < 0 {
+		m.logger.Error("lentCount going negative, likely a bug in borrow/recycle or reserve/release callbacks",
+			"current", m.lentCount, "delta", n)
+	}
 	m.lentCount += n
 	if m.lentCount == 0 {
 		// Signal drain complete by closing the channel.
--- a/go/services/multipooler/poolerserver/pooler.go
+++ b/go/services/multipooler/poolerserver/pooler.go
@@ -92,6 +92,9 @@ func NewQueryPoolerServer(logger *slog.Logger, poolManager connpoolmanager.PoolM
 // OnStateChange transitions the query service to match the new serving state.
 // Implements PoolerController interface.
 //
+// This method is only called by the StateManager, which serializes calls behind
+// a mutex. Concurrent calls are not possible or expected.
+//
 // For NOT_SERVING transitions, this performs a two-phase graceful drain:
 //  1. Set shuttingDown=true to reject new requests (existing reserved connections continue)
 //  2. Wait for in-flight connections to drain (up to gracePeriod)


go/services/multipooler/poolerserver/pooler.go
```diff
@@ -28,6 +28,14 @@ import (
...
 )
 
+var (
+	// ErrNotServing is returned when a request is made while not serving.
+	ErrNotServing = errors.New("pooler is not serving")
+
+	// ErrShuttingDown is returned when a new request is made during graceful shutdown.
+	ErrShuttingDown = errors.New("pooler is shutting down")
+)
+
```
GuptaManan100, line 38: Don't worry about these errors, I'm gonna replace them with mterrors that start buffering in https://github.com/multigres/multigres/pull/690.

docs/query_serving/graceful_drain_design.md
```diff
@@ -0,0 +1,205 @@
+# Graceful Drain via Connection Pool Manager
+
+## Overview
+
+When multipooler transitions to `NOT_SERVING` (e.g., during demotion
+from primary to replica), it must reject new queries immediately while
+allowing existing reserved connections — transactions, portals, COPY
+operations — to finish cleanly. Without graceful drain, a sudden
+cutover would abort in-flight transactions, causing client errors and
+potential data inconsistency.
```
mkindahl: Hmm... cutting the connection is bad, so good to avoid it, but what is the case where data can become inconsistent as a result of this behavior?

docs/query_serving/graceful_drain_design.md
```diff
@@ -0,0 +1,205 @@
...
+meant a `NOT_SERVING` transition would immediately flip the status,
+but in-flight queries on reserved connections would continue running
+with no coordination — and new queries would only be rejected if
+individual gRPC handlers checked the serving status (which they did
+not).
+
+PostgreSQL clients using transactions or extended query protocol
+expect that once a `BEGIN` succeeds, subsequent statements on that
+connection will execute until `COMMIT`/`ROLLBACK`. Abruptly rejecting
+mid-transaction queries violates this contract.
```
mkindahl: Hmm... if the server crashes, the transaction will be abruptly aborted as well, which is also part of the contract (on how to deal with crashes and disconnects), so it would not break the contract. Keeping it to a minimum is of course desirable, but it does not break the contract.

You might want to reword this.

go/services/multipooler/connpoolmanager/manager.go
```diff
@@ -440,6 +462,58 @@ type ManagerStats struct {
...
+	if m.lentCount == 0 {
+		// Signal drain complete by closing the channel.
+		select {
+		case <-m.zeroCh:
+			// Already closed, nothing to do.
+		default:
+			close(m.zeroCh)
+		}
+	} else if m.lentCount == n && n > 0 {
+		// Transitioned from 0 to positive: create a new open channel.
```
mkindahl, line 486: Defensively, you might want to make sure that the channel really is closed and/or raise an error if it is not. A bug in the code could result in m.lentCount being zero when entering this function but there are waiters on an unclosed channel.

go/services/multipooler/pools/reserved/reserved_pool.go
```diff
@@ -221,6 +230,9 @@ func (p *Pool) KillConnection(ctx context.Context, connID int64) error {
 	p.mu.Unlock()
 
 	p.killCount.Add(1)
+	if p.config.OnRelease != nil {
+		p.config.OnRelease()
+	}
```
haritabh17: Firing OnRelease here but never setting rc.released. If the connection has an in-flight request, it'll finish with a backend error and call rc.Release() — CompareAndSwap(false, true) succeeds, pool.release() fires OnRelease again, double-decrementing lentCount. CompareAndSwap on rc.released at the top of KillConnection before proceeding??

go/services/multipooler/poolerserver/pooler.go
```diff
@@ -66,25 +85,81 @@ func NewQueryPoolerServer(logger *slog.Logger, poolManager connpoolmanager.PoolM
...
+		return nil
+	}
+
+	// NOT_SERVING: begin graceful drain
+	s.shuttingDown = true
+	s.mu.Unlock()
+
+	// Wait for in-flight connections to drain (best effort, bounded by gracePeriod).
+	if s.poolManager != nil {
+		drainCtx, cancel := context.WithTimeout(ctx, s.gracePeriod)
```
haritabh17: gracePeriod=0 creates an already-expired context — drain is silently disabled, every NOT_SERVING transition force-kills all reserved connections immediately. Either validate the flag or default to some interval when zero here??

docs/query_serving/graceful_drain_design.md
GuptaManan100: Yep, incorrect documentation, fixed

docs/query_serving/graceful_drain_design.md
GuptaManan100: Yep, done

go/services/multipooler/connpoolmanager/manager.go
GuptaManan100: Fixed

go/services/multipooler/pools/reserved/reserved_pool.go
GuptaManan100: Fixed

go/services/multipooler/poolerserver/pooler.go
GuptaManan100: Fixed

PR #757, opened 2026-03-23
fix(multipooler): separate pooler consolidator to prevent cross-gateway name collisions

GuptaManan100:
## Summary

- Create `PoolerConsolidator` that deduplicates prepared statements by `(query, paramTypes)` instead of by incoming statement name
- Replace the gateway `Consolidator` (misused with `connId=0`) in the multipooler executor
- Fix gateway `Consolidator` to include `paramTypes` in its dedup key (same query with different type hints are distinct statements)
- Update design doc to document why the gateway and pooler need separate consolidators

## Problem

Each multigateway replica has an independent `Consolidator` that assigns canonical names starting from `stmt0`. When multiple gateways send requests to the same multipooler, they can send the same canonical name (e.g., `stmt0`) for completely different queries. The multipooler's `ensurePrepared` looked up by this incoming name using `connId=0`, causing it to return the wrong prepared statement entry. This led to PostgreSQL `42P05: prepared statement already exists` errors, which pgx treated as fatal connection errors, cascading into 90% error rates.

A secondary issue: the gateway `Consolidator` deduplicated by query text alone, ignoring `paramTypes`. The same SQL with different type hints (e.g., `SELECT $1` with `INT4` vs `TEXT`) should produce different prepared statements.

## Solution

Introduced `PoolerConsolidator` — a purpose-built consolidator for the multipooler that:
- Ignores incoming statement names entirely (they're gateway artifacts)
- Deduplicates by `(query text, param types)` to produce stable canonical names (`ppstmt0`, `ppstmt1`, ...)
- Has no connection tracking, reference counting, or lifecycle management (per-connection state is handled by `connstate.ConnectionState`)

Also fixed the gateway `Consolidator`'s `stmts` map to key by `dedupKey(query, paramTypes)` instead of `query` alone.

### DIFF
--- a/go/common/preparedstatement/consolidator_test.go
+++ b/go/common/preparedstatement/consolidator_test.go
@@ -276,9 +276,9 @@ func TestConsolidator_RemoveConnection_NonExistent(t *testing.T) {
 }
 
 // TestConsolidator_SameQueryDifferentParamTypes verifies that the
-// Consolidator incorrectly consolidates statements with the same query text
-// but different parameter types. PostgreSQL treats these as distinct prepared
-// statements (different plans, different type coercion).
+// Consolidator does not incorrectly consolidate statements with the same query
+// text but different parameter types. PostgreSQL treats these as distinct
+// prepared statements (different plans, different type coercion).
 func TestConsolidator_SameQueryDifferentParamTypes(t *testing.T) {
 	consolidator := NewConsolidator()
 
@@ -299,6 +299,35 @@ func TestConsolidator_SameQueryDifferentParamTypes(t *testing.T) {
 		"same query with different paramTypes should get different canonical names")
 	require.Equal(t, textOID, psi2.ParamTypes,
 		"psi2 should have TEXT param types, not INT4")
+}
+
+func TestConsolidator_RemoveSameQueryDifferentParamTypes(t *testing.T) {
+	consolidator := NewConsolidator()
+
+	query := "SELECT $1"
+	int4OID := []uint32{23}
+	textOID := []uint32{25}
+
+	psiInt4, err := consolidator.AddPreparedStatement(1, "s1", query, int4OID)
+	require.NoError(t, err)
+
+	psiText, err := consolidator.AddPreparedStatement(2, "s2", query, textOID)
+	require.NoError(t, err)
+	require.NotEqual(t, psiInt4.Name, psiText.Name)
+
+	// Remove the INT4 variant — the TEXT variant must survive.
+	consolidator.RemovePreparedStatement(1, "s1")
+
+	require.Nil(t, consolidator.GetPreparedStatementInfo(1, "s1"))
+	require.Equal(t, psiText, consolidator.GetPreparedStatementInfo(2, "s2"))
+
+	// The TEXT entry should still be in stmts.
+	_, exists := consolidator.stmts[dedupKey(query, textOID)]
+	require.True(t, exists)
+
+	// The INT4 entry should be gone.
+	_, exists = consolidator.stmts[dedupKey(query, int4OID)]
+	require.False(t, exists)
 }
 
 func TestConsolidator_Stats(t *testing.T) {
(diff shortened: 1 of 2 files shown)


go/common/preparedstatement/pooler_consolidator.go
```diff
@@ -0,0 +1,80 @@
...
+	pc.mu.Lock()
+	defer pc.mu.Unlock()
+
+	if name, ok := pc.stmts[key]; ok {
+		return name
+	}
+
+	name := fmt.Sprintf("ppstmt%d", pc.lastID)
+	pc.lastID++
+	pc.stmts[key] = name
```
haritabh17, line 64: Are we never removing entries from this map?

go/common/preparedstatement/pooler_consolidator.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// dedupKey builds a deduplication key from query text and param type OIDs.
+func dedupKey(query string, paramTypes []uint32) string {
+	if len(paramTypes) == 0 {
+		return query
+	}
+	var b strings.Builder
+	b.WriteString(query)
+	b.WriteByte(0) // null byte separator — cannot appear in valid SQL
```
haritabh17: Covering escape string literals (e.g E'\x00'), bytea contexts?? Can "SELECT E'\x00'" + no params collide with "SELECT E'" + paramTypes [0]? Very unlikely to hit in practice but maybe a length-prefix scheme (fmt.Sprintf("%d:", len(query)) before the query) would make it tighter?

go/common/preparedstatement/consolidator_test.go
```diff
@@ -275,6 +275,32 @@ func TestConsolidator_RemoveConnection_NonExistent(t *testing.T) {
 	consolidator.RemoveConnection(999)
 }
 
+// TestConsolidator_SameQueryDifferentParamTypes verifies that the
+// Consolidator incorrectly consolidates statements with the same query text
```
haritabh17: nit: Consolidator _does_ _not_ incorrectly consolidate them ... ??

go/common/preparedstatement/consolidator.go
```diff
@@ -173,7 +174,7 @@ func (psc *Consolidator) RemovePreparedStatement(connId uint32, name string) {
 	if exists {
 		psc.usageCount[psi] -= 1
 		if psc.usageCount[psi] == 0 {
-			delete(psc.stmts, psi.Query)
+			delete(psc.stmts, dedupKey(psi.Query, psi.ParamTypes))
```
haritabh17, line 177: There's no test that adds two statements with the same query but different paramTypes, removes one, and checks the other survives. Would be good to add — this is the most likely regression point if dedupKey or ParamTypes storage changes.

go/common/preparedstatement/pooler_consolidator.go
GuptaManan100: Correct, entries are never removed. The set of unique `(query, paramTypes)` pairs is bounded by the application's query surface (finite number of SQL statements), so in practice this doesn't grow without bound. Added a comment to the type doc explaining this design choice.

go/common/preparedstatement/pooler_consolidator.go
GuptaManan100: Good catch. Switched to a length-prefix scheme (`fmt.Sprintf("%d:", len(query))` before the query text) so the key is unambiguously parseable regardless of query content. The null byte separator is gone.

go/common/preparedstatement/consolidator_test.go
GuptaManan100: Fixed, changed to "does not incorrectly consolidate".

go/common/preparedstatement/consolidator.go
GuptaManan100: Added `TestConsolidator_RemoveSameQueryDifferentParamTypes`, adds two statements with the same query but different paramTypes, removes one, and verifies the other survives with correct stmts map entries.

PR #758, opened 2026-03-23
refactor(poolerwatch): unified lifecycle-aware PoolerCache for multigateway + multiorch

dweitzman:
## Summary

Consolidates pooler discovery and lifecycle tracking across multigateway and multiorch into a single typed, lifecycle-aware `poolerwatch.PoolerCache[T]`. Both services used to maintain their own ad-hoc caches over slightly different etcd watch wrappers; that duplication is gone.

Mostly a refactor, with two intentional behavior changes detailed below.

## What's worth knowing for review

**Cache shape.** `PoolerCache[T]` is generic over a per-pooler "rider" type. The consumer attaches state via three hooks bound at `Start` (`OnLive`, `OnUpdate`, `OnGone`). Two grace windows (`ShutdownGrace`, `MissingGracePeriod`) govern how long entries survive after lifecycle SHUTDOWN or NoNode events respectively. Tombstones (post-SHUTDOWN records of departed poolers) are retained outside the read-visible map for external etcd-cleanup loops.

**Concurrency model.** `c.mu` is always released before invoking any hook (so consumer-side locks can't deadlock against the cache lock). Multiorch's rider (`*store.Pooler`) uses `atomic.Pointer` + a per-rider mutex for copy-on-write semantics on `Mutate`, fixing a race the refactor surfaced under `go test -race`.

**Behavior change 1: cohort drift / tombstone REMOVE.** `CohortMismatchAnalyzer` now proposes REMOVE for cohort members that have either tombstoned (unconditional) or disappeared from the cache (gated by a new `commonconsensus.IsCohortMemberRemovalSafe` check against the rule's durability policy). One REMOVE per cycle per shard, with the existing `ExpectedOutgoingRule` CAS pinning. Before this, cohorts could carry stale entries indefinitely.

**Behavior change 2: missing-grace eviction reshape.** Main's orch had `forgetLongUnseenInstances`, a bookkeeping pass that evicted any pooler whose `LastSeen` exceeded 4h — health-driven, ignored topology. The refactor moves eviction into the cache. To avoid regressing on the "etcd lost an entry, pooler still healthy" case, `Config.LastReachedTimestamp` lets the consumer report the most recent wire contact; the cache's sweep deadline becomes `max(disposeAfter, lastReached + MissingGracePeriod)`. Net effect vs main: stricter — entries now need to be **both** missing from etcd **and** silent on the wire to be evicted.

## Known follow-ups

- **`PoolerHealthState` proto elimination.** It's never on the wire and doesn't need to be a proto. Code TODO + memory entry; will fold the `MultiPooler` rider/cache duplication into the same PR.
- **Broaden `LastReachedTimestamp` to count error responses**, not just successful health snapshots. Today's wiring is "last *successful* contact"; an error response from the pooler also proves it exists, and we should count it. Tracked in memory.


go/services/multiorch/recovery/pooler_watcher.go
dweitzman, line 1: Making this file smaller is one of the goals

go/services/multigateway/discovery.go
dweitzman, line 1: Making this file smaller is one of the goals

dweitzman (commented):
Left some comments on what motivated this refactor

go/services/multiorch/init.go
```diff
@@ -213,7 +213,7 @@ func (mo *MultiOrch) Init() error {
 func (mo *MultiOrch) Shutdown() {
 	mo.senv.GetLogger().Info("multiorch shutting down")
 	if mo.recoveryEngine != nil {
-		mo.recoveryEngine.Stop()
+		mo.recoveryEngine.Shutdown()
```
dweitzman, line 216: Renamed some Stop() methods to Shutdown() where it clarifies that something can be shut down but not necessarily restarted

go/services/multiorch/recovery/analysis/cohort_mismatch_analyzer.go
```diff
@@ -116,6 +126,72 @@ func (a *CohortMismatchAnalyzer) Analyze(sa *ShardAnalysis) ([]types.Problem, er
...
+		}
+		// Tombstones are KNOWN dead — they aren't contributing to the cohort
+		// anyway, so the durability-policy safety check (which asks "would
+		// removal + another failure leave a recruitable quorum?") doesn't
+		// apply. Removing a tombstone makes the rule accurate without
+		// degrading operational fault-tolerance beyond what already exists.
+		//
+		// For non-tombstone missing-from-cache cases the pooler may still be
+		// alive somewhere; gate those on the durability policy.
+		if !isTombstone && !commonconsensus.IsCohortMemberRemovalSafe(sa.HighestShardRule, id) {
```
dweitzman, line 159: With the distinction between known dead poolers (they published "SHUTDOWN" lifecycle stage to etcd) vs just missing or, in the future, potentially unhealthy ones, we can:

1) avoid removing suspected dead or unnecessary poolers if doing so would put the ability to do a failover at risk
2) only remove one dead or missing pooler at a time so the "can we safely remove this pooler" check doesn't fight with itself under concurrency

The removal itself should be a compare-and-swap, so it there shouldn't be correctness errors if for multiple orch are competing and make different choices, for example

go/common/topoclient/poolerwatch/cache.go
```diff
@@ -0,0 +1,1030 @@
...
+//     ShutdownGrace so external cleanup can find it.
+//   - MissingFromTopo → Live (recovery): OnUpdate fires (rider was preserved through
+//     grace; the pooler was never "gone" from caller's POV).
+//   - MissingFromTopo → SHUTDOWN: OnGone(Shutdown) fires (first time for this entry);
+//     entry removed, tombstone retained.
+//   - Tombstone → Live (restart-from-shutdown): OnLive(p, zero) fires fresh —
+//     the previous rider was already released through OnGone; the tombstone
+//     is dropped.
+//   - Tombstone → SHUTDOWN: tombstone timestamp refreshed; no hooks.
+func (c *PoolerCache[T]) applyUpsert(pooler *clustermetadatapb.MultiPooler) {
```
Verolop, line 748: if `Filter` changes dynamically, can an existing entry get stuck here? e.g. orch watch targets shrink, `reconcileCell` still marks this ID as `seen`, so it doesn't get deleted, but then `applyUpsert` returns before updating/evicting it. That seems like the old health stream entry could survive for a shard this orch no longer watches.

dweitzman: Added a comment that `Filter` changes aren't supported at the moment since gates connect to all poolers in a shard and orchs can't dynamically reconfigure what shards they follow

We can implement a SetFilter or RecalculateFilter method in the future if needed for dynamic scenarios

Verolop (approved):
solid! This one looked scarier than it is, but I like the lifecycle tightening.

LGTM :)

PR #760, opened 2026-03-23
feat(multiorch): add recovery control RPCs and optimize end-to-end tests

dweitzman:
Adds four new RPCs to `MultiOrchService` to allow endtoend tests to force an immediate on-demand recovery or pause recovery, which seems cleaner than stopping & starting multiorch and waiting an unknown amount of time to allow recovery to take action

### DIFF
--- a/go/test/endtoend/multiorch/fix_replication_test.go
+++ b/go/test/endtoend/multiorch/fix_replication_test.go
@@ -122,7 +122,7 @@ func TestFixReplication(t *testing.T) {
 	// Trigger recovery to fix the replication
 	// Use longer timeout for first recovery since multiorch needs to discover poolers
 	t.Log("Triggering recovery to fix replication...")
-	setup.TriggerRecoveryNow(t, "multiorch", 60*time.Second)
+	setup.TriggerRecoveryNow(t, "multiorch", 10*time.Second)
 
 	// Verify data IS now visible on replica after fix
 	t.Log("Verifying data IS now visible on replica after fix...")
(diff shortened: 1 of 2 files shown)


go/test/endtoend/multiorch/demote_stale_primary_test.go
```diff
@@ -266,7 +266,12 @@ func waitForDivergenceRepaired(t *testing.T, setup *shardsetup.ShardSetup, oldPr
 	oldPrimary := setup.GetMultipoolerInstance(oldPrimaryName)
 	require.NotNil(t, oldPrimary, "old primary instance should exist")
 
-	shardsetup.EventuallyPoolerCondition(t, []*shardsetup.MultipoolerInstance{oldPrimary}, timeout, 2*time.Second,
+	// Trigger recovery and wait for it to complete
+	t.Log("Triggering recovery to detect and repair stale primary...")
+	setup.TriggerRecoveryNow(t, "multiorch", timeout)
```
dweitzman, line 271: Here's an example of a test that triggers recovery on-demand instead of just waiting

go/test/endtoend/multiorch/fix_replication_test.go
```diff
@@ -140,18 +141,16 @@ func TestFixReplication(t *testing.T) {
...
 	t.Log("First fix completed successfully, breaking replication again...")
 
-	// Stop multiorch before breaking replication again
-	t.Log("Stopping multiorch before breaking replication again...")
-	multiorch := setup.GetMultiOrch("multiorch")
-	require.NotNil(t, multiorch, "multiorch instance should exist")
-	multiorch.Stop()
+	// Disable recovery while we break replication and assert the broken state,
+	// preventing the background loop from racing against our assertions.
+	enableRecovery := setup.DisableRecovery(t, "multiorch")
```
dweitzman, line 153: Here's an example of a test disabling orch recovery to verify that the shard is in a broken state before triggering an on-demand recovery

dweitzman:
Reviving this from an old experimental branch in preparation to use it for a pg_rewind-related test in an upcoming PR: I want to verify that nothing changes after orch runs, so it's useful to be able to do an on-demand recovery run and assert that nothing changed instead of waiting some arbitrary amount of time to make sure that nothing changed

PR #761, opened 2026-03-24
fix: extended query protocol fixes and k8s demo configuration

GuptaManan100:
## Summary

- Fix extended query protocol desynchronization with pgx and other pipelined clients (Parse/Bind errors no longer send premature ReadyForQuery)
- Always send ParameterDescription after Describe, even with zero parameters, per PostgreSQL protocol spec
- Allow re-parsing with the same statement name (replace instead of error), matching PostgreSQL behavior for retry scenarios
- Fix k8s demo: add auth credentials, set pooler connection capacity to 50, widen port range, add `sslmode=disable`

## Problem

The extended query protocol implementation had three issues causing failures with pg17 and pgx clients:

1. `handleParse` and `handleBind` sent `ReadyForQuery` on error, but in pipelined extended query protocol, `ReadyForQuery` must only come from `Sync`. This caused pgx to think the pipeline was complete, then read stale responses from subsequent messages (Describe, Sync), corrupting the next query's response stream.

2. `handleDescribe` skipped `ParameterDescription` when there were zero parameters. The PostgreSQL protocol requires it unconditionally after Describe of type 'S', causing clients to misinterpret the next message.

3. The prepared statement consolidator rejected re-parsing with an existing name. PostgreSQL allows this (replacing the old statement), which is needed when Parse succeeds but the subsequent Describe fails and the client retries.

Additionally, the k8s demo was missing authentication configuration and had suboptimal pooler/network settings.

## Solution

- Removed `writeReadyForQuery` calls from `handleParse` and `handleBind` error paths
- Made `writeParameterDescription` unconditional in `handleDescribe`
- Changed `AddPreparedStatement` to replace existing named statements instead of erroring
- Updated k8s manifests with correct auth, pooler capacity, and port range settings

### DIFF
--- a/demo/k8s/k8s-supafirehose.yaml
+++ b/demo/k8s/k8s-supafirehose.yaml
@@ -20,6 +20,10 @@ spec:
         - ip: "10.96.100.1"
           hostnames:
             - "multigateway"
+      securityContext:
+        sysctls:
+          - name: net.ipv4.ip_local_port_range
+            value: "1024 65535"
       containers:
         - name: supafirehose
           image: supafirehose:latest
@@ -38,15 +42,6 @@ spec:
             limits:
               memory: "2Gi"
               cpu: "2000m"
-          lifecycle:
-            postStart:
-              exec:
-                command:
-                  - /bin/sh
-                  - -c
-                  - echo "1024 65535" > /proc/sys/net/ipv4/ip_local_port_range
-          securityContext:
-            privileged: true
 ---
 apiVersion: v1
 kind: Service
--- a/go/common/preparedstatement/consolidator.go
+++ b/go/common/preparedstatement/consolidator.go
@@ -17,6 +17,7 @@ package preparedstatement
 import (
 	"errors"
 	"fmt"
+	"log/slog"
 	"sync"
 
 	"github.com/multigres/multigres/go/common/parser"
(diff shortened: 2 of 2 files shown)


demo/k8s/k8s-supafirehose.yaml
```diff
@@ -38,10 +38,15 @@ spec:
...
+              exec:
+                command:
+                  - /bin/sh
+                  - -c
+                  - echo "1024 65535" > /proc/sys/net/ipv4/ip_local_port_range
           securityContext:
-            capabilities:
-              add:
-                - SYS_RESOURCE
+            privileged: true
```
haritabh17: NET_ADMIN capability would be sufficient? Or an init container with sysctl — avoids giving the firehose container full host device access??

go/common/preparedstatement/consolidator.go
```diff
@@ -127,9 +127,18 @@ func (psc *Consolidator) AddPreparedStatement(connId uint32, name, queryStr stri
...
 
-	// If the name is non-empty, and a prepared statement for this name already exists on the connection, we throw an error.
-	if _, exists := psc.incoming[connId][name]; exists && name != "" {
-		return nil, errors.New("Prepared statement with this name exists")
+	// If the name is non-empty and a prepared statement for this name already exists
+	// on the connection, replace it. This matches PostgreSQL behavior where re-parsing
+	// with an existing name replaces the old statement. This is necessary to handle
+	// the case where Parse succeeds (adding to consolidator) but the subsequent
+	// Describe fails — the client retries Parse with the same name.
+	if existing, exists := psc.incoming[connId][name]; exists && name != "" {
```
haritabh17, line 136: Nit: this replacement is silent — if a named statement gets replaced due to an actual client bug (not the Parse-ok-Describe-fail retry), there's no signal to debug it. A slog.Debug here with the old and new query strings would be cheap and helpful for troubleshooting?

PR #762, opened 2026-03-24
feat(consensus): select promotion candidate by leadership_term then LSN

mkindahl:
Add read functions for multigres.leadership_history and use the highest replicated consensus term (leadership_term) as the primary criterion when selecting which standby to promote, with LSN as the tiebreaker.

Using leadership_term instead of raw LSN makes candidate selection protocol-native: a standby with a higher term has definitively applied more of the agreed consensus history, regardless of raw WAL position. leadership_history writes use RemoteOperationTimeout so synchronous standbys are guaranteed to have acknowledged them before a write returns.

Two-level ordering in selectCandidate:
1. leadership_term — highest consensus term replicated locally (primary)
2. lsn             — tiebreaker when terms are equal

timeline_id is still captured in WALPosition and included in the post-selection log message for observability, but does not affect which candidate is chosen.

When leadership_term is 0 for all candidates (empty leadership_history, e.g. a freshly bootstrapped cluster), LSN is the sole tiebreaker — identical to the previous behaviour.

Add queryLeadershipHistory and currentLeadershipRecord to pg_multischema.go, and populate LeadershipTerm / CohortMembers in executeRevoke after WAL replay stabilizes.

Fixes MUL-263

### DIFF
--- a/go/services/multiorch/consensus/leader_appointment.go
+++ b/go/services/multiorch/consensus/leader_appointment.go
@@ -125,103 +125,133 @@ func (c *Coordinator) discoverMaxTerm(cohort []*multiorchdatapb.PoolerHealthStat
 	return maxTerm, nil
 }
 
+// walPositionLSN extracts and parses the most relevant LSN from a WALPosition.
+// For a primary node CurrentLsn is used; for a standby, LastReceiveLsn.
+// Returns (0, false) when pos is nil, both LSN fields are empty, or the string
+// cannot be parsed.
+func walPositionLSN(pos *consensusdatapb.WALPosition) (pglogrepl.LSN, bool) {
+	if pos == nil {
+		return 0, false
+	}
+
+	// Get either the CurrentLsn (for primaries) or LastReceiveLsn (for
+	// standbys). Both field may be empty if the server is not a primary or
+	// standby, or if the pooler hasn't fully initialized and fetched WAL
+	// position yet.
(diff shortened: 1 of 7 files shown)


go/services/multiorch/consensus/leader_appointment.go
```diff
@@ -125,103 +125,123 @@ func (c *Coordinator) discoverMaxTerm(cohort []*multiorchdatapb.PoolerHealthStat
...
+// walPositionLSN extracts and parses the most relevant LSN from a WALPosition.
+// For a primary node CurrentLsn is used; for a standby, LastReceiveLsn.
+// Returns (0, false) when pos is nil, both LSN fields are empty, or the string
+// cannot be parsed.
+func walPositionLSN(pos *consensusdatapb.WALPosition) (pglogrepl.LSN, bool) {
+	if pos == nil {
+		return 0, false
+	}
+	lsnStr := pos.CurrentLsn
+	if lsnStr == "" {
```
dweitzman, line 142: Are you expecting CurrentLSN to be empty in any particular situation?

mkindahl: It can happen if the server is not in a state where it is either a primary or an active standby. It should not normally happen, but defensively checking it makes sure that we do not act on inaccurate information.

mkindahl: I added some comments to make the situation clear.

PR #763, opened 2026-03-24
feat(multigateway): add plan-level instrumentation to executor, spans, and query logs

haritabh17:
## Summary

- Add plan type and plan timing to the multigateway query observability path (Phase 2 observability, workstreams 2+3)
- Executor captures planning duration and selected primitive, surfaces both through `ExecuteResult` to the handler
- Handler enriches `gateway.query` spans with `db.plan.type` and `db.tables_used` attributes
- Structured query logs gain `duration.plan`, `db.plan.type`, and `db.tables_used` fields

## Description

Operators previously could not distinguish whether a slow query was spending time in planning or execution, or filter traces by plan type. This PR threads plan metadata from the planner through the executor to the handler's observability layer.

**engine/plan.go** — `Type string` field on `Plan` struct, populated by the planner.

**planner/planner.go** — `primitiveName()` maps all 7 primitive types (Route, Transaction, CopyStatement, ApplySessionState, GatewaySessionState, GatewayShowVariable, Sequence) to string labels. `plan.Type` is set after planning.

**executor/executor.go** — Plan timing captured around `planner.Plan()` in `StreamExecute` and around `planner.PlanPortal()` in `PortalStreamExecute` (both plan and no-plan paths). `PlanType` and `PlanTime` populated in `ExecuteResult`.

**handler/handler.go** — `recordQueryCompletion` takes `*ExecuteResult` instead of `[]string`, extracts plan metadata, calls `setSpanPlanAttributes`, and passes plan fields to query log. Multi-statement batches get nil result (per-table metrics already emitted per-statement inside `executeWithImplicitTransaction`).

**handler/tracing.go** — `setSpanPlanAttributes()` sets `db.plan.type` and `db.tables_used` on the active span.

**handler/querylog.go** — `PlanDuration`, `PlanType`, `TablesUsed` added to `queryLogEntry`. Emitted as `duration.plan`, `db.plan.type`, `db.tables_used`.

No new metrics introduced. Test added for `primitiveName()` and `newTestHandler` hardened with metrics initialization.

Closes MUL-264

  ### Traces (Grafana Tempo)                                                                                                                      
Screenshot 2026-03-24 at 13 47 25
                                               
                                                                      
  ### Logs (Grafana Loki)                              
Screenshot 2026-03-24 at 13 57 45

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,6 +45,7 @@ require (
 	go.opentelemetry.io/otel/sdk/log v0.15.0
 	go.opentelemetry.io/otel/sdk/metric v1.40.0
 	go.opentelemetry.io/otel/trace v1.40.0
+	go.opentelemetry.io/proto/otlp v1.9.0
 	go.uber.org/goleak v1.3.0
 	golang.org/x/crypto v0.48.0
 	golang.org/x/mod v0.33.0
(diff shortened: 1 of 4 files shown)


haritabh17:
> Can we add some e2e tests to see we get the required traces/spans?

Done now ✅✅

PR #766, opened 2026-03-25
feat(multiorch): add max_cycles to TriggerRecoveryNow and refactor test helpers

dweitzman:
These are mostly cleanups, plus TriggerRecoveryNow() is set up so it can do exactly one recovery cycle in preparation to add a test that verifies orch doesn't do anything or detect anything after running for only one cycle (vs running for a test amount of time)

### DIFF
--- a/go/test/endtoend/queryserving/buffer_test.go
+++ b/go/test/endtoend/queryserving/buffer_test.go
@@ -366,7 +366,7 @@ func triggerFailover(t *testing.T, setup *shardsetup.ShardSetup) {
 	t.Logf("BeginTerm accepted, emergency demotion triggered")
 
 	// Trigger immediate recovery to elect a new primary and fully stabilize the cluster.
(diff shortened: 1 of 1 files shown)


proto/multiorchservice.proto
```diff
@@ -119,7 +119,11 @@ message GetRecoveryStatusResponse {
 }
 
 message TriggerRecoveryNowRequest {
-  // Empty - timeout controlled by gRPC request deadline
+  // max_cycles limits the number of recovery cycles to run.
+  // If 0 (default), cycles continue until all problems are resolved or the deadline expires.
+  // If 1, exactly one cycle is run and the call returns immediately with any remaining problem codes.
+  // Values greater than 1 are rejected with INVALID_ARGUMENT.
+  int32 max_cycles = 1;
```
cuongdo: because negative values seem nonsensical, should this be a `uint32`?

dweitzman: Done

PR #767, opened 2026-03-25
test(multiorch): add TestRewindDivergedReplica

dweitzman:
TestRewindDivergedReplica exercises FixReplicationAction.tryPgRewind(), the code path taken when a replica has timeline divergence. It promotes a replica to timeline 2, writes a diverging row, restarts it as a standby, then verifies orch detects the not-replicating problem, runs pg_rewind via RewindToSource RPC, and the replica rejoins with consistent data.


dweitzman:
The first commit is just https://github.com/multigres/multigres/pull/766, so only the new file (go/test/endtoend/multiorch/rewind_diverged_replica_test.go) needs to be reviewed here

mkindahl (approved):
LGTM

go/services/multiorch/recovery/actions/fix_replication.go
```diff
@@ -368,6 +368,17 @@ func (a *FixReplicationAction) verifyReplicaNotReplicating(
 		return true, status, nil
 	}
 
+	// Check that the WAL receiver is actually streaming. primary_conninfo being set
+	// is not sufficient — after a failed or partial pg_rewind attempt the conninfo
+	// may be preserved on disk while the WAL receiver is not running. Without this
+	// check orch would consider the problem resolved and never retry the fix.
+	if status.WalReceiverStatus != "streaming" {
```
dweitzman: Made a small fix to the code based on test failures

go/services/multipooler/internal/manager/rpc_manager.go
```diff
@@ -922,23 +926,43 @@ func (pm *MultiPoolerManager) restartAsStandbyLocked(
...
+		//
+		// Gated on wantRewind rather than rewindPerformed: backup_label from a
+		// prior interrupted rewind attempt may still be present even when the
+		// current dry-run detects no divergence (already rewound).
+		//
+		// This must succeed before rewindPending is cleared. If it fails, the
+		// caller should retry: pg_rewind is idempotent on an already-rewound
+		// target (dry-run detects no divergence and skips the actual rewind),
+		// so the retry will reach this step again.
+		if err := pm.setAutoConfRecoveryTimeline(ctx, "current"); err != nil {
```
dweitzman: New change: before restarting postgres, `recovery_target_timeline` is set to `current`, and after restarting that gets cleared out

`latest` works well for a newly-started replica but can fail on an existing replica that had divergence.

dweitzman:
@mkindahl Re-requesting review because there's a substantive change: flipping `recovery_target_timeline` from `latest` to `current` while recovery runs

go/services/multipooler/internal/manager/rpc_manager.go
mkindahl: Good catch, but you might want to point out that the main reason is that we use a shared archive for `pgbackrest`.

This creates problems because recovery finds a history file for the newly started TLI, but that is not relevant in this particular case and it should start streaming WAL to move over to the next TLI rather than trying to switch TLI during recovery (Postgres handles this in general, but not for this particular case).

If we didn't have a shared archive for recovery, this would not be a problem (but we would have other problems).

mkindahl (approved):
Still LGTM. It was an interesting edge-case that you discovered.

go/services/multipooler/internal/manager/rpc_manager.go
NikolayS: +1 to what @mkindahl says. I wonder if the use of archive could be completely avoided during standby startup. This would help completely avoid the problem and there would be no need in flipping recovery_target_timeline.

PR #769, opened 2026-03-26
feat(multigateway): add prepared statement handling in simple protocol

GuptaManan100:
## Summary

- Add support for `PREPARE`, `EXECUTE`, and `DEALLOCATE` (including `DEALLOCATE ALL`) in the simple query protocol path
- Introduce `PreparedStatementPrimitive` in the engine that delegates to existing extended query protocol handlers (`HandleParse`, `HandleBind`, `HandleClose`)
- Add `HandleCloseAll` to the `Handler` interface for `DEALLOCATE ALL` support
- Add planner methods to route these statement types to the new primitive

## Description

Clients using the simple query protocol (e.g. `database/sql` Go driver, `psql` without explicit extended protocol) send `PREPARE`/`EXECUTE`/`DEALLOCATE` as plain SQL text. Previously the gateway didn't recognize these statements and would forward them directly to the backend, bypassing the prepared statement consolidator.

This change teaches the planner to recognize these statement types and creates a `PreparedStatementPrimitive` that reuses the existing extended query protocol machinery. `PREPARE` calls `HandleParse`, `EXECUTE` calls `HandleBind` + `PortalStreamExecute`, and `DEALLOCATE` calls `HandleClose`. This keeps prepared statement state consistent regardless of which protocol path the client uses.

`DEALLOCATE ALL` has no single-statement extended protocol equivalent, so a new `HandleCloseAll` method is added to the `Handler` interface that removes all statements for the connection from the consolidator.


go/services/multigateway/engine/prepared_statement_primitive.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+}
+
+// executePrepare delegates to HandleParse to register the statement in the consolidator.
+func (p *PreparedStatementPrimitive) executePrepare(
+	ctx context.Context,
+	conn *server.Conn,
+	callback func(context.Context, *sqltypes.Result) error,
+) error {
+	if err := conn.Handler().HandleParse(ctx, conn, p.stmtName, p.innerQuery, nil); err != nil {
```
haritabh17: The nil drops the parameter type OIDs, means the backend has to infer types that the client explicitly declared — this can produce wrong results with overloaded functions or polymorphic types??

go/services/multigateway/engine/prepared_statement_primitive.go
```diff
@@ -0,0 +1,279 @@
...
+
+// executeDeallocate delegates to HandleClose for named statements.
+// DEALLOCATE ALL uses the consolidator directly since there is no
+// single-statement extended protocol equivalent.
+func (p *PreparedStatementPrimitive) executeDeallocate(
+	ctx context.Context,
+	conn *server.Conn,
+	callback func(context.Context, *sqltypes.Result) error,
+) error {
+	if err := conn.Handler().HandleClose(ctx, conn, 'S', p.stmtName); err != nil {
```
haritabh17: silently no-ops if the statement doesn't exist. PostgreSQL returns does not exist error for DEALLOCATE nonexistent stmts. Should we match that behavior? Have e2e test for this case?

go/common/pgprotocol/server/conn.go
```diff
@@ -215,6 +215,17 @@ func (c *Conn) ConnectionID() uint32 {
...
 }
 
+// Handler returns the protocol handler for this connection.
+func (c *Conn) Handler() Handler {
+	return c.handler
+}
+
+// SetHandler sets the protocol handler for this connection.
+// This is intended for testing and for connections that need to override the handler.
+func (c *Conn) SetHandler(h Handler) {
```
haritabh17: handler was previously a private field set once at connection startup in the listener. Exposing it with a public getter/setter can create a circular dependency path??

go/services/multigateway/engine/prepared_statement_primitive.go
GuptaManan100: Good catch, fixed

go/services/multigateway/engine/prepared_statement_primitive.go
GuptaManan100: Good catch, fixed

go/common/pgprotocol/server/conn.go
GuptaManan100: Yep, I simplified to remove this method

go/common/mterrors/code.go
```diff
@@ -196,6 +197,20 @@ func NewUnrecognizedParameter(name string) *PgDiagnostic {
 		fmt.Sprintf("unrecognized configuration parameter %q", name), "")
 }
 
+// NewInvalidPreparedStatementError creates a PgDiagnostic for a reference to
+// a nonexistent prepared statement. SQLSTATE 26000 (invalid_sql_statement_name).
+func NewInvalidPreparedStatementError(name string) *PgDiagnostic {
+	return NewPgError("ERROR", PgSSInvalidSQLStatementName,
```
haritabh17, line 203: nit: Do you mean `InvalidPreparedStatementname` ?

GuptaManan100: I don't follow. What do you want me to change? The function name?

PR #771, opened 2026-03-26
refactor(multipooler): replace applicationName with poolerID type

dweitzman:
Replace the applicationName string type with a poolerID struct that holds both the proto ID and its derived PostgreSQL application_name. This eliminates repeated back-and-forth conversions between application names and proto IDs, and provides a single validated identity type callers can pass around.

### DIFF
--- a/go/services/multipooler/manager/pg_multischema.go
+++ b/go/services/multipooler/manager/pg_multischema.go
@@ -337,7 +337,7 @@ func (pm *MultiPoolerManager) insertDurabilityPolicy(ctx context.Context, policy
 // This is used for both promotion events and replication config changes.
 // This operation uses the remote-operation-timeout and will fail if it cannot complete within
 // that time. A timeout typically indicates that synchronous replication is not functioning.
-func (pm *MultiPoolerManager) insertHistoryRecord(ctx context.Context, termNumber int64, eventType string, leaderID applicationName, coordinatorID *clustermetadatapb.ID, walPosition, operation, reason string, cohortMembers, acceptedMembers []applicationName, force bool) error {
+func (pm *MultiPoolerManager) insertHistoryRecord(ctx context.Context, termNumber int64, eventType string, leaderID poolerID, coordinatorID *clustermetadatapb.ID, walPosition, operation, reason string, cohortMembers, acceptedMembers []poolerID, force bool) error {
 	if force {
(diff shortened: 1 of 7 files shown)


go/services/multipooler/manager/pg_multischema.go
```diff
@@ -337,7 +337,7 @@ func (pm *MultiPoolerManager) insertDurabilityPolicy(ctx context.Context, policy
 // This is used for both promotion events and replication config changes.
 // This operation uses the remote-operation-timeout and will fail if it cannot complete within
 // that time. A timeout typically indicates that synchronous replication is not functioning.
-func (pm *MultiPoolerManager) insertHistoryRecord(ctx context.Context, termNumber int64, eventType string, leaderID applicationName, coordinatorID *clustermetadatapb.ID, walPosition, operation, reason string, cohortMembers, acceptedMembers []applicationName, force bool) error {
+func (pm *MultiPoolerManager) insertHistoryRecord(ctx context.Context, termNumber int64, eventType string, leaderID poolerID, coordinatorID *clustermetadatapb.ID, walPosition, operation, reason string, cohortMembers, acceptedMembers []poolerID, force bool) error {
```
dweitzman, line 340: I think this PR is a reasonable cleanup on its own, but one of the motivations is so we can have some kind of a "Rules" type and have methods like this one easily return the full rules

It gets confusing when some method deal in proto3 ID values and some deal in applicationNames, so my solution is just to bundle the two together and say a method like this one treats the two as interchangeable as long as you've prevalidated that an ID is applicationName safe (not too long, no invalid characters, etc.)

go/services/multipooler/manager/pg_multischema_test.go
```diff
@@ -377,17 +384,25 @@ func TestInitializeMultischemaData(t *testing.T) {
 	}
 }
 
+func mustPoolerID(cell, name string) poolerID {
```
mkindahl, line 387: Why "must" PoolerID? Did you mean `makePoolerID`? I have no strong opinion on the name, so either one is fine, but wondering if it is an unintentional typo.

dweitzman: "must" because it panic()s if validation fails

Similar to something like "MustCompile" in the standard library: https://pkg.go.dev/regexp#MustCompile

PR #774, opened 2026-03-27
fix(etcd): use /health instead of /readyz for etcd health checks

mkindahl:
/readyz is only served on etcd's metrics listener (--listen-metrics-urls), which is not configured in local/test setups. This caused waitForServiceReady to time out after 60 attempts with 404 responses. /health is always available on the client listener in etcd 3.x.

Also adds explicit tests for checkEtcdHealth and WaitForReady. checkEtcdHealth had zero test coverage; without it a regression like reverting /health back to /readyz would go undetected. WaitForReady was only exercised implicitly via StartEtcd

Fixes MUL-272

### DIFF
--- a/go/common/topoclient/etcdtopo/test_helpers.go
+++ b/go/common/topoclient/etcdtopo/test_helpers.go
@@ -43,7 +43,7 @@ func checkPortAvailable(port int) error {
 	return nil
 }
 
-// WaitForReady waits for etcd to be ready by querying its /readyz endpoint in a loop.
+// WaitForReady waits for etcd to be ready by querying its /health endpoint in a loop.
 // The loop will retry until the endpoint returns 200 OK or the context is cancelled.
 // Callers should pass a context with an appropriate timeout.
 func WaitForReady(ctx context.Context, clientAddr string) error {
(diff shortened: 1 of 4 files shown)


cuongdo:
This may cause a regression in our end to end tests, as well as k8s health checking.

See PR #348

Verolop:
I think we should keep /readyz as the readiness contract and fix the etcd startup/config path instead of changing this check to /health.

#348  already moved multigres toward the kube-style readiness model, and multigres-operator is already on that same contract today with /readyz for readiness and /livez for liveness. So the right fix is to make the local/test etcd setup conform to that model, so multigres and the Operator stay aligned.

mkindahl:
> I think we should keep /readyz as the readiness contract and fix the etcd startup/config path instead of changing this check to /health.
> 
> #348 already moved multigres toward the kube-style readiness model, and multigres-operator is already on that same contract today with /readyz for readiness and /livez for liveness. So the right fix is to make the local/test etcd setup conform to that model, so multigres and the Operator stay aligned.

I hadn't noticed the switch to `/readyz` and I agree on using the k8s style conventions, that is, using a separate metrics port rather than sending requests to the client port. Closing this PR and created #775 instead, which fixes the provisioner and the demo to configure the metrics port.

PR #776, opened 2026-03-27
fix(provisioner): skip pgbackrest config when backup is disabled

mkindahl:
In a minimal local setup with backup.type: none, provisionMultipooler was unconditionally passing pgbackrest TLS cert paths to the multipooler process and fetching pgctld config solely to get the pgbackrest port. This fails when pgbackrest is not provisioned.

- Guard the pgbackrest args block in provisionMultipooler with `if p.pgBackRestCertPaths != nil`, matching the existing pattern in pgctld.go
- Skip generatePgBackRestCertsOnce in ProvisionDatabase when backup.type is "none"
- Add "none" as a valid backup type in buildBackupLocation and buildBackupConfig

Fixes MUL-275

### DIFF
--- a/go/provisioner/local/config.go
+++ b/go/provisioner/local/config.go
@@ -203,6 +203,8 @@ func buildBackupConfig(backupConfig map[string]string, baseDir string) BackupCon
 	config := BackupConfig{Type: backupType}
 
 	switch backupType {
+	case "none":
+		// No backup configured
 	case "local":
 		path := backupConfig["path"]
 		if path == "" {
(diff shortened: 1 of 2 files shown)


mkindahl:
Closing this — after re-tracing against current `main`, the premise ("this fails / panics when pgbackrest is not provisioned") doesn't hold:

- `generatePgBackRestCertsOnce` runs unconditionally in `ProvisionDatabase` and always sets `pgBackRestCertPaths` *before* the only caller of `provisionMultipooler`, so the `p.pgBackRestCertPaths.ServerCertFile` deref can never be nil through that path. There is no reachable panic.
- The nil only appears *as a consequence of this PR's own cert-skip change*, which the added `if p.pgBackRestCertPaths != nil` guard then defends against. So the guard is coupled to a problem the PR itself introduces, not a latent crash in `main`.

What *is* real: setting `backup.type: none` today produces a confusing error — `buildBackupLocation` has no `none` case and falls through to `default:`, returning `unknown backup type: none`. That's a poor UX for "backups disabled," but it's a clean error, not a crash.

I'll address the `backup.type: none` handling (making it a supported, first-class "backups disabled" option) in a separate change rather than carrying this stale branch (448 commits behind).

PR #804, opened 2026-04-02
fix(multigateway): allow ROLLBACK TO SAVEPOINT in aborted transaction state

haritabh17:
## Summary

- Allow `ROLLBACK TO SAVEPOINT`, `COMMIT`, and `ROLLBACK` to proceed when a transaction is in the aborted (failed) state, matching PostgreSQL behavior
- Transition `TxnStatus` from `Failed` back to `InBlock` after a successful `ROLLBACK TO SAVEPOINT`
- Add e2e tests for savepoint error recovery, nonexistent savepoint, and multiple recovery cycles

## Problem

When a statement fails inside an explicit transaction, multigres correctly enters the aborted state but rejects `ROLLBACK TO SAVEPOINT` with "current transaction is aborted, commands ignored until end of transaction block". PostgreSQL allows `ROLLBACK TO SAVEPOINT` in aborted state — this is the primary mechanism for error recovery within transactions.

```sql
BEGIN;
SAVEPOINT sp1;
SELECT 1/0;          -- error: division by zero
ROLLBACK TO sp1;     -- ERROR: current transaction is aborted (should succeed)
```

This is one of the largest source of cascading failures in the pgregress compatibility tests.

## Solution

- Add `IsAllowedInAbortedTransaction()` AST helper that returns true for `ROLLBACK`, `ROLLBACK TO SAVEPOINT`, and `COMMIT` (the three statements PostgreSQL permits in aborted state)
- Update both simple query (`HandleQuery`) and extended query (`HandleExecute`) gates to use the new helper instead of `IsRollbackStatement`
- Add explicit `TRANS_STMT_ROLLBACK_TO` case in `TransactionPrimitive.StreamExecute` that passes through to the backend and transitions `TxnStatus` from `Failed` to `InBlock` on success

### DIFF
--- a/go/common/parser/ast/helper.go
+++ b/go/common/parser/ast/helper.go
@@ -42,6 +42,23 @@ func IsRollbackStatement(stmt Stmt) bool {
 	return txStmt.Kind == TRANS_STMT_ROLLBACK
 }
 
+// IsAllowedInAbortedTransaction returns true if the statement may proceed when
+// the transaction is in the aborted (failed) state. PostgreSQL allows ROLLBACK,
+// ROLLBACK TO SAVEPOINT, and COMMIT (which it converts to ROLLBACK with a
+// WARNING) in this state. All other statements are rejected with SQLSTATE 25P02.
+func IsAllowedInAbortedTransaction(stmt Stmt) bool {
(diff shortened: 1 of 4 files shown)


GuptaManan100 (approved):
Love it!

PR #809, opened 2026-04-02
refactor(multipooler): merge ReserveStreamExecute into StreamExecute

haritabh17:
## Summary

- Merge `ReserveStreamExecute` into `StreamExecute` by adding `reservation_reasons` and `reservation_begin_query` to `ExecuteOptions`
- Eliminate the inline `BEGIN; <query>` SQL concatenation hack that caused incorrect reservation reason tracking on the multipooler
- Remove `ReserveStreamExecute` RPC, request/response messages, and all implementations across the stack

## Description

`ReserveStreamExecute` was a separate RPC that created a reserved connection and executed a query. This PR folds that functionality into `StreamExecute` via two new fields on `ExecuteOptions`:

- `reservation_reasons` (uint32 bitmask): when set with no `reserved_connection_id`, creates a new reservation; when set with an existing ID, adds reasons to the existing reservation
- `reservation_begin_query` (string): the BEGIN statement to execute when `ReasonTransaction` is in the reasons

This also fixes a design issue where pinned sessions prepended `BEGIN; <query>` as a single SQL string to avoid an extra RPC. The multipooler couldn't properly track `ReasonTransaction` since BEGIN was sent inline. Now the multipooler executes `BeginWithQuery` explicitly before the query, with correct reason tracking.

**Net -529 lines** across 15 files.


proto/query.proto
```diff
@@ -288,4 +288,15 @@ message ExecuteOptions {
...
+  // reservation_reasons is a bitmask of ReservationReason values. When non-zero and
+  // reserved_connection_id is 0, StreamExecute creates a new reserved connection.
+  // When non-zero and reserved_connection_id is set, the reasons are OR'd into
+  // the existing reservation.
+  uint32 reservation_reasons = 6;
+
+  // reservation_begin_query is the original BEGIN statement text to use when
+  // creating a transaction reservation (e.g., "BEGIN ISOLATION LEVEL SERIALIZABLE").
+  // If empty and the transaction reason is set, defaults to "BEGIN".
+  string reservation_begin_query = 7;
```
GuptaManan100: Instead of putting this in execute options, let's use reservation options that already exists. That keeps the separation between the two.

ExecuteOptions is used in many RPCs, but ReservationOptions will only be used in some of them. We can even move ReservationOptions to this file, so that it stays in the same place.

Not super important, we can go either way

go/services/multipooler/executor/executor.go
GuptaManan100: Stale comment needs fixing

go/services/multigateway/scatterconn/scatter_conn.go
```diff
@@ -219,14 +205,14 @@ func (sc *ScatterConn) StreamExecute(
...
 		// so the multipooler preserves transaction options instead of using plain "BEGIN".
 		if state.PendingBeginQuery != "" {
-			reservationOpts.BeginQuery = state.PendingBeginQuery
+			eo.ReservationBeginQuery = state.PendingBeginQuery
 			state.PendingBeginQuery = ""
 		}
-		reservedState, err := sc.gateway.ReserveStreamExecute(ctx, target, sql, eo, reservationOpts, callback)
+		reservedState, err := sc.gateway.StreamExecute(ctx, target, sql, eo, callback)
 		if err != nil {
 			return fmt.Errorf("reserve stream execute failed: %w", err)
```
GuptaManan100: Error message needs fixing

GuptaManan100 (approved):
Test coverage for Case 1 + ReservationReasons: The new code path in executor.go:193-207 (adding BEGIN to an existing reserved connection) is the most novel logic. I don't see a dedicated unit test for this path in the diff. The integration tests presumably cover it, but a unit test in executor_test.go that exercises StreamExecute with both ReservedConnectionId > 0 and ReservationReasons != 0 would be valuable.

Rest LGTM

proto/query.proto
haritabh17: Ack, updated

go/services/multigateway/scatterconn/scatter_conn.go
haritabh17: Ack, updated

go/services/multipooler/executor/executor.go
haritabh17: Ack, updated

Issue #723, opened 2026-03-12
Stale postmaster.pid on retained PVC causes pgctld to skip PostgreSQL startup


fernando-villalba:
### Summary

During rolling restarts with PVC retention (`whenScaled: Retain`), replacement pool pods can permanently fail to start PostgreSQL. The pod appears fully healthy — Kubernetes shows `2/2 Running`, `Ready=True`, and the CRD reports `phase: Healthy` — but no PostgreSQL process exists. All SQL connections to the affected pod fail silently.

### Root Cause

A stale `postmaster.pid` file from the previous pod incarnation remains on the retained PVC. When the replacement pod starts, `pgctld`'s `isPostgreSQLRunning()` check reads the PID from this file and calls `kill(pid, 0)` to verify the process exists. Due to PID reuse inside the container namespace, the stale PID happens to match a running pgctld thread, so the check returns `true`. `StartPostgreSQLWithResult()` then returns early with `AlreadyRunning = true`, **never calling `pg_ctl start`**.

**Code path** (`go/cmd/pgctld/command/start.go`):

```
StartPostgreSQLWithResult() (line 121)
  → isPostgreSQLRunning(dataDir) (line 125)
    → readPostmasterPID(dataDir) → reads PID from stale postmaster.pid (line 238)
    → isProcessRunning(pid) (line 243)
      → os.FindProcess(pid) + process.Signal(syscall.Signal(0)) (lines 420-428)
      → Returns true because the PID exists (it's pgctld, not postgres)
  → Returns early: AlreadyRunning = true, never calls pg_ctl start (lines 126-135)
```

The `isProcessRunning()` function only checks whether a process with that PID exists — it does **not** verify the process is actually `postgres`. In a container, PIDs are small integers (typically < 100), so PID collision between the stale `postmaster.pid` value and a pgctld/pgbackrest thread is very likely.

### Compounding Issue: Readiness Probe

`pgctld` serves HTTP health endpoints (`/live`, `/ready`) that respond `200 OK` as soon as the gRPC/HTTP servers start, independently of PostgreSQL state. This is by design (`server.go` line 251: "server should be able to start even with uninitialized data directory"), but it means Kubernetes marks the pod as `Ready` even though the database never started.

### Impact

- **Silent data unavailability**: PostgreSQL is not running. Multipooler continuously logs `connection refused` to the Unix socket. Multiorch detects `StalePrimary` but defers recovery behind a grace period that never resolves.
[...]

PR #812, opened 2026-04-03
fix(pgctld): stale postmaster pid

BhautikChudasama:
### Fixes
- https://github.com/multigres/multigres/issues/723

### Changes
- Previously, we checked whether a stale process (mentioned in `postmaster.pid`) was still running by sending signal 0 and assuming the process existed if no error occurred. Now, instead of relying on that approach, we verify whether PostgreSQL is actually running using `pg_isready`.

### Reproduce

```shell
➜ /Users/bhautik/workspace/src/multigres/bin/pgctld start \
    --pooler-dir /tmp/local-test \
    --pg-port 5433

...
PostgreSQL server started successfully (PID: 58207)

➜ kill -9 58207

➜ echo $$
75110

# Write obtained process ID of current shell in postmaster.pid

# Start again
➜ /Users/bhautik/workspace/src/multigres/bin/pgctld start \
    --pooler-dir /tmp/local-test \
    --pg-port 5433
# Instead of saying postgresql is running, It should remove pid and start postgres
```

### DIFF
--- a/go/cmd/pgctld/command/start.go
+++ b/go/cmd/pgctld/command/start.go
@@ -122,7 +122,7 @@ func StartPostgreSQLWithResult(logger *slog.Logger, config *pgctld.PostgresCtlCo
 	result := &StartResult{}
 
 	// Check if PostgreSQL is already running
-	if isPostgreSQLRunning(config.PostgresDataDir) {
+	if isPostgreSQLRunning(config) {
 		logger.Info("PostgreSQL is already running")
 		result.AlreadyRunning = true
 		result.Message = "PostgreSQL is already running"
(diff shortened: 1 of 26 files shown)


go/cmd/pgctld/command/start.go
```diff
@@ -416,6 +429,20 @@ func readPostmasterPID(dataDir string) (int, error) {
 	return pid, nil
 }
 
+func IsPostmasterProcessRunning(config *pgctld.PostgresCtlConfig) bool {
```
mkindahl, line 432: I think we have a function for this already, so it might make sense to use that. However, the usage is correct, so we can clean up this after merging.

mkindahl (approved):
Seems straightforward, so approving.

PR #824, opened 2026-04-08
feat(multipooler): add schema tracking for plan cache invalidation

GuptaManan100:
## Summary

- Add a schema tracker to multipooler that detects DDL changes via two mechanisms: a PostgreSQL event trigger (`ddl_command_end` → `pg_notify`) for near-instant detection, and fallback catalog polling (default 5 min) for changes missed during disconnects
- Broadcast a monotonically increasing `schema_version` counter in the health stream so multigateway clients can invalidate cached query plans when the schema changes
- Track schema version per pooler connection in the multigateway load balancer, firing a callback on version changes (plan cache hook is a TODO for a follow-up PR)
- Add `AwaitRunning` to the pubsub listener to safely coordinate cross-component startup ordering

## Description

The multipooler's new `schemaTracker` component implements `StateAware` and only runs when the pooler is PRIMARY+SERVING (replicas don't accept DDL). On startup it takes a baseline snapshot of `pg_catalog` and then detects creates, drops, and structural alterations by diffing subsequent snapshots. The event trigger installed in the multigres sidecar schema provides the fast path — each `ddl_command_end` fires a `pg_notify` that the tracker consumes via the existing pubsub listener, coalescing rapid DDL bursts into a single poll.

On the multigateway side, the load balancer compares incoming `schema_version` values against the last-seen version per connection. On reconnect the version is reset to `-1` so the first message always triggers an invalidation, ensuring no silent gaps.


go/services/multipooler/manager/schema_tracker.go
```diff
@@ -0,0 +1,326 @@
...
+            FROM   pg_catalog.pg_attribute a
+            WHERE  a.attrelid = c.oid
+              AND  a.attnum > 0
+              AND  NOT a.attisdropped
+        )
+    END                         AS definition
+FROM  pg_catalog.pg_class     c
+JOIN  pg_catalog.pg_namespace n ON n.oid = c.relnamespace
+WHERE c.relkind IN ('r', 'v', 'm', 'p')
+  AND n.nspname NOT IN ('pg_catalog', 'information_schema')
```
haritabh17: Skip multigres schema too??

GuptaManan100:
Closing this without merging. After talking it through, the multigateway planner doesn't currently consume schema information — routing is derived from the query text and the tablegroup topology, so a plan produced before a DDL statement is identical to one produced after. Invalidating the cache on schema change has no behavioral effect today, which means all of this plumbing (event trigger installed in every primary, extra field in the health stream, new component in the pooler lifecycle) is surface area without a consumer.

This becomes necessary when the planner starts to depend on schema metadata — `search_path` resolution, column-level validation, or shard-aware routing where the shard key's type affects primitive selection. At that point a cached plan built against an old schema could actually be wrong, and we'll need this signal.

PR #834, opened 2026-04-09
refactor(multipooler): introduce ruleStore for shard rule management

dweitzman:
Replace the insertHistoryRecord approach with a typed ruleStore
abstraction that owns the current_rule and rule_history tables.

- New ruleStorer interface (observePosition, updateRule) enables test
  isolation via fakeRuleStore, replacing direct DB mocks in manager tests
- ruleUpdateBuilder provides a builder API for constructing rule updates
  with optional fields (leader, cohort, WAL position, operation, etc.)
- buildNodePosition consolidates row-to-proto parsing shared by both the
  read (observePosition) and write (updateRule) paths
- newPoolerID/toPoolerIDs now return approximate values alongside errors,
  letting best-effort callers (e.g. BeginTerm revoke) use the result
  while strict callers (e.g. validateStandbyIDs) check the error
- Schema operations move from pg_multischema into rule_store where they
  logically belong

### DIFF
--- a/go/common/pgprotocol/client/extended.go
+++ b/go/common/pgprotocol/client/extended.go
@@ -369,9 +369,9 @@ func argToParam(arg any) ([]byte, error) {
 	}
 }
 
-// encodeStringArray encodes a []string as a PostgreSQL text-format array literal (e.g. {foo,bar}).
-// Elements that require quoting (containing braces, commas, backslashes, double-quotes, or whitespace)
-// are enclosed in double-quotes with internal double-quotes and backslashes escaped.
+// encodeStringArray encodes a []string as a PostgreSQL text-format array literal (e.g. {"foo","bar"}).
+// All elements are double-quoted with internal double-quotes and backslashes escaped.
+// Always quoting avoids edge cases (empty strings, NULL, whitespace, unicode) without loss of correctness.
 func encodeStringArray(elems []string) string {
 	if len(elems) == 0 {
 		return "{}"
@@ -382,33 +382,17 @@ func encodeStringArray(elems []string) string {
 		if i > 0 {
 			b.WriteByte(',')
 		}
-		if needsQuoting(e) {
-			b.WriteByte('"')
-			for _, c := range e {
-				if c == '"' || c == '\\' {
-					b.WriteByte('\\')
-				}
-				b.WriteRune(c)
+		b.WriteByte('"')
+		for _, c := range e {
+			if c == '"' || c == '\\' {
+				b.WriteByte('\\')
 			}
-			b.WriteByte('"')
-		} else {
-			b.WriteString(e)
+			b.WriteRune(c)
 		}
+		b.WriteByte('"')
 	}
 	b.WriteByte('}')
 	return b.String()
-}
-
-func needsQuoting(s string) bool {
-	if s == "" {
-		return true
-	}
-	for _, c := range s {
-		if c == '{' || c == '}' || c == ',' || c == '"' || c == '\\' || c == ' ' || c == '\t' || c == '\n' {
-			return true
-		}
-	}
-	return false
 }
 
 // processExecuteResponses processes responses to an Execute command.
--- a/go/common/pgprotocol/client/extended.go
+++ b/go/common/pgprotocol/client/extended.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/multigres/multigres/go/common/mterrors"
@@ -354,9 +355,55 @@ func argToParam(arg any) ([]byte, error) {
 	case time.Time:
 		// Use RFC3339 format which PostgreSQL understands.
 		return []byte(v.Format(time.RFC3339Nano)), nil
+	case []string:
+		// Encode as PostgreSQL array literal: {elem1,elem2,...}
+		// Elements containing commas, braces, backslashes, or whitespace are double-quoted.
+		return []byte(encodeStringArray(v)), nil
 	default:
 		return nil, fmt.Errorf("unsupported type: %T", arg)
 	}
+}
+
+// encodeStringArray encodes a []string as a PostgreSQL text-format array literal (e.g. {foo,bar}).
+// Elements that require quoting (containing braces, commas, backslashes, double-quotes, or whitespace)
+// are enclosed in double-quotes with internal double-quotes and backslashes escaped.
+func encodeStringArray(elems []string) string {
+	if len(elems) == 0 {
+		return "{}"
+	}
+	var b strings.Builder
+	b.WriteByte('{')
+	for i, e := range elems {
+		if i > 0 {
+			b.WriteByte(',')
+		}
+		if needsQuoting(e) {
+			b.WriteByte('"')
+			for _, c := range e {
+				if c == '"' || c == '\\' {
+					b.WriteByte('\\')
+				}
+				b.WriteRune(c)
+			}
+			b.WriteByte('"')
+		} else {
+			b.WriteString(e)
+		}
+	}
+	b.WriteByte('}')
+	return b.String()
+}
+
+func needsQuoting(s string) bool {
+	if s == "" {
+		return true
+	}
+	for _, c := range s {
+		if c == '{' || c == '}' || c == ',' || c == '"' || c == '\\' || c == ' ' || c == '\t' || c == '\n' {
+			return true
+		}
+	}
+	return false
 }
 
 // processExecuteResponses processes responses to an Execute command.
--- a/go/common/pgprotocol/client/extended.go
+++ b/go/common/pgprotocol/client/extended.go
@@ -240,7 +240,7 @@ func (c *Conn) PrepareAndExecute(ctx context.Context, name, queryStr string, par
 // QueryArgs executes a parameterized query using the extended query protocol.
 // This is a convenience method that accepts Go values as arguments and converts
 // them to the appropriate text format for PostgreSQL.
-// Supported argument types: nil, string, []byte, int, int32, int64, uint32, uint64,
+// Supported argument types: nil, string, []byte, int, int32, int64, *int64, uint32, uint64,
 // float32, float64, bool, and time.Time.
 func (c *Conn) QueryArgs(ctx context.Context, queryStr string, args ...any) ([]*sqltypes.Result, error) {
 	// Convert args to [][]byte
@@ -339,6 +339,11 @@ func argToParam(arg any) ([]byte, error) {
 		return []byte(strconv.FormatInt(int64(v), 10)), nil
 	case int64:
 		return []byte(strconv.FormatInt(v, 10)), nil
+	case *int64:
+		if v == nil {
+			return nil, nil // NULL
+		}
+		return []byte(strconv.FormatInt(*v, 10)), nil
 	case uint32:
 		return []byte(strconv.FormatUint(uint64(v), 10)), nil
 	case uint64:
--- a/go/common/pgprotocol/client/extended_test.go
+++ b/go/common/pgprotocol/client/extended_test.go
@@ -414,3 +414,39 @@ func TestWriteBindWithNullParams(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, []byte("hello"), p3)
 }
+
+func TestEncodeStringArray(t *testing.T) {
+	t.Run("empty slice", func(t *testing.T) {
+		assert.Equal(t, "{}", encodeStringArray([]string{}))
+	})
+
+	t.Run("single plain element", func(t *testing.T) {
+		assert.Equal(t, "{foo}", encodeStringArray([]string{"foo"}))
+	})
+
+	t.Run("multiple plain elements", func(t *testing.T) {
+		assert.Equal(t, "{foo,bar,baz}", encodeStringArray([]string{"foo", "bar", "baz"}))
+	})
+
+	t.Run("element with space is quoted", func(t *testing.T) {
+		assert.Equal(t, `{"foo bar"}`, encodeStringArray([]string{"foo bar"}))
+	})
+
+	t.Run("element with comma is quoted", func(t *testing.T) {
+		assert.Equal(t, `{"foo,bar"}`, encodeStringArray([]string{"foo,bar"}))
+	})
+
+	t.Run("element with double quote is quoted and escaped", func(t *testing.T) {
+		assert.Equal(t, `{"foo\"bar"}`, encodeStringArray([]string{`foo"bar`}))
+	})
+
+	t.Run("element with backslash is quoted and escaped", func(t *testing.T) {
+		assert.Equal(t, `{"foo\\bar"}`, encodeStringArray([]string{`foo\bar`}))
+	})
+
+	t.Run("pooler ID format round-trips through ParseTextArray", func(t *testing.T) {
+		elems := []string{"zone1_pooler-1", "zone1_pooler-2", "zone1_pooler-3"}
+		encoded := encodeStringArray(elems)
+		assert.Equal(t, "{zone1_pooler-1,zone1_pooler-2,zone1_pooler-3}", encoded)
+	})
+}
--- a/go/common/pgprotocol/client/extended_test.go
+++ b/go/common/pgprotocol/client/extended_test.go
@@ -421,11 +421,11 @@ func TestEncodeStringArray(t *testing.T) {
 	})
 
 	t.Run("single plain element", func(t *testing.T) {
-		assert.Equal(t, "{foo}", encodeStringArray([]string{"foo"}))
+		assert.Equal(t, `{"foo"}`, encodeStringArray([]string{"foo"}))
 	})
 
 	t.Run("multiple plain elements", func(t *testing.T) {
-		assert.Equal(t, "{foo,bar,baz}", encodeStringArray([]string{"foo", "bar", "baz"}))
+		assert.Equal(t, `{"foo","bar","baz"}`, encodeStringArray([]string{"foo", "bar", "baz"}))
 	})
 
 	t.Run("element with space is quoted", func(t *testing.T) {
@@ -447,6 +447,14 @@ func TestEncodeStringArray(t *testing.T) {
 	t.Run("pooler ID format round-trips through ParseTextArray", func(t *testing.T) {
 		elems := []string{"zone1_pooler-1", "zone1_pooler-2", "zone1_pooler-3"}
 		encoded := encodeStringArray(elems)
-		assert.Equal(t, "{zone1_pooler-1,zone1_pooler-2,zone1_pooler-3}", encoded)
+		assert.Equal(t, `{"zone1_pooler-1","zone1_pooler-2","zone1_pooler-3"}`, encoded)
+	})
+
+	t.Run("unicode elements are encoded correctly", func(t *testing.T) {
+		assert.Equal(t, `{"ᚼᛅᛁᛚ","ᚼᛅᛁᛗᚱ"}`, encodeStringArray([]string{"ᚼᛅᛁᛚ", "ᚼᛅᛁᛗᚱ"}))
+	})
+
+	t.Run("null string is quoted to avoid SQL NULL interpretation", func(t *testing.T) {
+		assert.Equal(t, `{"NULL"}`, encodeStringArray([]string{"NULL"}))
 	})
 }
--- a/go/common/sqltypes/arrays.go
+++ b/go/common/sqltypes/arrays.go
@@ -1,0 +1,57 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sqltypes
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ParseTextArray parses a PostgreSQL text-format array literal (e.g. {foo,bar}) into a []string.
+// Quoted elements (e.g. {"foo bar","baz,qux"}) and backslash escapes within quoted elements
+// are handled. Returns an error if the input is not a valid PostgreSQL array literal.
+func ParseTextArray(s string) ([]string, error) {
+	if len(s) < 2 || s[0] != '{' || s[len(s)-1] != '}' {
+		return nil, fmt.Errorf("not a PostgreSQL array literal: %q", s)
+	}
+	inner := s[1 : len(s)-1]
+	if inner == "" {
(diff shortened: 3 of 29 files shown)


go/common/pgprotocol/client/extended.go
```diff
@@ -354,11 +355,57 @@ func argToParam(arg any) ([]byte, error) {
 	case time.Time:
 		// Use RFC3339 format which PostgreSQL understands.
 		return []byte(v.Format(time.RFC3339Nano)), nil
+	case []string:
```
dweitzman, line 363: @GuptaManan100, added you as a reviewer because this PR implements support for `TEXT[]` type in postgres

go/services/multipooler/manager/rule_store.go
dweitzman, line 1: This file is the meat of the PR. Everything else is in support of the changes here

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,701 @@
...
+func (pm *MultiPoolerManager) createCurrentRuleTable(ctx context.Context) error {
+	execCtx, cancel := context.WithTimeout(ctx, 500*time.Millisecond)
+	defer cancel()
+	if err := pm.exec(execCtx, `CREATE TABLE IF NOT EXISTS multigres.current_rule (
+		shard_id                  BYTEA PRIMARY KEY,
+		coordinator_term          BIGINT NOT NULL,
+		rule_subterm              BIGINT NOT NULL,
+		leader_id                 TEXT,
+		coordinator_id            TEXT,
+		cohort_members            TEXT[] NOT NULL,
```
dweitzman, line 166: This `TEXT[]` for cohort is why I made a change to the protocol & executor

go/services/multipooler/manager/pg_replication.go
```diff
@@ -64,32 +64,49 @@ type poolerID struct {
...
 // - SetPrimaryConnInfo: standby's application_name when connecting to primary
 // - ConfigureSynchronousReplication: standby names in synchronous_standby_names
+//
+// On validation failure an approximate poolerID is returned alongside the error.
+// The approximate appName is "{cell}_{name}" with missing fields replaced by
+// "<unknown>". For overlong names the full (invalid) string is returned as-is.
+// Callers that require a strictly valid appName must check the error. Callers that
+// can tolerate an approximate name (e.g. for logging or informational responses)
+// may use the returned value regardless.
 func newPoolerID(id *clustermetadatapb.ID) (poolerID, error) {
```
dweitzman, line 74: This might be a little wacky, but it's potentially temporary: some RPCs are running the postgres application name instead of the topoclient `ID`, but it's theoretically possible to have an ID that can't be turned into a valid application name (we should block pooler startup in that case, ideally)

The hack here is to say that we return a conversion error but also return a placeholder non-empty value so that we can express the cohort for a rule where some of the members are invalid application names

This will be less of an issue later on after we migrate the multiorch to deal with rules instead of application names directly. Then we won't necessarily need to serialize bad IDs into application names anymore, we'll just reject requests to take action on a pooler if the rule is invalid

go/services/multipooler/manager/rpc_manager_test.go
```diff
@@ -932,18 +922,19 @@ func TestPromote_WithElectionMetadata(t *testing.T) {
 	pm.mu.Lock()
 	assert.Equal(t, clustermetadatapb.PoolerType_PRIMARY, pm.multipooler.Type)
 	pm.mu.Unlock()
-	assert.NoError(t, mockQueryService.ExpectationsWereMet())
 
-	// Note: We don't directly test that insertLeadershipHistory is called with the correct metadata
-	// because that would require mocking the database layer. The leadership history functionality
-	// will be tested by the actual implementation. This test verifies that the Promote method
-	// accepts the new parameters without error and completes successfully.
+	update := fakeRules.assertPromoteRecorded(t)
```
dweitzman, line 926: Unit tests have the option to use a fake rule store now instead of mocking SQL directly. The SQL for rule changes is complicated and returns quite a few columns, so this will hopefully be easier for unit tests to manage

go/common/pgprotocol/client/extended.go
```diff
@@ -354,11 +360,57 @@ func argToParam(arg any) ([]byte, error) {
...
+	if s == "" {
+		return true
+	}
+	for _, c := range s {
+		if c == '{' || c == '}' || c == ',' || c == '"' || c == '\\' || c == ' ' || c == '\t' || c == '\n' {
+			return true
+		}
+	}
+	return false
+}
```
mkindahl: I wonder if this is enough. PostgreSQL strings can contain control characters and general unicode code points, so just checking for the "obvious" ASCII characters would not be sufficient. See my comment above.

go/common/pgprotocol/client/extended.go
```diff
@@ -354,11 +360,57 @@ func argToParam(arg any) ([]byte, error) {
...
+	if len(elems) == 0 {
+		return "{}"
+	}
+	var b strings.Builder
+	b.WriteByte('{')
+	for i, e := range elems {
+		if i > 0 {
+			b.WriteByte(',')
+		}
+		if needsQuoting(e) {
```
mkindahl: There is the function `strings.ContainsAny`, but I do not think this logic is sufficient to figure out if quoting is needed since the `e` can be any unicode character, not just ASCII.

```suggestion
		if e == "" || strings.ContainsAny(e, `{},"\ `+"\t\n")
```

go/common/pgprotocol/client/extended.go
```diff
@@ -354,11 +360,57 @@ func argToParam(arg any) ([]byte, error) {
...
+		return []byte(encodeStringArray(v)), nil
 	default:
 		return nil, fmt.Errorf("unsupported type: %T", arg)
 	}
 }
 
+// encodeStringArray encodes a []string as a PostgreSQL text-format array literal (e.g. {foo,bar}).
+// Elements that require quoting (containing braces, commas, backslashes, double-quotes, or whitespace)
+// are enclosed in double-quotes with internal double-quotes and backslashes escaped.
+func encodeStringArray(elems []string) string {
```
mkindahl: I don't think this works correctly if you pass in arbitrary UTF-8 encoded unicode. I think that you do not need to check for the quote characters as you do below, you just need to check for backslash and quote and just copy the remaining bytes. You can also always quote the value, which makes the logic simpler.

go/common/pgprotocol/client/extended_test.go
```diff
@@ -414,3 +414,39 @@ func TestWriteBindWithNullParams(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, []byte("hello"), p3)
 }
+
+func TestEncodeStringArray(t *testing.T) {
```
mkindahl, line 418: Missing tests for unicode characters, for example:

```go
t.Run("slice with non-ascii runes", func(t *testing.T) {
		assert.Equal(t, `{"ᚼᛅᛁᛚ","ᚼᛅᛁᛗᚱ"}`, encodeStringArray([]string{"ᚼᛅᛁᛚ", "ᚼᛅᛁᛗᚱ"}))
})
```

go/common/sqltypes/arrays.go
```diff
@@ -0,0 +1,57 @@
...
+	for i < len(inner) {
+		c := inner[i]
+		switch {
+		case c == '"' && !quoted:
+			quoted = true
+		case c == '"' && quoted:
+			quoted = false
+		case c == '\\' && quoted && i+1 < len(inner):
+			i++
+			cur.WriteByte(inner[i])
```
mkindahl: What if `c == '\\'` but the next character does _not_ exist? Shouldn't that generate an error?

go/common/sqltypes/arrays_test.go
```diff
@@ -0,0 +1,90 @@
...
+		result, err := ParseTextArray(`{"foo,bar","baz"}`)
+		require.NoError(t, err)
+		assert.Equal(t, []string{"foo,bar", "baz"}, result)
+	})
+
+	t.Run("quoted element with backslash escape", func(t *testing.T) {
+		result, err := ParseTextArray(`{"foo\"bar"}`)
+		require.NoError(t, err)
+		assert.Equal(t, []string{`foo"bar`}, result)
+	})
```
mkindahl, line 61: Missing error test cases for strings like `{"foo\`, `{"foo\\,}`, and `{"foo","bar"`.

Also correct string cases that should work are `{foo, bar}` (note space).

go/services/multipooler/manager/manager.go
```diff
@@ -108,6 +108,7 @@ type MultiPoolerManager struct {
 	consensusState  *ConsensusState
 	topoLoaded      bool
 	consensusLoaded bool
+	rules           ruleStorer
```
mkindahl, line 111: Typo? Shouldn't it be `ruleStore`?

go/services/multipooler/manager/pg_replication_test.go
```diff
@@ -51,56 +51,70 @@ func TestNewPoolerID(t *testing.T) {
...
+			id:              &clustermetadatapb.ID{Cell: "us_west", Name: "replica-1"},
+			expectError:     true,
+			expectedAppName: "us_west_replica-1",
+		},
+		{
+			name:            "name contains underscore",
+			id:              &clustermetadatapb.ID{Cell: "zone1", Name: "replica_1"},
+			expectError:     true,
+			expectedAppName: "zone1_replica_1",
+		},
```
mkindahl, line 112: Are names ever "deconstructed" to get the `Cell` and the `Name`? If so, could it be that using an _ in the cell or the name could cause confusion as to what it part of the cell and what is part of the name?

So, for example, the `{Cell: "us_west", Name: "1_replica"}` will be indistinguishable from `{Cell: "us_west_1", Name: "replica"}`.

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,709 @@
...
+	"github.com/multigres/multigres/go/common/timeouts"
+	"github.com/multigres/multigres/go/common/topoclient"
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
+	"github.com/multigres/multigres/go/services/multipooler/executor"
+)
+
+// ruleStorer is the interface for reading and writing the current shard rule.
+// *ruleStore implements this; tests use fakeRuleStore.
+type ruleStorer interface {
```
mkindahl, line 33: The name is very close to the concrete structure `ruleStore` so there is going to be mixups. Maybe be very specific and call it `ruleStoreInterface`?

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,709 @@
...
+
+func newRuleUpdate(termNumber int64, coordinatorID *clustermetadatapb.ID, eventType, reason string, createdAt time.Time) *ruleUpdateBuilder {
+	return &ruleUpdateBuilder{
+		termNumber:    termNumber,
+		coordinatorID: coordinatorID,
+		eventType:     eventType,
+		reason:        reason,
+		createdAt:     createdAt,
+	}
+}
```
mkindahl, line 104: The normal case is probably to modify an existing rule, so you might want to have an interface where you can do something like

```go
rule.AsBuilder().withLeader(primary).create()
```

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,709 @@
...
+		previousTerm = &update.previousRule.coordinatorTerm
+		previousSubterm = &update.previousRule.ruleSubterm
+	}
+
+	// Use the remote operation timeout for history writes. This write validates that synchronous
+	// replication is functioning - it must wait long enough for standbys to connect and acknowledge.
+	execCtx, cancel := context.WithTimeout(ctx, timeouts.RemoteOperationTimeout)
+	defer cancel()
+
+	result, err := rs.queryService.QueryArgs(execCtx, `
```
mkindahl, line 364: The query is difficult to review since it is so long. I have to trust you that it does the right thing. Offhand I make the following observations:

1. Using `SELECT FOR UPDATE` and then adding one offers the same guarantees as using a sequence and inserting rows into a table, but without the potential gaps in the sequence. Is it important to not have gaps in the sequence of coordinate terms?
2. There is a `SELECT FOR UPDATE` which means that you expect to have multiple sessions running the same statement, but AFAICT, Multipooler would not run multiple such updates at the same time, so why is that `SELECT FOR UPDATE` needed?
3. The `locked` table is using the conditions `coordinator_term = ... AND rule_subterm = ...`, which means that if an error creeps into the logic or the table, this will return zero rows. How will that work with the query? What are the consequences of such a bug?
4. Everything you do in `next_sub` can be added to `locked` as instead, but this is mostly a matter of style.
5. The `shard_id` is always zero. Do we need it then? Can it ever be anything else, or is this just added to deal with shards in the future?

go/services/multipooler/manager/rule_store_test.go
mkindahl, line 1: I miss tests. For example, since you use `SELECT FOR UPDATE` it means that you expect conflicting updates, but there is nothing like that here.

mkindahl (commented):
Here are a few comments. The SQL statement for doing a rule update looks very complicated, so I wonder if it can be simplified.

We discussed having a single table and just append to the table instead. What are your thoughts around that?

go/common/pgprotocol/client/extended_test.go
dweitzman: Added unicode test

go/common/pgprotocol/client/extended.go
dweitzman: Changed to always quote

go/common/sqltypes/arrays_test.go
dweitzman: Added a bunch of test cases

go/common/pgprotocol/client/extended.go
dweitzman: Changed to always quote, so this is dead code

go/common/pgprotocol/client/extended.go
dweitzman: Removed this code in favor of always quoting

go/common/sqltypes/arrays.go
dweitzman: Added a test case

go/services/multipooler/manager/manager.go
dweitzman: This is the interface name, not the struct name. I don't have strong feelings on interface names, but an `-er` suffix seems consistent with Go conventions: https://go.dev/doc/effective_go#interface-names

go/services/multipooler/manager/pg_replication_test.go
dweitzman: I think the answer is that we sometimes maybe parse application names back to IDs, but as the refactoring proceeds that will become increasingly unnecessary to the point where it stops happening

My eventual goal is that we generally deal with the proto `ID` type and only convert it to an application name immediately before changing the GUC in postgres

When we get to that point, I can remove this code

In the meantime, I don't think anything that's supposed to work will break because of this change, I just want to avoid returning an empty cohort list just because orch is expecting information in the legacy appName format instead of IDs

go/services/multipooler/manager/rule_store.go
dweitzman: The `-er` naming is an attempt to follow Go conventions. I'm not personally that attached to Go conventions, but my default thinking is to try to follow them where practical

go/services/multipooler/manager/rule_store.go
dweitzman: I imagine something like `AsBuilder()` appearing in the future, but for this PR I wanted to minimize unused code since there's plenty interesting happening here already

go/services/multipooler/manager/rule_store.go
dweitzman: For now I feel most comfortable with `SELECT FOR UPDATE` on a single rules table because I'm not sure how to implement compare-and-swap without it and I predict that we'll want compare-and-swap (and put it into the builder interface).

dweitzman (commented):
I've also added a test that runs the SQL query against a real postgres instance

go/common/pgprotocol/client/extended_test.go
```diff
@@ -414,3 +414,47 @@ func TestWriteBindWithNullParams(t *testing.T) {
...
+		assert.Equal(t, `{"zone1_pooler-1","zone1_pooler-2","zone1_pooler-3"}`, encoded)
+	})
+
+	t.Run("unicode elements are encoded correctly", func(t *testing.T) {
+		assert.Equal(t, `{"ᚼᛅᛁᛚ","ᚼᛅᛁᛗᚱ"}`, encodeStringArray([]string{"ᚼᛅᛁᛚ", "ᚼᛅᛁᛗᚱ"}))
+	})
+
+	t.Run("null string is quoted to avoid SQL NULL interpretation", func(t *testing.T) {
+		assert.Equal(t, `{"NULL"}`, encodeStringArray([]string{"NULL"}))
+	})
```
mkindahl, line 459: That should not be possible since the extended protocol is used (parameters are passed directly as strings with no SQL-level parsing between), but it does not harm to test this.

go/common/sqltypes/arrays_test.go
```diff
@@ -0,0 +1,227 @@
...
+	t.Run("stray closing brace inside element", func(t *testing.T) {
+		_, err := ParseTextArray(`{foo}bar}`)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "unexpected '}'")
+	})
+
+	t.Run("multi-dimensional array", func(t *testing.T) {
+		_, err := ParseTextArray(`{{a,b},{c,d}}`)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "multi-dimensional")
```
mkindahl, line 225: Since a `\` will read one more rune after it would be good to test that a `\` at the end of the string does not cause problems.

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,713 @@
...
+// current_rule holds a single row per shard representing the current cluster rule.
+// It is used as a locking target (SELECT FOR UPDATE) to serialise concurrent
+// writes; rule_history provides the append-only audit log.
+//
+// coordinator_term=0 in the sentinel row means no rule has been applied yet.
+func (rs *ruleStore) createRuleTables(ctx context.Context) error {
+	execCtx, cancel := context.WithTimeout(ctx, 500*time.Millisecond)
+	defer cancel()
+
+	if _, err := rs.queryService.Query(execCtx, `CREATE TABLE IF NOT EXISTS multigres.current_rule (
```
mkindahl, line 160: This looks dangerous.

If you conditionally create the table, it might have a different definition compared to what this code expects (for example, as a result of a new version of the code). I think you should either get an error proper (and remove the `IF NOT EXISTS`) or drop the table before you create it if this is acceptable.

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,713 @@
...
+		created_at                TIMESTAMPTZ NOT NULL
+	)`); err != nil {
+		return mterrors.Wrap(err, "failed to create current_rule table")
+	}
+
+	if _, err := rs.queryService.QueryArgs(execCtx, `
+		INSERT INTO multigres.current_rule
+		  (shard_id, coordinator_term, rule_subterm, cohort_members, created_at)
+		VALUES ($1, 0, 0, '{}', now())
+		ON CONFLICT (shard_id) DO NOTHING`,
```
mkindahl, line 180: If the definition of the table does not match what you have above, this will fail despite using `ON CONFLICT ... DO NOTHING`. If you create the table above (by removing the `IF NOT EXISTS`), then the conflict resolution is not needed (but it does not hurt either).

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,713 @@
...
+		[]byte("0")); err != nil {
+		return mterrors.Wrap(err, "failed to initialize current_rule")
+	}
+
+	// Each row records a cluster state change (promotion, cohort membership, durability policy).
+	// The composite primary key (coordinator_term, rule_subterm) uniquely identifies each rule;
+	// rule_subterm is assigned by the application as MAX(rule_subterm)+1 within a coordinator_term.
+	if _, err := rs.queryService.Query(execCtx, `CREATE TABLE IF NOT EXISTS multigres.rule_history (
```
mkindahl, line 188: Same as above. If the table already exists, it could have a different definition compared to what the code expects, so this could cause problems. Better get a hard error here if the table should not exist, or drop the table if it is ok to lose the data in it.

go/common/sqltypes/arrays.go
mkindahl: I can't see a test case that parses a string that ends with a `\`. Did I miss it?

go/services/multipooler/manager/manager.go
mkindahl: Either one works for me. Just wondered if it was intentional.

go/services/multipooler/manager/rule_store.go
mkindahl: So you have multiple connections updating the row at the same time and you need to serialize these?

mkindahl (approved):
I think it makes more sense to remove the `IF NOT EXISTS` for the tables, or drop the tables prior to creating them, or we will run into weird errors later if the table already exists.

Approving since I do not think it will cause any issues currently, but probably something we should fix before releasing.

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,713 @@
...
+// concurrent writes at the database level in addition to the caller's action lock.
+//
+// Returns the node's position (rule + WAL LSN) at the time of the write,
+// or nil if force mode skipped the write.
+//
+// This operation uses the remote-operation-timeout and will fail if it cannot
+// complete within that time. A timeout typically indicates that synchronous
+// replication is not functioning.
+func (rs *ruleStore) updateRule(ctx context.Context, update *ruleUpdateBuilder) (*clustermetadatapb.NodePosition, error) {
+	if update.force {
```
dweitzman, line 309: In a follow-up PR I'm going to remove this

go/services/multipooler/manager/rule_store.go
```diff
@@ -0,0 +1,713 @@
...
+		  locked AS (
+		    -- FOR UPDATE serializes concurrent writes at the database level, complementing
+		    -- the action lock held by the caller.
+		    -- Returns zero rows (causing a no-op write and error return) when any condition fails.
+		    -- next_sub: 0 when starting a new coordinator term, otherwise increment within term.
+		    SELECT current_rule.coordinator_term, current_rule.rule_subterm,
+		           current_rule.leader_id, current_rule.cohort_members,
+		           CASE WHEN params.coordinator_term > current_rule.coordinator_term THEN 0
+		                ELSE current_rule.rule_subterm + 1
+		           END AS next_sub
```
dweitzman: I've renamed rule_subterm to leader_subterm and renamed next_sub to next_leader_subterm

go/services/multipooler/manager/rule_store.go
dweitzman: I think this is a reasonable change to make, but we should make it for all the other multischema DDLs as well. I'll send a follow-up PR

go/common/sqltypes/arrays_test.go
dweitzman: Done

PR #836, opened 2026-04-10
refactor(multipooler): eliminate disableMonitorInternal

dweitzman:
The postgres monitor previously used a stop/start mechanism (disableMonitorInternal/enableMonitorInternal) to suppress auto-recovery during emergency demotion. These days the monitor holds the action lock, so it doesn't fight with the actions themselves.

This PR also introduces a rewindPending atomic flag:
- EmergencyDemote sets the flag instead of stopping the runner
- monitorPostgresIteration checks the flag and skips early when set
- runPgRewind and RewindToSource clear the flag on success

For tests that need to crash a pooler and keep postgres from restarting, instead of disabling the monitor they rename the postgres config file which is a more targeted way of stopping postgres from starting


go/services/multiorch/recovery/engine.go
```diff
@@ -493,7 +493,7 @@ func (re *Engine) collectDetectedProblemsData() []DetectedProblemData {
 			AnalysisType: string(p.CheckName),
 			DBNamespace:  p.ShardKey.Database,
 			Shard:        p.ShardKey.Shard,
-			PoolerID:     topoclient.MultiPoolerIDString(p.PoolerID),
+			EntityID:     p.EntityID(),
```
dweitzman, line 496: This is actually a fix for a panic introduced by the previous PR to refactor shard analysis. PoolerID can be `nil` now

go/services/multipooler/manager/manager.go
```diff
@@ -1528,6 +1534,10 @@ type postgresState struct {
 	isPrimary        bool
 }
 
+// monitorInhibitFile is the name of a marker file that tests and demos can place in
+// the pooler directory to prevent the postgres monitor from auto-starting postgres.
+const monitorInhibitFile = "no-autostart"
```
dweitzman: This is an attempt to solve the problem of how a test case can prevent postgres monitor from auto-restarting postgres if we're trying to simulate postgres being stopped

The proposed solution here is to have a file called `no-autostart` in the pooler directory that indicates postgres monitor shouldn't restart postgres if it's stopped

It's a little risky because the file could be left behind on a demo cluster and prevent it from restart postgres in the future. An alternative approach to this would be to have an RPC that just sets an in-memory marker to avoid restarting postgres

go/services/multipooler/manager/manager.go
```diff
@@ -1554,6 +1564,12 @@ func (pm *MultiPoolerManager) monitorPostgresIteration(ctx context.Context) {
 		return
 	}
 
+	// Skip all action while awaiting a rewind after emergency demotion.
+	if pm.rewindPending.Load() {
```
dweitzman, line 1568: This rewindPending flag will go away when https://github.com/multigres/multigres/pull/837 is merged

go/services/multiorch/recovery/metrics.go
```diff
@@ -314,12 +314,12 @@ func (m *Metrics) RecordHealthCheckCycleDuration(ctx context.Context, duration t
...
 
 // DetectedProblemData represents a detected problem with its attributes for metric observation.
-// Each problem is tracked at per-pooler granularity.
+// Each problem is tracked per affected entity (pooler ID or shard key).
 type DetectedProblemData struct {
 	AnalysisType string
 	DBNamespace  string
 	Shard        string
-	PoolerID     string
+	EntityID     string
```
rafael, line 322: We already have the Shard. Maybe simpler to keep this as PoolerId and have the convention that this will be empty when the Scope is Shard.

The other question is that I wonder that even when  scope is Shard, the analysis comes from the view a pooler. So it might be useful to still propagate this information the way already is.

go/services/multipooler/manager/manager.go
rafael: Makes sense

go/services/multipooler/manager/rpc_backup.go
```diff
@@ -271,24 +271,11 @@ func (pm *MultiPoolerManager) RestoreFromBackup(ctx context.Context, backupID st
 	defer pm.actionLock.Release(ctx)
 
 	return telemetry.WithSpan(ctx, "restore-from-backup", func(ctx context.Context) error {
-		// Pause monitoring during restore to prevent interference
```
rafael, line 274: I don't remember why we added this one. Do you have context on this one?

go/services/multipooler/manager/rpc_initialization.go
```diff
@@ -57,13 +57,6 @@ func (pm *MultiPoolerManager) InitializeEmptyPrimary(ctx context.Context, req *m
 	}
 	defer pm.actionLock.Release(ctx)
 
-	// Pause monitoring during initialization to prevent interference
```
rafael, line 60: same here. Wondering if this could introduce flakes right now.

go/services/multipooler/manager/rpc_initialization_test.go
```diff
@@ -626,6 +626,7 @@ func TestTakeRemedialAction_StartPostgres(t *testing.T) {
 		logger:       slog.Default(),
 		actionLock:   NewActionLock(),
 	}
+	pm.postgresRestartsEnabled.Store(true)
```
rafael: why we need to set it to true? Shouldn't that be the default ?

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -1074,13 +1073,6 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
 		}, nil
 	}
 
-	// Pause monitoring during this operation to prevent interference
-	resumeMonitor, err := pm.PausePostgresMonitor(ctx)
-	if err != nil {
-		return nil, err
-	}
-	defer resumeMonitor(ctx)
-
```
rafael, line 1083: Isn't this going to introduce an issue?  If the monitor is running here, when we stop postgres for pgRewind. I wonder if you need to set the rewindPending signal here while this operation is happening

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -1364,13 +1356,6 @@ func (pm *MultiPoolerManager) RewindToSource(ctx context.Context, source *cluste
 	}
 	defer pm.actionLock.Release(ctx)
 
-	// Pause monitoring during this operation to prevent interference
-	resumeMonitor, err := pm.PausePostgresMonitor(ctx)
```
rafael, line 1368: Similar as the other comment, thinking that you need to set rewind pending while this is happening

go/test/endtoend/multiorch/dead_primary_recovery_test.go
```diff
@@ -89,11 +89,6 @@ func TestDeadPrimaryRecovery(t *testing.T) {
 			resp.Status.ReplicationStatus.PrimaryConnInfo.Port)
 	}
 
-	// Disable monitoring so multiorch orchestrates recovery instead
-	// of each multipooler's local postgres monitor. This could create races for
-	// this test.
-	disableMonitoringOnAllNodes(t, setup)
```
rafael, line 95: why don't we keep this the same way and just update / rename `disableMonitoringOnAllNodes` to be `disableRestartOnAllNodes` 

The reason of doing in it all nodes is that it was a convenience method for tests that are doing multiple failovers. I ran into issues where I just disabled in the first primary and then ran into issues later on.

So added this method, so in tests we keep it simple and disabled it in all poolers

go/services/multiorch/recovery/metrics.go
dweitzman: I experimented with saying that a missing pooler ID implies shard scope, but it hits a kind of an edge case for a dead primary

If there's a suspected dead primary, we have logic to avoid trying to contact it while re-polling all the pooler health statuses before taking action, and we identify the dead primary by pooler ID

However at a problem level, a dead primary is actually a shard-level problem even though it's caused by a single pooler

go/test/endtoend/multiorch/demote_stale_primary_test.go
```diff
@@ -74,16 +74,12 @@ func TestDemoteStalePrimary_SIGKILL(t *testing.T) {
 	oldPrimaryName := setup.PrimaryName
 	t.Logf("Initial primary: %s", oldPrimaryName)
 
-	// Disable monitoring on old primary so postgres is not restarted by pooler
```
rafael, line 77: I would argue this was not longer needed, because DemoteStale primary was Pausing the monitor internally. 

Related to my other comments, I think StalePrimary should set the signal itself and we shouldn't need to do this manually in the test .

go/test/endtoend/multipooler/setup_test.go
```diff
@@ -154,13 +153,6 @@ func WithResetGuc(gucs ...string) cleanupOption {
 	}
 }
 
-// WithEnabledMonitor returns a cleanup option that enables the PostgreSQL monitor.
-func WithEnabledMonitor() cleanupOption {
```
rafael, line 158: Don't we still want this but set the postgres restart? And I guess by default set it to false? 

Wondering if many tests are relying on this. I remember this was all tracking down flakes.

rafael (commented):
I did one pass. I think I have a meta question around Pausing the monitor.  Let's chat about that synchronously

rafael (approved):
There a few small minor comments, but now that I got the context from Pause, this change makes sense to me

go/services/multipooler/manager/rpc_initialization_test.go
dweitzman: Switched the code from postgresRestartsEnabled to postgresRestartsDisabled so the default value will do the correct thing here

This test isn't using the constructor so it depends on the struct defaults working

go/services/multipooler/manager/rpc_backup.go
dweitzman: This was from before we got the action lock during the postgres monitor. It does nothing now

go/services/multipooler/manager/rpc_initialization.go
dweitzman: Same thing: this isn't needed now that we use the action lock in postgres monitor

go/test/endtoend/multiorch/demote_stale_primary_test.go
dweitzman: This code will go away with https://github.com/multigres/multigres/pull/837 but I wanted to keep all the PRs separate

go/test/endtoend/multipooler/setup_test.go
dweitzman: I think it's reasonable for tests to opt-in to preventing postgres from restarting, but I don't have strong conviction. If this causes flakes, we can change the default

It's not obvious to me why a test would need postgres to stop and not be able to explicitly ask the pooler to keep it stopped

go/test/endtoend/multiorch/dead_primary_recovery_test.go
dweitzman: I introduced a ruleguard rule and a helper so that any endtoend test wanting to stop postgres will be gently encouraged to use a helper that both disables restarts and stops postgres

That makes the test behavior more explicit without changing the default vs production

I'm not strongly attached to that approach, though

go/services/multipooler/manager/rpc_manager.go
dweitzman: The action lock won't be released until resumeMonitor() is called, so postgres monitor is blocked anyway

go/services/multipooler/manager/rpc_manager.go
dweitzman: Same comment: RewindToSource already holds the action lock, so an additional block/lock isn't needed to prevent postgres monitor from acting

PR #837, opened 2026-04-10
feat(multipooler): EmergencyDemote restarts as standby with resignation signal

dweitzman:
`EmergencyDemote` previously stopped postgres permanently, leaving the node dead and requiring `pg_rewind` + operator intervention to rejoin the cluster. This PR makes it keep the node alive.

**Before:** `EmergencyDemote` stopped postgres and disabled the monitor.

**After:** `EmergencyDemote` restarts postgres as a standby, keeps the monitor running, and publishes a health signal to multiorch that this pooler is no longer able to serve as primary

### Changes

**EmergencyDemote behavior (`rpc_manager.go`)**
- Replace `stopPostgresForEmergencyDemote` with `restartPostgresAsStandby`
- Remove `disableMonitorInternal()` — the monitor stays running so the node auto-heals
- Set `resignedPrimaryAtTerm` after successful demotion

**Resignation signal (`AvailabilityStatus`)**

New `clustermetadata.AvailabilityStatus` proto message carries best-effort operational fitness signals that are in-memory (unlike `ConsensusStatus`, which is backed by postgres WAL and on-disk state). Coordinators treat absence as "unknown," not "unfit."

- `resigned_primary_at_term`: set when a primary voluntarily resigns. Signals the coordinator to trigger an immediate election rather than waiting for a heartbeat timeout. Cleared when the node is elected primary again.
- Exposed on `StatusResponse.availability_status` (field 13) so coordinators can read it via the existing status poll.

**Monitor integration (`manager.go`)**
- When the monitor detects postgres running as standby while the pooler type is PRIMARY (the normal post-demotion state), it sets `resignedPrimaryAtTerm` so the coordinator sees the signal even if the EmergencyDemote RPC response was missed.
- When the monitor transitions the pooler back to PRIMARY, it clears the signal.

**Recovery helper cleanup (`main_test.go`)**
- `restoreAfterEmergencyDemotion` no longer has a "re-enable monitor" step — the monitor is never disabled now.


go/services/multipooler/manager/manager.go
```diff
@@ -1695,6 +1677,7 @@ func (pm *MultiPoolerManager) takeRemedialAction(ctx context.Context, action rem
 		pm.setMonitorReason(ctx, reasonPostgresRunning, "MonitorPostgres: PostgreSQL is running")
 		pm.logger.InfoContext(ctx, "MonitorPostgres: PostgreSQL is running and primary")
 		pm.logger.InfoContext(ctx, "MonitorPostgres: Changing pooler type to primary")
+		pm.clearResignedPrimaryAtTerm()
```
rafael: This one makes me slightly nervous. Is this a valid state?

Wondering if, for the context of the monitor, it should be an invariant that if a pooler is starting as a primary, ResignedPrimaryAtTerm should already be cleared.

I'm worried about the case that a pooler was demoted, then for some unexpected reason starts as a primary and automatically clears this without a coordinator explicitly promoting it.

go/services/multiorch/recovery/types/pooler_signals.go
```diff
@@ -0,0 +1,50 @@
...
+// ensures we don't act on a REQUESTING_DEMOTION signal left over from a previous
+// election cycle that was never cleared (e.g. after a process restart).
+//
+// TODO: once the coordinator synthesizes AvailabilityStatus into
+// PoolerHealthState directly (see clustermetadata.proto TODO), read from
+// PoolerHealthState.AvailabilityStatus instead of ConsensusStatus so that
+// coordinator-synthesized signals (e.g. TEMPORARILY_UNAVAILABLE for unreachable
+// nodes) are handled here too.
+func PrimaryNeedsReplacement(p *multiorchdatapb.PoolerHealthState) bool {
+	ls := p.ConsensusStatus.GetAvailabilityStatus().GetLeadershipStatus()
```
rafael: nit - although correct, `ls` variable name made me read this code twice. `status` is more verbose, but seems clearer to me.

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -255,6 +256,43 @@ func (pm *MultiPoolerManager) executeRevoke(ctx context.Context, term int64, res
...
+func (pm *MultiPoolerManager) setResignedPrimaryAtTerm(term int64) {
+	pm.mu.Lock()
+	pm.resignedPrimaryAtTerm = term
+	pm.mu.Unlock()
+}
+
+// clearResignedPrimaryAtTerm clears the leadership demotion request, called when
+// this node is elected as primary again.
+func (pm *MultiPoolerManager) clearResignedPrimaryAtTerm() {
+	pm.mu.Lock()
```
rafael, line 306: shouldn't we call this during promote?

go/services/multipooler/manager/rpc_initialization_test.go
```diff
@@ -725,6 +725,99 @@ func TestTakeRemedialAction_LogDeduplication(t *testing.T) {
...
+	pm.setResignedPrimaryAtTerm(7)
+
+	lockCtx, err := pm.actionLock.Acquire(ctx, "test")
+	require.NoError(t, err)
+	defer pm.actionLock.Release(lockCtx)
+
+	pm.takeRemedialAction(lockCtx, remedialActionAdjustTypeToPrimary)
+
+	assert.Nil(t, pm.buildAvailabilityStatus(), "resignation signal must be cleared when transitioning back to primary")
+}
```
rafael, line 673: Claude simplify rec:

>   The three TestTakeRemedialAction_* tests (_SetsResignation, _NoResignationWhenNoPrimaryTerm, _ClearsResignation) share identical structure and differ only in setup/action/assertion. Consider
  consolidating into a single table-driven test per project convention.

proto/clustermetadata.proto
```diff
@@ -423,6 +423,97 @@ message HighestCoordinatorPromise {
   google.protobuf.Timestamp last_acceptance_time = 3;
 }
 
+// NodeSignal describes a node's general availability, independent of any specific
+// term or leadership role. 0 (UNKNOWN) means the field was not intentionally set;
+// coordinators treat it as "no concern known," not as "definitely healthy."
+enum NodeSignal {
```
rafael: Shouldn't this be Pooler ? Why introduce a new term? 

Let's talk more about this. I think now this can be confusing that now we have PoolerStatus and NodeStatus

go/services/multipooler/manager/manager.go
dweitzman: For availability signals, the nice thing is that they're best-effort: there's no correctness concern if resignation flaps on or off, we just want to make sure that a primary does consistently publish itself as resigned if it has no intention of acting as a primary again

At this moment:
- The action lock is held
- The rules table indicates that this pooler is supposed to be acting as primary

That gives this pooler the right to promote or re-promote. If/when someone wants to revoke primaryship, they'll need to wait until the action lock is free and undo what this remedial action did

dweitzman: There is something to be careful of: if we've made a revocation promise we need to make sure we don't accidentally re-enable writes. My other PR to always restart as standby would be defense against that, but this PR doesn't add any new risks because it's only changing published information, not actually running `pg_promote()`

go/services/multipooler/manager/rpc_consensus.go
dweitzman: It's not strictly necessary in the sense that this indicates what term we resigned at and establishing a new primary term at a higher number implicitly invalidates this. That said, it'd be fine to call this when setting the primary term

proto/clustermetadata.proto
dweitzman: Yeah, pooler makes sense for this. We're in an awkward transitional phase where there are old & new messages. Let's chat about goal naming for after the transition and what to do during the transition

go/services/multipooler/manager/manager.go
rafael: Let's talk more about this. This doesn't seem safe to me.

dweitzman: Removed this line to be safe

go/services/multipooler/manager/rpc_consensus.go
dweitzman: Seems harmless, done

go/services/multipooler/manager/rpc_initialization_test.go
dweitzman: Done

go/services/multiorch/recovery/types/pooler_signals.go
dweitzman: Done

proto/clustermetadata.proto
```diff
@@ -423,6 +423,97 @@ message HighestCoordinatorPromise {
...
+// LeadershipStatus is published only by nodes that are or have been primary.
+// It lets the coordinator distinguish an actively healthy primary, a primary
+// requesting demotion, and a node that has never held leadership.
+message LeadershipStatus {
+  // The primary_term at which this node was most recently elected.
+  // Non-zero only on nodes that have been elected primary.
+  int64 primary_term = 1;
+
+  LeadershipSignal signal = 2;
+}
```
mkindahl, line 468: This is slightly odd. If you only use this signal for primaries, you need to check if it is a primary before reading this, which makes it slightly awkward to use. In addition to that, does not standbys have status as well? Are you going to have a different set of status messages for that?

With three different enums that can mean different things, it is going to be complicated to deal with all combinations. Some of these are probably not going to make sense, but because of bugs we might end up there anyway, which will make it complicated to debug the system.

For example, if you read through a log and find a server where the leadership signal is `LEADERSHIP_SIGNAL_REQUESTING_DEMOTION` then you need to figure out if it is a primary before you can draw any conclusions. If it is not, then you will end up with a different set of questions that you need to figure out.

Might be better to just have a single status and make sure that each state is very clear.

proto/clustermetadata.proto
mkindahl: Node is a term used by Kubernetes as well, meaning essentially a physical machine. Might be better to not use that term.

go/services/multipooler/manager/manager.go
```diff
@@ -1703,6 +1685,19 @@ func (pm *MultiPoolerManager) takeRemedialAction(ctx context.Context, action rem
...
+		// TODO: Once ConsensusStatus is populated in StatusResponse, use the
+		// consensus term from there for a more precise resignation signal.
+		pm.mu.Lock()
+		cs := pm.consensusState
+		pm.mu.Unlock()
+		if cs != nil {
+			if term, err := cs.GetInconsistentTerm(); err == nil && term.GetPrimaryTerm() != 0 {
+				pm.setResignedPrimaryAtTerm(term.GetPrimaryTerm())
+			}
+		}
```
mkindahl: Unsure what the locks are used for here, and in particular why it is unlocked after. With this code, if the consensusTerm is changed after reading it, the code will act on something that is out of date.

I see it is used in this manner elsewhere though, but in some of the cases the lock is protecting the following call of a method (which looks correct), and the cases that release the lock early only check if `cs` is `nil`.

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -973,12 +968,23 @@ func (pm *MultiPoolerManager) emergencyDemoteLocked(ctx context.Context, consens
...
+	// primary_term (not the incoming consensusTerm) so the coordinator can
+	// correlate the signal with the term at which this node was elected.
+	pm.mu.Lock()
+	cs := pm.consensusState
+	pm.mu.Unlock()
+	if cs != nil {
+		if term, err := cs.GetTerm(ctx); err == nil && term.GetPrimaryTerm() != 0 {
+			pm.setResignedPrimaryAtTerm(term.GetPrimaryTerm())
+		}
+	}
```
mkindahl: Same as above. I think you unlock the `pm` too early and should at least keep if while using `cs`, possibly longer. You actually might want to create new methods that does this to avoid repetition and mistakes.

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -255,6 +256,44 @@ func (pm *MultiPoolerManager) executeRevoke(ctx context.Context, term int64, res
 	return nil
 }
 
+// buildAvailabilityStatus returns the current AvailabilityStatus for this node.
+// Leaders always publish a LeadershipStatus. Returns nil if no signals are set
+// and no leadership context exists.
+func (pm *MultiPoolerManager) buildAvailabilityStatus() *clustermetadatapb.AvailabilityStatus {
+	pm.mu.Lock()
+	resignedTerm := pm.resignedPrimaryAtTerm
+	pm.mu.Unlock()
```
mkindahl, line 270: This seems fine since resignedPrimaryAtTerm is copied by value.

go/services/multiorch/recovery/types/pooler_signals.go
```diff
@@ -0,0 +1,50 @@
...
+// coordinator-synthesized signals (e.g. TEMPORARILY_UNAVAILABLE for unreachable
+// nodes) are handled here too.
+func PrimaryNeedsReplacement(p *multiorchdatapb.PoolerHealthState) bool {
+	leadershipStatus := p.ConsensusStatus.GetAvailabilityStatus().GetLeadershipStatus()
+	if leadershipStatus == nil {
+		return false
+	}
+	if leadershipStatus.Signal != clustermetadatapb.LeadershipSignal_LEADERSHIP_SIGNAL_REQUESTING_DEMOTION {
+		return false
+	}
```
mkindahl, line 47: So unknown state will not trigger a replacement? Shouldn't you check the opposite, that is return false if it is `LEADERSHIP_SIGNAL_AVAILABLE`?

go/test/endtoend/multipooler/rpc_consensus_test.go
```diff
@@ -341,16 +340,12 @@ func TestConsensus_BeginTerm(t *testing.T) {
...
-		defer pgctldClient.Close()
-
+		// Verify PostgreSQL is running as standby (emergency demotion restarts as standby)
 		require.Eventually(t, func() bool {
-			statusResp, err := pgctldClient.Status(context.Background(), &pgctldpb.StatusRequest{})
-			return err == nil && statusResp.Status == pgctldpb.ServerStatus_STOPPED
-		}, 10*time.Second, 1*time.Second, "PostgreSQL should be stopped after emergency demotion from BeginTerm on pooler: %s", setup.PrimaryMultipooler.Name)
-		t.Log("Confirmed: PostgreSQL stopped after emergency demotion")
+			statusResp, err := primaryConsensusClient.Status(context.Background(), &consensusdatapb.StatusRequest{})
+			return err == nil && statusResp.Role == "replica"
```
mkindahl: You might want to use a constant for this to avoid misspellings and similar problems.

mkindahl (commented):
I think there are a few race conditions in the code.

proto/clustermetadata.proto
dweitzman: I have a different PR that does the naming change: https://github.com/multigres/multigres/pull/845

proto/clustermetadata.proto
dweitzman: Standby status can go directly into the AvailabilityStatus.

I made this a sub-message because the primary_term and signal sort of belong together: if you're resigning as leader, it's relative to a primary term.

This doesn't need to be a sub-message, but I was thinking it might be more clear to have leader-specific signals grouped together in the data model. I'm not attached to it, though

The reason I chose an enum was to avoid any mistake due to the way default values work in proto3. The default value of a boolean would be false or a number would be 0, and I wanted to defend against some sort of bug where the message isn't populated and we interpret that to mean someting like that we have or haven't resigned at term 0

I'm also not strongly attached to that, but guardrails was my motivation for both of these choices

go/services/multipooler/manager/rpc_manager.go
dweitzman: The pm locks are gone now -- I've modified this PR to make sure consensus state is initialized in the manager constructor so access to it requires no locking

For making sure term use is properly locked, `GetTerm()` asserts that the action lock is held. I've also updated `setResignedPrimaryAtTerm()` and the corresponding clear method to assert that the action lock is held

go/test/endtoend/multipooler/rpc_consensus_test.go
dweitzman: Done

go/services/multipooler/manager/manager.go
dweitzman: The locks are gone now that I've made it so `cs` can't ever be `nil`

go/services/multiorch/recovery/types/pooler_signals.go
dweitzman: Unknown state essentially means that the pooler didn't populate the leadership signal. It's not evidence that there's a problem with the leader, it's more evidence that there's a bug in our status code where it didn't explicitly inform multiorch of the leader's availability

go/services/multiorch/consensus/leader_appointment.go
```diff
@@ -201,12 +201,30 @@ func (c *Coordinator) selectCandidate(ctx context.Context, recruited []recruitme
 			"no recruited poolers available for candidate selection")
 	}
 
+	// Determine whether any non-resigned candidates exist. A node that has
+	// voluntarily requested demotion (REQUESTING_DEMOTION signal at its current
+	// primary term) should not be re-elected as long as there is another option.
+	hasNonResigned := false
```
dweitzman: I've also made changes here to avoid choosing a currently-resigned primary during leader appointment

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
```diff
@@ -88,7 +88,7 @@ func (a *PrimaryIsDeadAnalyzer) Analyze(sa *ShardAnalysis) ([]types.Problem, err
 	// yet). Suppressing would delay failover by up to the TCP keepalive
 	// interval (~30s).
 
-	if sa.ReplicasConnectedToPrimary {
+	if sa.ReplicasConnectedToPrimary && !sa.PrimaryHasResigned {
```
dweitzman, line 91: And a fix here to make sure that a resigned primary will be considered dead even if it has connected replicas

go/services/multiorch/consensus/leader_appointment.go
rafael: I'm nervous of still  choosing this server if there are no other options. I think I would rather that be a case where there is not viable candidate. 

It seems confusing that the system will choose again  a candidate that resigned

go/services/multiorch/consensus/leader_appointment_test.go
```diff
@@ -664,6 +664,150 @@ func TestSelectCandidate(t *testing.T) {
...
+			},
+		}
+
+		candidate, err := c.selectCandidate(ctx, recruited)
+		require.NoError(t, err)
+		require.Equal(t, "mp2", candidate.MultiPooler.Id.Name,
+			"must not re-elect the resigned primary; should pick the non-resigned node with highest LSN")
+	})
+
+	t.Run("selects resigned primary as last resort when it is the only candidate", func(t *testing.T) {
```
rafael: If we apply my other comment, I think in this case we bail and say that we have no viable candidates

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
rafael, line 91: could we update the unit tests to reflect this?

dweitzman: Done

go/services/multiorch/consensus/leader_appointment.go
dweitzman: Changed to remove a resigned primary from consideration for the next primary

go/services/multiorch/consensus/leader_appointment_test.go
dweitzman: Done

rafael (approved):
LGTM

proto/clustermetadata.proto
dweitzman: Dropped this message entirely for now. It's not needed yet

mkindahl (approved):
LGTM

PR #840, opened 2026-04-14
refactor(multipooler): remove IF NOT EXISTS from schema creation DDL

dweitzman:
Schema creation runs only once on a fresh database, so the IF NOT EXISTS guards are unnecessary and can mask errors if objects already exist unexpectedly.


mkindahl (approved):
LGTM

PR #842, opened 2026-04-14
feat(multipooler): publish topology state asynchronously

dweitzman:
Discovery information (pooler type, serving status) is eventually-consistent by design — consumers like multigateway already tolerate stale reads. Making multipooler block on etcd writes during failovers or type changes means etcd availability issues can slow down or fail operations that have already succeeded locally in PostgreSQL.

Introduce topoPublisher: a background goroutine that reflects in-memory state to etcd as soon as practical. Callers invoke Notify after each state transition; the goroutine wakes immediately and also retries on a 30s ticker to recover from missed writes during etcd outages.

### DIFF
--- a/go/services/multipooler/manager/topo_publisher.go
+++ b/go/services/multipooler/manager/topo_publisher.go
@@ -59,7 +59,15 @@ func newTopoPublisher(logger *slog.Logger, topoClient topoRegistrar) *topoPublis
 
 // Notify records mp as the desired topology state and schedules an immediate
 // publish attempt. mp is cloned so the caller may reuse the original.
-func (tp *topoPublisher) Notify(mp *clustermetadatapb.MultiPooler) {
+//
+// ctx must carry an action lock (see AssertActionLockHeld). The action lock
+// serialises state transitions, preventing concurrent calls from racing to
+// overwrite each other's desired state.
+func (tp *topoPublisher) Notify(ctx context.Context, mp *clustermetadatapb.MultiPooler) error {
+	if err := AssertActionLockHeld(ctx); err != nil {
+		return err
+	}
+
 	tp.mu.Lock()
 	tp.desired = proto.Clone(mp).(*clustermetadatapb.MultiPooler)
 	tp.mu.Unlock()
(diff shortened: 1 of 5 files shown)


rafael (approved):
LGTM

go/services/multipooler/manager/topo_publisher.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+// publishIfNeeded writes the desired state to etcd if it differs from the last
+// successfully published state. A no-op when state is already current.
+func (tp *topoPublisher) publishIfNeeded(ctx context.Context) {
+	tp.mu.Lock()
+	desired := tp.desired
+	lastPublished := tp.lastPublished
+	tp.mu.Unlock()
```
mkindahl, line 113: I think there is a race condition here, but it looks harmless.

1. T1 starts executing `publishIfNeeded`
2. T1 Locks the mutex and reads `desired` S1 and `lastPublished` S0.
3. T2 calls `Notify` and updates `desired` with S2.
4. T2 starts executing `publishIfNeeded`
5. T2 locks the mutex and reads `desired` S2 and `lastPublished` S0.
6. T2 succeeds in publishing S2
7. T2 locks mutex and set `lastPublished` to S2
8. T1 continues execution and publishes S1
9. T1 locks mutex and set `lastPublished` to S1

I think it is harmless because a later execution will notice the difference and fix it, but in the meantime some nodes might act on incorrect (old) information.

You can reorder the last four steps to find other interesting combinations.

mkindahl (approved):
There is a race condition, but it looks harmless, so approving.

go/services/multipooler/manager/topo_publisher.go
dweitzman: Good catch! I'm going to fix this by making `Notify()` assert that the action lock is held. It was already the case that the action lock was held whenever Notify() was called, but asserting it makes it explicitly and less error-prone

mkindahl: I don't think it solved it. The problem is that the lock is released between read and update, but it should probably be held for the duration to serialize the operations. Since an "external" call is made, this is probably not a good idea, so some thought need to be put into this to figure out a way to avoid the race.

However, as I said, it seems harmless.

PR #843, opened 2026-04-15
feat(multiorch): switch from polling to ManagerHealthStream

mkindahl:
Replace the queue-based healthcheck worker pool with per-pooler ManagerHealthStream gRPC server-streaming connections. Each multipooler pushes a full StatusResponse snapshot to multiorch on every state change and on a periodic 5s heartbeat, keeping the pooler store continuously up to date without periodic polling.

The old queue/worker/ticker architecture (discovery.go, healthcheck.go, queue.go) is removed. recheckProblem no longer needs to force-poll poolers before re-running analysis, and the post-recovery shard refresh is also removed — the streaming store makes both unnecessary.

applySnapshot now stores ConsensusTerm, AvailabilityStatus, and ConsensusStatus as separate top-level fields on PoolerHealthState so recovery code can access consensus metadata without going through the operational Status subtree.

Fix a bug in tryPgRewind: when pg_rewind fails or is not feasible, markPoolerDrained was returning nil, causing fixNotReplicating to call verifyReplicationStarted a second time and block the recovery cycle for an extra 5 seconds. tryPgRewind now returns errPoolerDrained so the caller returns immediately.

Also add a pre-check in fixNotReplicating that skips pg_rewind when the primary's postgres is not ready, preventing a scenario where stopping the replica's postgres against a dead source leaves two nodes down simultaneously.

Closes MUL-357


go/services/multiorch/recovery/engine.go
```diff
@@ -67,63 +66,36 @@ import (
...
-//	                                            │   Pooler Store       │
-//	                                            └──────────────────────┘
+//	- Clean up stale data from the in-memory state store
+//
+// StreamManager (per-pooler health streams):
+//
+//	Each pooler maintains a long-lived StreamManagerHealth gRPC stream to multiorch.
+//	Poolers push full health snapshots on every state change and on a periodic
+//	heartbeat (every 5s). On disconnect, the stream reconnects with exponential
+//	backoff (1s → 2s → … → 30s cap).
```
mkindahl: Question is if we need a health check every 5 seconds. That is quite often, especially considering that we get changes pushed immediately. Health checks would just stop coming if the pooler is stopped for some reason (or overloaded).

go/services/multiorch/recovery/recovery_loop.go
```diff
@@ -284,69 +282,26 @@ func (re *Engine) attemptRecovery(ctx context.Context, problem types.Problem) {
...
-		re.logger.InfoContext(ctx, "forcing refresh of all poolers post recovery",
-			"shard_key", problem.ShardKey.String(),
-		)
-		re.forceHealthCheckShardPoolers(ctx, problem.ShardKey, nil /* poolersToIgnore */)
-	}
 }
 
-// recheckProblem force re-polls the pooler and re-runs analysis
-// to check if the problem still exists.
+// recheckProblem re-runs analysis on the current store state to confirm the
```
rafael, line 288: Some context on the forcePoll: after a successful recovery action, we'd ensure the state reflected the post-remediation status check. Now that we have streaming, we're losing that guarantee. The latency is usually small, but wondering if we should still protect against it.

Thinking of an scenario where there is network weather right after the remediation and we don't get the most up to date view of the health, before the next recovery cycle (unlikely but seems plausible). 

So I think there is a small risk of acting on state that doesn't yet reflect what the coordinator just remediated. 

**One thing worth considering**: adding a targeted RefreshPoolerHealth(poolerID) to StreamManager that does a one-off Status RPC and updates the store synchronously, called only after successful recovery actions. Lighter than the old forceHealthCheckShardPoolers (one pooler vs. all shard poolers), and gives a hard guarantee we're seeing post-remediation state before the next cycle runs. I don't feel strongly about this, but sharing for consideration. I don't think this introduces a correctness problem, just that we could try to take actions on stale information that then would fail.

go/services/multiorch/recovery/stream_manager.go
```diff
@@ -0,0 +1,279 @@
...
+		sm.logger.WarnContext(ctx, "received snapshot with nil status, skipping",
+			"pooler_id", poolerID)
+		return
+	}
+
+	status := snapshot.Status.Status
+	now := timestamppb.Now()
+
+	// Fetch consensus status separately — non-critical, same as old pollPooler.
+	consensusState := sm.fetchConsensusStatus(ctx, poolerID, poolerHealth)
```
rafael: This might be incidental complexity that we added when we did not have all the interfaces. I can't think of a good reason on why this should be a separate RPC. 

I think we can drop  `fetchConsensusStatus` entirely. My mental model is that we are only using it to fetch `pooler.ConsensusStatus.CurrentTerm`, which can be replaced with `pooler.ConsensusTerm.TermNumber`

@dweitzman - could you double check my analysis here?

I think this will have even more cleanups soon. But for now, that would help us remove this extra RPC entirely

go/services/multiorch/recovery/stream_manager.go
```diff
@@ -0,0 +1,279 @@
...
+	if cancel, exists := sm.streams[poolerID]; exists {
+		cancel()
+		// The goroutine removes itself from sm.streams via its defer.
+	}
+}
+
+// runStream manages the lifecycle of one stream, reconnecting with backoff on failure.
+// It reads the latest MultiPooler metadata from the store on each reconnect attempt
+// so hostname/port changes are picked up automatically.
+func (sm *StreamManager) runStream(ctx context.Context, poolerID string) {
```
rafael: We already have a retry library. I think we can use that here. [Here](https://github.com/rafael/multigres/blob/743c0af05416f5ca9a0735acf94e71a65b08f1ae/go/cmd/pgctld/command/server.go#L432) a place where we use our retry library.

go/services/multipooler/manager/rpc_backup.go
```diff
@@ -385,7 +385,16 @@ func (pm *MultiPoolerManager) restoreFromBackupLocked(ctx context.Context, backu
 
 	// Mark as initialized after successful restore
 	return telemetry.WithSpan(ctx, "restore/mark-initialized", func(_ context.Context) error {
-		return pm.setInitialized()
+		if err := pm.setInitialized(); err == nil {
```
rafael: nit - I think it would be more ideomatic to have:

```

if err := pm.setInitialized(); err != nil {
   return err
 }
 pm.broadcastHealth()
 return nil
```

rafael (commented):
I couldn't finish a full pass. But have some initial comments so you can have tomorrow morning

go/services/multiorch/recovery/stream_manager.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		if resp.Snapshot != nil {
+			sm.applySnapshot(ctx, poolerID, poolerHealth, resp.Snapshot)
+		}
+	}
+}
+
+// applySnapshot writes health fields from a snapshot into the pooler store.
+// This mirrors the field writes performed by the old pollPooler function on success.
+func (sm *StreamManager) applySnapshot(ctx context.Context, poolerID string, poolerHealth *multiorchdatapb.PoolerHealthState, snapshot *multipoolermanagerdatapb.ManagerHealthSnapshot) {
```
dweitzman: This could be deferred to a follow-up PR, but I think we're going to want some kind of check here to compare the recency of snapshots

Then even if we get health information from multiple sources (like included as part of a different gRPC response) or if there's weird network and goroutine behavior, we can still have confidence that this health snapshot is always moving forwards in time and never back

go/services/multiorch/recovery/engine.go
```diff
@@ -584,10 +548,6 @@ func (re *Engine) TriggerRecoveryNow(ctx context.Context, maxCycles uint32) ([]D
 	re.logger.InfoContext(ctx, "TriggerRecoveryNow: forcing immediate recovery execution",
 		"max_cycles", maxCycles)
 
-	// Force health check all poolers to get fresh state
```
dweitzman, line 590: Removing this seems like it could cause a regression in endtoend tests.

The goal of TriggerRecoveryNow() is to allow endtoend tests to be faster and more deterministic by doing a health check & recovery on-demand instead of waiting for natural health checks to detect issues

The risk here is that trigger recovery now can be used to run exactly recovery cycle:
- See if any problem exists
- Try to fix them until they're fixed
- Return

If we don't force health checks here, we may fail to observe that a problem exists here and return immediately without actually waiting until the problem is fixed, and then when the test goes on to assert the the problem is fixed it might not be

go/services/multiorch/recovery/discovery.go
dweitzman, line 1: This is going to be a hell of a merge conflict with https://github.com/multigres/multigres/pull/758, but the functionality this adds some more immediately useful than the type safety that one adds

go/services/multiorch/recovery/stream_manager.go
```diff
@@ -0,0 +1,279 @@
...
+
+// StreamManager maintains one StreamManagerHealth stream per pooler.
+// It replaces the polling loop: instead of periodically calling the Status RPC,
+// each pooler pushes health snapshots to multiorch via a long-lived gRPC stream.
+//
+// When a snapshot arrives the StreamManager writes the same health fields into
+// the pooler store that the old pollPooler function wrote on success. On stream
+// disconnect the pooler is marked unreachable and reconnection is attempted with
+// exponential backoff (1s → 2s → … → 30s cap).
+type StreamManager struct {
```
dweitzman: In terms of code organization, my instinct is that it makes the most sense to align the health stream lifetime with orch's awareness of the pooler

That was the approach I took with gateway health checks: https://github.com/multigres/multigres/blob/3259d91f746c634eb5fec8a24b58515489fb6972/go/services/multigateway/poolergateway/pooler_connection.go#L207

The philosophy behind that is that we if we're tracking a pooler we should have one grpc client for it and streaming health for it, and if we decide to stop tracking it we should delete both

The advantage of that is that it feels obvious looking at the code that streams are live if and only if we're aware of a pooler. With the streams in a separate global here, it's not as obvious when and why streams are created and destroyed so it's necessary to read more of the code to build confidence about what it's doing

This isn't necessarily a blocking issue for me, but it's a strong preference. One option would actually be to refactor pooler_connection by pulling it out under go/common so the streaming logic is basically the same and the only difference is what information is requested or used. It doesn't feel like we need to have competing implementations of a pooler client with streaming health for gateway vs orch

dweitzman (commented):
I won't be able to read all of this today, but leaving a few comments on things I've seen so far

go/services/multiorch/recovery/discovery.go
mkindahl: One option could be to leave this file and not remove it yet. It is going to be a lot of dead code, but anyway.

go/services/multiorch/recovery/recovery_loop.go
mkindahl: It is a good point and similar to the point made in [the comment above](https://github.com/multigres/multigres/pull/843/changes#r3089610478). In some cases we want to ensure that we have a most recent state to act on.

go/services/multiorch/recovery/stream_manager.go
mkindahl: > This might be incidental complexity that we added when we did not have all the interfaces. I can't think of a good reason on why this should be a separate RPC.
> 
> I think we can drop `fetchConsensusStatus` entirely. My mental model is that we are only using it to fetch `pooler.ConsensusStatus.CurrentTerm`, which can be replaced with `pooler.ConsensusTerm.TermNumber`

It indeed looks like that. I removed it as a separate commit for now, but going to look into the other comments.

go/services/multipooler/manager/rpc_backup.go
mkindahl: Fixed. :)

go/services/multiorch/recovery/recovery_loop.go
mkindahl: I introduced a bidirectional stream with a Poll() method in #853 and will use it here.

go/services/multiorch/recovery/stream_manager.go
mkindahl: Added that to #853. Please take a look there. This PR will be on top of #853.

go/services/multiorch/recovery/engine.go
mkindahl: I have introduced a Poll method and a bidirectional stream in #853, which is used here.

go/services/multiorch/recovery/stream_manager.go
mkindahl: I think this is a good idea. It would also eliminate the need for separate RPCs between the orch and the pooler (but we might need some for other actors, such as the admin).

I changed to use a bidirectional stream for each pooler. The stream sends snapshots and health messages upstreams to the orch and you can send poll messages and other stuff downstream to control the pooler.

go/services/multiorch/recovery/recovery_loop.go
mkindahl: This is now rebased on the bidirectional stream.

go/services/multiorch/recovery/stream_manager.go
mkindahl: It is now used in this patch. Please tell me if I missed some place.

go/services/multiorch/recovery/stream_manager.go
mkindahl: This is now removed and we rely entirely on the health stream.

proto/multipoolermanagerdata.proto
```diff
@@ -241,6 +238,13 @@ message Status {
   // This is the combined check: process must exist AND respond to pg_isready.
   // Use postgres_running to check only if the process exists.
   bool postgres_ready = 14;
+
+  // True while pg_promote() has been called but postgres has not yet transitioned
+  // to primary mode (postgres_ready=false). Cleared once postgres accepts connections
+  // as primary, or if promotion fails or times out.
+  // Used by multiorch to suppress spurious PrimaryIsDead detection during the
+  // brief window when the promoted node is not yet accepting connections.
+  bool promotion_in_progress = 15;
```
mkindahl: This checks a particular status of the server, but it might be better to actually have an enum representing the status of the server, for example, recovering, promoting, etc. That will make the distinct status clear in the logs and also avoid weird combinations if we keep using bools for the server status.

It also aids in observability.

go/services/multiorch/recovery/discovery.go
mkindahl: This change should not be there any more.

proto/multipoolermanagerdata.proto
mkindahl: Adding this as a separate commit. PTAL.

proto/multipoolermanagerdata.proto
```diff
@@ -237,6 +234,23 @@ message Status {
...
+  // Cannot determine state: no data directory, process dead, or pgctld unreachable.
+  POSTGRES_STATUS_UNKNOWN = 0;
+  // Process is alive but not yet accepting connections (startup in progress).
+  POSTGRES_STATUS_STARTING = 1;
+  // pg_is_in_recovery() = true; replicating from primary.
+  POSTGRES_STATUS_STANDBY = 2;
+  // pg_promote() has been issued; postgres has not yet transitioned to primary mode.
+  POSTGRES_STATUS_PROMOTING = 3;
+  // pg_is_in_recovery() = false; accepting read-write connections.
+  POSTGRES_STATUS_PRIMARY = 4;
```
mkindahl, line 253: "Promoting" state is needed, but other names than "primary" and "standby" might help avoiding confusion between assigned role and actual status. Open to suggestions.

go/services/multiorch/recovery/analysis/generator_test.go
```diff
@@ -27,26 +27,21 @@ import (
...
-// primaryConsensusStatus builds a consensusdata.StatusResponse that names id as
-// the primary in its current rule with the given coordinator term. This is the
-// minimal fixture required for commonconsensus.IsPrimary to return true for a
-// given pooler.
-func primaryConsensusStatus(id *clustermetadatapb.ID, term int64) *consensusdatapb.StatusResponse {
-	return &consensusdatapb.StatusResponse{
+// primaryConsensusStatus builds a ConsensusStatus that names id as the primary
+// in its current rule with the given coordinator term. This is the minimal
+// fixture required for commonconsensus.IsPrimary to return true for a given pooler.
+func primaryConsensusStatus(id *clustermetadatapb.ID, term int64) *clustermetadatapb.ConsensusStatus {
```
rafael, line 38: nit - not for this PR, but I think I added something similar. Also, it looks like this was added to leader_appointment as well. We should probably consolidate and have a general helper that can be reused.

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
```diff
@@ -88,6 +88,20 @@ func (a *PrimaryIsDeadAnalyzer) Analyze(sa *ShardAnalysis) ([]types.Problem, err
...
 	// interval (~30s).
 
+	// Suppress while the topology primary is explicitly mid-promotion.
+	// Three guards are all required:
+	//   - PromotingPrimaryID != nil: multipooler has flagged pg_promote() is running
+	//   - PrimaryPoolerReachable: stream is live, so the flag is current (not stale)
+	//   - PrimaryPostgresRunning: postgres process is still alive
+	// If postgres crashes during promotion, PrimaryPostgresRunning=false and we fall through.
+	// If the multipooler crashes, PrimaryPoolerReachable=false and we fall through.
+	if sa.PromotingPrimaryID != nil && sa.PrimaryPoolerReachable && sa.PrimaryPostgresRunning {
```
rafael, line 98: nice

go/services/multiorch/recovery/engine.go
```diff
@@ -67,63 +68,30 @@ import (
...
-//	                                            │   Worker Pool       │
-//	                                            │   (N workers)       │
-//	                                            └──────────┬──────────┘
-//	                                                       │
-//	                                                       │ updates
-//	                                                       ▼
-//	                                            ┌──────────────────────┐
-//	                                            │   Pooler Store       │
-//	                                            └──────────────────────┘
+// HealthStream (per-pooler health streams):
```
rafael, line 79: Simpler and more powerful!

rafael (approved):
This is looking good to me.

proto/multipoolermanagerdata.proto
```diff
@@ -257,6 +271,26 @@ message StatusRequest {
 
 message StatusResponse {
   Status status = 1;
+
+  // Consensus term for this pooler. Moved here from Status (field 9) so that
+  // consensus metadata travels alongside but separate from operational state.
+  // Populated by the multipooler on every Status RPC call and health stream push.
+  ConsensusTerm consensus_term = 2;
```
dweitzman, line 284: FYI I'm deleting ConsensusTerm in https://github.com/multigres/multigres/pull/904 since all the information is already in ConsensusStatus, but I can wait for your PR and then deal with the merge conflict

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
dweitzman: Is this not already handled by whatever logic notices if another coordinator acted recently (based on the accepted time of consensus terms) and adds a grace period where other coordinators won't act until the previous coordinator has had time to finish whatever it was doing?

go/services/multiorch/recovery/pooler_watcher.go
```diff
@@ -245,12 +249,12 @@ func extractCellNameFromPath(watchPath string) string {
...
-	cell      string
-	targets   func() []config.WatchTarget
-	store     *store.PoolerStore
-	queue     *Queue
-	logger    *slog.Logger
+	topoStore   topoclient.Store
+	cell        string
+	targets     func() []config.WatchTarget
+	store       *store.PoolerStore
+	onNewPooler func(poolerID string)
```
dweitzman: Any reason not to used the typed proto3 ID type here (vs a string)?

go/services/multiorch/recovery/analysis/generator.go
```diff
@@ -433,8 +447,18 @@ func (g *AnalysisGenerator) computeShardLevelFields(sa *ShardAnalysis, poolers m
...
+		// PrimaryHasResigned: AvailabilityStatus and ConsensusTerm are populated from
+		// StatusResponse on every health stream snapshot, so PrimaryNeedsReplacement
+		// correctly detects BeginTerm REVOKE demotions without a separate RPC.
 		sa.PrimaryHasResigned = types.PrimaryNeedsReplacement(topologyPrimary)
-		sa.PrimaryReachable = topologyPrimary.IsLastCheckValid && topologyPrimary.GetStatus().GetPostgresReady() && !sa.PrimaryHasResigned
+		// PrimaryReachable requires the topology primary to be serving as PRIMARY and
+		// not have resigned. A resigned primary has voluntarily stepped down via
+		// BeginTerm REVOKE; treating it as reachable would prevent PrimaryIsDead
+		// detection even when postgres is still running on the demoted node.
+		sa.PrimaryReachable = topologyPrimary.IsLastCheckValid &&
```
dweitzman, line 458: Just as a note (not requesting any changes), sometime soon I still want to delete almost all the generator signals, or maybe just delete all of generator.go.

go/services/multiorch/recovery/engine.go
```diff
@@ -649,29 +612,69 @@ func (re *Engine) TriggerRecoveryNow(ctx context.Context, maxCycles uint32) ([]D
...
+				)
+				return
+			}
+			if ph, ok := re.poolerStore.Get(pb.id); ok && ph.StreamSnapshotsReceived > pb.baseline {
+				break
+			}
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(10 * time.Millisecond):
```
dweitzman, line 676: Not a blocker, but at some point I think a write-through cache would be simpler and cleaner than a 10ms polling interval

The PoolerConnection abstractions does this for multigateway's health streams, for example, which would make on-demand health checks a relatively simple and minimal change in the PoolerConnection abstraction model

go/services/multipooler/manager/manager.go
```diff
@@ -148,6 +148,11 @@ type MultiPoolerManager struct {
 	// rewind completes successfully. Set by emergencyDemoteLocked, cleared by RewindToSource.
 	rewindPending atomic.Bool
 
+	// promotionInProgress is set while pg_promote() has been called but postgres has not yet
+	// transitioned to primary mode. Cleared when promotion completes (success or failure).
+	// Reported in the health status so multiorch can suppress spurious PrimaryIsDead detection.
+	promotionInProgress atomic.Bool
```
dweitzman, line 154: I'm a little bit skeptical of this state since we presumably would need to time out its relevance to prevent blocking failover indefinitely, but if you think it's the right thing to do I'm we can discuss after this is merged.

go/services/multipooler/manager/manager.go
```diff
@@ -156,6 +161,11 @@ type MultiPoolerManager struct {
 	// pgMonitorLastLoggedReason tracks the last logged reason in the monitor to avoid duplicate logs.
 	pgMonitorLastLoggedReason string
 
+	// pgMonitorLastRunning tracks whether postgres was running in the last monitor iteration.
+	// Used to detect transitions so the healthStreamer can push an immediate snapshot to
+	// orchestrators when postgres starts or stops running.
+	pgMonitorLastRunning bool
```
dweitzman: This also feels a little off to me. In my mental model we should always tell the health stream what the current health is and if there's no material change it should just avoid broadcasting.

Happy to dicuss after the PR is merged though, if you think this it the right thing for now

dweitzman (commented):
Nothing blocking, but sending out some comments. Still need to read more of the test code

go/common/timeouts/rpc.go
```diff
@@ -35,3 +35,15 @@ const DefaultHealthHeartbeatInterval = 30 * time.Second
...
+// waiting for poll responses from connected poolers. A connected pooler should
+// respond within one network round-trip; this cap covers slow nodes and avoids
+// blocking TriggerRecoveryNow indefinitely.
+//
+// This value was picked to be long enough to allow a round-trip to a slow node,
+// but is shorter than the snapshot interval to ensure that we don't wait too
+// long for a response before triggering recovery. In practice, this should be
+// sufficient for most network conditions and allows for some variability in
+// response times without causing undue delays in recovery.
+const PollResponseWait = 500 * time.Millisecond
```
dweitzman, line 49: I have some doubts about whether our Status() handlers are reliably faster than this since they're inspecting live system state, but I guess we'll find out when the tests either get flakier or not

go/test/endtoend/multiorch/bootstrap_test.go
```diff
@@ -215,19 +206,24 @@ func TestBootstrapInitialization(t *testing.T) {
 		for _, inst := range setup.Multipoolers {
 			allInstances = append(allInstances, inst)
 		}
-		shardsetup.EventuallyPoolerCondition(t, allInstances, 30*time.Second, 1*time.Second,
-			func(name string, s *multipoolermanagerdatapb.Status) (bool, string) {
-				if !s.IsInitialized {
-					return false, "not yet initialized"
+		shardsetup.EventuallyPoolersCondition(t, allInstances, 30*time.Second, 1*time.Second,
```
dweitzman: I don't think the switch from EventuallyPoolerCondition() to EventuallyPoolersCondition() was necessary? It's still checking on a per-pooler basis, and the static typing is cleaner with EventuallyPoolerCondition()

go/test/endtoend/multiorch/bootstrap_test.go
```diff
@@ -361,18 +357,23 @@ func TestBootstrapInitialization(t *testing.T) {
...
-			func(_ string, s *multipoolermanagerdatapb.Status) (bool, string) {
-				if !s.IsInitialized {
-					return false, "not yet initialized"
-				}
-				if !s.PostgresReady {
-					return false, "postgres not running"
-				}
-				if s.ConsensusTerm == nil || s.ConsensusTerm.TermNumber == 0 {
-					return false, "consensus term not yet assigned"
+		shardsetup.EventuallyPoolersCondition(t, []*shardsetup.MultipoolerInstance{standbyInst}, 90*time.Second, 1*time.Second,
```
dweitzman: Same comment

go/test/endtoend/multiorch/dead_primary_recovery_test.go
```diff
@@ -563,44 +564,54 @@ func waitForNodeToRejoinAsStandby(t *testing.T, setup *shardsetup.ShardSetup, mu
 	primaryInst := setup.GetMultipoolerInstance(expectedPrimaryName)
 	require.NotNil(t, primaryInst, "primary %s should exist", expectedPrimaryName)
 
-	shardsetup.EventuallyPoolerCondition(t,
+	shardsetup.EventuallyPoolersCondition(t,
```
dweitzman: Same comment

go/services/multiorch/recovery/analysis/generator_test.go
```diff
@@ -475,6 +470,108 @@ func TestPopulatePrimaryInfo_PrimaryPostgresDown(t *testing.T) {
 	assert.False(t, sa.PrimaryReachable, "primary should NOT be reachable when postgres is down")
 }
 
+// TestPopulatePrimaryInfo_DemotedViaBegintermRevoke covers the scenario where a primary is
+// demoted via BeginTerm REVOKE and restarted as a standby (emergencyDemoteLocked behavior).
+// After restart, PoolerType=REPLICA in the health snapshot and primary_term stays > 0
+// (only DemoteStalePrimary clears it). PrimaryReachable must be false so PrimaryIsDead
+// triggers and a new primary can be elected.
+func TestPopulatePrimaryInfo_DemotedViaBegintermRevoke(t *testing.T) {
```
dweitzman: Nit: Beginterm -> BeginTerm

go/common/timeouts/rpc.go
mkindahl: I added the not just to point out what the purpose of this is. Note that it is shorter than the heartbeat interval so that we can move forward faster, but the heartbeat interval is the upper bound of what this value needs to be.

go/services/multiorch/recovery/analysis/generator.go
mkindahl: Yes, but I want to take things one step at a time so that we properly review each change. Bugs can easily creep in. It is also easier to review such a change.

go/services/multiorch/recovery/analysis/generator_test.go
mkindahl: Yes, there is a primaryRuleStatus but it was in a different file and local there.

go/services/multiorch/recovery/analysis/generator_test.go
mkindahl: Fixed.

go/services/multiorch/recovery/engine.go
mkindahl: Sounds like a good idea. We could probably make the cache generic and just use it here as well.

go/test/endtoend/multiorch/bootstrap_test.go
mkindahl: No, Claude is a little aggressive with name changes. I've already reverted a lot of such cases. I'll deal with this one as well

proto/multipoolermanagerdata.proto
mkindahl: Yes, but some information from ConsensusTerm is needed with the current code, so I opted to add this for the time being. Once we have the term number in ConsensusStatus we can remove this field.

mkindahl: Extended the comment.

go/services/multiorch/recovery/analysis/generator_test.go
mkindahl: I renamed this function to primaryRuleStatus to make it clear that they do the same thing. We should probably consolidate this into a package with test helpers, but I would like to take that as a separate PR to avoid overloading this PR.

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
mkindahl: Adding a grace period to allow the coordinator to finish the job introduces an unnecessary dependency, IMHO.

Once the coordinator finished the job, the server will no longer be promoting and hence this will not be blocked. This allows the analyzers to act only on status information from the poolers as well as some temporal properties regarding the status in previous snapshots. Although adding extra sources of information is possible, it makes analysis more complicated and risks introducing subtle problems.

In general, I try to keep subsystem dependencies down, and I think this is a clear and understandable way to prevent a promoting server from being deemed dead.

go/services/multiorch/recovery/pooler_watcher.go
mkindahl: We are storing the poolers in a map keyed by the pooler, so at the end we need a string. The conversion can be made inside the function or outside, and the previous `Start` function accepted a string rather than a `ID`.

I modified all three functions, Start, Stop, and Poll, to take an ID.

Note that we cannot use the pooler ID as key to the map (it's a pointer to a structure).

go/services/multiorch/recovery/engine.go
mkindahl: Added an issue to Linear for this.

go/services/multipooler/manager/manager.go
mkindahl: Right now, this is set inside `promoteStandbyToPrimary` while `pg_promote` is running and cleared after, so the code there is straightforward. Any timeouts needed could be added there and would naturally trigger a pick of a more suitable primary should the promotion fail.

go/services/multipooler/manager/manager.go
mkindahl: It makes sense to remove this field and compare states in the monitoring loop. I updated the logic to save the previous state, call the inner monitoring function, and then call `broadcastHealth` only if anything changed.

It also improved the code by not repeating the broadcast inside the monitoring iteration function and move it out to the caller, where it belongs.

go/test/endtoend/multiorch/bootstrap_test.go
mkindahl: Handled.

go/services/multiorch/consensus/leader_appointment_test.go
```diff
@@ -1508,12 +1479,15 @@ func TestAppointLeader(t *testing.T) {
 		// Create 3 nodes: mp1 (most advanced WAL), mp2, mp3
 		mp1 := createMockNode(fakeClient, "mp1", 5, "0/3000000", true, nil)
 		mp1.Status.PostgresReady = true
+		mp1.Status.PostgresRunning = true
```
mkindahl, line 1482: This is mostly for completeness. They are not used, but were missing before.

PR #844, opened 2026-04-15
feat(tests): add port pool script and update skill for integration tests

rafael:
# Desc

* This makes sure claude knows about the port allocator introduced in: https://github.com/multigres/multigres/pull/524/ 
* Add scripts/portpool.sh to coordinate port allocation across parallel integration tests, and update the mt-dev skill to use it automatically.


mkindahl (approved):
LGTM

PR #845, opened 2026-04-15
feat(consensus): propagate ConsensusStatus in BeginTerm and Status responses

dweitzman:
Rename HighestCoordinatorPromise → TermRevocation (semantics shift: revocation of participation in all terms below T). Rename NodePosition → PoolerPosition throughout proto and Go code.

Add ConsensusStatus (TermRevocation + CurrentPosition + HighestKnownRule) to BeginTermResponse and StatusResponse so multiorch has authoritative pooler state at decision points. Add a position cache to ruleStore so executeRevoke can populate ConsensusStatus without a second postgres round-trip.

Wire ConsensusStatus into the coordinator-side PoolerAnalysis so candidate selection can use committed rule numbers (coordinator_term + leader_subterm) rather than a flat leadership_term integer.


go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -257,9 +273,107 @@ func (pm *MultiPoolerManager) executeRevoke(ctx context.Context, term int64, res
...
+	pm.mu.Unlock()
+
+	var term *multipoolermanagerdatapb.ConsensusTerm
+	if cs != nil {
+		var err error
+		term, err = cs.GetTerm(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read consensus term: %w", err)
+		}
+	}
```
mkindahl: This smells like a race condition. The `consensusState` is read from `pm` and then used outside locking. In theory it can be modified as a result of some other thread making modifications to `pm`, but I cannot find a real case.

mkindahl (approved):
There are some suspicions uses of locks, but I cannot find a real case where this could cause a problem, so approving.

go/services/multipooler/manager/rpc_consensus.go
dweitzman: `.GetTerm(ctx)` asserts that the action lock is held, so there's some protection here against concurrency callers. The problem scenario would be if someone holding the action lock started a goroutine that also holds the action lock and races against itself, but that feels like an unexpected scenario

dweitzman: To add more context: this sort of concern is what motivated me originally to make `GetTerm()` assert the action lock is held

It feels suspicious to read the consensus term without holding a lock for any reason other than status reporting: if you read a term, you don't want it to change before you act on it. Status reporting and postgres monitor pre-checking (before getting the action lock) use `GetInconsistentTerm()` which is identical to `GetTerm()` aside from not asserting a lock is held

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -77,34 +77,33 @@ func (pm *MultiPoolerManager) BeginTerm(ctx context.Context, req *consensusdatap
 	// Term Acceptance (Consensus Rules)
 	// ========================================================================
 
-	pm.mu.Lock()
```
dweitzman, line 80: Deleting some no-longer-necessary locking while I'm in here, since my previous PR made it so `pm.consensusState` is only set during the constructor

PR #850, opened 2026-04-16
fix(shardsetup): stops postgres gracefully during integration tests shutdown

rafael:
# Desc
* One more cleanup related to how we clean up the cluster when exiting integration tests. We were not explicitly stopping Postgres. This had the side effect that Postgres was not being shut down cleanly, which sometimes leaked segments on macOS.
* Cleanup() now gracefully terminates all processes (pgctld, multipooler, multigateway, multiorch) before cancelling the context, ensuring pg_ctl stop runs and releases segments
* TerminateGracefully now stops PostgreSQL via gRPC for pgctld instances before sending SIGTERM, and accepts a logf function instead of *testing.T so it can be called from Cleanup() (which runs from
  TestMain without a t)
* Remove the `ProcessInstance.Stop()` footgun, this was effectively the same as `TerminateGracefully`, but it was not used consistently. 

# Tests
To verify:

Run this in main:
```
make build
go test -run TestBufferMultipleFailovers -count=5 ./go/test/endtoend/queryserving/...
```
Notice the output of:
```
ipcs -m
```
There should be some leaks.

Run the same in this branch:
```
make build
go test -run TestBufferMultipleFailovers -count=5 ./go/test/endtoend/queryserving/...
ipcs -m 
```
You shouldn't see any leaks.


go/test/endtoend/shardsetup/process.go
```diff
@@ -414,24 +414,6 @@ func (p *ProcessInstance) LogRecentOutput(t *testing.T, context string) {
...
-		return // Process not running
-	}
-
-	// If this is pgctld, stop PostgreSQL first via gRPC
-	if p.Binary == "pgctld" {
-		p.stopPostgreSQL()
-	}
-
-	// Then kill the process
-	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
```
dweitzman, line 430: Was this issue with this just that the timeout is too small?

dweitzman (approved):

go/test/endtoend/shardsetup/process.go
rafael: Yeah, I saw an issue with this too. It was too small and it did not finish cleanly

PR #851, opened 2026-04-16
test(shardsetup): improve checkBootstrapStatus log readability

rafael:
## Summary

- Reformats the `checkBootstrapStatus` log from a dense single-line summary into a labelled multi-line block
- Shows `<unknown>` when no primary has been identified yet, and `<none>` when no backup exists
- Each pooler status appears on its own indented line instead of being pipe-delimited inline

## Before
```
checkBootstrapStatus: SUMMARY primary=pooler-1 initialized=1/3 latest_backup="20260416-150038F" [pooler-1: queryable, type=PRIMARY, sync_replication_configured [...] | pooler-2: not_queryable | pooler-3: not_queryable]
```

## After
```
checkBootstrapStatus:
  primary:       pooler-1
  initialized:   1/3
  latest_backup: 20260416-150038F
  poolers:
    pooler-1: queryable, type=PRIMARY, sync_replication_configured [...]
    pooler-2: not_queryable
    pooler-3: not_queryable
```

### DIFF
--- a/go/test/endtoend/shardsetup/setup.go
+++ b/go/test/endtoend/shardsetup/setup.go
@@ -985,8 +985,17 @@ func checkBootstrapStatus(ctx context.Context, t *testing.T, setup *ShardSetup) 
 		attribute.StringSlice("pooler.statuses", poolerStatuses),
 	)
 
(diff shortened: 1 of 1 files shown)


dweitzman:
We discussed offline that maybe it'd reduce log spam to only print logs when a pooler or orch report different status than the previous state, which could maybe be determined by using proto deep equality checking after ignoring LSNs

PR #853, opened 2026-04-17
feat(multiorch): add ManagerHealthStream bidirectional stream

mkindahl:
Add a bidirectional gRPC health stream from multipooler to multiorch, replacing the periodic `Status` RPC poll. Poolers push a full health snapshot on connection, on every state change, and as a periodic heartbeat. The orchestrator can send a poll request to trigger an immediate snapshot.

Proto:
- Add `ManagerHealthStream` RPC to `MultiPoolerManager` (bidirectional)
- Add `ManagerHealthStreamClientMessage` (start / poll variants)
- Add `ManagerHealthStreamResponse` and `ManagerHealthSnapshot` message types
- Add SnapshotTrigger enum (`INITIAL`, `BROADCAST`, `POLL`, `HEARTBEAT`) so each snapshot carries the reason it was sent
- Add `stream_connected`, `stream_connected_since`, `stream_snapshots_received` fields to PoolerHealthState

Multipooler server (grpcmanagerservice):
- Implement `ManagerHealthStream` handler: reads start message, sends a full `Status()` snapshot on connection, on each `healthStreamer` broadcast, and on poll requests from the orchestrator
- Add `broadcastHealth()` nil-safe wrapper for pm.healthStreamer.Broadcast()
- Call `broadcastHealth()` after SetPrimaryConnInfo, `UpdateSynchronousStandbyList`, and restore/mark-initialized so role and replication changes propagate immediately rather than waiting for the next heartbeat

RPC client:
- Add ManagerHealthStream send/recv interface
- Add ManagerHealthStream to MultiPoolerClient interface and gRPC client
- Add FakeManagerHealthStream and ManagerHealthStream to FakeClient

Multiorch client (health_stream.go, new):
- HealthStream maintains one ManagerHealthStream stream per pooler
- Sends start message on connect, applies snapshots into PoolerStore
- Poll(poolerID) sends a poll request on the active stream
- Exponential backoff reconnect (1s -> 30s cap) on disconnect
- Application-level staleness watchdog: cancels the stream and triggers reconnect if no snapshot arrives within timeout_seconds (default 90s); catches "live TCP, silent server goroutine" failures that gRPC keepalive does not cover. Timeout is seeded from each snapshot timeout_seconds.

Tests:
- Unit tests for HealthStream: snapshot apply (primary/replica), Poll, disconnect/reconnect, concurrent store update, deleted pooler, postgres ready time preservation, type mismatch, staleness timeout
- Integration test TestManagerHealthStream_SnapshotOnSetPrimaryConnInfo: opens a real stream to a standby pooler, calls SetPrimaryConnInfo, and asserts a SNAPSHOT_TRIGGER_BROADCAST snapshot arrives promptly

Part of MUL-357

### DIFF
--- a/go/services/multiorch/recovery/actions/shard_init_action.go
+++ b/go/services/multiorch/recovery/actions/shard_init_action.go
@@ -169,6 +169,7 @@ func (a *ShardInitAction) getInitializedPoolers(shardKey commontypes.ShardKey) (
 		}
 		if len(pooler.CohortMembers) > 0 {
 			cohortEstablished = true
+			initialized = nil
 			return false // stop iteration
 		}
 		if pooler.IsInitialized {
--- a/go/services/multiorch/recovery/health_stream.go
+++ b/go/services/multiorch/recovery/health_stream.go
@@ -1,0 +1,410 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package recovery
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"sync"
+	"time"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/multigres/multigres/go/common/rpcclient"
+	"github.com/multigres/multigres/go/common/topoclient"
+	multiorchdatapb "github.com/multigres/multigres/go/pb/multiorchdata"
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+	"github.com/multigres/multigres/go/services/multiorch/store"
+	"github.com/multigres/multigres/go/tools/retry"
+)
+
+const (
+	// streamReconnectInitialBackoff is the initial wait before retrying a
+	// failed ManagerHealthStream connection.
+	streamReconnectInitialBackoff = 1 * time.Second
+
+	// streamReconnectMaxBackoff caps the exponential backoff between retries.
+	streamReconnectMaxBackoff = 30 * time.Second
+
+	// streamStalenessTimeout is the default maximum time to wait for a health
+	// snapshot before treating the stream as stale and reconnecting.
+	//
+	// This is an application-level watchdog that catches the "server goroutine
+	// stuck but TCP alive" failure mode that gRPC keepalive does not cover.
+	// gRPC keepalive (10s ping / 10s timeout) handles dead TCP connections;
+	// this constant handles a live connection whose send loop has deadlocked.
+	//
+	// The server guarantees a snapshot at least every managerHealthPollingInterval
+	// (5s), so messages normally arrive far more frequently than this. The value
+	// is seeded from the server's recommended timeout in each snapshot.
+	streamStalenessTimeout = 90 * time.Second
+)
+
+// streamEntry holds the lifecycle handles for one active stream goroutine.
+type streamEntry struct {
+	cancel context.CancelFunc
+
+	// mu protects stream; set to the live gRPC stream after the start message
+	// is sent, and cleared on stream exit.
+	mu     sync.Mutex
+	stream rpcclient.ManagerHealthStream
+}
+
+// HealthStream maintains one HealthStream stream per pooler. It replaces the
+// polling loop: instead of periodically calling the Status RPC, each pooler
+// pushes health snapshots to multiorch via a long-lived gRPC stream.
+//
+// When a snapshot arrives the HealthStream writes the same health fields into
+// the pooler store that the old pollPooler function wrote on success. On stream
+// disconnect the pooler is marked unreachable and reconnection is attempted
+// with exponential backoff (1s → 2s → … → 30s cap).
+type HealthStream struct {
+	logger    *slog.Logger
+	rpcClient rpcclient.MultiPoolerClient
+	store     *store.PoolerStore
+
+	// stalenessTimeout overrides streamStalenessTimeout if positive.
+	// Set via WithStalenessTimeout; used in tests.
+	stalenessTimeout time.Duration
+
+	// Active stream goroutines, keyed by pooler ID string.
+	mu      sync.Mutex
+	streams map[string]*streamEntry
+
+	// Parent context; cancelled by Stop().
+	ctx    context.Context
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+}
+
+// Option is a functional option for NewHealthStream.
+type Option func(*HealthStream)
+
+// WithStalenessTimeout overrides the default staleness timeout. Intended for tests.
+func WithStalenessTimeout(d time.Duration) Option {
+	return func(hs *HealthStream) {
+		hs.stalenessTimeout = d
+	}
+}
+
+// NewHealthStream creates a HealthStream.
+//
+// Call Start() for each pooler that should be monitored.
+func NewHealthStream(
+	ctx context.Context,
+	rpcClient rpcclient.MultiPoolerClient,
+	poolerStore *store.PoolerStore,
+	logger *slog.Logger,
+	options ...Option,
+) *HealthStream {
+	smCtx, cancel := context.WithCancel(ctx)
+	hs := &HealthStream{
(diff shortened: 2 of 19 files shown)


proto/multipoolermanagerdata.proto
```diff
@@ -291,6 +291,57 @@ message StatusResponse {
...
+// Sent immediately on connection, on any state change, and periodically as a
+// heartbeat when no state change has occurred.
+message ManagerHealthSnapshot {
+  // Full health state of the pooler.
+  StatusResponse status = 1;
+
+  // The duration the orchestrator should use to detect a stale or dead stream.
+  // If no message is received within this duration, the pooler should be marked
+  // unhealthy.
+  int64 timeout_seconds = 2;
```
dweitzman: This could be be a `google.protobuf.Duration timeout = 2;`

proto/multipoolermanagerdata.proto
```diff
@@ -291,6 +291,57 @@ message StatusResponse {
   Status status = 1;
 }
 
+// ManagerHealthStreamRequest carries the orchestrator's identity when opening a
+// health stream. Sent as the first message inside ManagerHealthStreamClientMessage.
+message ManagerHealthStreamRequest {
+  // Identifier for the orchestrator instance opening this stream.
+  string orch_id = 1;
```
dweitzman: Should this be the `ID` proto message?

proto/multiorchdata.proto
```diff
@@ -93,4 +93,16 @@ message PoolerHealthState {
...
+
+  // Whether a StreamManagerHealth stream is currently active for this pooler.
+  bool stream_connected = 20;
+
+  // The time at which the current stream was established.
+  google.protobuf.Timestamp stream_connected_since = 21;
+
+  // The total number of health snapshots received over the lifetime of all streams
+  // to this pooler (monotonically increasing).
+  int64 stream_snapshots_received = 22;
```
dweitzman: Is this for making health updates sequenced? If so, I don't think it'll work if the multipooler is restarted. Making it an opaque "health_version" or something like that would allow you to seed it with the current timestamp on pooler startup

go/services/multiorch/recovery/health_stream.go
```diff
@@ -0,0 +1,317 @@
...
+		existing.LastSeen = now
+		existing.IsUpToDate = true
+		existing.IsLastCheckValid = true
+		existing.IsPostgresReady = status.PostgresReady
+		if status.PostgresReady {
+			existing.LastPostgresReadyTime = now
+		}
+		// NOTE: when PostgresReady is false, LastPostgresReadyTime is intentionally
+		// left at its previous value so callers can reason about "last known good" time.
+		existing.IsPostgresRunning = status.PostgresRunning
```
dweitzman, line 367: Copying every thing indvidually here seems a little error prone. For example: `CohortMembers` is missing in this logic.

Would it make sense to use proto helpers to clone the previous status and then just update the fields that need updating (timestamps, etc.)?

go/services/multiorch/recovery/actions/shard_init_action.go
```diff
@@ -169,6 +169,7 @@ func (a *ShardInitAction) getInitializedPoolers(shardKey commontypes.ShardKey) (
 		}
 		if len(pooler.CohortMembers) > 0 {
 			cohortEstablished = true
+			initialized = nil
```
rafael, line 172: good catch. Wondering if we should add a unit test for this to have coverage

go/services/multiorch/recovery/health_stream.go
```diff
@@ -0,0 +1,317 @@
...
+	defer func() {
+		entry.mu.Lock()
+		entry.stream = nil
+		entry.mu.Unlock()
+	}()
+
+	hs.markConnected(poolerID)
+
+	for {
+		resp, err := stream.Recv()
```
rafael, line 215: Should we have a staleness check here that aligns with: The orchestrator MUST implement a staleness timeout? 

Also, thinking if we should set the grpc keep alive timeouts aligning with this as well.

proto/multipoolermanagerdata.proto
```diff
@@ -291,6 +291,57 @@ message StatusResponse {
...
+
+  // Last consensus term known to the orchestrator. The pooler may use this to
+  // determine whether to include term-specific context in the initial snapshot.
+  int64 known_term = 2;
```
rafael, line 302: I did not see these two fields being used yet. Should skip them for now and them later?

go/services/multipooler/grpcmanagerservice/service.go
```diff
@@ -322,3 +325,134 @@ func (s *managerService) RewindToSource(ctx context.Context, req *multipoolerman
 func (s *managerService) SetPostgresRestartsEnabled(ctx context.Context, req *multipoolermanagerdatapb.SetPostgresRestartsEnabledRequest) (*multipoolermanagerdatapb.SetPostgresRestartsEnabledResponse, error) {
 	return s.manager.SetPostgresRestartsEnabled(ctx, req)
 }
+
+// managerHealthPollingInterval is how often ManagerHealthStream proactively
+// polls Status() even without a broadcast notification. This guarantees that
+// postgres state changes (e.g. process death) reach the orchestrator within
+// this interval even when the multipooler's local monitor is disabled.
+const managerHealthPollingInterval = 5 * time.Second
```
rafael, line 333: I wonder if this should configurable

rafael (commented):
Left a few comments. One extra thought: Do you think it would be worth to add some integration tests in multipooler?   One test that opens a stream, triggers a state-change RPC (e.g. SetPrimaryConnInfo), and asserts a snapshot arrives reflecting the new state would exercise broadcast + send + proto wiring end-to-end in a single case.

go/services/multipooler/grpcmanagerservice/service.go
mkindahl: Agree. I made it a const rather than hard-coding I think that would make it easier to turn into an option. I'm actually considering if the orchestrator should send this to the pooler on connection so that it is entirely controlled from that side.

proto/multipoolermanagerdata.proto
mkindahl: Removed.

go/services/multiorch/recovery/actions/shard_init_action.go
mkindahl: I think we have a test for that: `TestShardInitAction_GetInitializedPoolers_CohortAlreadyEstablished`. The only problem is that it does not trigger all the time, just spuriously.

go/services/multiorch/recovery/health_stream.go
mkindahl: > Should we have a staleness check here that aligns with: The orchestrator MUST implement a staleness timeout?

Added a staleness timeout. It will capture the case when the pooler is stopped, but not the node (so the TCP connection is still up). It is set quite high, but I think we need to re-visit these settings since they are "all over the place."

> Also, thinking if we should set the grpc keep alive timeouts aligning with this as well.

I think we have one already?

The gRPC timeout is set in `newDial` in `go/common/rpcclient/conn_cache.go`. It is a lot lower than the staleness timeout so it should trigger as soon as the connection is lost. The staleness timeout above will only fire if the node is up but the pooler is not active for some reason.

mkindahl:
> Left a few comments. One extra thought: Do you think it would be worth to add some integration tests in multipooler? One test that opens a stream, triggers a state-change RPC (e.g. SetPrimaryConnInfo), and asserts a snapshot arrives reflecting the new state would exercise broadcast + send + proto wiring end-to-end in a single case.

I added an integration test for this. To make the tests more stable I have a "trigger" field in each snapshot that tells what triggered the sending of the snapshot. This allow us to distinguish heartbeat events from other kinds of events, e.g. due to a state change or a poll.

go/services/multipooler/grpcmanagerservice/service.go
mkindahl: Added #864 as a follow-up to this one.

rafael (approved):
LGTM

proto/multipoolermanagerdata.proto
mkindahl: We can actually skip this for now since it is not used. Otherwise, I agree.

proto/multiorchdata.proto
mkindahl: No, this is just for statistics. We might be better off not having it at all, so removing it. We can add it when we need it.

proto/multipoolermanagerdata.proto
mkindahl: Good point. Will change.

go/services/multiorch/recovery/health_stream.go
mkindahl: Agree, but this is a larger refactor, so I add it as a separate commit to be able to move forward with the PRs that are after this PR. To fix the current issue, I add `CohortMembers` here.

mkindahl: See #867

PR #858, opened 2026-04-17
fix(multipooler): initialize topoPublisher in newRemedialActionTestManager

dweitzman:
Notify() now dereferences the receiver to acquire its mutex before checking the action lock, so a nil topoPublisher panics. The test helper was constructing MultiPoolerManager without this field.

Looks like an uncaught merge conflict from https://github.com/multigres/multigres/pull/842


rafael (approved):
LGTM

PR #859, opened 2026-04-17
fix(consensus): fixes for consensus recruitment

rafael:
# Desc
The previous `ValidateQuorum` only checked candidacy (whether the recruited set has enough
members to form a new quorum). It did not check revocation, whether the un-recruited cohort
members could still form a parallel quorum on their own and commit writes under a different quorum. 

This PR closes that gap and refactors the durability-policy code to make the guarantees explicit.

TODO: We still need to think about how are we going to handle followers that are not part of the cohort.  How do we handle the case where a follower that is nor part of the cohort happened to have the most advanced timeline? 

## Changes

- **Revocation check added.** `CheckSufficientRecruitment` now requires:
  - **Candidacy:** recruited poolers can form a fresh quorum under the policy (forward progress).
  - **Revocation:** no subset of un-recruited cohort poolers can itself satisfy the policy (no
    parallel quorum can commit outside our recruitment).
- **`MULTI_CELL_AT_LEAST_N` revocation tightened to pooler-level.** The old cell-based check
  ("every cohort cell has ≥1 recruited pooler") is too lenient when cells hold multiple poolers:
  un-recruited poolers can still span enough cells to form a valid quorum. New rule: un-recruited
  cohort poolers must span fewer than N distinct cells. Added unit tests to cover this case.
- **Typed `DurabilityPolicy` interface introduced** under `go/common/consensus/`:
  - `CheckAchievable(proposedCohort) error` — pre-flight feasibility.
  - `CheckSufficientRecruitment(cohort, recruited) error` — candidacy + revocation together.
  - `Description() string`
  - Implementations: `AtLeastNPolicy`, `MultiCellPolicy`. `PolicyFromProto` is the factory.
- **Defensive subset check on every `CheckSufficientRecruitment` call** — `recruited` must ⊆
  `cohort`, so candidacy counts reflect only policy-eligible poolers.
- **Callers migrated:** `BeginTerm`, `preVote`, `ShardInitAction`,
  `ShardNeedsInitializationAnalyzer`. The old `Coordinator.ValidateQuorum` wrapper is deleted;
  callers invoke `policy.CheckSufficientRecruitment` directly.
- **`IsDurabilityPolicyAchievable` free function removed** — callers use `policy.CheckAchievable`
  instead, which actually inspects cell distribution for `MULTI_CELL`.

## Test plan

-  Existing tests still pass, added more coverage to cover in more depth CheckRevocation


go/common/consensus/durability.go
```diff
@@ -21,18 +21,6 @@ import (
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 )
 
-// IsDurabilityPolicyAchievable returns true if the given number of poolers is sufficient
```
rafael, line 24: @mkindahl - I think this should address one of the comments you mentioned in the bootstrap PR. Now this is not a standalone function

go/common/consensus/policy_at_least_n.go
```diff
@@ -0,0 +1,82 @@
...
+//     fresh quorum.
+//   - Revocation: fewer than N cohort poolers are absent from recruited, so no
+//     N-subset of the cohort avoids our recruitment,  no parallel quorum can
+//     still form outside our recruited set.
+func (p AtLeastNPolicy) CheckSufficientRecruitment(cohort, recruited []*clustermetadatapb.ID) error {
+	if err := validateRecruitedSubset(cohort, recruited); err != nil {
+		return err
+	}
+
+	if len(recruited) < p.N {
```
dweitzman, line 51: This is a nuanced point, but I don't think this is the right function to validate candidacy. It's embedding an assumption: that recruitment is only happening for the purpose of doing a failover while keeping the durability rules and cohort the same

I think the right place to validate candidacy is in the leader appointment code, since that's the code that knows what proposal is intended after recruitment

I view the purpose of this method as ensuring two things:
- freezing logical time so that a new proposal can be made. As for what the proposal is (failover, a cohort change etc.), I view that as a separate problem
- ensuring the uniqueness of the coordinator term number

go/common/consensus/policy_at_least_n.go
```diff
@@ -0,0 +1,82 @@
...
+			missing++
+		}
+	}
+	// If we have N or more nodes missing in the recruitment, they could form a quorum on their own.
+	// We can't allow that.
+	// Example: AT_LEAST_N with N=2 and a cohort of 5 poolers. After recruiting
+	// pooler-1, pooler-2, and pooler-3, we still need to recruit pooler-4 OR
+	// pooler-5 — otherwise the 2 unrecruited poolers could form their own
+	// 2-pooler quorum that still satisfies the durability policy.
+	if missing >= p.N {
```
dweitzman, line 69: This is necessary but insufficient if N is more than half the cohort size

If the cohort size is 10 and the policy is at least 8:
- we need at least 3 nodes recruited to revoke (implemented by this logic)
- but we also need at least 6 nodes recruited to ensure uniqueness for the coordinator term

go/common/consensus/policy_at_least_n.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// CheckSufficientRecruitment enforces both candidacy and revocation:
+//   - Candidacy: recruited has at least N poolers, so the new leader can form a
+//     fresh quorum.
+//   - Revocation: fewer than N cohort poolers are absent from recruited, so no
+//     N-subset of the cohort avoids our recruitment,  no parallel quorum can
+//     still form outside our recruited set.
+func (p AtLeastNPolicy) CheckSufficientRecruitment(cohort, recruited []*clustermetadatapb.ID) error {
+	if err := validateRecruitedSubset(cohort, recruited); err != nil {
```
rafael, line 47: This guardrail seems sensible. @dweitzman , curious of your thoughts. 

We shouldn't be calling CheckSufficientRecruitment, with recruited nodes that are not part of the cohort.

go/common/consensus/durability.go
```diff
@@ -46,3 +35,74 @@ func ParseUserSpecifiedDurabilityPolicy(name string) (*clustermetadatapb.Durabil
...
+//
+// It exposes two checks a coordinator needs to safely appoint a new leader
+// under the generalized-consensus model:
+//
+//  1. Achievability: a pre-flight feasibility gate. Checks if the proposed cohort could
+//     satisfy this policy's durability conditions.
+//  2. Sufficient recruitment: the recruited subset of a committed cohort can
+//     form a fresh quorum (candidacy) AND intersects every other quorum the
+//     cohort could form (revocation).
+type DurabilityPolicy interface {
```
dweitzman, line 49: I suspect in the long term we might want something more like `DurabilityPolicyPlugin` where there's only one instance and we can pass the policy parameters as the first argument to the methods, but I don't have any objections to doing this.

go/common/consensus/policy_multi_cell.go
```diff
@@ -0,0 +1,98 @@
...
+	recruitedKeys := poolerKeysOf(recruited)
+	unrecruitedCells := make(map[string]struct{})
+	for _, pooler := range cohort {
+		if _, ok := recruitedKeys[topoclient.ClusterIDString(pooler)]; ok {
+			continue
+		}
+		unrecruitedCells[pooler.GetCell()] = struct{}{}
+	}
+	if len(unrecruitedCells) >= p.N {
+		return fmt.Errorf("revocation not satisfied: un-recruited cohort poolers span %d cells, another possible quorum could be formed spanning %d cells (%s)",
```
dweitzman, line 77: Not a blocking comment, but hypothetically we could share code between at least for poolers and at least for cells by sharing a common helper with maybe generic typing that where the multi cell policy passes in cells and the at least policy passes in poolers

The generic helper would only need to have items that can be serialized to a unique string and it could return an typed error with a slice containing the set of of things that could form a rogue cohort, poolers for "at least" and cells for "at least cells"

The possible benefit would be that the math for this method is finicky, so having only one copy of it might be nice. I don't have strong feelings, though

dweitzman (commented):
Seems reasonable overall. Still need to read the tests

go/common/consensus/policy_at_least_n.go
rafael: Yeah, makes sense. In my mental model, the current code is correct, candidacy is checked just above this block, and together the two checks guarantee intersection, so concurrent recruitments can't both succeed without sharing a node. I added a test to cover this.

If we move the candidacy check out (per your other comment), we'd need to add an explicit majority check here.

go/common/consensus/policy_at_least_n.go
dweitzman: Agreed

go/common/consensus/durability.go
```diff
@@ -46,3 +35,118 @@ func ParseUserSpecifiedDurabilityPolicy(name string) (*clustermetadatapb.Durabil
...
+	//     commit outside our recruitment.
+	CheckSufficientRecruitment(cohort, recruited []*clustermetadatapb.ID) error
+
+	// Description returns a human-readable summary of the policy.
+	Description() string
+}
+
+// PolicyFromProto converts a proto DurabilityPolicy into a concrete
+// DurabilityPolicy implementation.
+func PolicyFromProto(policy *clustermetadatapb.DurabilityPolicy) (DurabilityPolicy, error) {
```
mkindahl: Nit: the general convention seems to be to call it "NewObjectFromWhatever".

go/common/consensus/durability.go
```diff
@@ -46,3 +35,118 @@ func ParseUserSpecifiedDurabilityPolicy(name string) (*clustermetadatapb.Durabil
...
+	}
+	return out
+}
+
+// validateRecruitedSubset returns an error if any recruited pooler is not a
+// member of the cohort. All durability policies assume recruited ⊆ cohort so
+// that candidacy counts reflect only policy-eligible poolers. This is a
+// defensive invariant check; call sites should already enforce it upstream.
+func validateRecruitedSubset(cohort, recruited []*clustermetadatapb.ID) error {
+	cohortKeys := poolerKeysOf(cohort)
```
mkindahl, line 113: Nit: alternative is `slices.Collect` instead of a new function.

go/common/consensus/policy_at_least_n.go
```diff
@@ -0,0 +1,90 @@
...
+		return err
+	}
+
+	// Revocation: if N or more cohort poolers are un-recruited, they could form a
+	// quorum on their own. Example: AT_LEAST_N with N=2 and a cohort of 5 poolers.
+	// After recruiting pooler-1, pooler-2, and pooler-3, we still need to recruit
+	// pooler-4 OR pooler-5 — otherwise the 2 unrecruited poolers could form their
+	// own 2-pooler quorum that still satisfies the durability policy.
+	unrecruited := unrecruitedKeys(cohort, recruited, topoclient.ClusterIDString)
+	if len(unrecruited) >= p.N {
```
mkindahl: Isn't it easier to just check `len(cohort) - len(recruited) >= p.N`? You're not using the `unrecruited` array for anything (except checking length).

go/common/consensus/policy_multi_cell.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+
+	// Revocation: if the un-recruited cohort poolers themselves span N or more cells, they could
+	// form a commit quorum on their own under this policy. We can't allow that.
+	// Example: MULTI_CELL_AT_LEAST_2 and a cohort of 6 poolers (2 per cell across 3 cells).
+	// Recruiting one pooler from each cell covers every cohort cell, but the 3 un-recruited
+	// poolers still span 3 cells — enough to form a separate 2-cell quorum on their own.
+	unrecruitedCells := unrecruitedKeys(cohort, recruited, func(id *clustermetadatapb.ID) string { return id.GetCell() })
+	if len(unrecruitedCells) >= p.N {
```
mkindahl: Isn't it sufficient to just compare the lengths? There are no additional checks inside `unrecruitedKeys` that would filter out any members from cohort or recruited, and the `unrecruitedKeys` are not used except to count the number of unrecruited members.
```suggestion
	if len(cohort) - len(recruited) >= p.N {
```

go/common/consensus/policy_multi_cell.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+)
+
+// MultiCellPolicy requires acknowledgement from poolers spanning at least N
+// distinct cells.
+type MultiCellPolicy struct {
+	N    int
+	Desc string
```
mkindahl: Description is only used in tests. Maybe stick to the generated names alone and add this later if we need it?

go/common/consensus/policy_at_least_n_test.go
mkindahl, line 1: Missing `Any(0)` tests.

mkindahl (approved):
Looks good. Some minor comments for your consideration.

go/common/consensus/policy_at_least_n_test.go
```diff
@@ -0,0 +1,274 @@
...
+		name       string
+		n          int
+		cohort     []*clustermetadatapb.ID
+		recruited  []*clustermetadatapb.ID
+		wantErrMsg string // empty means expect nil error
+	}{
+		{
+			name: "AT_LEAST_2 with all 3 cohort poolers recruited is sufficient",
+			n:    2,
+			cohort: []*clustermetadatapb.ID{
```
dweitzman, line 108: Should some test cases use different cells to show that the cell doesn't matter?

dweitzman (approved):

go/common/consensus/policy_at_least_n.go
rafael: This makes sense. Updated

go/common/consensus/policy_multi_cell.go
rafael: For MultiCell we count cells spanned by un-recruited poolers, not poolers. unrecruitedKeyCount will perform that grouping. Doing len(cohort) - len(recruited) would give a pooler-count, which is a  different threshold and would break tightened cell-level revocation (_I think_)

Does this makes sense to you as well?

mkindahl: Yes, that makes sense.

PR #860, opened 2026-04-17
feat(pgregresstest): patch-based verification replaces strict text diff

haritabh17:
## Summary

- Replaces `pg_regress`'s *any-diff-is-a-failure* verdict with a patch-aware pipeline. Per-test patches under `go/test/endtoend/pgregresstest/testdata/pg17/patches/<name>.patch` document accepted deviations between multigres and upstream PostgreSQL 17 — each patch is a reviewable unified diff.
- Two modes: `verify` (default, CI) strict-diffs the patched expected against actual; `generate` (developer) writes/updates patches so the same diff would produce `pass` next time.
- Ships with 10 seeded patches (boolean, enum, float4/8, int2/4, money, oid, txid, uuid) covering most of yesterday's MUL-358 regressions.
- Fixes MUL-358.

## Problem

The plan-cache feature (#814) started deparsing queries before sending them to the backend, which caused 50 pgregress tests to fail overnight. Every failure was a cosmetic mismatch between multigres's output and upstream's — error-message wording, caret positions, etc. — not a real regression.

Trying to fix the deparser to produce byte-for-byte-identical output with PostgreSQL is whack-a-mole: any textual deviation (typed literal → CAST, casing, whitespace) produces a `pg_regress` failure, and each kind needs its own code fix in the planner or parser.

## Solution

Move the policy decision out of code and into checked-in `.patch` files. Every accepted deviation is a documented unified diff that a reviewer can read and approve. Server code stays focused on correctness, not on matching upstream's text output.

### Pipeline per test

```
upstream/expected/<name>.out  ──► apply patches/<name>.patch ──► multigres-expected
                                                                          │
actual results/<name>.out  ───────────────────────────────────────────►   │
                                                                          ▼
                                                                diff -U3 ──► empty → PASS
                                                                             non-empty → FAIL (+ diff hunks)
```

Expected output is read from the PostgreSQL source tree (`prep_buildtree` only symlinks Makefiles, not `.out` files). Actual output is read from the build tree's `results/` directory populated by `pg_regress`.

### Modes

| Mode | Selected by | Behavior |
|---|---|---|
| verify | `PGREGRESS_PATCH_MODE=verify` (default) | Strict diff of patched expected vs actual. Missing patch + diff = fail. Stale patch (upstream moved) = fail with clear reason. |
| generate | `PGREGRESS_PATCH_MODE=generate`, or `make pgregress-update-patches` | Residual diffs rewrite `patches/<name>.patch`. Redundant patches (upstream already matches) are removed. |

### Results format

Each test's entry in `results.json`:

```json
{
  "name": "boolean",
  "status": "pass",
  "duration": "61ms",
  "patch_applied": true,
  "patch_path": "go/test/endtoend/pgregresstest/testdata/pg17/patches/boolean.patch"
}
```

Markdown compatibility report adds a **Patch** column with clickable links.

### Code layout

```
go/test/endtoend/pgregresstest/
  patch_verify.go           # VerifyTest, applyPatch, generateDiff
  patch_verify_test.go      # 6 unit tests, no cluster required
  postgres_builder.go       # VerifyWithPatches + pg_regress --use-existing
  pgregress_test.go         # calls VerifyWithPatches after RunRegressionTests
  testdata/pg17/
    README.md               # conventions + workflow
    patches/
      <name>.patch          # 10 seeded patches
```

### Makefile

- `make pgregress` — verify mode.
- `make pgregress-update-patches` — generate mode.

## Extras

- **`pg_regress --use-existing --dbname=postgres`**: `DROP DATABASE` + `CREATE DATABASE` are blocked by the unsafe-statement rejection landed in #849, so `pg_regress` can no longer manage its own `regression` database through multigateway. This switch runs all tests against the existing `postgres` database.
- **`buildTimeout 10m → 20m`**: Cold-cache first runs on slower developer machines (macOS) were hitting the 10-minute budget during `configure` + `make` + `install`. 20m gives headroom without affecting CI.

## Verification

- `go test -run 'TestVerify_|TestGenerate_' ./go/test/endtoend/pgregresstest/...` → 6/6 pass. Covers verify pass/fail, patch absorbs diff, stale patch detection, generate writes patch, generate removes redundant patch.
- Local end-to-end run on a 10-test subset (generate mode) produced the 10 seeded patches; a follow-up run in verify mode returns all 10 to pass.
- `golangci-lint run ./go/test/endtoend/pgregresstest/...` → 0 issues.

## Follow-up (not in this PR)

- Run generate mode over the remaining ~210 tests and review the resulting patches.
- Clean up state leakage between tests running against the persistent `postgres` database (add `DROP SCHEMA public CASCADE; CREATE SCHEMA public;` pre-step if cross-test state becomes a problem).
- Scaffold `testdata/pg18/` when PG is upgraded.

## Test plan

- [ ] CI nightly turns green on the next run after merge.
- [ ] `make pgregress-update-patches` locally to extend patch coverage to the full suite, review each generated patch in a follow-up PR.
- [ ] `make pgregress` returns all-pass with patch column visible in report.

### DIFF
--- a/go/test/endtoend/pgregresstest/patch_verify.go
+++ b/go/test/endtoend/pgregresstest/patch_verify.go
@@ -1,0 +1,306 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pgregresstest
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/multigres/multigres/go/tools/executil"
+)
+
+// PatchMode selects the behavior of VerifyTest:
+//   - PatchModeVerify: strict diff check; existing patches must produce a clean
+//     match (no residual diff), missing patches cause fail when a diff exists.
+//   - PatchModeGenerate: any residual diff is absorbed by rewriting the
+//     per-test patch file, so the test ends up passing. Stale patches (tests
+//     that now match upstream exactly) have their patch files deleted.
+type PatchMode string
+
+const (
+	PatchModeVerify   PatchMode = "verify"
+	PatchModeGenerate PatchMode = "generate"
+)
+
+// PatchModeEnv is the environment variable that selects the mode. When unset
+// or empty, verify mode is used.
+const PatchModeEnv = "PGREGRESS_PATCH_MODE"
+
+// GetPatchMode reads the mode from the environment. Defaults to verify.
+func GetPatchMode() PatchMode {
+	v := os.Getenv(PatchModeEnv)
+	switch PatchMode(v) {
+	case PatchModeGenerate:
+		return PatchModeGenerate
+	case "", PatchModeVerify:
+		return PatchModeVerify
+	default:
+		// Unknown value: treat as verify to stay strict by default.
+		return PatchModeVerify
+	}
+}
+
+// VerifyOutcome is the result of applying patch-based verification to one test.
+type VerifyOutcome struct {
+	Name         string `json:"name"`
+	Status       string `json:"status"` // "pass" | "fail"
+	PatchApplied bool   `json:"patch_applied"`
+	// PatchPath is the path to the patch file relative to the repo root, for
+	// links in status reports. Empty if no patch was used.
+	PatchPath string `json:"patch_path,omitempty"`
+	// Reason is a short human-readable description when Status == "fail".
+	Reason string `json:"reason,omitempty"`
+	// Diff is the residual unified diff on failure. Populated only on fail.
+	Diff string `json:"diff,omitempty"`
+}
+
+// VerifyInput describes one test's inputs to VerifyTest. All paths are absolute.
+type VerifyInput struct {
+	// Name is the bare test name (e.g. "boolean").
+	Name string
+	// ExpectedPath points at postgres's checked-in expected output
+	// (e.g. /tmp/multigres_pg_cache/source/postgres/src/test/regress/expected/boolean.out).
+	ExpectedPath string
+	// ActualPath points at the output produced by this test run
+	// (e.g. <buildDir>/src/test/regress/results/boolean.out).
+	ActualPath string
+	// PatchDir is the directory containing per-test patch files
+	// (e.g. <repo>/go/test/endtoend/pgregresstest/testdata/pg17/patches).
+	PatchDir string
+	// RepoRoot is used only to compute a nice relative PatchPath for reports.
+	RepoRoot string
+}
+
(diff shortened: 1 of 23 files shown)


go/services/multigateway/executor/executor.go
```diff
@@ -107,7 +107,16 @@ func (e *Executor) StreamExecute(
...
+	// When the cached plan is just a Route, send the client's original query
+	// bytes directly so the backend computes error positions against the same
+	// string psql renders the caret against. Going through Route.StreamExecute
+	// would deparse the normalized AST and subtly change whitespace/casing,
+	// shifting the caret off by one or more columns.
+	if route, ok := plan.Primitive.(*engine.Route); ok {
+		err = e.exec.StreamExecute(ctx, conn, route.TableGroup, route.Shard, queryStr, state, callback)
+	} else {
+		err = plan.StreamExecute(ctx, e.exec, conn, state, bindVars, callback)
+	}
```
GuptaManan100, line 119: Hmmmm, not sure about this one. What kind of issues does it create?

haritabh17: The root problem with going through `Route.StreamExecute` is that `ast.ReconstructSQL` produces bytes that differ from what the client sent, so the backend's error `Position` no longer lines up with what
 psql draws the ^ caret against. That shifts carets and breaks bunch of pgregress tests.

e.g. Typed literals become CAST: `bool 'XXX'` deparses to `CAST('XXX' AS boolean)`. The "invalid input syntax" positionis computed against the CAST form; psql renders the caret against `bool 'XXX'`. Same for `'NAN'::float4`, `'badguid'::uuid`, malformed range literals, etc.

go/test/endtoend/pgregresstest/patch_verify.go
```diff
@@ -0,0 +1,298 @@
...
+	aPath := filepath.Join(tmpDir, "a")
+	bPath := filepath.Join(tmpDir, "b")
+	if err := os.WriteFile(aPath, a, 0o644); err != nil {
+		return nil, err
+	}
+	if err := os.WriteFile(bPath, b, 0o644); err != nil {
+		return nil, err
+	}
+
+	cmd := executil.Command(ctx, "diff", "-U3", aPath, bPath)
```
mkindahl: Most of the changes you have a whitespace changes, so to avoid always having to update the patches on whitespace changes you might want to use `-b` (`--ignore-space-change`).

There is an option `-w` as well, but that might remove whitespace inside strings, so not sure about it. Might be fine if since changes we'll see are probably not going to be relevant even if inside a string.

mkindahl (approved):
Looks good to me, but having one suggestion to avoid having to update the diffs more often than necessary.

go/test/endtoend/pgregresstest/testdata/pg17/patches/boolean.patch
```diff
@@ -0,0 +1,11 @@
+--- /var/folders/pn/2dkqhb2x237_jnnqxfj12lgr0000gn/T/pgregress-diff-2614898860/a	2026-04-22 01:41:24
```
GuptaManan100: Patch files contain absolute temp-directory paths in headers. DOn't know if we can avoid them, but if we can maybe we should, because they'll create unnecessary diffs anytime we generate new patches.

GuptaManan100 (approved):
Rest seems good to me!

go/test/endtoend/pgregresstest/patch_verify.go
haritabh17: Good point, updated.

go/test/endtoend/pgregresstest/testdata/pg17/patches/boolean.patch
haritabh17: Ack, fixed

PR #864, opened 2026-04-20
feat(multiorch): make ManagerHealthStream timing configurable by orchestrator

mkindahl:
The orchestrator now negotiates snapshot_interval and staleness_timeout with the pooler on stream open. The pooler echoes back the actual values it will use in a ManagerHealthStreamStartResponse.

- snapshot_interval: how often the server proactively pushes snapshots when no state-change broadcast has occurred (default 5s)
- staleness_timeout: echoed in every snapshot so the client can arm a watchdog to detect a stuck stream (default 90s)

Fixes a nil pointer panic when the orchestrator omits timing fields from the start message (direct .Seconds field access on nil *durationpb.Duration).  Switches to AsDuration() which is nil-safe.

Consolidates staleness timeout, heartbeat interval, and snapshot interval constants into go/common/timeouts as DefaultHealthStreamStalenessTimeout, DefaultHealthHeartbeatInterval, and DefaultSnapshotInterval.

Closes MUL-360

### DIFF
--- a/go/services/multiorch/recovery/health_stream.go
+++ b/go/services/multiorch/recovery/health_stream.go
@@ -16,15 +16,18 @@ package recovery
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
 	"time"
 
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/multigres/multigres/go/common/rpcclient"
+	"github.com/multigres/multigres/go/common/timeouts"
 	"github.com/multigres/multigres/go/common/topoclient"
 	multiorchdatapb "github.com/multigres/multigres/go/pb/multiorchdata"
 	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
@@ -39,19 +42,6 @@ const (
 
 	// streamReconnectMaxBackoff caps the exponential backoff between retries.
 	streamReconnectMaxBackoff = 30 * time.Second
-
-	// streamStalenessTimeout is the default maximum time to wait for a health
-	// snapshot before treating the stream as stale and reconnecting.
-	//
-	// This is an application-level watchdog that catches the "server goroutine
-	// stuck but TCP alive" failure mode that gRPC keepalive does not cover.
-	// gRPC keepalive (10s ping / 10s timeout) handles dead TCP connections;
-	// this constant handles a live connection whose send loop has deadlocked.
-	//
-	// The server guarantees a snapshot at least every managerHealthPollingInterval
-	// (5s), so messages normally arrive far more frequently than this. The value
-	// is seeded from the server's recommended timeout in each snapshot.
-	streamStalenessTimeout = 90 * time.Second
 )
 
 // streamEntry holds the lifecycle handles for one active stream goroutine.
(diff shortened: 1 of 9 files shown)


rafael:
I did a quick pass of the second commit (we should maybe look into stacked PR's ). It is makins sense. I will take an extra pass after we land the main HealthStream PR.

One extra thought that could be another PR after this one is wiring now this configurable parameter into multiorch viper variables. So it is something that operators can set and change during the lifecycle of a coordinator.

proto/multipoolermanagerdata.proto
```diff
@@ -306,22 +306,54 @@ message ManagerHealthStreamClientMessage {
...
+  // a snapshot when no state-change broadcast has occurred.
+  // Zero means use the server default (currently 5s).
+  int64 snapshot_interval_seconds = 1;
+
+  // staleness_timeout_seconds is the window after which the client will treat
+  // the stream as stale and reconnect. The server echoes the actual value it
+  // will use in ManagerHealthStreamStartResponse and in every subsequent
+  // ManagerHealthSnapshot.timeout_seconds.
+  // Zero means use the server default (currently 90s).
+  int64 staleness_timeout_seconds = 2;
```
dweitzman, line 321: Thoughts on google.protobuf.Duration vs having `_seconds` in the name for all of these?

mkindahl: Right, intended to fix those as well given your last comment. Thanks for reminding me.

mkindahl: Fixed.

go/services/multiorch/recovery/health_stream.go
```diff
@@ -40,18 +42,18 @@ const (
...
 	// gRPC keepalive (10s ping / 10s timeout) handles dead TCP connections;
 	// this constant handles a live connection whose send loop has deadlocked.
 	//
-	// The server guarantees a snapshot at least every managerHealthPollingInterval
-	// (5s), so messages normally arrive far more frequently than this. The value
-	// is seeded from the server's recommended timeout in each snapshot.
-	streamStalenessTimeout = 90 * time.Second
+	// The actual value in use is confirmed by the server in its start response
+	// and then echoed in every snapshot's timeout_seconds.
+	defaultStalenessTimeout = 90 * time.Second
```
rafael: We have `go/common/timeouts/rpc.go`, I think we can add this default there and it can be shared between this package and the server.

go/services/multipooler/manager/health_provider.go
```diff
@@ -35,15 +35,10 @@ const (
...
-	// orchestrator receives no message within this window it should treat the
-	// pooler as unreachable and reconnect.
-	DefaultRecommendedStalenessTimeout time.Duration = defaultRecommendedStalenessTimeout
-
-	// defaultHealthHeartbeatInterval is the interval between periodic health
-	// broadcasts when no state changes occur.
-	defaultHealthHeartbeatInterval = 30 * time.Second
+	// DefaultHealthHeartbeatInterval is the interval between periodic health
+	// broadcasts when no state changes occur. Exported so the gRPC handler can
+	// include it in ManagerHealthStreamStartResponse for the orchestrator to observe.
```
rafael: Comment seems stale in regards to the include.

Same idea that we can probably move this one to the common package.

go/services/multiorch/recovery/health_stream.go
```diff
@@ -252,9 +285,9 @@ func (hs *HealthStream) streamOnce(ctx context.Context, poolerID string, poolerH
 				}
 				timer.Reset(d)
 			case <-timer.C:
-				hs.logger.WarnContext(ctx, "health stream stale: no snapshot received within timeout, reconnecting",
+				hs.logger.WarnContext(ctx, "health stream stale: no message received within timeout, reconnecting",
 					"pooler_id", poolerID,
-					"timeout", effectiveStalenessTimeout,
+					"timeout", initialStaleness,
```
rafael: this is a nit: it seems that if the negotiated staleness changed, this will still log the initial one right?

rafael (approved):
LGTM. Two very minor comments

go/services/multiorch/recovery/health_stream.go
mkindahl: Right. Fixed.

go/services/multiorch/recovery/health_stream.go
mkindahl: Moved. Question is if we shall move the other timeouts as well? They seem to be common as well.

go/services/multipooler/manager/health_provider.go
mkindahl: Fixed the comment, but I wonder if we should move this to the common place as well. It's not really an RPC timeout, but it might be good to have a single place for all components to read defaults from.

dweitzman (approved):

go/services/multipooler/manager/health_provider.go
mkindahl: Moved it.

PR #869, opened 2026-04-21
feat(multipooler): Crash protection during initial backup 

rafael:
## Summary                                                                                                                                                                                              
                                                                                                                                                                                                             
   - Write a `.multigres-bootstrap-in-progress` sentinel in `pooler_dir` before initdb and clear it after the final data-directory cleanup                                                                   
   - If the sentinel is present on startup, treat any `pg_data` as stale from a prior crashed attempt and remove it before retrying
   - Sentinel lives in `pooler_dir` (not `PGDATA`) so pgBackRest backups never capture it
   - Adds `shardsetup.WithDeferredMultipoolerStart` so integration tests can stage on-disk state before the multipooler starts

   ## Problem

   Follow-up to a review comment on #820 (from @mkindahl): if the first-backup flow crashes between `initdb` and the final cleanup, the pooler is left with a populated `pg_data` directory and
   `createFirstBackupAndInitializeLocked` refuses to proceed on restart — requiring manual intervention.

   ## Solution

   Use a sentinel file as an on-disk "bootstrap in progress" signal:
   - Written right before `InitDataDir`; removed after the happy-path `removeDataDirectory()` succeeds.
   - Removal is ordered (data dir first, sentinel second) so the invariant **data dir present ⇒ sentinel present** holds, and any retry that sees the sentinel can clean up idempotently.
   - Kept outside `PGDATA` so pgBackRest (which only archives `pg_data`) never picks it up — sidestepping the concern about the marker propagating into future restores.

## Tests
* Added relevant unit tests.
* The integration test `TestBootstrapSentinelCrashRecovery` stages `PG_VERSION` + sentinel in `pooler_dir` before the multipooler starts, then asserts via `GetBackups` that a completed backup appears and the sentinel is cleaned up.
* I also tested that without this changes the integration test fails, validating that it would have gotten stuck.


go/services/multipooler/manager/rpc_first_backup.go
```diff
@@ -46,6 +49,17 @@ func (pm *MultiPoolerManager) createFirstBackupAndInitializeLocked(ctx context.C
...
 	}
 
+	// A sentinel from a prior attempt means we crashed between initdb and the
+	// final data-directory cleanup. The data directory (if any) is stale and
+	// safe to remove so we can retry from scratch. The sentinel is recreated
+	// below before initdb, so we do not remove it here.
+	if pm.hasBootstrapSentinel() {
+		pm.logger.WarnContext(ctx, "Bootstrap sentinel from prior attempt detected; removing stale data directory before retry")
+		if err := pm.removeDataDirectory(); err != nil {
+			return false, false, mterrors.Wrap(err, "failed to remove stale data directory from prior bootstrap attempt")
```
mkindahl, line 66: An error can be raised if the data directory was removed below but it crashed before removing the sentinel file. This is a valid case and I do not think we need to print an error in this case.

However, we _should_ print an error if we cannot remove the data directory for other reasons, e.g., permissions, so the error matters only if the data directory is still there after the remove call.

go/services/multipooler/manager/rpc_first_backup.go
```diff
@@ -142,11 +168,41 @@ func (pm *MultiPoolerManager) createFirstBackupAndInitializeLocked(ctx context.C
...
 }
 
+func (pm *MultiPoolerManager) bootstrapSentinelPath() string {
+	return filepath.Join(pm.multipooler.PoolerDir, constants.BootstrapSentinelFile)
+}
+
+func (pm *MultiPoolerManager) hasBootstrapSentinel() bool {
+	_, err := os.Stat(pm.bootstrapSentinelPath())
+	return err == nil
+}
```
mkindahl: There is a difference between that the file does not exist and if there are permission issues, for example. If the file does not exist but there are other issues, this function will incorrectly state that the file exists. I am not sure about the consequences, but it might be better to return a bool plus an error and deal with the situation in the caller.

go/services/multipooler/manager/rpc_first_backup.go
```diff
@@ -142,11 +168,41 @@ func (pm *MultiPoolerManager) createFirstBackupAndInitializeLocked(ctx context.C
...
+
+func (pm *MultiPoolerManager) writeBootstrapSentinel() error {
+	path := pm.bootstrapSentinelPath()
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		return fmt.Errorf("failed to ensure pooler directory exists: %w", err)
+	}
```
mkindahl: The pooler dir should exist at this point, so it might be better to raise an error here rather than creating the directory. This would allow us to capture the actual issue.

go/services/multipooler/manager/rpc_first_backup.go
mkindahl, line 1: There seems to be duplication between the deferred cleanup in `createFirstBackupAndInitializeLocked` and the explicit cleanup last in the function (it is guarded with the `skipCleanup`).

I am not sure what the reason for the duplication is (the code looks the same), but it might be better to let the deferred function deal with the cleanup and not have cleanup code at the end of `createFirstBackupAndInitializeLocked`.

go/test/endtoend/multipooler/bootstrap_sentinel_recovery_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/multigres/multigres/go/test/utils"
+
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+)
+
+// TestBootstrapSentinelCrashRecovery verifies that a multipooler starting up
+// with a bootstrap sentinel + empty pg_data directory — the on-disk state a
+// prior crashed first-backup attempt would leave behind — recovers correctly:
+// it removes the stale directory, runs the first-backup flow, and ends up with
+// a completed backup.
```
mkindahl: There is one crash point after (or before) the data directory is created and the sentinel exists. Probably good to test this case as well.

mkindahl (commented):
Looks good. I have some suggestions for making it more robust to problems for your consideration. I cannot see a problem with keeping the marker file outside the data directory, so that should be fine.

go/services/multipooler/manager/rpc_first_backup.go
rafael: Good flag. I consolidated

mkindahl (approved):
LGTM

PR #871, opened 2026-04-21
test(multiorch): verify failover works with etcd down

dweitzman:
Stop etcd before running the dead primary recovery test iterations to verify that consensus (gRPC between multipoolers) and multigateway routing (PrimaryObservation streams) both work independently of etcd.

Adds StopEtcd to ShardSetup and WriteNow to WriterValidator. Uses WriteNow to assert per-iteration that multigateway routes writes to the new primary.

Also fixes appoint_leader.go to use the self-reported PoolerType from gRPC health checks rather than MultiPooler.Type from etcd, which goes stale when etcd is down and caused elections to be incorrectly skipped.


rafael (approved):
LGTM

PR #872, opened 2026-04-21
test(multiorch): fix flaky TestHealthStream_ConcurrentWatcherUpdate

dweitzman:
The test fired a goroutine with time.Sleep(5ms) to promote a pooler to PRIMARY, then checked the result after waiting for a snapshot to be applied. Under the race detector the snapshot was applied in under 5ms, so the goroutine fired after the final Get() and the assertion saw the original REPLICA type.

Fix by adding a WaitGroup so wg.Wait() runs before the final Get(), guaranteeing the promotion is applied regardless of ordering.

Also add TestProtoStore_DoUpdate_Concurrent to store_test.go for testing lost-update protection.

### DIFF
--- a/go/services/multiorch/recovery/health_stream_test.go
+++ b/go/services/multiorch/recovery/health_stream_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"log/slog"
 	"os"
+	"sync"
 	"testing"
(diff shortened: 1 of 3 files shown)


rafael (approved):
LGTM

PR #874, opened 2026-04-21
feat(multigateway): add per-query metrics tracking

haritabh17:
## Summary

- Normalize each query via the existing AST normalizer, hash the normalized SQL with xxh3, and attach the resulting 16-char `query.fingerprint` label to existing `mg.gateway.query.*` OTel metrics — so per-shape trends are queryable in Prometheus without putting query text in a label.
- Track aggregate per-fingerprint stats (calls, errors, min/avg/max latency, rows) in-process using a new registry backed by the existing Theine W-TinyLFU cache (with the bloom doorkeeper enabled). Cardinality is bounded by the cache; fingerprints outside the tracked set fold into `__other__`.
- Expose `GET /debug/queries` (HTML + `?format=json`) on the multigateway admin port with the tracked fingerprints, sortable by calls / total time / avg time / errors / last seen, auto-refreshing every 10s.
- Add two Grafana panels in the **Multigateway – Query Metrics** row: _Top Queries by p99 Latency_ (timeseries) and _Top Queries_ (table with fingerprint, calls/s, p99/p95/p50 ms, error rate; fingerprint column links to `/debug/queries`).

Closes MUL-212.

### Why bounded-cardinality fingerprint labeling

Raw SQL text has effectively unbounded cardinality (one series per literal), so using it as a Prometheus label would blow up the scrape. The registry wraps `theine.Store` with doorkeeper admission (first-touch ignored, admission on repeat) so one-off queries never enter the tracked set. `Labelize(fp)` returns either the fingerprint or `__other__`; that's the value that lands on the metric label, keeping cardinality bounded by the configured cache size.

### Feature wiring

- `ast.NormalizeResult.Fingerprint()` — stable `xxh3` hash over the existing normalized SQL; `ast.FingerprintSQL(s)` for callers that already have a normalized string (used on the extended-protocol / portal path).
- Executor populates `ExecuteResult.NormalizedSQL` + `.Fingerprint` on both simple and portal paths.
- `handler.recordQueryCompletion` calls `registry.Record(...)` and passes `registry.Labelize(fp)` as the `query.fingerprint` label to `QueryDuration`, `QueryErrors`, and `RowsReturned`.
- Non-cacheable statements (BEGIN / COMMIT / SET / DDL) label as `__utility__` instead — their space is small and bounded.
- Registry is shared between primary and replica handlers so a shape hits the same bucket regardless of which listener served it.
- `Close()` wired into `Shutdown`.

### Config flags

| Flag | Default | Notes |
|---|---|---|
| `--query-metrics-memory` | `2 MiB` | Registry byte budget. `0` disables per-query tracking (all other metrics still work). |
| `--query-metrics-sql-max-bytes` | `4096` | Truncation cap for the representative normalized SQL stored per fingerprint. |

## Test plan

- [x] Unit tests for `NormalizeResult.Fingerprint` (determinism, shape-invariance, collision-by-shape).
- [x] `queryregistry` unit tests covering disabled / nil registry, stats aggregation, sort orders, SQL truncation, empty-fingerprint no-op, concurrent writes (race-detector clean), **cardinality bound under 5k-fingerprint churn**, and TinyLFU doorkeeper admission.
- [x] HTTP handler tests for `/debug/queries` — JSON body, HTML default, nil-registry path.
- [x] Local end-to-end validation: `pgbench -c 4 -j 2 -T 300` against a local cluster populated nine distinct fingerprints (positive- and negative-delta pgbench variants) totalling ~295k calls; both `/debug/queries` and Grafana panels showed live data.

## Example dashboard
Screenshot 2026-04-22 at 18 19 44

### DIFF
--- a/go/common/parser/ast/normalizer.go
+++ b/go/common/parser/ast/normalizer.go
@@ -14,6 +14,12 @@
 
 package ast
 
+import (
+	"fmt"
+
+	"github.com/cespare/xxhash/v2"
+)
+
 // NormalizeResult holds the output of AST normalization.
 type NormalizeResult struct {
 	// NormalizedSQL is the SQL string with literals replaced by $1, $2, ...
@@ -32,6 +38,22 @@ type NormalizeResult struct {
 // WasNormalized reports whether any literals were replaced during normalization.
 func (r *NormalizeResult) WasNormalized() bool {
 	return len(r.BindValues) > 0
+}
+
+// Fingerprint returns a stable 16-character hex hash of the normalized SQL.
+// Two queries with the same NormalizedSQL produce identical fingerprints,
+// which makes the fingerprint suitable as a per-query-shape identifier for
+// metrics aggregation and cache keys.
+func (r *NormalizeResult) Fingerprint() string {
+	return FingerprintSQL(r.NormalizedSQL)
+}
+
+// FingerprintSQL returns a stable 16-character hex hash of the given SQL
+// string. Useful for callers that already have a normalized or
+// placeholder-form SQL (e.g. extended-protocol queries whose Parse text
+// already contains $N placeholders) and don't need to run the normalizer.
+func FingerprintSQL(sql string) string {
+	return fmt.Sprintf("%016x", xxhash.Sum64String(sql))
 }
 
 // Normalize replaces literal A_Const values in the AST with ParamRef ($1, $2, ...)
--- a/go/common/web/templates/queries_debug.html
+++ b/go/common/web/templates/queries_debug.html
@@ -56,11 +56,13 @@
       <section>
         <h4>Sort</h4>
         <p class="sort-links">
-          <a href="?sort=calls&limit={{.Limit}}"{{if eq .SortBy "calls"}} class="active"{{end}}>calls</a>
-          <a href="?sort=total_time&limit={{.Limit}}"{{if eq .SortBy "total_time"}} class="active"{{end}}>total time</a>
-          <a href="?sort=avg_time&limit={{.Limit}}"{{if eq .SortBy "avg_time"}} class="active"{{end}}>avg time</a>
-          <a href="?sort=errors&limit={{.Limit}}"{{if eq .SortBy "errors"}} class="active"{{end}}>errors</a>
-          <a href="?sort=last_seen&limit={{.Limit}}"{{if eq .SortBy "last_seen"}} class="active"{{end}}>last seen</a>
+          {{range .SortLinks}} {{if .Active}}
+          <a class="active" href="?sort={{.Key}}&amp;limit={{$.Limit}}"
+            >{{.Label}}</a
+          >
+          {{else}}
+          <a href="?sort={{.Key}}&amp;limit={{$.Limit}}">{{.Label}}</a>
+          {{end}} {{end}}
         </p>
       </section>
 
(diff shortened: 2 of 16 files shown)


go/common/web/templates/queries_debug.html
```diff
@@ -0,0 +1,126 @@
+<!doctype html>
```
rafael, line 1: I happened to take a quick peek at this PR. Something that we should discuss. Should this be here? Or in admin-web? I'm kind of the mindset that the pooler should only expose the raw data, but everything the UI itself should be exposed through the web-admin UI.

haritabh17: Ack, will move this and another /debug page in a follow up

go/common/parser/ast/normalizer.go
```diff
@@ -34,6 +40,22 @@ func (r *NormalizeResult) WasNormalized() bool {
 	return len(r.BindValues) > 0
 }
 
+// Fingerprint returns a stable 16-character hex hash of the normalized SQL.
+// Two queries with the same NormalizedSQL produce identical fingerprints,
+// which makes the fingerprint suitable as a per-query-shape identifier for
+// metrics aggregation and cache keys.
+func (r *NormalizeResult) Fingerprint() string {
+	return FingerprintSQL(r.NormalizedSQL)
+}
```
GuptaManan100, line 49: Why not use the NoramlizedSQL directly?

go/services/multigateway/handler/queryregistry/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/multigres/multigres/go/common/cache/theine"
+)
+
+// maxSQLLabelLen caps the normalized SQL exposed as a Prometheus label so that
+// downstream label storage and Grafana legends stay readable. The full text
+// remains available via /debug/queries.
+const maxSQLLabelLen = 256
```
GuptaManan100: maxSQLLabelLen = 256 in metrics.go shadows MaxSQLLength = 4096 from config — there are now two truncation points with different caps:
- QueryStats.normalizedSQL stored at up to 4 KB (config).
- RegisterMetrics callback re-truncates to 256 B for the info-gauge label.

GuptaManan100 (approved):
Rest LGTM!

go/services/multigateway/handler/queryregistry/metrics.go
haritabh17: Ack, updating

go/common/parser/ast/normalizer.go
haritabh17: Thinking that using fingerprint keeps the hot-path counters cheap: duration, errors, and rows_returned emit many series per fingerprint (buckets × status × protocol × instance), and the label lands on every one of them. A 16-char hex stays compact; normalized SQL averages a few hundred bytes and can be multi-KB for ORM-generated queries — rough math on a typical OLTP app could be ~10 MB of label bytes per scrape vs. ~100 KB with the info-gauge. 

The full SQL isn't lost — mg_gateway_query_info exposes the fingerprint→SQL mapping once per tracked fingerprint, and dashboards join via group_left.

PR #875, opened 2026-04-21
feat(pgctld): add --init-db-sql-file flag to run SQL after initdb

rafael:
## Summary

- Adds `--init-db-sql-file` to `pgctld` (repeatable / CSV StringSlice) that runs each .sql file against the target database during `pgctld init` and the gRPC `InitDataDir` RPC.
- Mirrors Vitess's `--init_db_sql_file`, scoped to per-instance bootstrap (users, GRANTs, extensions, seed data) — the single-file design is extended to support multiple files run in order.
- Exposes the same option in shard-setup integration tests as `WithInitDbSQLFiles(...)`, with a new end-to-end test that asserts init-script artifacts exist on the elected primary after bootstrap.

## Description

Each file is executed via `psql -v ON_ERROR_STOP=1 -f <file>` against `--pg-database` using the transient PostgreSQL instance already used for custom-database creation. If both a non-default database and init SQL files are set, the transient instance starts once and handles both steps. Files are validated with `os.Stat` up front so operator misconfiguration (missing path) surfaces a clear error at the CLI boundary rather than buried inside psql stderr.

Refactor note: `setupDatabase` is split into `createDatabaseOnInstance(pg, database)` so it can share the transient instance with the new `runInitDbSQLFiles` helper, avoiding a double start/stop when both features are in use.

## Test plan

- [x] Unit: `TestInitDbSQLFilesFlag` (parsing: default / repeated / CSV)
- [x] Unit: `TestRunInitDbSQLFiles` (order, missing-file error, abort-on-first-failure with mock psql)
- [x] Integration: `TestShardSetup_InitDbSQLFilesExecuted` — bootstraps a shard with two init SQL files, verifies the created table row and cluster-wide role exist on the primary

---
Fixes: MUL-366

### DIFF
--- a/go/cmd/pgctld/command/init.go
+++ b/go/cmd/pgctld/command/init.go
@@ -110,11 +110,11 @@ func InitDataDirWithResult(logger *slog.Logger, poolerDir string, cfg PgCtldServ
 		return nil, fmt.Errorf("failed to create postgres config: %w", err)
 	}
 
-	// If the target database is not the default "postgres" (always created by initdb),
-	// start PostgreSQL transiently and create it — same as docker-library/postgres does.
-	if cfg.Database != constants.DefaultPostgresDatabase {
-		if err := setupDatabase(logger, cfg); err != nil {
-			return nil, fmt.Errorf("failed to create database %q: %w", cfg.Database, err)
+	// Post-initdb steps that need a running server (custom DB creation, init
+	// SQL files) share a single transient PostgreSQL instance.
+	if cfg.Database != constants.DefaultPostgresDatabase || len(cfg.InitDbSQLFiles) > 0 {
+		if err := postInitdbSetup(logger, cfg); err != nil {
+			return nil, err
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


go/cmd/pgctld/command/init.go
```diff
@@ -124,13 +124,61 @@ func InitDataDirWithResult(logger *slog.Logger, poolerDir string, cfg PgCtldServ
...
+		if err := createDatabaseOnInstance(logger, pg, cfg.Database); err != nil {
+			return fmt.Errorf("failed to create database %q: %w", cfg.Database, err)
+		}
+	}
+
+	if runSQL {
+		if err := runInitDbSQLFiles(logger, pg, cfg.Database, cfg.InitDbSQLFiles); err != nil {
+			return err
+		}
+	}
```
mkindahl: Nit (feel free to ignore): If there are no files, the for-loop will not iterate, so it is strictly speaking not necessary to have an if-statement around this.

go/cmd/pgctld/command/root.go
```diff
@@ -114,6 +115,11 @@ func GetRootCommand() (*cobra.Command, *PgCtlCommand) {
 			EnvVars:  []string{constants.PgInitdbArgsEnvVar},
 			Dynamic:  false,
 		}),
+		initDbSQLFiles: viperutil.Configure(reg, "init-db-sql-file", viperutil.Options[[]string]{
+			Default:  []string{},
+			FlagName: "init-db-sql-file",
+			Dynamic:  false,
+		}),
```
mkindahl, line 123: We might want to make this possible to configure using an environment variable. No strong opinion, and I don't think we need it right now, but it seems useful in docker containers.

go/cmd/pgctld/command/root.go
```diff
@@ -164,6 +170,7 @@ management for PostgreSQL servers.`,
 	root.PersistentFlags().String("pg-hba-template", pc.pgHbaTemplate.Default(), "Path to custom pg_hba.conf template file")
 	root.PersistentFlags().String("postgres-config-template", pc.postgresConfigTmpl.Default(), "Path to custom postgresql.conf template file")
 	root.PersistentFlags().String("pg-initdb-args", pc.pgInitdbArgs.Default(), "Extra arguments passed to initdb (overrides "+constants.PgInitdbArgsEnvVar+" env var)")
+	root.PersistentFlags().StringSlice("init-db-sql-file", pc.initDbSQLFiles.Default(), "Path to a .sql file to run against the target database after data directory initialization. Repeat the flag to run multiple files in order.")
```
mkindahl:
```suggestion
	root.PersistentFlags().StringSlice("init-db-sql-file", pc.initDbSQLFiles.Default(), "Path to an .sql file to run against the target database after data directory initialization. Repeat the flag to run multiple files in order.")
```

mkindahl (approved):
LGTM.

cuongdo (approved):

go/cmd/pgctld/command/root.go
rafael: Good idea. Added

PR #876, opened 2026-04-21
refactor(consensus): move HA RPCs to ConsensusService, delete unused RPCs

dweitzman:
Move EmergencyDemote, DemoteStalePrimary, Promote, and UpdateConsensusRule from MultiPoolerManager to MultiPoolerConsensus service. Remove 12 RPCs that were deleted from the manager proto: StandbyReplicationStatus, ResetReplication, ConfigureSynchronousReplication, PrimaryStatus, PrimaryPosition, StopReplicationAndGetStatus, GetDurabilityPolicy, CreateDurabilityPolicy, ChangeType, GetFollowers, UndoDemote, and State.

Drop StandbyUpdateOperation_STANDBY_UPDATE_OPERATION_REPLACE; only ADD and REMOVE are supported. Remove GetLeadershipView and CanReachPrimary from ConsensusService (their data is now exposed via ManagerHealthStream).

Update all callers, tests, and rpcclient facades to use the new locations.


go/test/endtoend/multiorch/bootstrap_test.go
```diff
@@ -67,19 +68,20 @@ func TestBootstrapInitialization(t *testing.T) {
 			status, err := client.Manager.Status(ctx, &multipoolermanagerdatapb.StatusRequest{})
 			require.NoError(t, err, "should get status from %s", name)
 
-			t.Logf("Node %s Status: IsInitialized=%v, HasDataDirectory=%v, PostgresReady=%v, PostgresRole=%s, PoolerType=%s",
+			t.Logf("Node %s Status: IsInitialized=%v, HasDataDirectory=%v, PostgresReady=%v, PostgresRole=%s, PoolerType=%s, CohortMembers=%d",
 				name, status.Status.IsInitialized, status.Status.HasDataDirectory,
-				status.Status.PostgresReady, status.Status.PostgresRole, status.Status.PoolerType)
-
-			// Nodes should be completely uninitialized (no data directory at all)
```
dweitzman, line 74: This test logic has probably been broken since the bootstrap change, but it seems like this PR might make pooler initialization faster or something so the race condition is more likely to trigger

Poolers can initialize without orch now, so it's better to check that there's no cohort instead of asserting that there's no backup

proto/consensusservice.proto
```diff
@@ -17,18 +17,49 @@ syntax = "proto3";
...
+  // EmergencyDemote demotes the current leader server
+  rpc EmergencyDemote(multipoolermanagerdata.EmergencyDemoteRequest)
+    returns (multipoolermanagerdata.EmergencyDemoteResponse);
+
+  // Promote promotes a replica to leader (Multigres-level operation)
+  rpc Promote(multipoolermanagerdata.PromoteRequest)
+    returns (multipoolermanagerdata.PromoteResponse);
+
+  // UpdateConsensusRule updates the synchronous standby list (quorum membership)
+  rpc UpdateConsensusRule(multipoolermanagerdata.UpdateSynchronousStandbyListRequest)
```
dweitzman, line 55: Note: I renamed the UpdateSynchronousStandbyList RPC to UpdateConsensusRule, but I haven't renamed the request and response types

go/services/multipooler/grpcconsensusservice/service.go
```diff
@@ -63,20 +65,75 @@ func (s *consensusService) Status(ctx context.Context, req *consensusdata.Status
...
+		req.Force)
+	if err != nil {
+		return nil, mterrors.ToGRPC(err)
+	}
+	return &multipoolermanagerdatapb.SetPrimaryConnInfoResponse{}, nil
+}
+
+// RewindToSource performs pg_rewind to synchronize this server with a source
+func (s *consensusService) RewindToSource(ctx context.Context, req *multipoolermanagerdatapb.RewindToSourceRequest) (*multipoolermanagerdatapb.RewindToSourceResponse, error) {
+	return s.manager.RewindToSource(ctx, req.Source)
```
cuongdo: when there's an error, should the return value be wrapped with `mterrors.ToGRPC()`?

cuongdo (commented):
I left one comment, and I'll leave it to others with more context do a proper review. I love the net reduction of code though!

rafael (approved):
The rename and removing of unused methods makes sense to me.

go/services/multipooler/grpcconsensusservice/service.go
dweitzman: Good catch!

PR #880, opened 2026-04-22
feat(multigateway): walk expression trees to reject unsafe FuncCalls and track set_config

GuptaManan100:
## Summary

- Walks every `FuncCall` in a statement's expression trees at plan time and rejects built-ins that bypass the pooler boundary: `dblink*`, `pg_read_file*`, `pg_ls_dir`, `pg_stat_file`, `lo_import/export`, `pg_execute_server_program`, and the `{query,table,cursor}_to_xml*` helpers. Schema-qualified variants (`pg_catalog.dblink`) resolve to the same entry.
- Tracks `set_config(name, value, false)` calls that sit directly as a top-level SELECT target-list entry, so the pooler's `SessionSettings` picks up changes that previously leaked. A SELECT with tracked set_configs plans as `Sequence[silent ApplySessionState..., Route]`: the silent primitives update the tracker, the Route sends the unmodified query to PG which executes set_config and streams the result back.
- Rejects `set_config` in any other position (WHERE, subqueries, CTEs, DEFAULT expressions, nested inside another function, INSERT/UPDATE) and non-literal arguments anywhere. `set_config(..., true)` passes through — transaction-scoped, nothing to track.

## Description

The existing Tier 2 filter blocks whole statements (LOAD, ALTER SYSTEM, CREATE DATABASE, …) but does not look inside expressions. `SELECT set_config('work_mem', '10GB', false)` is functionally equivalent to `SET work_mem = '10GB'` but bypasses the `VariableSetStmt` planner, so the gateway's `SessionSettings` diverges from the backend's actual state and the change leaks when the pool reuses that connection. Similar bypass vectors exist via `dblink`, `pg_read_file`, and friends.

The walker runs inside `Plan()` (and `PlanPortal()`, via a shared `planUnsupportedConstructs` helper), not in the executor. Putting it in the planner means the plan cache short-circuits it: a cached plan is by construction safe. The normalizer is configured to preserve literals inside `set_config(...)` so the walker still sees the original `A_Const` arguments despite normalization happening earlier in the pipeline.

When a SELECT's target list contains accepted set_configs, `planSelectStmt` builds a `Sequence[silent ApplySessionState per call, Route]`. The silent primitives update `SessionSettings` without emitting anything; the Route sends the original SQL to PG, which executes set_config normally and returns the full result (set_config column + any other target columns + table rows). No fast-path for the bare `SELECT set_config(...)` case — one PG round-trip is worth the uniform planning logic.

Out of scope (covered by follow-up tickets):
- Calls inside PL/pgSQL bodies (DO, CREATE FUNCTION LANGUAGE plpgsql) — waiting on body-analysis work.
- Dynamic SQL (`EXECUTE 'SELECT '||var`) — inherently unanalyzable.

## Test plan

- [x] Unit: rejection list (bare + schema-qualified + embedded contexts), set_config accepted positions, set_config rejected positions, non-literal arg rejection, `pg_catalog` name resolution, plan construction for bare / mixed / multi-set_config
- [x] E2E: SELECT set_config persists across 100 pool rotations (simple + extended protocol), `pg_catalog.set_config`, mixed `SELECT set_config(...), id, name FROM t ORDER BY id` returns rows AND updates tracker, multiple set_configs in one target list, `set_config(..., true)` passes through inside a transaction, blocklist functions rejected with SQLSTATE 0A000, connection healthy after rejections
- [x] Regression: existing session settings / SET-RESET / Tier 2 rejection E2E suites still pass


go/services/multigateway/planner/planner.go
```diff
@@ -148,10 +151,11 @@ func (p *Planner) Plan(
 		plan, err = p.planDefault(sql, stmt, conn)
 
 	case ast.T_SelectStmt:
-		if ss := stmt.(*ast.SelectStmt); ss.IntoClause != nil && ss.IntoClause.Rel != nil && ss.IntoClause.Rel.RelPersistence == ast.RELPERSISTENCE_TEMP {
+		ss := stmt.(*ast.SelectStmt)
+		if ss.IntoClause != nil && ss.IntoClause.Rel != nil && ss.IntoClause.Rel.RelPersistence == ast.RELPERSISTENCE_TEMP {
```
haritabh17, line 155: collectTopLevelSetConfigs doesn't gate on IntoClause, so a SELECT set_config('work_mem','256MB',false), * INTO TEMP foo FROM t should collect and validate the set_config, but this branch takes the temp-table fast-path and throws exprResult.SetConfigs away. PG still runs the set_config on the backend, but the gateway tracker stays stale — isn't that the exact pool-rotation leak this PR closes? Should we either reject this shape or thread setConfigs through to planTempTableCreation?

go/common/parser/ast/normalizer.go
```diff
@@ -79,9 +86,13 @@ func Normalize(stmt Stmt) *NormalizeResult {
 
 		// Skip subtrees where literal values carry semantic meaning that
 		// affects planning (e.g., SET timezone = 'UTC') — don't normalize them.
-		switch node.(type) {
+		switch n := node.(type) {
 		case *VariableSetStmt, *VariableShowStmt, *DefElem:
 			return false
+		case *FuncCall:
+			if isPlannerLiteralFunc(n.Funcname) {
+				return false
```
haritabh17: Do we need to skip normalization when is_local=true? Those calls aren't tracked or rewritten, so the literals don't need preserving — the PostgREST idiom is select set_config('request.jwt.claims', '<dynamic JSON>', true) once per request, which would produce a unique fingerprint every time under this rule and churn the plan cache. Would gating the skip on an is_local=false literal avoid that?

go/services/multigateway/planner/unsafe_funccall.go
```diff
@@ -0,0 +1,313 @@
...
+	s, ok := c.Val.(*ast.String)
+	if !ok {
+		return "", false
+	}
+	return s.SVal, true
+}
+
+// constBoolArg returns the underlying boolean value if n is a boolean-valued
+// A_Const literal. PostgreSQL parses `true`/`false` as A_Const{Val: Boolean}.
+func constBoolArg(n ast.Node) (bool, bool) {
```
haritabh17, line 362: These do a direct c.Val.(*ast.String|*ast.Boolean) with no TypeCast unwrap, so set_config('work_mem', '256MB'::text, false) or set_config('x','y','t'::bool) gets rejected as "non-literal arguments" even though the user did pass literals. Same for numeric literals where PG would implicitly cast (set_config('x', 100, false)). Worth peeling one layer of TypeCast before the assertion?

go/services/multigateway/planner/unsafe_funccall.go
```diff
@@ -0,0 +1,313 @@
...
+
+// validateAcceptedSetConfig verifies that an allowed-position set_config
+// call has the expected literal arguments, and returns its (name, value)
+// when is_local=false. Returns (nil, nil) when is_local=true — allowed but
+// not tracked.
+func validateAcceptedSetConfig(fc *ast.FuncCall) (*setConfigCall, error) {
+	name, value, isLocal, ok := parseSetConfigArgs(fc)
+	if !ok {
+		return nil, mterrors.NewFeatureNotSupported(
+			"set_config with non-literal arguments is not supported through the connection pooler")
```
haritabh17: Extended-protocol clients that Parse "SELECT set_config($1,$2,$3)" then Bind will hit this with ParamRef args. I assume the rejection is intentional (can't track what we don't see at plan time), but could the message distinguish "bound parameter" from "non-literal expression" so the client knows the fix is to inline the literals rather than rewrite the expression?

go/services/multigateway/planner/planner.go
GuptaManan100: Good catch — that was a real gap. Threading the set_configs into `planTempTableCreation` would have to invent a Sequence that mixes the silent-tracker primitives with the temp-table reservation primitive, and the failure modes there (tracker applied but reservation refused, or vice versa) felt like more surface than this corner case warranted. So I went with rejection: `collectTopLevelSetConfigs` in `unsafe_funccall.go` now bails out when `ss.IntoClause != nil`, which surfaces the same "set_config is only supported as a top-level SELECT target list entry" error users already see for set_config in INSERT VALUES / CTEs / subqueries. New rejection coverage is in `unsafe_funccall_test.go` (`TestInspectExpressionFuncCalls_SetConfigRejected/set_config_in_SELECT_INTO_TEMP_target_list`).

go/common/parser/ast/normalizer.go
GuptaManan100: Agreed, this matters for any per-request `set_config(..., true)` pattern. Normalizer now only skips when `is_local` is *not* literal `true`

go/services/multigateway/planner/unsafe_funccall.go
GuptaManan100: Done. Added `unwrapTypeCast` (loops to handle stacked casts like `'t'::text::bool`) and call it at the top of both `constStringArg` and `constBoolArg`. Its not perfect since we don't have the type casting code from postgres, but it's good enough for now.

go/services/multigateway/planner/unsafe_funccall.go
GuptaManan100: rejection is intentional, but you're right that the diagnostic was unhelpful. Fixed

go/services/multigateway/planner/unsafe_funccall.go
```diff
@@ -0,0 +1,371 @@
...
+// metadata-query helpers.
+//
+// Keys are lowercased unqualified function names. Schema-qualified calls
+// (e.g. pg_catalog.dblink) resolve to the same entry via resolveFuncName.
+var funcBlocklist = map[string]string{
+	// Outbound database connections.
+	"dblink":           "dblink is not supported: outbound database connections are not permitted through the connection pooler",
+	"dblink_exec":      "dblink_exec is not supported: outbound database connections are not permitted through the connection pooler",
+	"dblink_connect":   "dblink_connect is not supported: outbound database connections are not permitted through the connection pooler",
+	"dblink_connect_u": "dblink_connect_u is not supported: outbound database connections are not permitted through the connection pooler",
```
haritabh17: Should we add dblink_open, dblink_fetch, dblink_close, dblink_send_query, and dblink_get_result to the blocklist for completeness.

go/test/endtoend/queryserving/unsafe_funccall_test.go
```diff
@@ -0,0 +1,301 @@
...
+		},
+		{
+			"set_config in CTE",
+			"WITH c AS (SELECT set_config('work_mem','256MB',false)) SELECT * FROM c",
+			"set_config is only supported as a top-level SELECT target list entry",
+		},
+		{
+			"non-literal set_config value",
+			"SELECT set_config('work_mem', name, false) FROM (SELECT 'x' AS name) s",
+			"non-literal arguments",
```
haritabh17: This `"non-literal arguments"` doesn't match any error 

`"non-literal arguments"` → `"set_config value argument must be a literal constant"`.

go/services/multigateway/planner/select_stmt.go
```diff
@@ -0,0 +1,67 @@
...
+	setConfigs []setConfigCall,
+) (*engine.Plan, error) {
+	if len(setConfigs) == 0 {
+		return p.planDefault(sql, stmt, conn)
+	}
+
+	primitives := make([]engine.Primitive, 0, len(setConfigs)+1)
+	for _, sc := range setConfigs {
+		primitives = append(primitives, engine.NewApplySessionStateSilent(sql, syntheticSetStmt(sc)))
+	}
```
haritabh17, line 50: `SELECT set_config('x','y',false), 1 AS one` This kind of statements may fail

go/services/multigateway/planner/planner.go
```diff
@@ -212,8 +216,13 @@ func (p *Planner) PlanPortal(
...
 
-	// Reject Tier 2 statements (server-level operations unsafe for a hosted pooler).
-	if err := planUnsupportedStmt(stmt); err != nil {
+	// Non-cacheable extended-protocol statements reach PlanPortal directly
+	// (cacheable ones go through resolvePortalPlan → Plan, which does the
+	// same checks), so both paths must share the same pre-dispatch rejection.
+	// We throw away the set_config result here: PlanPortal only routes
+	// gateway-local statement types, none of which are SELECTs that could
+	// carry tracked set_configs.
+	if _, err := planUnsupportedConstructs(stmt); err != nil {
```
haritabh17, line 225: On the extended-protocol path, when a cacheable SELECT contains set_config(...,false), the planner builds the right plan (Sequence = ApplySessionStateSilent + Route) but the executor never runs it — it only extracts routing (which falls through for Sequence) and forwards the raw portal.

go/services/multigateway/engine/apply_session_state.go
```diff
@@ -61,6 +69,18 @@ func NewApplySessionState(sql string, stmt *ast.VariableSetStmt) *ApplySessionSt
 	}
 }
 
+// NewApplySessionStateSilent creates an ApplySessionState that updates the
+// tracker without emitting anything to the client. Intended for use inside a
+// Sequence where a Route primitive owns the client-facing response — see
+// planner.planSelectStmt for the `SELECT set_config(...), * FROM t` case.
+func NewApplySessionStateSilent(sql string, stmt *ast.VariableSetStmt) *ApplySessionState {
+	return &ApplySessionState{
+		VariableStmt:   stmt,
```
haritabh17, line 79: executeReset unconditionally calls callback regardless of SilentTracking. And here we do not restrict the VariableSetStmt.Kind, so a future caller passing a RESET-kind statement could emit a stray CommandComplete("RESET") into the protocol stream ahead of the Route's real response, corrupting the client's result parsing?

haritabh17 (changes requested):

go/services/multigateway/planner/unsafe_funccall.go
GuptaManan100: Yes, added all five to `funcBlocklist` in `unsafe_funccall.go`.

go/services/multigateway/engine/apply_session_state.go
GuptaManan100: Agreed, defensive fix applied.

go/test/endtoend/queryserving/unsafe_funccall_test.go
GuptaManan100: Good catch, that string was left over from the pre-rewrite error message. Updated the e2e case in `unsafe_funccall_test.go` to expect `"set_config value argument must be a literal constant"`, which is what `validateAcceptedSetConfig` actually emits now for the `name` column inside the subquery.

go/services/multigateway/planner/planner.go
GuptaManan100: Fixed by adding PortalStreamExecute to the primitive interface 😄

go/services/multigateway/planner/select_stmt.go
GuptaManan100: I want to make sure I'm fixing the failure mode you actually observed. On the simple-protocol path I exercised this exact shape end-to-end (see `TestMultiGateway_SetConfigRoutedAsSET/SELECT set_config alongside table read` in `unsafe_funccall_test.go`. It runs `SELECT set_config('work_mem','384MB',false), id, name FROM set_config_mix_t` and verifies both the row stream and the tracker update), so I don't have a reproducer there.

haritabh17 (approved):
🚀

PR #881, opened 2026-04-22
test(sqllogictest): add differential suite vs upstream corpus

GuptaManan100:
## Summary

- Add `go/test/endtoend/queryserving/sqllogictest/` — a label-gated differential suite that runs the upstream SQLite sqllogictest corpus against direct PostgreSQL and Multigres multigateway, under both simple and extended wire protocols.
- Extract the Postgres-from-source build logic out of `pgregresstest/` into a new shared `pgbuilder/` package that both suites now consume. `pgbuilder.StartStandalone` is new: initdb + launch a postgres server on a free port, used by the sqllogictest differential target.
- Install `sqllogictest-rs` v0.29.1 via `make tools`, SHA-pinned for darwin/linux × arm64/amd64.
- Add `.github/workflows/test-sqllogictest.yml` — label-gated (`Run SqlLogicTest Tests`) + nightly cron at 03:30 UTC + weekly Slack summary + `workflow_dispatch`. Results are cached as a baseline; regressions surface via Slack, reusing `.github/scripts/detect-regressions.sh` unchanged.

## Description

**Corpus.** Shallow-clones [gregrahn/sqllogictest](https://github.com/gregrahn/sqllogictest) (git mirror of the SQLite fossil repo) at a pinned SHA into `/tmp/multigres_slt_cache`. Upstream [`COPYRIGHT.md`](https://github.com/gregrahn/sqllogictest/blob/master/COPYRIGHT.md) offers quad-licensing (GPL / BSD / MIT / CC0 — "use whichever applies best"); we consume the corpus under MIT. `SLT_CORPUS_DIR` overrides for bring-your-own checkouts.

**Tracker, not a gate.** The Go test never fails on divergences or corpus failures — it records per-file pass/fail for both targets under both protocols (4 invocations per file) and writes `results.json` + markdown summary to `GITHUB_STEP_SUMMARY`. The JSON shape mirrors pgregresstest's so `.github/scripts/detect-regressions.sh` handles regression detection unchanged. Each protocol produces its own suite entry (`SQLLogicTest-simple`, `SQLLogicTest-extended`), so regressions are tracked independently per protocol.

**Reset between runs.** Each file runs 4× against a shared database; the public schema is dropped and recreated before each invocation so .test files can assume a clean slate.

**Opt-in.** Set `RUN_SQLLOGICTEST=1` to run. Skips cleanly if `postgres`/`initdb`/`pg_ctl`/`pg_isready` aren't on PATH (same pattern as other e2e suites).

**Shared pgbuilder.** `pgregresstest/postgres_builder.go` now wraps `pgbuilder.Builder` — the source checkout, configure/make/install, and cache-dir management moved into `go/test/endtoend/pgbuilder/`. Regression-suite-specific helpers (pg_regress wrappers, TAP parsing, markdown/JSON reports) stayed behind. pgregresstest's public API is unchanged.

## Test plan

- [x] Local smoke test: `RUN_SQLLOGICTEST=1 SLT_CORPUS_GLOB='test/select1.test' /mt-dev integration queryserving/sqllogictest` — clones corpus, runs 1 file × 2 protocols × 2 targets, emits results.json with `failed_both: 1` on SQLite-integer-semantics mismatch (correctly categorized, no proxy divergence), Go test exits 0.
- [x] `make tools` downloads and SHA-verifies sqllogictest-bin 0.29.1 on darwin-arm64.
- [x] Compile: `go build ./...` and `go vet ./go/test/endtoend/...` clean.
- [x] pgregresstest still builds after the `pgbuilder` extraction.
- [ ] CI run on this PR (label this PR `Run SqlLogicTest Tests` to trigger).


go/test/endtoend/queryserving/sqllogictest/sqllogictest_test.go
```diff
@@ -0,0 +1,279 @@
...
+//	SLT_CORPUS_DIR=<dir>      — use an external corpus instead of testdata/
+//	SLT_CORPUS_GLOB=<glob>    — scope which corpus files run (default: **/*.slt)
+//	SLT_PER_FILE_TIMEOUT=<d>  — Go duration, e.g. "90s" (default: 5m)
+func TestPostgreSQLSqlLogicTest(t *testing.T) {
+	if os.Getenv("RUN_SQLLOGICTEST") != "1" {
+		t.Skip("skipping sqllogictest: set RUN_SQLLOGICTEST=1 to run")
+	}
+
+	if !utils.HasPostgreSQLBinaries() {
+		t.Skip("skipping: PostgreSQL binaries not found on PATH")
```
haritabh17: Should we remove lines this since this doesn't apply to a suite that builds PG from source as done in a later block

go/test/endtoend/queryserving/sqllogictest/corpus.go
```diff
@@ -0,0 +1,201 @@
...
+	var stderr bytes.Buffer
+	cloneCmd.Stderr = &stderr
+	if err := cloneCmd.Run(); err != nil {
+		return "", fmt.Errorf("clone corpus: %w (stderr: %s)", err, stderr.String())
+	}
+
+	checkoutCmd := executil.Command(ctx, "git", "-C", dir, "checkout", CorpusCommit)
+	checkoutCmd.Stderr = &stderr
+	if err := checkoutCmd.Run(); err != nil {
+		return "", fmt.Errorf("checkout pinned corpus commit: %w (stderr: %s)", err, stderr.String())
```
haritabh17: Using a single bytes.Buffer (stderr) for both the git clone command and the subsequent git checkout command without calling stderr.Reset().

go/test/endtoend/queryserving/sqllogictest/sqllogictest_test.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	if len(files) == 0 {
+		t.Fatalf("no files matched SLT_CORPUS_GLOB in %s", corpusRoot)
+	}
+	t.Logf("Phase 3: corpus = %d files from %s (glob=%q, commit=%s)",
+		len(files), corpusRoot, firstNonEmpty(os.Getenv("SLT_CORPUS_GLOB"), DefaultCorpusGlob), CorpusCommit)
+
+	// Phase 4: start the baseline standalone postgres.
+	t.Logf("Phase 4: starting standalone PostgreSQL...")
+	standalone, err := pgbuilder.StartStandalone(t, buildCtx, builder, standalonePassword)
```
haritabh17: Using the same ctx till here might hit timeouts

.github/workflows/test-sqllogictest.yml
```diff
@@ -0,0 +1,251 @@
...
+        if: ${{ !cancelled() && steps.results.outputs.path != '' }}
+        env:
+          RESULTS_PATH: ${{ steps.results.outputs.path }}
+        run: |
+          mkdir -p /tmp/sqllogictest-baseline
+          cp "$RESULTS_PATH" /tmp/sqllogictest-baseline/results.json
+
+      - name: Save current results as new baseline
+        if: ${{ !cancelled() && steps.results.outputs.path != '' }}
+        uses: actions/cache/save@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
```
haritabh17: Should we ensure that only the nightly runs are considered as baseline and not runs done manually / via PRs??

haritabh17 (approved):
Rest looks good 🚀

go/test/endtoend/queryserving/sqllogictest/sqllogictest_test.go
GuptaManan100: Fixed

go/test/endtoend/queryserving/sqllogictest/corpus.go
GuptaManan100: Fixed

go/test/endtoend/queryserving/sqllogictest/sqllogictest_test.go
GuptaManan100: Fixed

.github/workflows/test-sqllogictest.yml
GuptaManan100: Fixed

PR #882, opened 2026-04-22
feat(multipooler): implement SCRAM key passthrough to backing PostgreSQL

haritabh17:
## Summary

- Per-user backend pools now authenticate to PostgreSQL as the real end user using SCRAM-SHA-256 keys derived during the client's handshake at multigateway and forwarded on every query RPC.
- No plaintext password ever lives in multigres process memory; session-scoped keys are captured on the gateway `Conn`, zeroized on close, and consumed only at dial time on multipooler.
- `pg_hba_template.conf` no longer grants blanket `trust` on the local socket — `local all all` now requires `scram-sha-256`. A narrow admin-user trust line is preserved as the admin-pool bootstrap path and tagged as a follow-up.
- Closes MUL-192.

## Description

**Proto + wire.** New `query.UserAuth { client_key, server_key }` embedded on `ExecuteOptions` so every query-path RPC (`StreamExecute`, `PortalStreamExecute`, `Describe`, `CopyBidiExecute`, `ConcludeTransaction`, `DiscardTempTables`, `ReleaseReservedConnection`) carries the session's keys. Internal gRPC is mTLS-protected from MUL-191, so keys are confidential in flight.

**Capture on multigateway.** After ``ScramAuthenticator.HandleClientFinal`` succeeds, ``pgprotocol/server.Conn`` records the ``ClientKey`` / ``ServerKey`` alongside ``user`` and ``database``. ``Conn.Close`` scrubs the bytes in place and drops the slices. A private ``userAuthFrom(conn)`` helper in ``scatterconn`` wires keys into all 11 ``ExecuteOptions`` construction sites and copies the bytes defensively so lazy / retried gRPC marshalling cannot race ``Conn.Close``.

**Consume on multipooler.** ``pgprotocol/client.Config`` carries optional ``ScramClientKey`` / ``ScramServerKey``; the auth dispatcher picks ``newScramClientWithKeys`` when present and falls back to the password path otherwise. ``connpoolmanager`` gains ``*WithAuth`` variants on ``GetRegularConn``, ``GetRegularConnWithSettings``, and ``NewReservedConn`` that route keys into a new ``buildUserClientConfig`` on first-time pool creation (keys are deterministic per user+password, so caching at pool creation is correctness-safe). The existing signatures are preserved as thin wrappers so callers and tests that don't care about auth don't change.

**Executor threading.** ``scramKeysFromOptions(options)`` helper in executor; seven pool-acquire sites migrated to ``*WithAuth`` variants. ``portalExecuteWithRegular`` signature gains ``clientKey, serverKey`` (caller updated).

**pg_hba.** SCRAM passthrough is unconditional — there is no opt-out flag. ``pg_hba_template.conf`` replaces the blanket ``local all all trust`` with ``local all all scram-sha-256``; a narrow ``local all <admin> trust`` line is preserved with an inline TODO to close once pgctld provisions the admin password. ``Manager.Open`` logs a clear warning when the admin password is empty so operators see the remaining gap. (An earlier draft of this PR shipped a ``--multipooler-scram-passthrough`` toggle; it was removed in 656c3af0 — once the same change made ``pg_hba.conf`` enforce ``scram-sha-256`` on the local socket, flipping the flag to ``false`` could not actually re-enable trust auth, so the runtime toggle was strictly worse than none.)

**Password-rotation self-heal.** ``withReopenRetry`` detects class-28 SQLSTATE on the dial path, evicts the stale user pool from the snapshot, and retries once using the triggering session's keys — which are known-current, derived seconds earlier during the gateway-side SCRAM handshake against the live ``pg_authid`` verifier. Concurrent sessions racing the same rotation each find the eviction already done and retry against the fresh pool. A retry cap surfaces a clean class-28 error if the retrier itself is holding old-password keys, so the manager cannot loop indefinitely.

**Design doc.** ``docs/query_serving/scram_passthrough_design.md`` describes the final state including the preserved admin-user trust exception as a tracked open item, and documents the password-rotation self-heal flow.

## Test plan

- [x] Unit tests cover: SCRAM key capture + zeroize on close + safe close when no keys; ``scramKeysFromOptions`` (nil / missing / populated / empty); ``buildUserClientConfig`` (with keys / nil keys); ``userAuthFrom`` defensive copy; ``withReopenRetry`` evicts a stale user pool and retries once on class-28; ``withReopenRetry`` surfaces a persistent class-28 instead of looping.
- [x] ``go test -short ./go/...`` passes.
- [ ] Follow-up: integration test against a real PostgreSQL with ``pg_hba`` rendered from the updated template, exercising the end-to-end passthrough dial (MUL-192 unblocks this; tracked alongside harness updates).
- [ ] Follow-up: close the admin-user trust exception after pgctld provisions the admin password during ``initdb``.

### DIFF
--- a/docs/query_serving/scram_passthrough_design.md
+++ b/docs/query_serving/scram_passthrough_design.md
@@ -1,0 +1,101 @@
+# SCRAM Key Passthrough: MultiPooler → PostgreSQL
+
+## Overview
+
+MultiPooler authenticates to its backing PostgreSQL as the real end user, on every backend connection, using SCRAM-SHA-256 — without ever handling a plaintext password. The cryptographic material needed to authenticate to PostgreSQL is derived during the client's SCRAM handshake to MultiGateway, forwarded to MultiPooler on each query RPC, and used when MultiPooler dials a new backend connection.
+
+This document describes the design as a whole: where the keys live, how they travel, how they are consumed, what the security properties are, and what the boundaries of the approach are.
+
+## Background
+
+### Why SCRAM keys can be passed through
+
+PostgreSQL's SCRAM-SHA-256 verifier stores only `StoredKey = H(ClientKey)` and `ServerKey`. The client proves knowledge of `ClientKey` by sending `ClientProof = ClientKey XOR ClientSignature`, where `ClientSignature = HMAC(StoredKey, AuthMessage)`. A verifier that sees the proof can validate it *and* recover the plaintext `ClientKey` by XORing the proof with the signature it just computed.
+
+This property — that a successful SCRAM handshake reveals `ClientKey` to the verifier — is what makes passthrough possible. With `ClientKey` plus `ServerKey` (which MultiGateway already has, because MultiGateway fetched the verifier from `pg_authid` via MultiPooler's admin pool as part of the incoming handshake), MultiGateway can impersonate the user against any other SCRAM-SHA-256 verifier of the same password — including the backing PostgreSQL.
+
+Our SCRAM library exposes this via `scram.ScramAuthenticator.ExtractedKeys()`, and the wire-protocol client accepts pre-computed keys via `client.newScramClientWithKeys(...)`.
+
+### Why this matters
+
+Three things are true in the final design that are not true of a trust-on-the-socket design:
+
+1. **Identity reaches PostgreSQL.** `current_user`, `pg_stat_activity`, audit logs, row-level security predicates — all see the real user. PostgreSQL is no longer asked to authorize on the basis of "some trusted local process asserted a user name."
+2. **`pg_hba.conf` is authoritative.** Access decisions live in one place. There is no configuration path that lets an unauthenticated local process log into PostgreSQL (except replication and the configured superuser, preserved intentionally).
+3. **Plaintext passwords never exist in MultiGateway or MultiPooler memory.** Only derived `ClientKey`/`ServerKey`, session-scoped, zeroized at client session end.
+
+## Goals and Non-Goals
+
+### Goals
+
+- MultiPooler authenticates to PostgreSQL as the real user on every backend connection opened for that user's pool.
+- Neither MultiGateway nor MultiPooler ever holds a plaintext password.
+- Keys are scoped to the client session, never persisted, zeroized on close, never logged.
+- MultiPooler's admin pool (used for bootstrap operations such as reading `pg_authid`) continues to work using its configured superuser credential, independently of per-user passthrough.
+
+### Non-Goals
+
+- **Client-certificate authentication.** Cert-authenticated sessions do not produce SCRAM keys; how MultiPooler dials for them is a separate design.
+- **SCRAM-SHA-256-PLUS (channel binding).** Internal gRPC between MultiGateway and MultiPooler is mTLS-protected, which covers the bulk of the MITM story for the wire segment where keys are exposed.
+- **Per-user key caching on MultiPooler.** Possible optimization; not required for correctness.
+- **Credential storage outside PostgreSQL.** All verifiers continue to live in `pg_authid`.
+- **Fallback to MD5 or cleartext auth.** Both are explicitly rejected by the wire-protocol client.
+
+## Design
+
+### Where keys live
+
+Keys live on the MultiGateway-side client `Conn` struct (`pgprotocol/server/conn.go`), alongside `user`, `database`, and `params`. They are set exactly once, immediately after `ScramAuthenticator.HandleClientFinal` confirms authentication, by calling `ExtractedKeys()`. They are zeroized exactly once, inside `Conn.Close`, after handler state is released.
+
+Keys are session-scoped. A PostgreSQL password change invalidates only future client sessions; already-open backend connections remain usable until they close through their normal lifecycle.
+
+Keys are never written to disk, never included in logs, never surfaced in error messages. Any new log sites that touch the auth path are audited for leakage.
+
+### Wire format
+
+A `query.UserAuth` message holds `client_key` and `server_key` as bytes. It is embedded as an optional field on `query.ExecuteOptions`. Because `ExecuteOptions` rides on every query-path RPC — `StreamExecute`, `PortalStreamExecute`, `Describe`, `CopyBidiExecute`, `ConcludeTransaction`, `DiscardTempTables`, `ReleaseReservedConnection` — a single field covers the full query path.
+
+MultiGateway populates `UserAuth` from the session's captured keys on every outbound query RPC. Internal gRPC is mTLS-protected, so keys in flight are encrypted on the wire.
+
+### MultiPooler dial logic
+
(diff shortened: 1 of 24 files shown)


docs/query_serving/scram_passthrough_design.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Wire format
+
+A `query.UserAuth` message holds `client_key` and `server_key` as bytes. It is embedded as an optional field on `query.ExecuteOptions`. Because `ExecuteOptions` rides on every query-path RPC — `StreamExecute`, `PortalStreamExecute`, `Describe`, `CopyBidiExecute`, `ConcludeTransaction`, `DiscardTempTables`, `ReleaseReservedConnection` — a single field covers the full query path.
+
+MultiGateway populates `UserAuth` from the session's captured keys on every outbound query RPC. Internal gRPC is mTLS-protected, so keys in flight are encrypted on the wire.
+
+### MultiPooler dial logic
+
+`pgprotocol/client.Config` carries optional `ClientKey` and `ServerKey`. The client-side auth dispatch in `client/startup.go:handleAuthenticationRequest` branches on their presence: when set, it constructs a SCRAM client via `newScramClientWithKeys(user, clientKey, serverKey)`; otherwise it uses the password path.
```
GuptaManan100: Field name drift: the doc says "`pgprotocol/client.Config` carries optional `ClientKey` and `ServerKey`" but the actual fields are `ScramClientKey` / `ScramServerKey`. Please update so future readers grepping for the names can find them.

proto/query.proto
```diff
@@ -301,6 +301,28 @@ message ExecuteOptions {
...
+  UserAuth user_auth = 7;
+}
+
+// UserAuth carries cryptographic material extracted from the client's SCRAM
+// handshake at multigateway. The pair (client_key, server_key) is sufficient
+// to authenticate as the user to any SCRAM-SHA-256 verifier of the same
+// password, without ever holding the plaintext password.
+message UserAuth {
+  // client_key is the SCRAM ClientKey recovered from the client's proof
+  // during authentication. 32 bytes for SCRAM-SHA-256.
```
GuptaManan100, line 320: Defensive redaction: please add a custom `String()` method on `UserAuth` that returns "UserAuth{redacted}" (or similar). Today no log site dumps it, but proto messages get printed accidentally via `%v`, debug middleware, or future tracing — and the bytes here are password-equivalent. One-line guardrail for a high-value target. (Implement by hand on a non-generated companion file so it survives `make proto`.)

go/services/multipooler/connpoolmanager/interface.go
```diff
@@ -58,17 +58,31 @@ type PoolManager interface {
 	// GetRegularConn acquires a regular connection for the specified user.
 	GetRegularConn(ctx context.Context, user string) (regular.PooledConn, error)
 
+	// GetRegularConnWithAuth is GetRegularConn that also carries SCRAM
+	// passthrough keys from the caller's session. Keys are consumed only when
+	// this call triggers first-time user-pool creation.
+	GetRegularConnWithAuth(ctx context.Context, user string, clientKey, serverKey []byte) (regular.PooledConn, error)
```
GuptaManan100: Shouldn't this replace the other GetRegularConn? Like isn't auth required all the time now?

go/services/multipooler/connpoolmanager/manager_test.go
```diff
@@ -338,6 +340,125 @@ func TestWithReopenRetry_OnlyRetriesOnce(t *testing.T) {
...
+		}
+		return 42, nil
+	})
+
+	require.NoError(t, err)
+	assert.Equal(t, 42, result)
+	assert.Equal(t, 2, calls, "should retry exactly once after auth failure")
+	require.Len(t, seenPools, 2)
+	assert.NotSame(t, seenPools[0], seenPools[1], "retry must run against a freshly-built pool, not the evicted one")
+	assert.False(t, manager.HasUserPool("testuser") && manager.userPoolsSnapshot.Load() != nil && (*manager.userPoolsSnapshot.Load())["testuser"] == seenPools[0],
```
GuptaManan100: Let's unwrap this into multiple assert.False and assert.True statements

GuptaManan100 (commented):
The "Flag + pg_hba" paragraph still describes `--multipooler-scram-passthrough` defaulting to true with an opt-out env var. The flag was removed in commit 656c3af0. Please update the PR description before merge so the historical record matches the shipped code.


Another thing that I;m wondering, is do we really need it? Like multigateway is authenticating the user right now already, and then why do we need multipooler to do so again against postgres?

docs/query_serving/scram_passthrough_design.md
haritabh17: Fixed — updated to ScramClientKey / ScramServerKey so it matches the code.

go/services/multipooler/connpoolmanager/manager_test.go
haritabh17: Done — split into require.NotNil on the snapshot, assert.True on HasUserPool, and assert.NotSame against the stale pool.

haritabh17:
> Another thing that I;m wondering, is do we really need it? Like multigateway is authenticating the user right now already, and then why do we need multipooler to do so again against postgres?

Fair question. The reason is that PostgreSQL itself is still the auth boundary that matters: pg_hba.conf is enforced regardless of what happens upstream, so the pooler→Postgres hop has to satisfy it on its own terms. Passthrough lets us keep normal SCRAM in pg_hba.conf while preserving the real user identity end-to-end, so RLS, pg_stat_activity, and audit logs all see the right principal, and Postgres stays in the chain as defense-in-depth if multigateway is ever breached or bypassed. The pooler never holds the plaintext password — the SCRAM keys are derived from the gateway-side handshake.

go/services/multipooler/connpoolmanager/interface.go
haritabh17: Good call — collapsed the duplicate. GetRegularConn, GetRegularConnWithSettings, and NewReservedConn now each take clientKey, serverKey []byte directly; the *WithAuth siblings are gone. Session callers pass real keys; admin/internal paths (internal_query, pubsub/listener) pass nil, nil.

proto/query.proto
haritabh17: Couldn't take this literally — tested it: stock protoc-gen-go always emits func (x *UserAuth) String(), so a same-package companion file gives method UserAuth.String already declared. The wider Go ecosystem handles this via the [debug_redact = true] field option: Kubernetes uses it for AuthConfig password fields, and upstream protobuf-go has the runtime path tracked but not yet shipped — once it lands, generated String() will redact automatically.

Took the same approach: marked client_key / server_key [debug_redact = true] and added a descriptor-reflection lockdown test that fails if either field loses the annotation. Today the annotation is advisory in Go (so String() still emits raw bytes); it becomes load-bearing the moment upstream lands the runtime path.

GuptaManan100 (approved):

proto/query.proto
GuptaManan100: Okay, that works!

PR #884, opened 2026-04-22
fix(multipooler): retry ensurePreparedWithName on stale reserved connection

dweitzman:
When the pool hands out a connection that postgres has silently closed (e.g. idle timeout), the first write to a new reserved connection fails with broken pipe. Non-reserved queries recover via retryOnConnectionError, but the reserved connection setup path had no equivalent retry.

If ensurePreparedWithName fails with a connection error before any BEGIN has been sent (no session state to preserve), release the stale connection and retry up to 3 times with a fresh connection from the pool.

Fixes flaky TestWrappedPreparedStatementExecution/multigateway/create_temp_table_as_execute.

### DIFF
--- a/go/services/multipooler/executor/executor.go
+++ b/go/services/multipooler/executor/executor.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"time"
 
 	"github.com/multigres/multigres/go/common/mterrors"
 	"github.com/multigres/multigres/go/common/pgprotocol/protocol"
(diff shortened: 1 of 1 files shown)


GuptaManan100:
Thanks for digging into this, the stale-reserved-conn failure mode you identified is real and the reproducer is useful.

Closing in favor of a pool-layer fix. On review we noticed the same silent-close failure mode exists at two other first-write points in reserveAndStreamExecute (the BEGIN at executor.go:315 and the query itself at executor.go:362), and at the other NewReservedConn call sites (executor.go:520, executor.go:774). Rather than layering ad-hoc retries at each site, we're going to add a validate-with-retry primitive to reserved.Pool.NewConn, a WithValidate(func(ctx, *regular.Conn) error) option that plumbs through Manager/UserPool/reserved.Pool, retries on IsConnectionError via pooled.Taint(), and lets callers opt in per site.

For this bug specifically, we'll pass ensurePreparedWithName as the validate closure at executor.go:293. That removes the retry loop from the executor entirely and (as a bonus) makes the retry work even in the beginTx case that this PR had to bail on, since Parse can safely move before BEGIN.

New branch going up shortly.

GuptaManan100:
Closed in favour of https://github.com/multigres/multigres/pull/890

PR #885, opened 2026-04-22
fix(multigateway): rollback implicit transaction before returning isolation level error

rachit77:
## Problem

When a batch like `SELECT 1; BEGIN ISOLATION LEVEL SERIALIZABLE` arrives, Multigres injects a silent `BEGIN` to wrap the batch in an implicit transaction. When the user's `BEGIN ISOLATION LEVEL SERIALIZABLE` is then rejected (SQLSTATE 25001 isolation level must be set before any query), the implicit transaction was never rolled back. The PostgreSQL connection returned to the pool with an open transaction, risking dirty state for the next client that received it.

Root cause: `isImplicitTx` was set to `false` at the top of the `BEGIN` handler then immediately checked so the rollback branch was dead code.

## Fix

Save the flag before clearing it (`wasImplicitTx := isImplicitTx`) and use the saved value for the rollback check. No rollback is issued for explicit transactions (client-owned) only for implicit ones (Multigres-owned).

## Tests

- `BeginWithIsolationAfterQuery_RollsBackImplicitTx` verifies ROLLBACK is sent before the error is returned for an implicit transaction
- `BeginWithIsolationInExplicitTx_DoesNotRollback` verifies Multigres does not touch a client-owned explicit transaction

### DIFF
--- a/go/services/multigateway/handler/transaction_helpers.go
+++ b/go/services/multigateway/handler/transaction_helpers.go
@@ -127,6 +127,7 @@ func (h *MultiGatewayHandler) executeWithImplicitTransaction(
 		//     error out matching PostgreSQL behavior — SET TRANSACTION ISOLATION LEVEL
 		//     must be called before any query.
 		if ast.IsBeginStatement(stmt) {
+			wasImplicitTx := isImplicitTx
 			isImplicitTx = false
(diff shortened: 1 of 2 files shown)


GuptaManan100 (approved):
Thank you for your contribution to Multigres!

PR #888, opened 2026-04-23
refactor(consensus): remove and consolidate StatusResponse fields

dweitzman:
StatusResponse previously carried several redundant or overlapping fields (current_term, wal_position, role, is_healthy, is_eligible, pooler_id, cell, timeline_info). These are removed in favour of the existing nested types:

- LSN and current rule now live in ConsensusStatus.current_position
- Term revocation in ConsensusStatus.term_revocation
- Node identity in a new top-level id field, always set even when postgres is unavailable and ConsensusStatus cannot be built

On the consumer side:
- IsPrimary() helper added to go/common/consensus for checking whether a node holds the primary role according to the current agreed rule
- Test checks for postgres role are replaced with direct pg_is_in_recovery() queries, since consensus tracks agreed state across nodes, not per-node liveness signals
- createMockNode() in leader_appointment_test.go now accepts an explicit ShardRule so the incumbent primary is unambiguous in each test scenario


proto/consensusdata.proto
```diff
@@ -127,29 +120,9 @@ message StatusRequest {
 }
 
 message StatusResponse {
```
dweitzman, line 122: These deletions are the meat of the PR, and most other things are just consequences of that

proto/clustermetadata.proto
```diff
@@ -469,6 +471,10 @@ message ConsensusStatus {
   // ahead of current_position when the pooler has forward knowledge of an
   // upcoming rule that has not yet been replicated or written.
   HighestKnownRule highest_known_rule = 3;
+
+  // id identifies the pooler that produced this status. Makes ConsensusStatus
+  // self-describing when passed around without a surrounding envelope.
+  ID id = 4;
```
dweitzman, line 477: Adding the pooler ID into consensus status feels like it'll be helpful in the future, but I'm open to disagreement on that

go/common/consensus/pooler.go
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+package consensus
+
+import clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+
+// IsPrimary reports whether the pooler identified by cs is the designated
+// primary according to its highest committed rule. Returns false when cs, its
+// ID, or the current rule is absent.
+func IsPrimary(cs *clustermetadatapb.ConsensusStatus) bool {
```
dweitzman, line 22: If consensus status reports what pooler should be primary and the ID of the current pooler, this helper can determine whether the status in question belongs to a pooler that is or isn't the current consensus primary

That avoids the need for an `is_primary` field on the proto3 messages

go/services/multiorch/consensus/leader_appointment.go
```diff
@@ -426,30 +426,21 @@ func (c *Coordinator) EstablishLeadership(
...
-				waitReq := &multipoolermanagerdatapb.WaitForLSNRequest{
-					TargetLsn: expectedLSN,
-				}
-				waitCtx, waitCancel := context.WithTimeout(ctx, timeouts.RemoteOperationTimeout)
-				defer waitCancel()
-				if _, err := c.rpcClient.WaitForLSN(waitCtx, candidate.MultiPooler, waitReq); err != nil {
-					return mterrors.Wrapf(err, "candidate failed to replay WAL to %s", expectedLSN)
-				}
-			}
+	expectedLSN := status.GetConsensusStatus().GetCurrentPosition().GetLsn()
```
rafael, line 427: Note for self. This works because now GetCurrentPosition works regardless of the state.

go/services/multiorch/consensus/leader_appointment.go
```diff
@@ -426,30 +426,21 @@ func (c *Coordinator) EstablishLeadership(
...
+	if expectedLSN != "" && !commonconsensus.IsPrimary(status.GetConsensusStatus()) {
```
rafael, line 428: I have some thoughts here. Let's sync about this.

go/services/multiorch/recovery/actions/fix_replication_test.go
```diff
@@ -491,15 +495,17 @@ func TestFixReplicationAction_FailsWhenReplicationDoesNotStart(t *testing.T) {
...
-				TimelineId: 2, // Primary on timeline 2
+			ConsensusStatus: &clustermetadatapb.ConsensusStatus{
+				TermRevocation: &clustermetadatapb.TermRevocation{
+					RevokedBelowTerm: 1,
+				},
 			},
 		},
 		"multipooler-cell1-replica1": {
-			CurrentTerm: 1,
-			TimelineInfo: &consensusdatapb.TimelineInfo{
```
rafael, line 501: we removed the TImelineInfo. Why this did not break the tests?

go/services/multipooler/grpcconsensusservice/service_test.go
```diff
@@ -219,12 +219,8 @@ func TestConsensusService_Status(t *testing.T) {
 		// Should succeed
 		assert.NoError(t, err)
 		assert.NotNil(t, resp)
-		assert.Equal(t, "test-service", resp.PoolerId)
-		assert.Equal(t, "zone1", resp.Cell)
-		// Without database, should not be healthy
```
rafael, line 224: Is this hte idea here that these Availability signals?

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -405,74 +405,18 @@ func (pm *MultiPoolerManager) ConsensusStatus(ctx context.Context, req *consensu
...
-	localCurrentTerm := int64(0)
-	if term != nil {
-		localCurrentTerm = term.GetTermNumber()
-	}
 	localPrimaryTerm := int64(0)
 	if term != nil {
 		localPrimaryTerm = term.GetPrimaryTerm()
 	}
 
-	// Check if database is healthy by attempting a simple query
```
rafael, line 419: where did we moved this now?

rafael (commented):
It mostly looks good to me, but few things were not obvious:
- what happened with the timeline divergence field? 
- What exactly replaces isHealthy. 

For the other fields it was more obvious.

rafael (approved):
LGTM

go/services/multiorch/consensus/leader_appointment.go
dweitzman: To follow-up on discussion elsewhere: this code seems like it can probably be deleted, but I'm planning to make that a separate PR and keep this one as much of a pure refactoring as practical.

go/services/multiorch/recovery/actions/fix_replication_test.go
dweitzman: This wasn't used in production

go/services/multipooler/grpcconsensusservice/service_test.go
dweitzman: To the direct question, yes I do think of this as an availability signal

Of more immediate importance: this signal wasn't used. We might want to eventually, but I think we should add it if/when we're ready to use it

go/services/multipooler/manager/rpc_consensus.go
dweitzman: It's not moved, just deleted. The `isHealthy` signal wasn't consumed by anything. We may add this back later, but my preference is to wait until signals are needed to implement them

PR #890, opened 2026-04-23
fix(multipooler): retry on stale-socket connection errors during reserved-conn acquisition

GuptaManan100:
## Summary

- Adds a pool-layer retry primitive for reserved-connection acquisition so silent stale-socket failures (idle-timed-out PG connections) no longer surface to callers
- Closes the same gap on the regular pool path (`GetWithSettings` retry when `ApplySettings` hits a dead socket)
- Wires the new `WithValidate` hook into all three executor `NewReservedConn` call sites — `reserveAndStreamExecute`, `portalExecuteWithReserved`, and `CopyInitiate`
- Fixes a latent pgprotocol bug where `InitiateCopyFromStdin` left the conn in a dirty state on `ErrorResponse` (didn't drain the trailing RFQ)

## Problem

The reserved connection pool can hand out a regular pooled conn that PostgreSQL silently closed (idle session timeout). The first user-issued write on that socket — a `Parse`, `BEGIN`, or `InitiateCopyFromStdin` — fails with `EPIPE` / `EOF` / `FATAL 57P0x`. The non-reserved query path already recovers via `retryOnConnectionError` (in-place reconnect), but the reserved acquisition path had no equivalent. A flaky integration test (`TestWrappedPreparedStatementExecution/multigateway/create_temp_table_as_execute`) hit this against `ensurePreparedWithName` in `executor.reserveAndStreamExecute`.

A second, narrower variant of the same bug existed at `connpool.GetWithSettings`: when applying SETs failed with a connection error, the connpool closed the conn but did not retry, so callers saw the failure even though a fresh socket would have succeeded.

## Solution

Three layered fixes that share constants and retry semantics:

1. **`reserved.Pool.NewConn` accepts `ReservedConnOption`s, including `WithValidate`.** The validate callback runs on the underlying `*regular.Conn` before the reserved conn is registered. Connection-class errors taint the pooled conn (closing the socket immediately) and trigger a retry on a fresh one, up to `constants.MaxConnPoolRetryAttempts` total with `constants.ConnPoolRetryBackoff` between them. Plumbed through `Manager.NewReservedConn` and `UserPool.NewReservedConn` as a variadic option, so existing call sites are unchanged.

2. **`regular.Pool.GetWithSettings` retries on its own when settings are non-empty.** Same retry regime, same shared constants. Every existing `GetRegularConnWithSettings` caller benefits transparently — no API change.

3. **Executor wires `WithValidate` at all three new-reserved-conn call sites.** The first write (Parse / BEGIN+InitiateCopy / ensurePrepared) moves into the validate hook so the pool can retry it transparently. Existing-reserved-conn paths (`GetReservedConn`) are not exposed because those conns are actively held and never sit idle.

A side fix in `pgprotocol/client.Conn.InitiateCopyFromStdin`: on `MsgErrorResponse` we now drain to `ReadyForQuery` (via the existing `waitForReadyForQuery` helper) instead of returning early. The previous behavior left an unconsumed RFQ in the read buffer, which corrupted the next user of that socket. The original COPY-init call site masked this by always tainting on error; with the validate hook now recycling on non-connection errors, the protocol-level bug had to be fixed at the source.

Shared constants live in a new `go/common/constants/connpool.go` so the regular and reserved pools can't drift out of lockstep.

Docs (`connection_pooling.md`, `prepared_statements_design.md`) updated to describe the new retry primitive and the now-stale interface signature.

### DIFF
--- a/go/services/multipooler/pools/reserved/reserved_pool.go
+++ b/go/services/multipooler/pools/reserved/reserved_pool.go
@@ -204,49 +204,53 @@ func (p *Pool) NewConn(ctx context.Context, settings *connstate.Settings, opts .
 }
 
 // acquireValidated borrows a regular connection from the underlying pool
-// and, if validate is non-nil, runs it against the connection. Any
-// connection-classified error encountered along the way — whether from
-// GetWithSettings (stale socket exposed while applying SETs) or from
-// validate (stale socket exposed by the first user-issued write) —
-// triggers a retry on a fresh socket, up to constants.MaxConnPoolRetryAttempts. The
-// connpool itself only self-heals across ResetAllSettings failures
-// (connReopen at connpool/pool.go), not ApplySettings failures, so the
-// retry loop here closes the gap for new reserved-conn acquisition.
+// and, if validate is non-nil, runs it against the connection. A
+// connection error from validate (stale socket exposed by the first
+// user-issued write) taints the socket and triggers another attempt on
+// a fresh one, up to constants.MaxConnPoolRetryAttempts.
+//
+// Connection errors from GetWithSettings itself (stale socket exposed
+// while applying SETs) are handled inside regular.Pool.GetWithSettings,
+// which retries with the same regime; this layer does not double-retry.
 //
 // Non-connection errors propagate unchanged, with the pooled conn
-// recycled (or no conn to clean up if Get failed).
+// recycled.
 func (p *Pool) acquireValidated(
 	ctx context.Context,
 	settings *connstate.Settings,
 	validate func(context.Context, *regular.Conn) error,
 ) (regular.PooledConn, error) {
+	if validate == nil {
+		pooled, err := p.conns.GetWithSettings(ctx, settings)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get connection: %w", err)
+		}
+		return pooled, nil
+	}
+
 	var lastErr error
 	for attempt := 1; attempt <= constants.MaxConnPoolRetryAttempts; attempt++ {
 		pooled, err := p.conns.GetWithSettings(ctx, settings)
 		if err != nil {
-			if !mterrors.IsConnectionError(err) {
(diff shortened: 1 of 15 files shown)


go/services/multipooler/pools/reserved/reserved_pool.go
```diff
@@ -191,6 +203,72 @@ func (p *Pool) NewConn(ctx context.Context, settings *connstate.Settings) (*Conn
...
+		}
+
+		validateErr := validate(ctx, pooled.Conn)
+		if validateErr == nil {
+			return pooled, nil
+		}
+		if !mterrors.IsConnectionError(validateErr) {
+			pooled.Recycle()
+			return nil, validateErr
+		}
```
haritabh17: In the CopyReady new-conn branch, the validate closure runs BEGIN then InitiateCopyFromStdin on the raw *regular.Conn. If BEGIN succeeds (connection now in txn state 'T') but InitiateCopyFromStdin fails with a non-connection error (e.g. table not found, permission denied), the wrapped error reaches here which calls pooled.Recycle() without Taint — returning the connection to the pool still stuck in failed-transaction-error state ('E').

haritabh17 (approved):
Lets address the one point, rest looks good 🚀

go/services/multipooler/pools/reserved/reserved_pool.go
GuptaManan100: Good catch — fixed in b148c6ca. Changed `acquireValidated` to always taint on any validate error (not just connection errors). All three call sites (reserveAndStreamExecute, portalExecuteWithReserved, CopyInitiate) perform state-modifying work in validate, so partial-success failure (BEGIN → InitiateCopy fails) leaves the conn in a state we can't safely reuse. This also matches the pre-primitive behavior at every call site (each did `Release(ReleaseError)` on failure before this PR). Test and doc updated to match.

PR #893, opened 2026-04-23
fix(multipooler): use pguser in conninfo builder

soedirgo:
MultiPooler is using the database name as the user name in `setPrimaryConnInfoLocked()` - pass the `PgUser()` instead.


GuptaManan100 (approved):
Thank you for contributing to Multigres!

PR #899, opened 2026-04-24
fix: pass ctx to `filterAndPrioritize` for correct trace propagation

rachit77:
## Issue
- `filterAndPrioritize` was using `re.shutdownCtx` (the engine's root context) for a debug log call instead of the current recovery cycle context. This disconnected the log entry from the active `recovery/cycle` and `recovery/attempt` trace spans

## Proposed Solution
- Pass `ctx` through from `processShardProblems` to fix trace propagation


rachit77:
@GuptaManan100  Could you please share your thoughts on this small improvement? If it looks good to you, I’ll go ahead and fix the CI to make it green

PR #901, opened 2026-04-24
feat(pgctld): add /ready probe that dials postgres and gRPC sockets

niconosenzo:
## Description

This PR implements the /ready endpoint for pgctld.

Ready means postgres is accepting connections on its unix socket and the gRPC server is accepting connections. Replication health is intentionally excluded, a pod with degraded replication must stay in DNS so
multiorch can reach it to initiate repair. 

The check uses an actual dial rather than a socket file-existence test.

## Changes

- Added /ready HTTP handler in pgctld that dials the postgres Unix socket and the gRPC endpoint on every request.
- Added SocketFile() accessor on GrpcServer so the server command can pass the socket path to the handler without reaching into unexported fields.
- Wired the handler in server.go with pgSocketPath, grpcSocketPath, grpcBindAddress, and grpcPort. Matching the two gRPC listener modes (--grpc-socket-file vs --grpc-port).
- gRPC check prefers the Unix socket when configured and falls back to a TCP dial on the configured bind address. When neither is configured the check is skipped (mirrors GrpcServer.IsEnabled()).
- Wildcard bind addresses ("", 0.0.0.0, ::) are mapped to "localhost" for the TCP probe so the dial works on IPv4-only, IPv6-only, and dual-stack hosts. Specific addresses (e.g. ::1) are dialed directly during the probe.

## Testing

ToDo: field tests.

Ran the new TestPgctldReadyHandler suite covering both sockets up, each socket missing independently, stale socket files with no listener, TCP wildcard bind, IPv6 loopback bind, and the socket-preferred-over-TCP
precedence rule. The IPv6 case is guarded with t.Skip so it doesn't flake on IPv4-only CI runners.

### DIFF
--- a/go/cmd/pgctld/command/ready.go
+++ b/go/cmd/pgctld/command/ready.go
@@ -1,0 +1,71 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package command
+
+import (
+	"net"
+	"strconv"
+
+	"github.com/multigres/multigres/go/common/timeouts"
+)
+
+// grpcAccepting verifies the gRPC server is accepting connections. Prefers
+// the Unix socket, and falls back to a TCP dial on the configured bind
+// address when only the port is configured.
+// When neither is configured, gRPC is not in use and the check is skipped.
+func grpcAccepting(socketPath, bindAddress string, port int) bool {
+	if socketPath != "" {
+		return unixSocketAccepting(socketPath)
+	}
+	if port != 0 {
+		return tcpAccepting(bindAddress, port)
+	}
+	return true
+}
+
+// unixSocketAccepting returns true if a process is currently accepting
+// connections on the given Unix socket path. A successful Dial means the
+// kernel handed off to a listener, distinguishing a live listener from a
+// stale socket file left behind (by a crash, for example).
+// Any dial error (e.g. ENOENT, ECONNREFUSED, EACCES, timeout) returns false.
+func unixSocketAccepting(path string) bool {
+	conn, err := net.DialTimeout("unix", path, timeouts.ReadyDialTimeout)
+	if err != nil {
+		return false
+	}
+	_ = conn.Close()
+	return true
+}
+
+// tcpAccepting returns true if a TCP listener is accepting on the given
+// bind address and port.
+// If bindAddress is a wildcard ("", "0.0.0.0", or "::"), we dial "localhost"
+// instead: Go's resolver picks 127.0.0.1 and/or ::1 based on the host's
+// stack, so the probe works on IPv4-only, IPv6-only, and dual-stack hosts.
+// If bindAddress is a specific address (e.g. "127.0.0.1", "::1", an
+// interface IP) we dial that address directly.
+// See unixSocketAccepting for the error-handling contract.
+func tcpAccepting(bindAddress string, port int) bool {
+	host := bindAddress
+	if host == "" || host == "0.0.0.0" || host == "::" {
+		host = "localhost"
+	}
+	conn, err := net.DialTimeout("tcp", net.JoinHostPort(host, strconv.Itoa(port)), timeouts.ReadyDialTimeout)
+	if err != nil {
+		return false
+	}
+	_ = conn.Close()
+	return true
+}
--- a/go/cmd/pgctld/command/ready_test.go
+++ b/go/cmd/pgctld/command/ready_test.go
@@ -1,0 +1,130 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package command
+
+import (
+	"net"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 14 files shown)


go/cmd/pgctld/command/ready.go
```diff
@@ -0,0 +1,110 @@
...
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/multigres/multigres/go/common/web"
+)
+
+// readyDialTimeout bounds each probe dial. Local dials complete in microseconds;
+// the timeout is a safety guarantee for a kernel under extreme load, not an expected wait.
+const readyDialTimeout = 500 * time.Millisecond
```
mkindahl: How was this timeout value picked?

I agree that it should be set as high as possible and that it does not represent the normal expected wait, but I could imagine that some callers have a shorter timeout and the caller will timeout before this timeout fires. This would be harmless.

I could also imagine that some callers are fine with a longer timeout and that if we set it too short it will just cause spurious failures. Probably not very likely either, but something we might want to think about.

If there is no particular reason why this value was picked? It might be good to add a comment about the reason for this value so that we know if we can experiment with it or if there is a particular reason for this value.

go/cmd/pgctld/command/ready.go
```diff
@@ -0,0 +1,110 @@
...
+func pgctldReadyHandler(pgSocketPath, grpcSocketPath, grpcBindAddress string, grpcPort int) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if !unixSocketAccepting(pgSocketPath) {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			_ = web.Templates.ExecuteTemplate(w, "isok.html", false)
+			return
+		}
+		if !grpcAccepting(grpcSocketPath, grpcBindAddress, grpcPort) {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			_ = web.Templates.ExecuteTemplate(w, "isok.html", false)
```
mkindahl: The error is ignored here, but I think it _might_ be ok.

It can fail if the server is misconfigured and, for example, the template file moved out of sync with the code. In that case, we could return an error instead (like 500), but it would only serve the purpose of allowing us to see that in logs and fix the issue. It would shorten the time to find the error in production, but I do not think it is critical for this PR.

go/cmd/pgctld/command/ready_test.go
```diff
@@ -0,0 +1,201 @@
...
+	dir, err := os.MkdirTemp("/tmp", "p")
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = os.RemoveAll(dir) })
+	return dir
+}
+
+// listenUnix starts a Unix socket listener at path and registers cleanup.
+func listenUnix(t *testing.T, path string) string {
+	t.Helper()
+	l, err := net.Listen("unix", path)
```
mkindahl, line 41: It's just a test, but you might want to check that the 104-limit length is not exceeded before calling, to avoid spurious and "hard to figure" out errors. It shouldn't be exceeded, but if it does it will behave weird and bugs do creep into code. Would save us some time each time we make a mistake.

go/cmd/pgctld/command/ready.go
```diff
@@ -0,0 +1,110 @@
...
+// configured and the check is skipped (this matches GrpcServer.IsEnabled()).
+func pgctldReadyHandler(pgSocketPath, grpcSocketPath, grpcBindAddress string, grpcPort int) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if !unixSocketAccepting(pgSocketPath) {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			_ = web.Templates.ExecuteTemplate(w, "isok.html", false)
+			return
+		}
+		if !grpcAccepting(grpcSocketPath, grpcBindAddress, grpcPort) {
+			w.WriteHeader(http.StatusServiceUnavailable)
```
mkindahl: As discussed separately: if `pgctld` fails causing the service to be marked as non-ready, we might stop sending traffic to the pod even though `postgres` process is still running and active. It should not normally happen unless we take a single such `pgctld` failure as a signal that the pod is not ready, but just want to make sure to mention it.

mkindahl (approved):
LGTM. Added a few comments that you can deal with anyway you like. I do not think they are in any sense critical.

go/cmd/pgctld/command/ready.go
niconosenzo: Ack. Just one thought though, by the time we call `ExecuteTemplate` the status code header is already written,   hence failing the whole probe because of a cosmetic issue (the html template) feels like a bit out of the scope of this change. Adding a comment like:
```
// status code already committed; body is cosmetic, error is intentionally ignored
```
would help ?

go/cmd/pgctld/command/ready.go
niconosenzo: It was chosen somewhat arbitrarily. The only consideration taken was to stay within the [default value ](https://github.com/multigres/multigres/pull/901) set by `TimeoutSeconds`. Will add that on the comment.

go/cmd/pgctld/command/ready.go
```diff
@@ -0,0 +1,89 @@
...
+// grpcBindAddress: the gRPC listener's bind address (value of --grpc-bind-address).
+// grpcPort: gRPC TCP port (value of --grpc-port).
+//
+// If grpcSocketPath is empty and grpcPort is zero, gRPC is treated as not
+// configured and the check is skipped (this matches GrpcServer.IsEnabled()).
+
+// readyDialTimeout bounds each probe dial. Local dials complete in microseconds,
+// 500ms stays well within Kubernetes' default probe timeoutSeconds: 1, leaving
+// headroom for HTTP overhead. It is a safety net for a kernel under extreme load.
+const readyDialTimeout = 500 * time.Millisecond
```
rafael: small nit - this could be moved to the common timeouts package.

go/cmd/pgctld/command/server.go
```diff
@@ -164,6 +164,28 @@ func (s *PgCtldServerCmd) runServer(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Register /ready probe: postgres socket accepting + gRPC accepting.
+	// Replication health is intentionally excluded (see pgctldReadyHandler).
+	pgSocketPath := fmt.Sprintf("%s/.s.PGSQL.%d", pgctld.PostgresSocketDir(poolerDir), s.pgCtlCmd.pgPort.Get())
```
rafael: nit - most places in the codebase we use `filepath.Join`. Just for consistency, we can do it that way.

go/cmd/pgctld/command/server.go
```diff
@@ -164,6 +164,28 @@ func (s *PgCtldServerCmd) runServer(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Register /ready probe: postgres socket accepting + gRPC accepting.
+	// Replication health is intentionally excluded (see pgctldReadyHandler).
```
rafael, line 168: nit - looks like the pgctldReadyHandler comment is stale

rafael (approved):
LGTM - left some very minor nit comments

PR #902, opened 2026-04-24
feat(multiorch,multipooler): graceful primary shutdown and Shutdown RPC

mkindahl:
On SIGTERM, a PRIMARY multipooler now signals PoolerType_STOPPING via the health stream so multiorch can orchestrate a controlled failover (PrimaryIsStoppingAnalyzer → ShutdownPrimaryAction → EmergencyDemote + AppointLeader) before the process exits. It waits up to 45 s for postgres to stop and falls back to a direct stop on timeout. REPLICA poolers stop postgres directly with no failover coordination.

A new Shutdown RPC on MultiPoolerManager triggers the same sequence programmatically by sending SIGTERM to the process, reusing the existing OnTermSync handler with no code duplication.

Also fix a port collision in the localprovisioner integration test: etcd was launching with --listen-metrics-urls on client-port+2, colliding with the MultiadminHTTPPort allocation. EtcdMetricsPort is now allocated independently via the port pool server.

Closes MUL-36

### DIFF
--- a/go/services/multiorch/recovery/shutdown_primary.go
+++ b/go/services/multiorch/recovery/shutdown_primary.go
@@ -1,0 +1,377 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package recovery
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"google.golang.org/protobuf/types/known/durationpb"
+
+	"github.com/multigres/multigres/go/common/eventlog"
+	"github.com/multigres/multigres/go/common/mterrors"
+	"github.com/multigres/multigres/go/common/topoclient"
+	commontypes "github.com/multigres/multigres/go/common/types"
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
+	multiorchpb "github.com/multigres/multigres/go/pb/multiorch"
+	multiorchdatapb "github.com/multigres/multigres/go/pb/multiorchdata"
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+)
+
+// shutdownPrimaryCallback wraps ShutdownPrimary with the signature expected by
+// analysis.ShutdownPrimaryFunc so it can be injected into the action factory without
+// exposing the full proto request type to the analysis sub-package.
+func (re *Engine) shutdownPrimaryCallback(
+	ctx context.Context,
+	primaryID *clustermetadatapb.ID,
+	database, tableGroup, shard string,
+) error {
+	_, err := re.ShutdownPrimary(ctx, &multiorchpb.ShutdownPrimaryRequest{
+		Database:              database,
+		TableGroup:            tableGroup,
+		Shard:                 shard,
+		PrimaryId:             primaryID,
+		DrainTimeout:          durationpb.New(10 * time.Second),
+		StandbyCatchupTimeout: durationpb.New(30 * time.Second),
+		Reason:                "SIGTERM",
+	})
+	return err
+}
+
+// isPrimaryCandidate returns true if the pooler type is currently reported to
+// be in PRIMARY or REPLICA state. Both types are valid candidates for promotion
+// to PRIMARY during the ShutdownPrimary flow, which excludes the old primary
+// from consideration (since the pooler type is STOPPING).
+//
+// We deliberately exclude UNKNOWN (or any other unknown types that we don't know
+// are valid) to be conservative in the face of missing or stale health data. If
+// a pooler has no health status, we don't want to risk electing it as the new
+// primary only to discover it's unhealthy when we try to promote it.
+func (re *Engine) isPrimaryCandidate(pooler *multiorchdatapb.PoolerHealthState) bool {
+	if pooler.MultiPooler == nil {
+		return false
+	}
+
+	switch pooler.GetStatus().GetPoolerType() {
+	case clustermetadatapb.PoolerType_PRIMARY, clustermetadatapb.PoolerType_REPLICA:
+		return true
+	case clustermetadatapb.PoolerType_STOPPING, clustermetadatapb.PoolerType_DRAINED:
+		return false
+	default:
+		re.logger.Warn("isPrimaryCandidate: unknown pooler type, skipping",
+			"pooler_id", topoclient.MultiPoolerIDString(pooler.MultiPooler.Id),
+			"pooler_type", pooler.GetStatus().GetPoolerType())
+		return false
+	}
+}
+
+// ShutdownPrimary orchestrates a graceful primary switchover for the given shard.
+//
+// Triggered when a primary multipooler signals PoolerType_STOPPING on the health
+// stream (e.g. on SIGTERM). The pooler has already stopped accepting writes by
+// the time this runs; EmergencyDemote drains remaining connections and stops
+// PostgreSQL. After stopping the primary, it is marked DRAINED in topology so
+// getpoolers never shows two PRIMARY nodes simultaneously and the node is
+// excluded from future elections.
+//
+// See the ShutdownPrimary RPC in multiorchservice.proto for the full sequence.
+func (re *Engine) ShutdownPrimary(ctx context.Context, req *multiorchpb.ShutdownPrimaryRequest) (_ *multiorchpb.ShutdownPrimaryResponse, retErr error) {
+	shardKey := &clustermetadatapb.ShardKey{
+		Database:   req.Database,
+		TableGroup: req.TableGroup,
+		Shard:      req.Shard,
+	}
+	primaryIDStr := topoclient.MultiPoolerIDString(req.PrimaryId)
+	start := time.Now()
+
+	re.logger.InfoContext(ctx, "ShutdownPrimary: starting graceful switchover",
+		"shard_key", commontypes.FormatShardKey(shardKey),
+		"primary", primaryIDStr,
+		"drain_timeout", req.DrainTimeout.AsDuration(),
+		"standby_catchup_timeout", req.StandbyCatchupTimeout.AsDuration(),
+		"reason", req.Reason)
+
+	// Step 1: Validate that the given pooler is currently PRIMARY or STOPPING.
+	primary, err := re.shutdownValidatePrimary(shardKey, req.PrimaryId)
+	if err != nil {
+		return nil, err
+	}
+
+	// Telemetry: record duration and emit demotion events.
+	eventlog.Emit(ctx, re.logger, eventlog.Started, eventlog.PrimaryDemotion{
+		NodeName: primaryIDStr,
+		Reason:   "shutdown",
+	})
+	defer func() {
+		duration := time.Since(start)
+		status := RecoveryActionStatusSuccess
+		if retErr != nil {
+			status = RecoveryActionStatusFailure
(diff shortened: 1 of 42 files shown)


go/services/multipooler/manager/rpc_shutdown.go
```diff
@@ -0,0 +1,40 @@
...
+
+// InitiateGracefulShutdown triggers the same graceful shutdown sequence as
+// receiving SIGTERM. The caller returns before the shutdown begins; the process
+// exits asynchronously after the graceful sequence completes.
+func (pm *MultiPoolerManager) InitiateGracefulShutdown(ctx context.Context, attr slog.Attr) {
+	pm.logger.InfoContext(ctx, "Shutdown RPC received. Initiating graceful shutdown", attr)
+	go func() {
+		time.Sleep(shutdownRPCDelay)
+		if err := syscall.Kill(os.Getpid(), syscall.SIGTERM); err != nil {
+			pm.logger.Error("Shutdown RPC: failed to send SIGTERM to self", "error", err)
```
cuongdo: There's no error propagation here. Does that mean that it's up to the initiator of this action to act if the graceful termination doesn't stop within an allotted time?

go/services/multipooler/manager/shutdown.go
```diff
@@ -0,0 +1,83 @@
...
+	"time"
+
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
+	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
+)
+
+// primaryShutdownSignalTimeout is how long we wait for multiorch to call
+// EmergencyDemote after we signal STOPPING. If it doesn't arrive in time we
+// stop postgres ourselves so the process always exits cleanly.
+const primaryShutdownSignalTimeout = 45 * time.Second
```
cuongdo: nit: we've tended to put constants in `go/common/constants`

proto/clustermetadata.proto
```diff
@@ -250,6 +250,11 @@ enum PoolerType {
 
   // DRAINED is used for poolers that are temporarily removed from serving traffic
   DRAINED = 3;
+
+  // STOPPING is used for a primary that is being gracefully shut down.
+  // The primary has been demoted and a new primary is being elected.
+  // Analyzers skip STOPPING poolers to prevent spurious recovery attempts.
```
cuongdo, line 273: It would be good to add comments that explain the difference between `DRAINED` and `STOPPING`

proto/multiorchservice.proto
```diff
@@ -49,6 +50,47 @@ service MultiOrchService {
...
+  //   1. Fences writes via EmergencyDemote. This ensures no new transactions
+  //      start on the primary after this point. The primary is still
+  //      responsible for allowing in-flight transactions to complete, but no
+  //      new ones will be accepted.
+  //   2. Waits for all standbys to replay to the primary's final LSN. This
+  //      ensures the new primary will have all transactions before promotion,
+  //      minimizing potential data loss and ensuring a smooth transition. If
+  //      the timeout expires, the switchover proceeds anyway to avoid hanging
+  //      indefinitely.
+  //   3. Clears synchronous_standby_names on the old primary. Since writes are
```
cuongdo: I think we also want to clear `archive_command` on the old primary. We don't want conflicting timelines to exist in the backup repo.

cuongdo (commented):
I took a pass at reviewing

go/services/multipooler/manager/shutdown.go
mkindahl: There were no suitable file for pure pooler constants so added one and moved it there.

go/services/multiorch/recovery/shutdown_primary.go
```diff
@@ -0,0 +1,337 @@
...
+			"ShutdownPrimary", "ShutdownPrimary",
+			status,
+			shardKey.Database, shardKey.Shard)
+	}()
+
+	// Step 3: Read the primary's current WAL LSN while it is still running.
+	// The gateway has already stopped routing new writes (STOPPING signal), so
+	// this LSN is a safe target for standby catchup. If the Status call fails
+	// we skip the standby wait and proceed directly to EmergencyDemote.
+	currentLSN := re.shutdownGetPrimaryLSN(ctx, primary.MultiPooler, primaryIDStr)
```
dweitzman: What would break if we just ran EmergencyDemote() with some kind of graceful shutdown timeout and then did a normal failover?

go/services/multipooler/manager/rpc_shutdown.go
mkindahl: Kill cannot really fail in this case, so this is just an extra guard to ensure that the process shuts down regardless.

mkindahl: I rewrote the code slightly, but the kill call is still there.

proto/clustermetadata.proto
mkindahl: The comment about DRAINED is actually wrong. It is used when the server is considered permanently out of commission, e.g., when `pg_rewind` fails. We should probably change the name to something like "DECOMMISSIONED" or "INOPERATIVE" (or even "KAPUT" or "OUT_OF_KILTER" if we want something less mainstream).

I updated the comment but kept the name.

go/services/multiorch/recovery/shutdown_primary.go
mkindahl: When running EmergencyDemote(), it does not wait for the standbys to catch up and rather "cut the ropes" directly. This ensure that we get the changes replayed on the standbys before doing that so that we are not missing any changes and can fail over to the new primary without data loss.

proto/multiorchservice.proto
mkindahl: Not sure that is necessary. We are setting `archive_mode` to `on`, so it means it is only active on the primary, not on the standby, and not during recovery either. We should not be able to get conflicting timelines in the backups.

go/services/multiorch/recovery/shutdown_primary.go
dweitzman: EmergencyDemote() does the restart in `fast` mode which I believe should ensure the shutdown checkpoint reaches quorum? And after that, I believe any replicas can still receive transactions by connecting to the former primary while it's a standby

So my understanding is:
- EmergencyDemote does a graceful shutdown
- With no waiting after that, orch should be able to discover a most advanced new primary to promote during revocation

go/services/multiorch/recovery/shutdown_primary.go
```diff
@@ -0,0 +1,376 @@
...
+	// Step 6: Elect a new primary from the cohort.
+	re.logger.InfoContext(ctx, "ShutdownPrimary: electing new primary",
+		"shard_key", shardKey.String(),
+		"cohort_size", len(filteredCohort))
+	if err := re.coordinator.AppointLeader(ctx, shardKey.Shard, filteredCohort, shardKey.Database, "ShutdownPrimary"); err != nil {
+		return nil, mterrors.Wrap(err, "AppointLeader failed")
+	}
+
+	// Step 7: Find the newly elected primary and return its ID.
+	newPrimary, err := re.shutdownFindNewPrimary(ctx, shardKey, primaryIDStr)
```
cuongdo, line 200: Is there a chance that the health streaming hasn't had a chance to propagate information about the new primary by the time `shutdownFindNewPrimary` is called here?

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
```diff
@@ -57,6 +57,11 @@ func (a *PrimaryIsDeadAnalyzer) Analyze(sa *ShardAnalysis) ([]types.Problem, err
 		return nil, nil
 	}
 
+	// Primary is gracefully stopping — ShutdownPrimary handles the election.
+	if sa.PrimaryIsStopping {
```
cuongdo: Should this be the following?

```go
if sa.PrimaryIsStopping && sa.PrimaryPoolerReachable {
	return nil, nil
}
```

That second clause may make a difference if the primary:
1. receives a SIGTERM
2. sets its STOPPING status
3. crashes (e.g. SIGKILL, OOM) before graceful shutdown can occur

cuongdo (commented):
Did another pass and left some comments. @rafael and/or @dweitzman should review since this touches consensus.

proto/clustermetadata.proto
mkindahl: I modified this and used STOPPING while the pooler is gracefully shutting down and DRAINED once it is finally shut down.

go/services/multiorch/recovery/analysis/primary_is_dead_analyzer.go
mkindahl: Good point. Added the extra condition so that we report an error if it was marked stopped but is not reachable.

go/services/multiorch/recovery/shutdown_primary.go
mkindahl: I'm a little unsure, but I changed it to use the wait version of the poll, so that should ensure that it waits for the status to be updated and sent back.

proto/multiorchservice.proto
```diff
@@ -127,3 +158,32 @@ message TriggerRecoveryNowResponse {
   // remaining_problem_codes lists problem codes still detected (e.g., "REPLICA_NOT_REPLICATING")
   repeated string remaining_problem_codes = 4;
 }
+
+message ShutdownPrimaryRequest {
+  // Shard to perform the switchover on
+  string database = 1;
```
dweitzman, line 164: You can use the new ShardKey! 🙂

proto/multiorchservice.proto
```diff
@@ -127,3 +158,32 @@ message TriggerRecoveryNowResponse {
...
+  // Required: must be a positive duration.
+  google.protobuf.Duration standby_catchup_timeout = 6;
+
+  // Human-readable reason for the shutdown (e.g. "SIGTERM", "operator request").
+  // Used for logging and audit trails only.
+  string reason = 7;
+}
+
+message ShutdownPrimaryResponse {
+  // The newly elected primary after the switchover.
```
dweitzman, line 183: Is there any plan for how to use the "new_primary_id" here? Maybe we can just remove if?

If there are plans to use the result, I feel like it'd make the most sense to introduce new message like a "leadership" and say that we always describe a leadership as a leader ID + a rule number

proto/clustermetadata.proto
```diff
@@ -261,8 +261,17 @@ enum PoolerType {
...
+  // be removed from the cluster. This is set when a pooler is permanently
+  // removed from consideration, for example, if pg_rewind fails and the server
+  // can no longer be part of the cluster. Analyzers skip DRAINED poolers to
+  // prevent spurious recovery attempts.
   DRAINED = 3;
+
+  // STOPPING is used for a primary that is in the process of being gracefully
+  // shut down. The primary has been demoted and a new primary is being elected.
+  // Analyzers skip STOPPING poolers to prevent spurious recovery attempts.
+  STOPPING = 4;
```
dweitzman, line 274: Just to save some context from discussion elsewhere, we'd talked about maybe this being some kind of a "lifecycle stage" rather than a pooler type?

cc @rafael

proto/multiorchservice.proto
```diff
@@ -49,6 +50,36 @@ service MultiOrchService {
...
+  //   5. Builds the election cohort from shard poolers, excluding STOPPING and
+  //      DRAINED poolers. This prevents the old primary from being re-elected
+  //      and prevents irreparable replicas from becoming candidates.
+  //   6. Elects a new primary via AppointLeader.
+  //   7. Polls shard poolers to confirm the new primary and returns its ID.
+  //
+  // Returns once a new primary has been elected. The caller can then shut down
+  // PostgreSQL cleanly. Errors are non-fatal: the caller should log and proceed
+  // with shutdown regardless.
+  rpc ShutdownPrimary(ShutdownPrimaryRequest) returns (ShutdownPrimaryResponse) {}
```
dweitzman, line 82: Here's my feeling on this (although I'm open other perspectives): this should not be a multiorch method

If someone wants to shutdown a primary, they'll either ask the primary directly or ask multiadmin service to ask the primary

The reason a gRPC service was added to multiorch was to help with endtoend tests. It's not intended to be an administrative service, and "I'm shutting down" seems like a decision a pooler can and sometimes must make independently without orch involvement, so it feels more consistent to stay that orch learns of that lifecycle state the same way it learns of any other lifecycle state, through health checks

mkindahl:
Closing this in favor of #969.

PR #903, opened 2026-04-24
refactor(consensus): derive primary_term from committed rule

rafael:
## Summary

- Removes the persisted `primary_term` field from `ConsensusTerm`, `PrimaryStatus`, and `StatusResponse`. Primary term is now derived from the committed rule at read time via `commonconsensus.PrimaryTerm(cs)`, which returns the rule's `coordinator_term` when the pooler holds the primary role per `IsPrimary(cs)` and 0 otherwise.
- Updates analyzers, recovery actions, and 4 end-to-end tests to read primary term through the helper instead of the removed proto field.

## Problem

Now that we have consensus_status rule, we are making that the canonical source of truth for everything consensus related. PrimaryTerm is not longer needed and can be derived from the consensus status. 

## Solution

`commonconsensus.PrimaryTerm(cs)` derives the term from the rule when the pooler is primary, returning 0 otherwise. Callers that previously read `ConsensusTerm.primary_term` or `PrimaryStatus.primary_term` now call this helper (or `primaryTermLocked` on the multipooler manager when an action lock is held for a consistent read).

In `BeginTerm`, standbys are built by filtering recruited poolers through `types.PrimaryNeedsReplacement` — the same predicate already used to skip resigned candidates in `selectCandidate`. Leaving the resigned pooler untouched lets the stale-primary analyzer detect the mismatched rule on the next pass and drive the proper recovery flow.

## Test plan

- I went through all the tests and updated them accordingly to the new changes. This is the part where I would like some extra eyes as these changes made some tests obsolete.

### DIFF
--- a/go/services/multiorch/consensus/leader_appointment.go
+++ b/go/services/multiorch/consensus/leader_appointment.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/multigres/multigres/go/common/mterrors"
 	"github.com/multigres/multigres/go/common/timeouts"
+	"github.com/multigres/multigres/go/services/multiorch/recovery/types"
 	"github.com/multigres/multigres/go/tools/pgutil"
 
 	commonconsensus "github.com/multigres/multigres/go/common/consensus"
@@ -95,12 +96,19 @@ func (c *Coordinator) BeginTerm(ctx context.Context, shardID string, cohort []*m
 		return nil, nil, 0, mterrors.Wrapf(err, "candidacy validation failed for shard %s", shardID)
 	}
 
-	// Separate candidate from standbys
+	// Resigned poolers are excluded: they may still carry a stale primary rule
+	// that needs pg_rewind, which the stale-primary analyzer owns.
 	var standbys []*multiorchdatapb.PoolerHealthState
 	for _, pooler := range recruitedPoolers {
-		if pooler.MultiPooler.Id.Name != candidate.MultiPooler.Id.Name {
-			standbys = append(standbys, pooler)
+		if pooler.MultiPooler.Id.Name == candidate.MultiPooler.Id.Name {
+			continue
 		}
+		if types.PrimaryNeedsReplacement(pooler) {
+			c.logger.InfoContext(ctx, "Skipping resigned pooler from standbys",
+				"pooler", pooler.MultiPooler.Id.Name)
+			continue
+		}
+		standbys = append(standbys, pooler)
 	}
 
 	return candidate, standbys, proposedTerm, nil
@@ -226,7 +234,7 @@ func (c *Coordinator) selectCandidate(ctx context.Context, recruited []recruitme
 		// unconditionally avoids confusing re-elections of a node that just
 		// stepped down. If all candidates are resigned the election is deferred
 		// until a non-resigned candidate is available.
-		if poolerRequestingDemotion(r.pooler) {
+		if types.PrimaryNeedsReplacement(r.pooler) {
 			c.logger.InfoContext(ctx, "Skipping resigned candidate during selection",
 				"pooler", r.pooler.MultiPooler.Id.Name)
 			continue
@@ -284,18 +292,6 @@ func (c *Coordinator) selectCandidate(ctx context.Context, recruited []recruitme
 type recruitmentResult struct {
 	pooler      *multiorchdatapb.PoolerHealthState
 	walPosition *consensusdatapb.WALPosition
-}
-
-// poolerRequestingDemotion reports whether the pooler's cached health state shows it has
-// voluntarily requested to be replaced (REQUESTING_DEMOTION signal at its current primary term).
-// This duplicates the logic in recovery/types.PrimaryNeedsReplacement to avoid a circular import
-// (recovery imports consensus; consensus cannot import recovery).
-func poolerRequestingDemotion(pooler *multiorchdatapb.PoolerHealthState) bool {
-	ls := pooler.GetConsensusStatus().GetAvailabilityStatus().GetLeadershipStatus()
-	return ls != nil &&
-		ls.Signal == clustermetadatapb.LeadershipSignal_LEADERSHIP_SIGNAL_REQUESTING_DEMOTION &&
-		ls.PrimaryTerm != 0 &&
-		ls.PrimaryTerm == pooler.GetConsensusStatus().GetPrimaryTerm()
 }
 
 // recruitNodes sends BeginTerm RPC to all poolers in parallel and returns those that accepted.
--- a/go/services/multiorch/consensus/leader_appointment.go
+++ b/go/services/multiorch/consensus/leader_appointment.go
@@ -98,6 +98,8 @@ func (c *Coordinator) BeginTerm(ctx context.Context, shardID string, cohort []*m
 
 	// Resigned poolers are excluded: they may still carry a stale primary rule
 	// that needs pg_rewind, which the stale-primary analyzer owns.
+	// TODO: once poolers self-rewind after emergency demotion, this exclusion
+	// becomes unnecessary — the resigned pooler can rejoin as a standby directly.
 	var standbys []*multiorchdatapb.PoolerHealthState
 	for _, pooler := range recruitedPoolers {
 		if pooler.MultiPooler.Id.Name == candidate.MultiPooler.Id.Name {
--- a/go/services/multiorch/consensus/leader_appointment_test.go
+++ b/go/services/multiorch/consensus/leader_appointment_test.go
@@ -682,12 +682,21 @@ func TestSelectCandidate(t *testing.T) {
 			logger:        logger,
 		}
 
+		resignedPoolerID := &clustermetadatapb.ID{Name: "mp1-resigned"}
 		resignedPooler := &multiorchdatapb.PoolerHealthState{
 			MultiPooler: &clustermetadatapb.MultiPooler{
-				Id: &clustermetadatapb.ID{Name: "mp1-resigned"},
+				Id: resignedPoolerID,
 			},
 			ConsensusStatus: &consensusdatapb.StatusResponse{
-				PrimaryTerm: 4,
+				ConsensusStatus: &clustermetadatapb.ConsensusStatus{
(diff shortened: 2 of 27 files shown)


go/services/multipooler/manager/manager.go
```diff
@@ -715,12 +716,17 @@ func (pm *MultiPoolerManager) checkAndSetReady() {
...
 		}
 
-		// Set initial primary observation from loaded consensus state.
-		if term, _ := pm.consensusState.GetInconsistentTerm(); term != nil && term.GetPrimaryTerm() > 0 {
-			pm.healthStreamer.UpdatePrimaryObservation(&poolerserver.PrimaryObservation{
-				PrimaryID:   pm.serviceID,
-				PrimaryTerm: term.GetPrimaryTerm(),
-			})
+		// Set initial primary observation from the highest known rule.
+		// TODO: This publishes PrimaryID as ourselves without first verifying that
```
dweitzman: I don't think this TODO is needed?...

I think this is on startup, so presumably no new WAL entries are flowing in yet and whatever `getInconsistentConsensusStatus()` returns is relatively accurate?

go/services/multipooler/manager/manager.go
```diff
@@ -1696,11 +1728,10 @@ func (pm *MultiPoolerManager) takeRemedialAction(ctx context.Context, action rem
...
 		// election. Use the primary_term (the term at which we were elected) since
 		// the coordinator uses this to decide whether the signal is still active.
-		// TODO: Once ConsensusStatus is populated in StatusResponse, use the
-		// consensus term from there for a more precise resignation signal.
-		if term, err := pm.consensusState.GetInconsistentTerm(); err == nil && term.GetPrimaryTerm() != 0 {
-			if err := pm.setResignedPrimaryAtTerm(ctx, term.GetPrimaryTerm()); err != nil {
+		if state.primaryTerm != 0 {
+			if err := pm.setResignedPrimaryAtTerm(ctx, state.primaryTerm); err != nil {
 				pm.logger.ErrorContext(ctx, "MonitorPostgres: failed to set resigned primary term", "error", err)
+				return
```
dweitzman, line 1733: This `return` statement looks like a change in behavior. Was there a reason for not wanting to make sure we also ensure the pooler type is set to replica?

My general feeling is still that pooler type doesn't mean anything, but I'm wondering if this was a specific motivation for not wanting to ensure it's updated

go/services/multiorch/recovery/actions/demote_stale_primary.go
```diff
@@ -203,11 +204,8 @@ func (a *DemoteStalePrimaryAction) findCorrectPrimary(shardKey commontypes.Shard
 		}
 
 		if poolerType == clustermetadatapb.PoolerType_PRIMARY {
-			// Get its PrimaryTerm (not consensus term)
-			var primaryTerm int64
-			if ct := pooler.GetStatus().GetConsensusTerm(); ct != nil {
-				primaryTerm = ct.PrimaryTerm
-			}
+			// Get its PrimaryTerm (not consensus term) from its current rule.
```
dweitzman: Nit: I feel like the wording of this comment could be misleading, since the primary term kind of is the consensus term, just zero'd out if this pooler is not the indicated primary

go/services/multiorch/consensus/leader_appointment_test.go
```diff
@@ -1253,6 +1280,68 @@ func TestBeginTerm(t *testing.T) {
 		require.Contains(t, err.Error(), "recruitment validation failed",
 			"should fail recruitment validation when only 1 node recruited but need 2")
 	})
+
+	t.Run("excludes resigned pooler from standbys so stale-primary recovery can handle it", func(t *testing.T) {
+		// A pooler that emergency-demoted (REQUESTING_DEMOTION signal with a
+		// matching primary term) must not land in the standbys list returned by
+		// BeginTerm. If it did, the standby-config step would call
+		// SetPrimaryConnInfo on it directly, bypassing the stale-primary flow
```
dweitzman, line 1288: Note: I'm hoping this will be a temporary state and poolers will start to self-rewind in the near future

go/services/multipooler/manager/manager_test.go
```diff
@@ -363,6 +363,8 @@ func TestValidateAndUpdateTerm(t *testing.T) {
 			mockQueryService := mock.NewQueryService()
 			mockQueryService.AddQueryPattern("SELECT pg_is_in_recovery", mock.MakeQueryResult([]string{"pg_is_in_recovery"}, [][]any{{"f"}}))
 			manager.qsc = &mockPoolerController{queryService: mockQueryService}
+			manager.rules = newRuleStore(logger, mockQueryService)
+			manager.rules = newRuleStore(logger, mockQueryService)
```
dweitzman: This has the same line twice in a row?...

go/services/multipooler/manager/rpc_consensus_test.go
```diff
@@ -38,6 +39,45 @@ import (
 	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
 )
 
+// observePositionRow builds a mock result row for the observePosition query
+// where the rule names primaryAppName (e.g. "zone1_stale-primary") as the
+// primary with the given coordinator term.
+func observePositionRow(primaryAppName string, coordinatorTerm int64) ([]string, [][]any) {
```
dweitzman, line 45: Would using `fakeRuleStore` make more sense here instead of mocking the SQL queries directly?

go/services/multipooler/manager/rpc_initialization_test.go
```diff
@@ -54,6 +55,10 @@ func NewTestMultiPoolerManager(t *testing.T) *MultiPoolerManager {
 	}
 	pm, err := NewMultiPoolerManager(slog.Default(), mp, &Config{})
 	require.NoError(t, err)
+	// Replace the rule store's backing query service with a mock — the real
+	// one requires a running postgres, and tests that exercise rule reads
+	// (via observePosition) crash on nil otherwise.
+	pm.rules = newRuleStore(pm.logger, mock.NewQueryService())
```
dweitzman: Similar comment: would `fakeRuleStore` make sense here?

dweitzman (approved):

go/services/multiorch/consensus/leader_appointment.go
```diff
@@ -95,12 +96,19 @@ func (c *Coordinator) BeginTerm(ctx context.Context, shardID string, cohort []*m
...
 	}
 
-	// Separate candidate from standbys
+	// Resigned poolers are excluded: they may still carry a stale primary rule
+	// that needs pg_rewind, which the stale-primary analyzer owns.
 	var standbys []*multiorchdatapb.PoolerHealthState
 	for _, pooler := range recruitedPoolers {
-		if pooler.MultiPooler.Id.Name != candidate.MultiPooler.Id.Name {
-			standbys = append(standbys, pooler)
+		if pooler.MultiPooler.Id.Name == candidate.MultiPooler.Id.Name {
```
rafael, line 105: This was a subtle bug: the resigned primary was being excluded from the candidate list but still added to the standby list. We'd call SetPrimaryConnInfo, it would start replicating from the new primary, and learn through replication that it was no longer primary, so DemoteStalePrimary never ran.

Prior to changes in this PR, this worked anyway because PrimaryTerm on disk made the node report as a stale primary, so the demotion path still fired. We want DemoteStalePrimary to run when the node was emergency demoted.

This fix makes sure we exclude this node from the standby list. I think we should also add a guardrail in SetPrimaryConnInfo: an emergency-demoted primary should reject that RPC until the pg_rewind flag is cleared. That can be a follow-up PR.

go/services/multiorch/consensus/leader_appointment.go
```diff
@@ -286,18 +294,6 @@ type recruitmentResult struct {
 	walPosition *consensusdatapb.WALPosition
 }
 
-// poolerRequestingDemotion reports whether the pooler's cached health state shows it has
-// voluntarily requested to be replaced (REQUESTING_DEMOTION signal at its current primary term).
-// This duplicates the logic in recovery/types.PrimaryNeedsReplacement to avoid a circular import
```
rafael, line 291: This circular depedency no longer happens, we can remove this

go/services/multiorch/consensus/leader_appointment_test.go
```diff
@@ -1253,6 +1280,68 @@ func TestBeginTerm(t *testing.T) {
 		require.Contains(t, err.Error(), "recruitment validation failed",
 			"should fail recruitment validation when only 1 node recruited but need 2")
 	})
+
+	t.Run("excludes resigned pooler from standbys so stale-primary recovery can handle it", func(t *testing.T) {
```
rafael, line 1284: test for the bug fix

go/services/multipooler/manager/manager.go
rafael: This was a judgement call. My reasoning was that if we fail to advertise the resigned primary, we should not advertise the change type yet.  It seemed off and I was thinking that better to return early and a later run of the monitor would either succeed at both or keep re-trying.

go/services/multiorch/consensus/leader_appointment_test.go
rafael: Added a comment for this

go/services/multipooler/manager/rpc_consensus_test.go
rafael: I changed it for rpc_initialization, but this one seems harder to swap, because the mocks are more involved. It's not obvious I can achieve the same with the fakeRulesStore yet.

go/services/multipooler/manager/rpc_consensus_test.go
```diff
@@ -981,6 +1020,13 @@ func TestDemoteStalePrimary_UpdatesConsensusTerm(t *testing.T) {
 				}
 			},
 			setupQueryMock: func(m *mock.QueryService) {
+				// observePosition reports this pooler as the current primary, so
```
rafael, line 1027: I think this is wrong. I need to double check this test.

PR #907, opened 2026-04-24
feat(consensus): add BuildSafeProposal library for Recruit protocol

dweitzman:
Adds `go/common/consensus` library that validates Recruit responses and builds a safe promotion proposal for the Recruit→Propose→Inform protocol.


rafael (approved):
I reviewed this on Friday and took one more quick pass today. The API makes sense and the overall direction. I did not look in detail to the tests. 

Looks good to merge and we could iterate forward.

PR #908, opened 2026-04-24
feat(multigateway,multiadmin-web): move per-gateway debug pages to multiadmin-web

haritabh17:
## Summary

- Follow-up to #874. Backs out the multigateway's server-rendered `/debug/queries` and `/debug/consolidator` HTML pages and re-implements them as React drilldowns inside multiadmin-web. The multigateway now emits raw data only — multiadmin-web owns rendering.
- Adds a new `MultiGatewayManager` gRPC service mirroring the `MultiPoolerManager` pattern: `GetQueryRegistry` and `GetConsolidatorStats` returning structured proto snapshots. Multiadmin proxies them via grpc-gateway at `/api/v1/gateways/{cell}/{name}/queries` and `.../consolidator`.
- The "Per-Query Performance" page in the frontend matches the screenshot reference shared in the [thread](https://supabase.slack.com/archives/C0916SZD47P/p1776870574219509?thread_ts=1776859311.043469): one row per fingerprint, columns for query / % of runtime / count / total time / p50 / p99 / rows, with per-cell sparklines stacked below the scalar value.

### Backend

The `queryregistry` package picks up two new pieces of state needed for the new UI:

- A bucketed duration histogram per fingerprint, with bucket boundaries matching the existing OTel `mg.gateway.query.duration` metric so registry-derived percentiles agree with Prometheus-derived ones. `Snapshot` now carries `P50Duration` and `P99Duration` derived from the histogram.
- A rolling sampler goroutine that captures one trend sample per fingerprint every `SampleInterval` (default 10s) into a fixed-capacity ring (default 60 samples → 10-minute window). Five metrics are tracked: `calls/s`, `total time ms/s`, `p50 ms`, `p99 ms`, `rows/s`. Trends are exposed on `Snapshot` as oldest-to-newest float slices, ready to feed a sparkline.

Sampler is opt-in — setting `SampleInterval` or `TrendWindowSamples` to zero disables sampling and keeps the existing memory profile.

The histogram + trend-ring state grows the per-fingerprint `CachedSize` from ~520 B to ~3 KB at default `TrendWindowSamples=60` (5 rings × 60 × 8 B = 2400 B charged eagerly so theine's admission policy isn't fooled). To keep the admitted-fingerprint count roughly equivalent to the pre-trend layout, the default `--query-metrics-memory` budget is bumped from 2 MiB to 8 MiB. Operators who want the old footprint can set `--query-metrics-memory=2097152` (or disable trend sampling entirely).

A new `MultiGatewayManager` gRPC service on the multigateway exposes those snapshots:

- `proto/multigatewaymanagerdata.proto` defines `QueryStatSnapshot` (with the trend arrays), `QueryRegistrySnapshot`, and `ConsolidatorStats`.
- `proto/multigatewaymanagerservice.proto` defines `GetQueryRegistry` + `GetConsolidatorStats`.
- `go/services/multigateway/manager.go` wires the registry + handler into the RPC handlers; the new service is registered alongside the existing `CancelManager` on the gateway's gRPC server.

Multiadmin gains two proxy methods on `MultiAdminService` (`GetGatewayQueries`, `GetGatewayConsolidator`) annotated with `google.api.http` so grpc-gateway exposes them at `/api/v1/gateways/{cell}/{name}/queries` and `.../consolidator`. The implementation resolves the gateway via topology, dials its gRPC port through a pluggable dialer (kept lightweight for v1; the existing `rpcclient` is multipooler-only), and forwards the response. The pluggable dialer makes the unit tests reproducible without standing up real services.

The HTML pages are gone: `handleQueriesDebug` / `handleConsolidatorDebug`, their view-model types, the `queries_debug.html` and `consolidator_debug.html` templates, and `status_test.go` are all removed. The gateway's index page no longer lists "Queries" or "Consolidator" links. Index page, `/ready`, `/live`, and `/config` stay — those are out of scope per the [scope confirmation in the thread](https://supabase.slack.com/archives/C0916SZD47P/p1777062700696329?thread_ts=1776859311.043469).

### Frontend (`web/multiadmin/`)

Two new drilldown routes nested under the existing multigateways list:

- `/dashboard/multigateways/[cell]/[name]/queries` — Per-Query Performance table fed by `getGatewayQueries`. Columns: query (wrapped SQL), % of runtime (gradient bar gauge), count/s, total time (s), p50 ms, p99 ms, rows/s, last seen. Each numeric cell stacks the scalar value above a sparkline rendered from the matching trend slice in the response. Free-text filter on normalized SQL; auto-refresh every 10s.
- `/dashboard/multigateways/[cell]/[name]/consolidator` — prepared-statement consolidator stats fed by `getGatewayConsolidator`. Summary cards (unique statements / total references / connection count) and a filtered (name, query, refs) table.

Supporting pieces: typed methods on `MultiAdminClient` mirroring `getPoolerStatus`, request/response types matching the new proto, a small `Sparkline` recharts wrapper sized for table cells, and a new "Diagnostics" column on the multigateways table with `Queries` + `Consolidator` drilldown links per row.

### Tests

- `histogram_test.go` covers the bucket layout, percentile interpolation, and end-to-end percentile derivation through `Record`.
- `trend_test.go` covers the ring buffer (push, snapshot, wrap) and the sampler's delta computation across multiple manual `sampleAll` calls.
- `server_gateway_test.go` stands up an in-process gRPC server with a fake `MultiGatewayManager` and a custom dialer to exercise the multiadmin proxy methods end-to-end: validation, happy-path proxy, and upstream error mapping.

## Dashboard
Screenshot 2026-04-25 at 02 10 28

Screenshot 2026-04-25 at 02 33 47

## Test plan

- [x] `go test ./go/services/multigateway/... ./go/services/multiadmin/...` clean
- [x] `go build ./...` clean
- [x] `npx tsc --noEmit` clean from `web/multiadmin`
- [x] `npx next build` clean from `web/multiadmin`
- [x] Manual smoke test: bring up a fresh local cluster, run a pgbench workload, navigate to `/dashboard/multigateways/<cell>/<name>/queries` and verify the table populates with sparklines that grow over the 10-minute window.
- [x] Manual smoke test: open `/dashboard/multigateways/<cell>/<name>/consolidator` while a client holds prepared statements; verify summary cards + statement table.
- [ ] Confirm `/debug/queries` and `/debug/consolidator` are 404 on the gateway after redeploy.


go/services/multiadmin/server.go
```diff
@@ -416,3 +427,88 @@ func (s *MultiAdminServer) SetPostgresRestartsEnabled(ctx context.Context, req *
...
+	if err != nil {
+		s.logger.ErrorContext(ctx, "Failed to get query registry from gateway", "gateway_id", req.GatewayId, "error", err)
+		return nil, status.Errorf(codes.Unavailable, "failed to get query registry from gateway: %v", err)
+	}
+	return &multiadminpb.GetGatewayQueriesResponse{Snapshot: resp.Snapshot}, nil
+}
+
+// GetGatewayConsolidator proxies a prepared-statement consolidator snapshot
+// from the target multigateway's MultiGatewayManager.GetConsolidatorStats RPC.
+func (s *MultiAdminServer) GetGatewayConsolidator(ctx context.Context, req *multiadminpb.GetGatewayConsolidatorRequest) (*multiadminpb.GetGatewayConsolidatorResponse, error) {
```
GuptaManan100, line 458: Thank you for including this too!

go/services/multigateway/handler/queryregistry/registry.go
```diff
@@ -56,17 +58,42 @@ type QueryStats struct {
...
+// strings and the trend ring buffers (5 rings × cap × 8 bytes/float64),
+// charged eagerly at admit time so theine's admission policy isn't fooled
+// into over-admitting once the sampler warms up.
 func (s *QueryStats) CachedSize(_ bool) int64 {
-	return 256 + int64(len(s.fingerprint)) + int64(len(s.normalizedSQL))
+	base := int64(384) + int64(len(s.fingerprint)) + int64(len(s.normalizedSQL))
+	if s.trends.initialized {
+		base += int64(5) * int64(len(s.trends.callRate.buf)) * 8
+	}
+	return base
```
GuptaManan100, line 90: The cached size is increasing significantly, so at default MaxMemoryBytes=2MiB, the registry now admits ~5× fewer fingerprints than before. Worth either bumping the default budget or calling this out in the changelog.

go/services/multigateway/handler/queryregistry/registry.go
```diff
@@ -56,17 +58,42 @@ type QueryStats struct {
...
+
+	// Trend bookkeeping — guarded by trendMu. Last-* fields hold the
+	// counter / histogram values at the previous sample so the sampler can
+	// compute per-interval deltas. Without this delta-of-histograms the
+	// trend percentile would be cumulative-since-admission and would
+	// flat-line under sustained load.
+	trendMu                   sync.Mutex
+	trends                    trendBuffers
+	lastSampleCalls           uint64
+	lastSampleTotalDurationNs int64
```
GuptaManan100, line 74: In the future we could implement ddsketch or some other algorithm that are able to give p99 and p95 values not exactly but within a better range, rather than assuming a normal distribution.

GuptaManan100 (approved):
Rest LGTM

go/services/multigateway/handler/queryregistry/registry.go
haritabh17: Good catch. Bumped the default --query-metrics-memory from 2 MiB → 8 MiB

PR #909, opened 2026-04-24
feat(multigateway): enforce rolcanlogin and rolvaliduntil during SCRAM auth

haritabh17:
Fixes MUL-385.

## Summary

- Multigateway's SCRAM path now fetches `rolcanlogin` and the `rolvaliduntil` predicate alongside `rolpassword` from `pg_authid`, rejecting NOLOGIN and password-expired roles with the same error strings and SQLSTATEs native PostgreSQL emits.
- Previously, a role with `rolvaliduntil` in the past authenticated successfully through multigateway (real bug, since downstream trust-over-Unix-socket bypasses PG's own password-expiry check). NOLOGIN was only enforced incidentally via the lazy per-user pool connection, with a mis-timed error.

## Problem

The admin query at `pools/admin/admin_conn.go` returned only `rolpassword`, so multigateway's SCRAM verification had no visibility into `rolcanlogin` or `rolvaliduntil`. Two consequences:

- **Expired passwords authenticated.** `VALID UNTIL` in the past only affects password-based auth. The multipooler → PG connection uses the local trust entry in `pg_hba.conf`, so PG never checks expiry downstream either. Result: a role disabled via `ALTER ROLE ... VALID UNTIL '<past>'` still logged in through multigateway, diverging from native PG.
- **`NOLOGIN` rejection had wrong timing.** It worked only because the per-user pool connection to PG failed on first query, so the client saw `AuthenticationOk` followed by an error on the first statement — not an authentication-time rejection.

## Solution

The admin query now evaluates both predicates server-side:

```sql
SELECT rolpassword, rolcanlogin,
       (rolvaliduntil IS NULL OR rolvaliduntil > now()) AS password_valid
FROM pg_catalog.pg_authid WHERE rolname = $1 LIMIT 1
```

Two new sentinel errors (`admin.ErrLoginDisabled`, `admin.ErrPasswordExpired`) propagate through distinct gRPC codes (`PermissionDenied`, `Unauthenticated`) to the gateway, which maps them to new scram sentinels and emits the native-PG FATAL errors:

- `rolcanlogin=false` → SQLSTATE **28000**, `role "X" is not permitted to log in`
- `rolvaliduntil` past → SQLSTATE **28P01**, `password authentication failed for user "X"` (opaque, matching PG's convention)

Precedence matches PostgreSQL's `CheckPasswordAuth` (`src/backend/libpq/auth.c`): login-disabled is checked before password validity, so operators see consistent error codes regardless of which mechanism disabled the role.

## Validation

End-to-end matrix against a local cluster, comparing multigateway against direct PG over TCP (`scram-sha-256`):

| # | Scenario | Multigateway | Direct PG | Match |
|---|---|---|---|---|
| 1 | Normal login | succeeds | succeeds | ✅ |
| 2 | `NOLOGIN` role | `role "X" is not permitted to log in` | same | ✅ |
| 3 | `VALID UNTIL` past | `password authentication failed for user "X"` | same | ✅ |
| 4 | `VALID UNTIL` future | succeeds | succeeds | ✅ |
| 5 | No password set | `password authentication failed for user "X"` | same | ✅ |
| 6 | Wrong password | `password authentication failed for user "X"` | same | ✅ |
| 7 | Nonexistent role | `password authentication failed for user "X"` | same | ✅ |

Byte-for-byte identical in all 7 cases.

## Test plan

- [x] `go test ./go/...` — all 95 packages pass
- [x] New unit tests cover `rolcanlogin=false`, `rolvaliduntil` past, login-disabled-beats-expired precedence, and the gateway-side gRPC→scram mappings
- [x] `golangci-lint` clean on all touched packages
- [x] Manual end-to-end validation against a fresh cluster (matrix above)


go/common/pgprotocol/scram/authenticator.go
```diff
@@ -170,8 +181,12 @@ func (a *ScramAuthenticator) HandleClientFirst(ctx context.Context, clientFirstM
...
-		if errors.Is(err, ErrUserNotFound) {
-			return "", ErrUserNotFound
+		// Forward sentinel errors unwrapped so the caller can match on them
+		// without peeling wrapping layers.
+		if errors.Is(err, ErrUserNotFound) ||
+			errors.Is(err, ErrLoginDisabled) ||
+			errors.Is(err, ErrPasswordExpired) {
+			return "", err
 		}
 		return "", fmt.Errorf("auth: failed to get password hash: %w", err)
```
GuptaManan100: Errors.Is unwraps errors, so if we want we can still just call fmt.Errorf and it would work fine, without needing this if branch.

GuptaManan100 (approved):
LGTM!

go/common/pgprotocol/scram/authenticator.go
haritabh17: Ack updated.

PR #910, opened 2026-04-25
docs(skills): drive pgbench across all multigateways by default

haritabh17:
## Summary

- Updates the `mt-local-cluster` skill so "push traffic" / "generate load" distributes pgbench across **all** multigateway instances using libpq `load_balance_hosts=random` (PG 16+), instead of single-targeting one gateway via `-h localhost -p 15432`.
- Adds a short rationale: each gateway maintains independent in-memory state (query registry, consolidator, connection pool), so single-port runs leave the other gateways' diagnostic pages empty and hide per-instance bugs.
- Single-port use stays allowed when the user is reproducing an instance-specific bug, but the skill now requires the choice be explicit rather than the default.

## Description

Previous example single-targeted zone1:

```bash
PGPASSWORD=postgres pgbench -h localhost -p 15432 -U postgres -c 4 -j 2 -T 300 -P 5 postgres
```

New default uses libpq multi-host conninfo:

```bash
CONNSTR='host=localhost,localhost,localhost port=15432,15433,15434 dbname=postgres user=postgres password=postgres load_balance_hosts=random'
PGPASSWORD=postgres pgbench -i "$CONNSTR"
pgbench -c 9 -j 3 -T 300 -P 5 "$CONNSTR"
```

`-c 9` is a multiple of the gateway count for even sticky-connection distribution; `load_balance_hosts=random` is supported on libpq 16+. The "intent mapping" section is also updated so future "push traffic" invocations produce the multi-host conninfo by default.

## Test plan

- [ ] Run pgbench against the new conninfo on a 3-cell local cluster and verify each multigateway shows non-zero traffic via `curl /api/v1/gateways/{cell}/{name}/queries` (the new endpoint introduced in #908).

### DIFF
--- a/.claude/skills/mt-local-cluster/SKILL.md
+++ b/.claude/skills/mt-local-cluster/SKILL.md
@@ -222,13 +222,29 @@ demo/local/multigres-with-otel.sh cluster start --config-path <config-path>
 
 Run pgbench init synchronously first, then start the workload in a **background Agent** so the user sees the output when it completes (do NOT use `run_in_background` on Bash — that hides output).
 
-```bash
-# Step 1: Init (synchronous)
-PGPASSWORD=postgres pgbench -h localhost -p 15432 -U postgres -i postgres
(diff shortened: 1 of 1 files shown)


rafael (approved):
LGTM

PR #911, opened 2026-04-25
fix(multigateway): honor SET LOCAL on gateway-managed GUCs

haritabh17:
## Summary

- `SET LOCAL statement_timeout` was silently ignored: the value never reached multigateway's GUC state because the planner short-circuited LOCAL to a backend pass-through *before* the gateway-managed dispatch ran. SHOW kept returning the session-level value.
- Adds a third tier (`localValue` / `isLocalSet`) to `GatewayManagedVariable[T]` so the gateway tracks PostgreSQL's local/session/default GUC semantics for variables it owns. Effective priority: local > session > default.
- Wires `ResetAllLocalGUCs()` into `executeCommit` and `executeRollback` so LOCAL overrides clear at the transaction boundary, matching PG.

## Problem

Closes [MUL-390](https://linear.app/supabase/issue/MUL-390).

`statement_timeout` is intercepted by multigateway as a gateway-managed GUC: `SET` updates `MultiGatewayConnectionState`, `SHOW` reads it back. The planner's LOCAL handling looked like this:

```go
if stmt.IsLocal {
    return p.planDefault(...)             // pass-through to backend
}
if isGatewayManagedVariable(stmt.Name) {
    // gateway-managed dispatch (only reached for non-LOCAL)
}
```

So `SET LOCAL statement_timeout = '40ms'` was forwarded to the backend without updating gateway state, while `SHOW statement_timeout` continued reading from the (un-updated) gateway state. Subsequent SHOW returned the session-level default rather than the LOCAL override.

Other `SET LOCAL` values (`DateStyle`, `role`, `search_path`, `request.jwt.claims`) were unaffected — they're not gateway-managed, so the pass-through path produced correct round-trip behavior. Verified manually.

PostgREST issues `SET LOCAL statement_timeout` per request to enforce per-request query timeouts, so this bug specifically broke timeout enforcement for any deployment fronting PostgREST through multigateway.

## Solution

Three coordinated changes:

1. `GatewayManagedVariable[T]` gains `localValue` / `isLocalSet` plus `SetLocal` / `ResetLocal`, and `GetEffective()` resolves `local > session > default`. RESET clears only the session layer (matches PG: RESET inside a transaction does not undo a prior SET LOCAL).
2. The planner reorders its check: `isGatewayManagedVariable` runs *before* the IsLocal pass-through, threading `stmt.IsLocal` into the gateway-managed primitive. Non-gateway-managed SET LOCAL keeps passing through to the backend.
3. `transaction_primitive.go` calls `state.ResetAllLocalGUCs()` from both `executeCommit` and `executeRollback`, at the same boundary as the existing LISTEN/NOTIFY commit hook.

`SET LOCAL` outside a transaction is the no-op PG warns on; multigateway's existing autocommit handling already triggers an immediate commit, so the LOCAL value is cleared right after the statement that set it — matching PG semantics.

### Known limitation

`ROLLBACK TO SAVEPOINT` does not restore the pre-savepoint LOCAL value because no per-savepoint snapshot is kept. Tracked separately as [MUL-392](https://linear.app/supabase/issue/MUL-392) (savepoint-aware GUC stack), which covers both gateway-managed and backend-managed variables.

## Verification

End-to-end against a local cluster, all six acceptance scenarios pass:

| Scenario | Result |
|---|---|
| `BEGIN; SET LOCAL statement_timeout='40ms'; SHOW` | `40ms` inside, default after COMMIT |
| `pg_sleep(1)` with `100ms` LOCAL | `ERROR: canceling statement due to statement timeout` |
| ROLLBACK clears LOCAL | reverts to session/default |
| Session SET + LOCAL coexistence | LOCAL wins inside; session restored after COMMIT |
| RESET inside txn does not undo LOCAL | matches PG |
| Backend-managed `SET LOCAL` (DateStyle, role) | regression-free, still passes through |

13 new unit tests in `handler/statement_timeout_test.go` cover the GUC tier semantics and the `MultiGatewayConnectionState` integration. All multigateway test packages green.

## Test plan

- [x] CI green on this PR.
- [ ] Manual repro on a fresh cluster: `BEGIN; SET LOCAL statement_timeout = '40ms'; SHOW statement_timeout; SELECT pg_sleep(1); COMMIT;` — SHOW returns `40ms` and pg_sleep is canceled.
- [ ] Apply the `Run Extended Query Serving Tests` label to confirm pgregress doesn't regress (statement_timeout is not directly exercised by the regression or isolation suites; this is a no-regression check only).

### DIFF
--- a/go/services/multigateway/engine/gateway_session_state.go
+++ b/go/services/multigateway/engine/gateway_session_state.go
@@ -36,6 +36,7 @@ type GatewaySessionState struct {
 	sql      string // Original SQL for debugging
 	variable string // Variable name (e.g., "statement_timeout")
 	isReset  bool   // true for RESET, false for SET
+	isLocal  bool   // true for SET LOCAL (transaction-scoped); ignored for RESET
 
 	// Typed fields for each gateway-managed variable.
 	// Only the field matching `variable` is used.
@@ -44,11 +45,14 @@ type GatewaySessionState struct {
 
 // NewStatementTimeoutSet creates a primitive that SETs `statement_timeout`.
 // The value is pre-parsed and stored in the appropriate typed field.
-func NewStatementTimeoutSet(sql string, statementTimeout time.Duration) *GatewaySessionState {
+// When isLocal is true, the value is treated as a transaction-local override
+// (SET LOCAL), cleared on COMMIT/ROLLBACK.
+func NewStatementTimeoutSet(sql string, statementTimeout time.Duration, isLocal bool) *GatewaySessionState {
 	return &GatewaySessionState{
 		sql:              sql,
 		variable:         "statement_timeout",
 		statementTimeout: statementTimeout,
+		isLocal:          isLocal,
 	}
 }
 
@@ -76,9 +80,15 @@ func (g *GatewaySessionState) StreamExecute(
 	}
 	switch g.variable {
 	case "statement_timeout":
-		if g.isReset {
+		switch {
+		case g.isReset:
+			// RESET clears the session-level override. The transaction-local
+			// override (if any) stays in effect — this matches PostgreSQL,
+			// where RESET inside a transaction doesn't undo a prior SET LOCAL.
 			state.ResetStatementTimeout()
-		} else {
+		case g.isLocal:
+			state.SetLocalStatementTimeout(g.statementTimeout)
+		default:
 			state.SetStatementTimeout(g.statementTimeout)
 		}
 	default:
--- a/go/services/multigateway/engine/gateway_session_state.go
+++ b/go/services/multigateway/engine/gateway_session_state.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/multigres/multigres/go/common/mterrors"
 	"github.com/multigres/multigres/go/common/parser/ast"
 	"github.com/multigres/multigres/go/common/pgprotocol/server"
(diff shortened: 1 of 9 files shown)


go/services/multigateway/engine/gateway_session_state.go
```diff
@@ -74,11 +95,42 @@ func (g *GatewaySessionState) StreamExecute(
...
+	// SET LOCAL outside a transaction block is a no-op in PostgreSQL — it emits
+	// a WARNING with SQLSTATE 25P01 (no_active_sql_transaction) and the value is
+	// discarded immediately by the implicit-autocommit boundary. We mirror that
+	// here: skip the state mutation and surface the WARNING as a NoticeResponse.
+	// Without this guard, isLocalSet would persist for the lifetime of the
+	// connection because no COMMIT/ROLLBACK ever fires to clear it.
+	if g.isLocal && !conn.IsInTransaction() {
+		warning := mterrors.NewPgError("WARNING", "25P01",
+			"SET LOCAL can only be used in transaction blocks", "")
+		warning.MessageType = 'N'
```
GuptaManan100: Let's create a new function called `mterrors.NewPgNotice` instead of creating an error and flipping the message type.

GuptaManan100 (approved):
Rest LGTM!

go/services/multigateway/engine/gateway_session_state.go
haritabh17: Ack, updating

PR #912, opened 2026-04-25
fix(pgctld): retry crash recovery during orphan-cleanup window

haritabh17:
## Summary

- After a postmaster crash (`kill -9` of a single PID, OOM, segfault), orphaned worker processes hold the SHM segment for ~1–5s while they detect parent death; during that window `postgres --single` fails with the same "lock file `postmaster.pid` already exists" error postgres reports when it is genuinely running.
- The previous `runCrashRecovery` matched that pattern and silently returned `nil`, so crash recovery never ran and `pg_control` stayed in `in production` — surfacing as intermittent CI failures in `TestPgRewind_AfterCrash` and a real production hazard for fast pg_rewind-driven failovers.
- Adds a bounded retry loop (10 × 500ms ≈ 5s) on the lock-file pattern before falling back to today's "already running" no-op. Non-lock errors still fail fast; ctx cancellation aborts the loop.
- Refactored into a testable inner loop with five new unit tests covering the orphan-cleanup race, retry exhaustion, hot path, non-retryable errors, and ctx cancellation.

## Problem

The test `TestPgRewind_AfterCrash` flaked intermittently on the Code Coverage / Subprocess coverage job (and on `main`). Tracing the failure: the test SIGKILLs the postmaster, waits for the postmaster PID to disappear, then calls `PgRewind`, which internally invokes `runCrashRecovery`. The orphan workers re-parented to init still hold the SHM segment, so `postgres --single` returns `FATAL: lock file "postmaster.pid" already exists`. The previous code matched that pattern and treated it as "postgres is already running" — a silent no-op. `pg_control` was never transitioned to `shut down in recovery`, and the assertion failed.

The same race is reachable in production: multiorch / operator tooling that invokes `pg_rewind` quickly after detecting a primary crash can hit the orphan-cleanup window, causing crash recovery to silently no-op and leaving the data directory inconsistent for the next caller.

A previous test-side fix (PR #911) that killed the postmaster's process group was reverted because it masked the real bug — single-PID SIGKILL is the realistic representation of OOM / segfault / `kill -9`, the most common production crash modes.

## Solution

`runCrashRecovery` now retries on the lock-file pattern with bounded backoff (10 × 500ms ≈ 5s) before falling back to the historical "postgres already running" branch. Within the retry window the orphans clean up, the lock releases, and recovery completes. If the lock genuinely doesn't clear after the window, behavior is identical to today (log + return `nil`).

Refactored into `runCrashRecoveryAttempts` (testable inner loop) plus `runSingleUserPostgres` so unit tests can simulate the orphan-cleanup window without a real postgres binary.

## Test plan

- [x] New unit tests pass (`TestRunCrashRecovery_*` — 5 cases covering retry-on-orphan-cleanup, retry-exhaustion-returns-nil, first-attempt-succeeds, non-lock-error-fails-fast, ctx-cancel-aborts).
- [x] Existing `TestPostgresAlreadyRunningPattern` still passes.
- [x] `TestPgRewind_AfterCrash` runs deterministically — 5/5 passes locally (3.4–5.0s each) without modifying the test's `kill -9 <postmaster_pid>` semantics.
- [x] Full `./go/cmd/pgctld/command/...` unit suite green.

Closes MUL-394.


go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -21,15 +21,28 @@ import (
...
+// walreceiver, bgwriter) keep the SHM segment attached for ~1-5s while they
+// detect parent death via PostmasterIsAlive() and exit. The same error is
+// reported during that window even though postgres is not actually running.
 var postgresAlreadyRunningPattern = regexp.MustCompile(`lock file ".*" already exists`)
 
+// Bounds the retry window used to wait out the orphan-cleanup race after a
+// postmaster crash. Suggested by MUL-394: ~5s covers the worst-case worker
+// PostmasterIsAlive() detection latency observed in practice.
+const (
+	crashRecoveryMaxAttempts = 10
```
cuongdo: These should probably go in `go/common/constants/postgres.go`

go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -66,39 +79,69 @@ func extractClusterState(output string) string {
...
+	ctx context.Context,
+	logger *slog.Logger,
+	run func(context.Context) ([]byte, error),
+	retryDelay time.Duration,
+) error {
+	logger.InfoContext(ctx, "Starting single-user crash recovery")
 
-	cmd.SetStdin(devNull)
+	var lastOutput string
+	for attempt := 1; attempt <= crashRecoveryMaxAttempts; attempt++ {
```
cuongdo: nit: it would be a little cleaner to use `Attempts` in the `tools.retry` package

cuongdo (approved):

go/cmd/pgctld/command/crash_recovery.go
haritabh17: Done — moved both

go/cmd/pgctld/command/crash_recovery.go
haritabh17: Ack. Switched to retry.New(...).Attempts(ctx)

PR #913, opened 2026-04-26
feat(multipooler): add Recruit RPC and ValidateRevocation helper

dweitzman:
Adds the Recruit RPC handler plus a ValidateRevocation() helper under `go/common/consensus`.


go/common/consensus/revocation.go
```diff
@@ -0,0 +1,100 @@
...
+//     strictly less than revoked_below_term. If the node has already applied
+//     WAL at or beyond the revocation term, the revocation has no authority
+//     over those writes.
+//
+//  2. Coordinator idempotency: if the node already accepted a revocation at
+//     the same term, the request must come from the same coordinator. A
+//     different coordinator at the same term is refused; the same coordinator
+//     is re-accepted (idempotent success).
+//
+//  3. Recruitment idempotency: when the coordinator and term match, the
```
rafael, line 45: Maybe you could expand on why this important, as this might not be obvious to future readers (i.e recovering from a coordinator forgetting it did something already )

go/common/consensus/revocation.go
```diff
@@ -0,0 +1,100 @@
...
+//     coordinator_initiated_at timestamp must also match. A differing
+//     timestamp means a distinct recruitment round at the same term number,
+//     which is treated as a conflict.
+//
+// A nil status or nil current_position means there is no WAL position to
```
rafael: under which scenarios we can a no WAL position? 

I can't think of any case outside bootstrap. In bootstrap we are not revoking. So it seems like you should always have a WAL with non nil stauts?

go/common/consensus/revocation.go
```diff
@@ -0,0 +1,100 @@
...
+			storedCoord := topoclient.ClusterIDString(stored.GetAcceptedCoordinatorId())
+			reqCoord := topoclient.ClusterIDString(revocation.GetAcceptedCoordinatorId())
+			if storedCoord != reqCoord {
+				return fmt.Errorf(
+					"cannot accept revocation: already accepted term %d from coordinator %s, requested by %s",
+					storedTerm, storedCoord, reqCoord,
+				)
+			}
+			// Same coordinator, same term: verify the recruitment round matches.
+			if !proto.Equal(stored.GetCoordinatorInitiatedAt(), revocation.GetCoordinatorInitiatedAt()) {
```
rafael, line 106: nice

go/common/consensus/revocation_test.go
```diff
@@ -0,0 +1,199 @@
...
+			},
+			revocation: &clustermetadatapb.TermRevocation{
+				RevokedBelowTerm:       5,
+				AcceptedCoordinatorId:  coordA,
+				CoordinatorInitiatedAt: ts2,
+			},
+			wantErr: "different coordinator_initiated_at",
+		},
+		{
+			name: "SameTerm_BothNilTimestamp_Idempotent",
```
rafael: should we guard against the timestamp not being nil? The contract enforces timestamp being present?

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -404,6 +405,154 @@ func (pm *MultiPoolerManager) clearResignedPrimaryAtTerm(ctx context.Context) er
...
+
+	// Step 3: Re-check with the stable (post-stop) status.
+	// This catches the rare race where a WAL rule entry arrived between the
+	// sanity check and actually stopping replication.
+	stableStatus, err := pm.getConsensusStatus(ctx)
+	if err != nil {
+		eventlog.Emit(ctx, pm.logger, eventlog.Failed, termEvent, "error", err)
+		return nil, mterrors.Wrap(err, "failed to read stable status after stopping replication")
+	}
+	if err := consensus.ValidateRevocation(stableStatus, revocation); err != nil {
```
rafael: I wonder if to start with it's simpler just to error here and make sure that re-recruit would succeed.  Specially for the primary. I get nervous with the re-promote.

rafael (approved):
Few small comments. Overall, LGTM

go/common/consensus/revocation.go
dweitzman: Changed to require a non-nil WAL position

go/common/consensus/revocation.go
dweitzman: Done

go/common/consensus/revocation_test.go
dweitzman: Done

go/services/multipooler/manager/rpc_consensus.go
dweitzman: Removed re-promotion for now with a comment left behind. Kept re-following the primary for replicas

PR #915, opened 2026-04-27
feat(benchmarking): make pgbench runner capture CPU/memory profiles and pick targets

GuptaManan100:
## Summary

- Make `TestPgBench` configurable: select any subset of `postgres,multigateway,pgbouncer` via `PGBENCH_TARGETS`, filter `simple`/`extended` via `PGBENCH_PROTOCOLS`, skip churn via `PGBENCH_NO_CHURN`.
- Capture CPU profiles from multigateway + primary multipooler during each scenario when `CAPTURE_PPROF=1`, and heap / allocs / goroutine snapshots before+after each scenario when `CAPTURE_HEAP=1`.
- Always sample per-process CPU usage (postgres process tree, multigateway, primary multipooler) for every scenario × target run and persist as JSON.
- Optionally bump postgres `max_connections` and restart pgctld via `PGBENCH_PG_MAX_CONNECTIONS` so we can drive direct-postgres or pgbouncer at high client counts without hitting the default 60-connection cap.

## Description

The existing pgbench suite ran a fixed cross-product of scenarios against postgres + multigateway + pgbouncer with no profiling hooks, which made it hard to drill into where multigateway's CPU goes. This PR adds the knobs needed to turn the suite into a usable perf workbench locally without changing what CI runs by default.

New env vars (all default to today's behavior except `PGBENCH_TARGETS`, which now defaults to `multigateway` so a quick local run only spins up what's actually under measurement):

| Variable | Default | Purpose |
| --- | --- | --- |
| `PGBENCH_TARGETS` | `multigateway` | subset of `postgres,multigateway,pgbouncer` |
| `PGBENCH_PROTOCOLS` | `simple,extended` | restrict to one protocol |
| `PGBENCH_NO_CHURN` | unset | skip the `-C` connection-churn variants |
| `PGBENCH_PG_MAX_CONNECTIONS` | unset | bump postgres `max_connections` and restart pgctld |
| `CAPTURE_PPROF` | unset | fetch `/debug/pprof/profile` from multigateway + primary multipooler during each scenario |
| `CAPTURE_HEAP` | unset | fetch `/debug/pprof/{heap,allocs,goroutine}` before+after each scenario |

New supporting files:

- `cpu_sampler.go` — `ps`-based CPU sampler. Postgres backends are enumerated each tick from `postmaster.pid` + `pgrep -P`, so the sampler is resilient to backend churn. multigateway and multipooler use static-pid enumerators.
- `profile_capture.go` — small HTTP fetcher for the four pprof endpoints we care about (CPU, heap, allocs, goroutine).
- `pg_config.go` — appends a `max_connections` override into each pgctld's `postgresql.auto.conf` and restarts via the pgctld gRPC `Restart` RPC. Disables the multipooler monitor's auto-restart for the duration so it doesn't race.

`pgbench.md` is updated with the new envs and the additional output paths.

Smoke-tested locally with all features on (`PGBENCH_TARGETS=postgres,multigateway,pgbouncer`, both protocols, sustained+churn, `CAPTURE_PPROF=1 CAPTURE_HEAP=1 PGBENCH_PG_MAX_CONNECTIONS=300`, 5s × 1,5 clients) — all 24 scenarios PASS, every artifact produced.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,4 @@ demo/k8s/data/
 # Generated PR description
 pr-description.md
 
+docs/query_serving/benchmarking/results/*
--- a/docs/query_serving/benchmarking/pgbench.md
+++ b/docs/query_serving/benchmarking/pgbench.md
@@ -100,11 +100,17 @@ the test is skipped.
 
 ### Environment variables
 
-| Variable           | Default   | Description                                                                  |
-| ------------------ | --------- | ---------------------------------------------------------------------------- |
-| `RUN_BENCHMARKS`   | (unset)   | Set to `1` to enable benchmark tests. Matches the "Run Benchmarks" PR label. |
-| `PGBENCH_DURATION` | `30`      | Seconds per scenario                                                         |
-| `PGBENCH_CLIENTS`  | `1,10,50` | Comma-separated client counts                                                |
+| Variable                     | Default           | Description                                                                                      |
+| ---------------------------- | ----------------- | ------------------------------------------------------------------------------------------------ |
+| `RUN_BENCHMARKS`             | (unset)           | Set to `1` to enable benchmark tests. Matches the "Run Benchmarks" PR label.                     |
+| `PGBENCH_DURATION`           | `30`              | Seconds per scenario                                                                             |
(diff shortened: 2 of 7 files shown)


Verolop (approved):
💖

PR #916, opened 2026-04-27
refactor(consensus): move sync replication config onto DurabilityPolicy

rafael:
## Description

The motivation is locality: each policy already owns its quorum semantics (achievability, recruitment intersection), so it should also own the leader-side Postgres config those semantics imply. Pulling the function up onto the interface removes the `policy.QuorumType` branch in the old free function and makes adding a third policy a matter of implementing one method. In the future this would also enable us to have more flexibility into how this configuration gets built for more complex durability policies. 

`LeaderDurabilityPostgresConfig` carries only the durability-meaningful outputs — `SyncCommit`, `SyncMethod`, `NumSync`, and the eligible standby IDs. `

The standby list semantics are preserved: the full cohort (including the candidate) appears in `SyncStandbyIDs` for `AT_LEAST_N`, and `MULTI_CELL_AT_LEAST_N` excludes same-cell members (which incidentally drops the candidate). `num_sync = N-1` continues to account for the primary's local write.

Tests migrated from the old `sync_replication_test.go` into `policy_at_least_n_test.go` and `policy_multi_cell_test.go` (the natural homes given each test exercises a single policy). The `mustPolicy` helper that other multiorch tests depend on moved into `leader_appointment_test.go`. One assertion in `leader_appointment_test.go` was tightened: it had been verifying that `ReloadConfig=true` was set on the wire request — that's now a literal at the call site, so the assertion is dropped at the policy level and verified by the integration-style test that builds the full `PromoteRequest`.

## Next Steps 
- Once we start using the Proposal type, `LeaderDurabilityPostgresConfig` can be refactored to take a proposal instead of candidate/stanbdy ids. 

## Changes
- Moved `BuildSyncReplicationConfig` to be part of consensus common durability policy. 
- Introduced `LeaderDurabilityPostgresConfig` as the policy's return type so the policy layer doesn't leak `multipoolermanagerdata.ConfigureSynchronousReplicationRequest` (an RPC shape with fields like `ReloadConfig` that aren't durability-meaningful). The proto is assembled at the call site in `EstablishLeadership`.
- `EstablishLeadership` now takes the typed `commonconsensus.DurabilityPolicy` directly; the redundant `NewPolicyFromProto` round-trip inside it is gone. `AsyncFallback` now lives on the Go policy structs and is populated in `NewPolicyFromProto`.

## Test plan

- Unit tests are passing and audited that relevant tests were migrated to the new package
- Integration tests continue to pass


go/common/consensus/durability.go
```diff
@@ -79,13 +115,19 @@ func NewPolicyFromProto(policy *clustermetadatapb.DurabilityPolicy) (DurabilityP
 		if policy.RequiredCount < 1 {
 			return nil, fmt.Errorf("AT_LEAST_N requires RequiredCount >= 1, got %d", policy.RequiredCount)
 		}
-		return AtLeastNPolicy{N: int(policy.RequiredCount)}, nil
+		return AtLeastNPolicy{
+			N:             int(policy.RequiredCount),
+			AsyncFallback: policy.AsyncFallback,
```
rafael: We are not using this AsyncFallback and seems dangerous. I wonder if we should remove this. Could be a foot-gun.

go/common/consensus/durability.go
```diff
@@ -62,10 +64,44 @@ type DurabilityPolicy interface {
...
+	// impossible.
+	//
+	// A nil result means the policy does not require leader-side sync
+	// enforcement — async replication is sufficient, either because the
+	// policy is trivially satisfied (e.g., RequiredCount==1) or because the
+	// configured AsyncFallback is ALLOW and no eligible standbys remain.
+	BuildLeaderDurabilityPostgresConfig(
+		logger *slog.Logger,
+		cohort []*clustermetadatapb.ID,
+		candidate *clustermetadatapb.ID,
```
dweitzman: I feel like "primary" is a more accurate name. At the point we're setting GUC it's because we're trying to configure a pooler that's acting as a primary, and whether that's because it's a candidate for a proposal or a decided leader feels beyond the scope for this method.

go/common/consensus/durability.go
dweitzman: I'm 100% fine with removing it

go/services/multiorch/consensus/leader_appointment.go
```diff
@@ -504,11 +504,28 @@ func (c *Coordinator) EstablishLeadership(
...
+			cohortIDs = append(cohortIDs, p.MultiPooler.Id)
+		}
+	}
+	leaderCfg, err := policy.BuildLeaderDurabilityPostgresConfig(c.logger, cohortIDs, candidate.MultiPooler.Id)
 	if err != nil {
 		return mterrors.Wrap(err, "failed to build synchronous replication config")
 	}
+	var syncConfig *multipoolermanagerdatapb.ConfigureSynchronousReplicationRequest
+	if leaderCfg != nil {
+		syncConfig = &multipoolermanagerdatapb.ConfigureSynchronousReplicationRequest{
```
dweitzman: I think the other code's comments were saying that if leaderCfg is `nil` it means we still do `ConfigureSynchronousReplicationRequest` but configure an async policy?

dweitzman (approved):

PR #917, opened 2026-04-27
perf(pgwire): in-place packet encoding on read and write paths

GuptaManan100:
## Summary

- Rewrites the pgwire read and write paths on both sides of the proxy (`go/common/pgprotocol/server/` for multigateway, `go/common/pgprotocol/client/` for multipooler) to encode/decode packets in place against `bufio.Reader`/`bufio.Writer`'s internal buffers, eliminating per-message allocations and the multi-`bufio.Write` pattern.
- Local microbenchmarks show **−86% time geomean on the server write path**, **−93% on the client write path**, **−75% on the server read path**, and **−22% on the client read path**, with allocations dropping from 4–742 to 0 across most hot writers and 1 alloc/op (the body buffer) remaining on the client read.
- Consolidates onto a single encoding API (`startPacket` / `writePacket` / `writeXxxAt` encoders) across both sides — removes the older `writeMessage` / `writeMessageNoFlush` / `writeByte` / `writeUint32` `*Conn` methods and the production `MessageWriter` type (kept only as a test fixture).
- Adds a new architecture doc at `docs/query_serving/pgwire_io_path.md` covering the design, the two non-obvious decisions (encoding into `bufferedWriter.AvailableBuffer()`, holding `bufMu` across encoding to defend against the async notification pusher), the server/client threading differences, the per-connection memory model, the slow-path implicit-flush behavior, and the read-path fixes.

### Write path

The previous write path used a sequence of small `bufio.Writer.Write` calls per protocol field — a 1-column DataRow was 6+ separate writes, and a typical pgbench TPC-B response was ~28 small writes plus 4 mutex acquisitions per query. We replaced this with in-place encoding into the `bufio.Writer`'s `AvailableBuffer()` (Go 1.18+): a body-size pre-pass reserves space, the `writeXxxAt` helpers fill in fields directly into the buffered writer's internal storage, and a single `bufferedWriter.Write` commits the packet (a self-copy on the fast path that just advances the cursor). The slow path (oversize packets, no buffered writer attached) borrows from a `bufpool.Pool` — server-side per-listener, client-side package-level since clients have no listener context.

`bufMu` now spans the `startPacket → encode → writePacket` window on the server side so the async notification pusher can't interleave bytes inside a packet body. The client side has no equivalent pusher, so the high-level operation's existing `bufMu` is sufficient and `startPacket` doesn't relock. Client startup/SCRAM helpers (`sendStartupMessage`, `writeSSLRequest`, `sendClientFirst`, `sendClientFinal`) run during single-threaded connection setup; the invariant is now documented at each call site.

### Read path

Three orthogonal fixes, applied independently and benchmarked between:

1. **`Peek` / `Discard` for type and length** instead of `io.ReadFull(io.Reader, []byte)`. The interface dispatch in `io.ReadFull` forced the stack-local slice header to the heap on every call; using `bufio.Reader`'s concrete-typed `Peek` and `Discard` avoids the escape entirely. Drops 2 allocs/op.
2. **Server-side body bufpool with `inboundPoolBuf` field on `Conn`** to track the pool's `*[]byte`. The previous `returnReadBuffer(buf []byte)` shape took `&buf` of a parameter, which forced the slice header to the heap on every call (same trap we hit on the write side). The parameterless `returnReadBuffer()` reads the field instead. Drops 1 alloc/op. `handleDescribe`/`handleClose` were also routed through the pooled `readMessageBody` for consistency.
3. **`MessageReader` returns by value** rather than `*MessageReader`. Returning `&local` from `NewMessageReader` was a textbook escape; returning the struct lets it live on the caller's stack. Methods keep pointer receivers (Go auto-addresses). Helpers that walk the SAME reader as the caller take `*MessageReader` to avoid silent cursor-loss bugs from by-value parameter copying — documented in `NewMessageReader`'s comment.

Also pulled in two latent bug fixes: `handleDescribe`/`handleClose` were using `bufferedReader.Read(nameBuf)` (which can short-read) — switched to `io.ReadFull` (now subsumed by the `readMessageBody` refactor). `Conn.Close()` defensively calls `returnReadBuffer()` so a handler panic between `readMessageBody` and the deferred return doesn't leak a pool buffer.

### Why the client read path keeps `make([]byte, length)` per body

`parseDataRow` returns `sqltypes.Value` slices that alias INTO the body buffer, and rows accumulate across multiple `readMessage` calls before the user callback fires. Pooling the body would require copying every column value out before recycling, which is strictly more allocations than the current `make` per body on typical workloads. Vitess (`~/vitess/go/mysql/conn.go`) reaches the same conclusion and ships a two-tier API (pool for command packets where parsing copies values out, fresh-alloc for streaming result rows where Values alias). We documented the asymmetry in `client/packet.go`'s `readMessageBody` comment.

### Microbenchmark numbers (Apple M4, parallel, b.RunParallel)

Server write path:
- WriteDataRow_4col: 22.7 ns / 7 allocs → 3.5 ns / 0 allocs (6.5×)
- WriteCommandComplete: 11.9 ns / 4 → 1.7 ns / 0 (7.0×)
- PgbenchResponse_4col: 158.8 ns / 49 allocs → 12.6 ns / 0 (12.6×)
- geomean: −86%

Client write path:
- ClientWriteParse: 48.1 ns / 2 → 2.0 ns / 0 (24×)
- ClientWriteBind_4param: 50.6 ns / 2 → 4.4 ns / 0 (11.5×)
- ClientPgbenchPipeline (Parse+Bind+Describe+Execute+Sync+flush): 200.6 ns / 9 → 8.6 ns / 0 (23×)
- geomean: −93%

Server read path:
- ReadMessage_Sync: 14.7 ns / 32 B / 3 → 1.2 ns / 0 / 0 (12×)
- ReadMessage_Bind_4param: 20.9 ns / 32 B / 3 → 5.6 ns / 0 / 0 (3.7×)
- geomean: −75%

Client read path:
- ClientReadMessage_DataRow_4col: 20.7 ns / 68 B / 2 → 15.7 ns / 64 B / 1
- geomean: −22%

### Files

Major changes: `go/common/pgprotocol/server/{packet,query,conn,startup}.go`, `go/common/pgprotocol/client/{packet,query,extended,conn,startup,scram}.go`. New benchmark files: `*_bench_test.go` and `*_read_bench_test.go` on both sides. New tests: `TestWritePacketSizeMismatchPanics` and `TestReturnReadBufferIsIdempotent` in `server/packet_test.go`. New doc: `docs/query_serving/pgwire_io_path.md`. Test-only `MessageWriter` moved to `test_helpers_test.go` on both sides.

## Test plan

- [x] `go test ./go/common/pgprotocol/...` (passes locally)
- [x] `go test -race ./go/common/pgprotocol/...` (passes locally)
- [x] Local benchmark suites (`go test -bench '^Benchmark(Write|Read|Pgbench|ClientWrite|ClientRead)' ./go/common/pgprotocol/...`) confirm the numbers above


go/common/pgprotocol/server/startup.go
```diff
@@ -49,7 +49,7 @@ func (c *Conn) readAndDispatchStartup() error {
 	if err != nil {
 		return fmt.Errorf("failed to read startup packet: %w", err)
 	}
-	defer c.returnReadBuffer(buf)
+	defer c.returnReadBuffer()
```
haritabh17, line 52: Looks like this may leak one 16 KB pool buffer per SSL- or GSS-negotiated connection??

go/common/pgprotocol/server/startup.go
haritabh17, line 514: While at it, lets cover this too - Every other readMessageBody caller in the server package pairs with defer c.returnReadBuffer(). This one seems to be missing.

go/common/pgprotocol/server/packet.go
```diff
@@ -24,70 +24,92 @@ import (
...
-// Returns a buffer from the pool that must be returned using returnReadBuffer.
+// Returns a slice that must be released by calling returnReadBuffer.
+//
+// On the pool path, the underlying *[]byte is stashed in
+// c.inboundPoolBuf so returnReadBuffer can return it without taking
+// the address of a stack-local slice (which would force the slice
+// header to the heap on every call). On the make-fallback path
+// (no listener), there's nothing to recycle and inboundPoolBuf
+// stays nil.
 func (c *Conn) readMessageBody(length int) ([]byte, error) {
```
haritabh17, line 74: Now we are unconditionally overwriting `inboundPoolBuf` without first checking/returning any existing buffer. This means a single missed returnReadBuffer call can orphan the previous buffer permanently. A defensive returnReadBuffer at the start of readMessageBody would make the design more robust against pairing errors, though it would mask bugs.

haritabh17 (changes requested):

go/common/pgprotocol/server/startup.go
GuptaManan100: Confirmed, good catch. The leak path is `readAndDispatchStartup` recursing through `handleSSLRequest` / `handleGSSENCRequest` to read the next startup packet: the recursive `readMessageBody` call overwrote `inboundPoolBuf` without releasing the outer one, so the outer `defer returnReadBuffer()` no-oped against the wrong pointer and the original buffer leaked.  Fixed at the source by adding a defensive `c.returnReadBuffer()` at the top of `readMessageBody` in `server/packet.go` — any stale buffer is returned to the pool before a new one is grabbed. This also addresses your packet.go:69 comment about unconditional overwrites. Added `TestReadMessageBodyReleasesStaleInbound` in `server/packet_test.go` to lock down the behavior, and the existing SSL/GSS tests (`TestSSLRequest`, `TestGSSENCRequest`, `TestGSSENCRequest_ThenSSLRequest_WithTLS`, etc.) all still pass.

go/common/pgprotocol/server/startup.go
GuptaManan100: Confirmed, `readSASLInitialResponse` was missing the pair, so SCRAM auth leaked one body buffer per connection. Added `defer c.returnReadBuffer()` immediately after the `readMessageBody` call. The body is only read through a `MessageReader` and then `string(data)` (which copies), so it's safe to return the buffer at function return. Added `TestReadSASLInitialResponseReleasesBuffer` in `server/packet_test.go` that asserts `inboundPoolBuf` is nil after the call returns.

go/common/pgprotocol/server/packet.go
GuptaManan100: Took the defensive route. The SSL/GSS recursive-dispatch case (your startup.go:52 comment) is real production behavior, not a bug — by the time the recursion reads the next packet, the outer caller has already extracted everything it needs from the first body. The defensive `c.returnReadBuffer()` at the top of `readMessageBody` makes that flow safe without callers having to remember to release before recursing.

haritabh17 (approved):
🚀

PR #918, opened 2026-04-28
feat(benchmarking): add sysbench oltp_point_select harness

GuptaManan100:
## Summary

- Adds `TestSysBench`, a manual perf harness that drives sysbench `oltp_point_select` against multigateway (and optionally postgres/pgbouncer) over the pgsql driver.
- Reuses the shared cluster setup and reporting types from `TestPgBench`; emits per-scenario CPU stats, optional CPU pprofs, and a sysbench-shaped markdown report.
- Runs both `prepared` (`--db-ps-mode=auto`) and `simple` (`--db-ps-mode=disable`) modes so parser-per-query cost can be isolated from the rest of the proxy/pooler path.

## Description

`TestSysBench` lives next to `TestPgBench` in `go/test/endtoend/queryserving/benchmarking` and follows the same skip-by-default convention: it requires `RUN_BENCHMARKS=1`, and self-skips with an actionable message if `sysbench` is missing or was built without the pgsql driver.

`ScenarioResult` and `EnvironmentInfo` in `pgbench_runner.go` grew a few `omitempty` fields (`PSMode`, `SysBenchVersion`, `MultigresVersion`, `Arch`, `PlanCacheBytes`) so each tool's `results.json` only carries fields relevant to that tool. Reporting is split: `WriteSysBenchMarkdownReport` renders the sysbench-shaped table (per-scenario rows grouped by `ps_mode`) rather than overloading the pgbench renderer, since the scenario shapes don't line up.

`SYSBENCH.md` documents the env knobs (`SYSBENCH_CLIENTS`, `SYSBENCH_DURATION`, `SYSBENCH_PS_MODES`, `SYSBENCH_TARGETS`, etc.), the output bundle layout, the loud-failure conditions (zero TPS, missing summary lines, empty pprof), and known caveats around sub-millisecond p99 reporting.

### DIFF
--- a/go/test/endtoend/queryserving/benchmarking/pgbench_runner.go
+++ b/go/test/endtoend/queryserving/benchmarking/pgbench_runner.go
@@ -45,20 +45,22 @@ type ScenarioConfig struct {
 
 // ScenarioResult holds computed metrics from a pgbench or sysbench run.
 //
-// Protocol/ConnChurn are pgbench-only fields; PSMode is sysbench-only.
-// They're emitted as `omitempty` so each tool's results.json contains only
-// the fields relevant to that tool.
+// Protocol/ConnChurn/LatencyP50 are pgbench-specific but kept non-omitempty
+// so existing downstream tooling (e.g. the weekly pgbench workflow that
+// filters with `select(.connection_churn == false)`) keeps working — their
+// zero values for sysbench rows are harmless. PSMode is sysbench-only and
+// stays `omitempty` so it doesn't pollute pgbench output.
 type ScenarioResult struct {
 	Target       string  `json:"target"`
 	Scenario     string  `json:"scenario"`
 	TPS          float64 `json:"tps"`
 	LatencyAvg   float64 `json:"latency_avg_ms"`
-	LatencyP50   float64 `json:"latency_p50_ms,omitempty"`
+	LatencyP50   float64 `json:"latency_p50_ms"`
 	LatencyP99   float64 `json:"latency_p99_ms"`
 	Clients      int     `json:"clients"`
 	Duration     int     `json:"duration_seconds"`
-	Protocol     string  `json:"protocol,omitempty"`
-	ConnChurn    bool    `json:"connection_churn,omitempty"`
+	Protocol     string  `json:"protocol"`
+	ConnChurn    bool    `json:"connection_churn"`
 	PSMode       string  `json:"ps_mode,omitempty"`
 	Transactions int     `json:"transactions"`
 }
--- a/go/test/endtoend/queryserving/benchmarking/pgbench_runner.go
+++ b/go/test/endtoend/queryserving/benchmarking/pgbench_runner.go
@@ -43,34 +43,44 @@ type ScenarioConfig struct {
 	ConnChurn bool   // -C flag (new connection per transaction)
 }
 
-// ScenarioResult holds computed metrics from a pgbench run.
+// ScenarioResult holds computed metrics from a pgbench or sysbench run.
+//
+// Protocol/ConnChurn are pgbench-only fields; PSMode is sysbench-only.
+// They're emitted as `omitempty` so each tool's results.json contains only
+// the fields relevant to that tool.
 type ScenarioResult struct {
 	Target       string  `json:"target"`
 	Scenario     string  `json:"scenario"`
 	TPS          float64 `json:"tps"`
 	LatencyAvg   float64 `json:"latency_avg_ms"`
-	LatencyP50   float64 `json:"latency_p50_ms"`
+	LatencyP50   float64 `json:"latency_p50_ms,omitempty"`
 	LatencyP99   float64 `json:"latency_p99_ms"`
 	Clients      int     `json:"clients"`
 	Duration     int     `json:"duration_seconds"`
-	Protocol     string  `json:"protocol"`
-	ConnChurn    bool    `json:"connection_churn"`
+	Protocol     string  `json:"protocol,omitempty"`
+	ConnChurn    bool    `json:"connection_churn,omitempty"`
+	PSMode       string  `json:"ps_mode,omitempty"`
 	Transactions int     `json:"transactions"`
 }
 
 // BenchmarkReport is the top-level report containing all results.
 type BenchmarkReport struct {
 	Timestamp   string           `json:"timestamp"`
+	LoadTool    string           `json:"load_tool,omitempty"` // "pgbench" or "sysbench"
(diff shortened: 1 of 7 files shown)


go/test/endtoend/queryserving/benchmarking/pgbench_runner.go
```diff
@@ -43,34 +43,44 @@ type ScenarioConfig struct {
...
 	TPS          float64 `json:"tps"`
 	LatencyAvg   float64 `json:"latency_avg_ms"`
-	LatencyP50   float64 `json:"latency_p50_ms"`
+	LatencyP50   float64 `json:"latency_p50_ms,omitempty"`
 	LatencyP99   float64 `json:"latency_p99_ms"`
 	Clients      int     `json:"clients"`
 	Duration     int     `json:"duration_seconds"`
-	Protocol     string  `json:"protocol"`
-	ConnChurn    bool    `json:"connection_churn"`
+	Protocol     string  `json:"protocol,omitempty"`
```
Verolop: I think we still need `connection_churn: false` to be present for pgbench results. The weekly pgbench workflow uses that field to pick out sustained runs (`select(.connection_churn == false)`), so omitting it here means those rows won’t make it into the Slack summary/overhead calculation. Could we leave this field non-omitempty for pgbench, or give sysbench its own result shape?
Ignore all of this if you have different plans :)

go/test/endtoend/queryserving/benchmarking/sysbench_runner.go
```diff
@@ -0,0 +1,425 @@
...
+		// 1.0.x defaults to 95 (which then ends up labelled "95th
+		// percentile:" in the output and fails our regex).
+		"--percentile=99",
+		"--db-ps-mode="+psFlag,
+		"run",
+	)
+
+	out, err := executil.Command(ctx, r.binary, args...).CombinedOutput()
+
+	logPath := filepath.Join(r.OutputDir, "logs", "sysbench-"+sc.Name+".txt")
```
Verolop: these artifacts probably need the target name in the path too. With `SYSBENCH_TARGETS=postgres,multigateway`, both targets write to the same `logs/sysbench-sustained_1c_prepared.txt`, so the second run hides the first one. Same issue for `cpu/<scenario>/processes.json`. Could we nest these by target, like the pgbench harness does?

Verolop (commented):
I left two comments around result/artifact compatibility: one pgbench JSON field that existing CI still depends on, and one multi-target artifact overwrite case. Both look pretty contained. If you have different plans, it's all safe to ignore though.

go/test/endtoend/queryserving/benchmarking/pgbench_runner.go
GuptaManan100: Good catch! fixed. I dropped `omitempty` from the three pgbench-shaped fields

go/test/endtoend/queryserving/benchmarking/sysbench_runner.go
GuptaManan100: Fixed. The sysbench bundle now nests by target everywhere there's a per-target artifact.

go/test/endtoend/queryserving/benchmarking/sysbench_test.go
```diff
@@ -0,0 +1,283 @@
...
+	captureHeap := os.Getenv("CAPTURE_HEAP") == "1"
+	if capturePprof {
+		t.Logf("CAPTURE_PPROF=1: CPU profiles will be captured during each scenario")
+	}
+
+	mgwPid := setup.Multigateway.Process.Process.Pid
+	pgPidEnum := postgresPidEnumerator(setup.GetPrimary(t).Pgctld.PoolerDir + "/pg_data")
+
+	var allResults []ScenarioResult
+	for i, scenario := range scenarios {
```
haritabh17: Quiesce currently runs only between scenarios, not between targets within scenario. Back-to-back targets can leak buffer cache state, replication lag, or background autovacuum from previous run, biasing later targets. Should we move quiesce inside target loop so each target starts from clean baseline?

go/test/endtoend/queryserving/benchmarking/sysbench_test.go
```diff
@@ -0,0 +1,283 @@
...
+	// Prepare runs once against postgres directly — sbtestN tables are
+	// shared across all targets because the same backend serves them all.
+	t.Run("prepare", func(t *testing.T) {
+		if err := runner.Prepare(ctx, t, pgTarget); err != nil {
+			t.Fatalf("sysbench prepare failed: %v", err)
+		}
+	})
+
+	capturePprof := os.Getenv("CAPTURE_PPROF") == "1"
+	captureHeap := os.Getenv("CAPTURE_HEAP") == "1"
```
haritabh17, line 130: pgbench path sets enable-log when CAPTURE_HEAP=1, sysbench path does not. If heap profile gets captured without matching workload log, diagnosing regression from heap alone is hard, should sysbench mirror pgbench here?

go/test/endtoend/queryserving/benchmarking/sysbench_test.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// Fail loudly if any scenario landed zero TPS — the parser explicitly
+	// guards against this, but defence in depth.
+	for _, r := range allResults {
+		if r.TPS <= 0 {
+			t.Fatalf("scenario %s/%s recorded zero TPS", r.Target, r.Scenario)
+		}
+	}
+
+	report := &BenchmarkReport{
```
haritabh17, line 171: LoadTool field set on pgbench result struct but not on sysbench one, so emitted JSON lacks tool provenance. Downstream parsers that key on tool tag would silently misclassify sysbench runs as default/pgbench. Worth setting symmetrically?

haritabh17 (approved):
Rest looks good 🚀

go/test/endtoend/queryserving/benchmarking/sysbench_test.go
GuptaManan100: Fixed

go/test/endtoend/queryserving/benchmarking/sysbench_test.go
GuptaManan100: Fixed

go/test/endtoend/queryserving/benchmarking/sysbench_test.go
GuptaManan100: Fixed

PR #919, opened 2026-04-28
perf(pgwire): coalesce writes into one syscall per pipelined batch

GuptaManan100:
## Summary

- Move pgwire write buffering out of every per-handler function and into the `serve()` loop, so a pipelined extended-protocol batch (Parse + Bind + Describe + Execute + Sync) flushes once instead of once per inbound message.
- Drops `syscall.write` count from ~4 per sysbench prepared transaction to 1, and `syscall.syscall` cum CPU on multigateway from ~58% to ~45% at 32 clients prepared.
- Sysbench `oltp_point_select` TPS improves **+46% to +76%** on prepared mode and **+14% to +53%** on simple mode (varies by client count).

## Problem

Each pgwire handler (`handleParse`, `handleBind`, `handleDescribe`, `handleExecute`, `handleClose`, `handleSync`) used to wrap its body in `c.startWriterBuffering()` / `defer c.endWriterBuffering()` and call `c.flush()` at the end. Since the pgwire client pipelines its commands and only blocks reading after `Sync`, this meant a single sysbench prepared transaction issued ~4 separate `write(2)` syscalls — one per inbound message — when one would have sufficed. The kernel round-trip cost showed up as the largest entry in the multigateway CPU profile.

## Solution

Hoist the buffering window to the `serve()` loop. The window opens lazily via `startWriterBuffering()` before each `handleMessage` (idempotent: no-op if already attached) and is released back to the writer pool only at flush boundaries — `Sync` and `Query` flush+release, `Flush` flushes without releasing, the error path flushes+releases. Per-handler bodies just write packets and return; the buffer accumulates across pipelined messages until `Sync`.

This matches what upstream PostgreSQL does (errors and intermediate replies stay in `PqSendBuffer` until `pq_flush` is called at Sync/Flush) and what vitess does for its mysql wire layer (`startWriterBuffering` / `endWriterBuffering` per command).

The client side (`go/common/pgprotocol/client/extended.go`) was already coalesced — `BindAndExecute` and friends already write all packets before flushing once — so no change needed there.

## Tests

`buffering_test.go` adds a `countingNetConn` that records every underlying `Write()` call and pins the new behavior with four tests:

- `TestBufferingWindow_OneSyscallPerBatch` — full Parse + Bind + Describe + Execute + Sync sequence produces exactly **1** socket Write.
- `TestBufferingWindow_PerHandlerFlushBaseline` — same fixtures but with explicit `flush()` between handlers produces **4** Writes (regression sentinel).
- `TestBufferingWindow_SimpleQueryFlushesOnce` — simple-`Q` reply is one Write.
- `TestBufferingWindow_ParseErrorDeferredToSync` — a Parse error stays buffered and ships together with the trailing Sync's `ReadyForQuery`, matching upstream PostgreSQL's deferred error semantics.

All `pgprotocol/...` unit and integration tests pass with `-race`.

## Benchmarks

End-to-end sysbench `oltp_point_select` against multigateway, 60s/scenario, 5s quiesce, `darwin/arm64`, GOMAXPROCS=10, M4. Same fixtures, back-to-back runs.

### TPS — `ps_mode=prepared` (the path the optimization targets)

| clients | BEFORE TPS | AFTER TPS | **Δ** |
|---:|---:|---:|---:|
| 1   | 1,586  | **2,788**  | **+76%** |
| 8   | 7,963  | **12,797** | **+61%** |
| 32  | 12,131 | **17,664** | **+46%** |

Average latency at 32 clients prepared: 2.64 ms → **1.81 ms** (-31%).

### TPS — `ps_mode=simple`

| clients | BEFORE TPS | AFTER TPS | **Δ** |
|---:|---:|---:|---:|
| 1   | 3,320  | **5,093**  | **+53%** |
| 8   | 14,082 | **19,608** | **+39%** |
| 32  | 17,194 | **19,646** | **+14%** |

### CPU profile (32 clients prepared, multigateway)

| metric | BEFORE | AFTER |
|---|---:|---:|
| `syscall.syscall` cum | **57.7%** | **45.1%** |
| `bufio.Writer.Flush` cum | 38.4% (per-handler `flush`) | 21.9% (only at `endWriterBuffering`) |

### Microbench (`BenchmarkExtendedReply_*`, M4)

Drives the same 5-message extended-protocol reply through the buffered-writer path with a counting `net.Conn`:

| shape | ns/op | **writes/op** |
|---|---:|---:|
| per-handler flush (pre-PR) | ~13 ns | **4** |
| coalesced (post-PR) | ~12 ns | **1** |

The wall-time delta on a discard sink is small because there's no real syscall cost; the **writes/op** drop is what translates into the kernel-side wins above on a real socket.

## Test plan

- [x] `go test -race ./go/common/pgprotocol/...` (unit + integration)
- [x] Sysbench `oltp_point_select` BEFORE vs AFTER, prepared and simple, 1/8/32 clients, with CPU pprof captured
- [x] Reviewer to verify on Linux loopback that the syscall-count reduction translates similarly (Darwin syscall costs are higher, so Linux gains may be smaller in % but still material in absolute CPU)


go/common/pgprotocol/server/conn.go
```diff
@@ -500,7 +536,10 @@ func (c *Conn) handleMessage(msgType byte) error {
 		return c.handleSync()
 
 	case protocol.MsgFlush:
-		return c.flush()
+		// Flush itself is handled by the serve() loop after this
+		// switch returns — see the post-handleMessage flush-boundary
+		// dispatch. Returning nil here avoids a redundant flush().
+		return nil
```
haritabh17: Returning nil without reading the mandatory 4-byte length field from the wire? The serve() loop reads the 'H' type byte via ReadMessageType, then dispatches to handleMessage which immediately returns nil. The 4 length bytes (0x00 0x00 0x00 0x04) remain unconsumed in the read buffer. Should we add something like:

```
	case protocol.MsgFlush:
		// Read and discard the Flush message length (Int32, always 4).
		if _, err := c.ReadMessageLength(); err != nil {
			return fmt.Errorf("failed to read Flush message length: %w", err)
		}
		// The actual flush is handled by the serve() loop after this
		// switch returns — see the post-handleMessage flush-boundary
		// dispatch. Returning nil here avoids a redundant flush().
		return nil
```

go/common/pgprotocol/server/conn.go
```diff
@@ -468,10 +488,26 @@ func (c *Conn) serve() error {
...
+		switch msgType {
+		case protocol.MsgSync, protocol.MsgQuery:
+			// End of a batch. Flush and release the writer back to
+			// the pool — the connection might idle here.
+			c.endWriterBuffering()
+		case protocol.MsgFlush:
+			// Client asked for an explicit mid-pipeline flush.
+			// Push bytes out but keep the writer attached: more
+			// pipelined messages typically follow.
+			_ = c.flush()
```
haritabh17: The serve() loop now uses c.endWriterBuffering() (which discards flush errors). In the old code, each handler returned c.flush() errors, which propagated to the serve() loop's error handler, triggering immediate connection closure and proper logging. Now, if a socket write fails during flush (e.g., broken pipe), looks like the error may get silently swallowed, and the loop would continue until the next ReadMessageType() fails — losing the original write-error context and delaying cleanup by one iteration?

---

Also the async notification goroutine `EnableAsyncNotifications ` writes to c.bufferedWriter while serve() writes to it without holding bufMu. This switch widens the window — the writer now stays attached across the whole pipelined batch instead of being detached per-handler.

go/common/pgprotocol/server/buffering_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+	assert.Equal(t, int64(4), netConn.writes.Load(),
+		"per-handler flush should land 4 socket Writes — got %d", netConn.writes.Load())
+}
+
+// TestBufferingWindow_SimpleQueryFlushesOnce drives the simple-Q path
+// (a single 'Q' message produces RowDescription + DataRow +
+// CommandComplete + ReadyForQuery) and asserts the reply is one
+// syscall.
+func TestBufferingWindow_SimpleQueryFlushesOnce(t *testing.T) {
```
haritabh17, line 222: Calls handleQuery() which invokes testHandler.HandleQuery. That mock returns nil without calling the callback, meaning no result messages (RowDescription, DataRow, CommandComplete) are written — only ReadyForQuery.

go/common/pgprotocol/server/conn.go
GuptaManan100: Good catch, confirmed that's a real desync. To be clear: this is a pre-existing bug on `main`, not something this PR introduced. The pre-PR `MsgFlush` case did `return c.flush()` which also never read the 4-byte length; the next iteration of `serve()` would have read the high-byte of the length as a bogus message-type. Glad we're fixing it now.

Why we never noticed: no client we exercise in e2e actually sends a standalone `H` (Flush) message — libpq doesn't, pgx doesn't except in `pgx.Batch` flows we don't test, and our integration suites are all libpq/pgx-driven. `grep -rn 'MsgFlush' go/test/ go/services/` returns zero hits for outbound sends. Code review caught what test coverage didn't.

go/common/pgprotocol/server/conn.go
GuptaManan100: Right. Old per-handler `return c.flush()` propagated the error; the new shape was discarding it at the dispatch. Fixed.

go/common/pgprotocol/server/buffering_test.go
GuptaManan100: Fixed

go/common/pgprotocol/server/conn.go
GuptaManan100: Good catch, but to clarify: this is a pre-existing race on main, not introduced by this PR. Synchronous write helpers (`writeRowDescription`, `writeDataRow`, …) call `getWriter()` once and then write the body without holding `bufMu`, while the async pusher does hold `bufMu` across its full notification packet. So both goroutines can call `bufferedWriter.Write` concurrently, that's the actual race, and it's there on main today.

I'll fix it properly in a follow-up, most likely by giving the synchronous helpers a `withBufferedWriter(func(*bufio.Writer) error) error` wrapper that holds bufMu across the whole encode, mirroring what the async pusher already does. Keeping it out of this PR so the syscall-coalescing change stays focused.

haritabh17 (approved):
🚀

PR #923, opened 2026-04-28
fix(multigateway): forward bindVars through Sequence to children

GuptaManan100:
## Summary

- Follow-up to #880 — we missed this while addressing the review comments on that PR.
- `Sequence.StreamExecute` was passing `nil` for `bindVars` to its children, so the trailing `Route` in a `[silent ApplySessionState..., Route]` plan couldn't reconstruct the original SQL from the normalized AST. The placeholder `$N` reached PG unbound and the backend errored with `there is no parameter $1`.
- Reproducer reported by @haritabh17 via psql: `SELECT set_config('work_mem','256MB',false), 42 AS num` — fails on simple protocol, works on plain PG.

## Problem

The cacheable simple-protocol path normalizes `42` to `$1` and stores the literal in `bindValues`. The planner emits a Sequence whose trailing Route holds the normalized SQL plus the `NormalizedAST` so it can substitute `bindVars` back at execution time. But Sequence.StreamExecute was iterating its children with `p.StreamExecute(ctx, exec, conn, state, nil, callback)` — dropping `bindVars`. Route then sent the normalized `$1` form to PG, which has no bind values for it.

The bug only fires when (a) the SELECT has at least one `set_config(..., false)` target and (b) some other target in the same SELECT is a literal that the normalizer would parameterize. A bare `SELECT set_config(...)` happens to work because there are no literals outside the (skipped) set_config subtree.

## Solution

Forward `bindVars` to every child in `Sequence.StreamExecute`. Silent `ApplySessionState` steps ignore them (their `VariableSetStmt` is fully synthesized at plan time from the literal set_config args), Route uses them.

Three layers of regression coverage:

- **Engine unit** — existing `TestSequence_PortalStreamExecute_*` covers the portal path; the StreamExecute path is now validated by the executor-level test below.
- **Executor unit** — new `TestStreamExecute_SetConfigWithSiblingLiteral` runs `SELECT set_config('work_mem','256MB',false), 42 AS num` through the executor and asserts the backend SQL contains `42` and not `$1`.
- **End-to-end** — new `SELECT set_config alongside sibling literal` subtest under `TestMultiGateway_SetConfigRoutedAsSET` runs the failing query through real psql against multigateway and verifies both the row data and the post-rotation tracker state (50 SHOW iterations).

### DIFF
--- a/go/services/multigateway/engine/sequence.go
+++ b/go/services/multigateway/engine/sequence.go
@@ -39,17 +39,25 @@ func NewSequence(primitives []Primitive) *Sequence {
 }
 
 // StreamExecute executes each primitive in order, stopping on first error.
+//
+// bindVars are forwarded to every child. The current Sequence shape used by
+// planSelectStmt is [silent ApplySessionState..., Route]: the silent steps
+// ignore bindVars (their VariableSetStmt is fully synthesized at plan time
+// from the literal set_config args), while the trailing Route uses bindVars
+// + its NormalizedAST to reconstruct the original literals before sending
+// the query to the backend. Dropping bindVars here breaks that
+// reconstruction — the normalized `$N` placeholders reach PG unbound and
+// fail with "there is no parameter $N".
 func (s *Sequence) StreamExecute(
 	ctx context.Context,
 	exec IExecute,
 	conn *server.Conn,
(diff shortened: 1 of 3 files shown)


haritabh17 (approved):
🚀

PR #924, opened 2026-04-29
perf(pgwire): fuse Describe(P)+Execute into one round trip on prepared path

GuptaManan100:
## Summary

- Detects libpq's `Bind+Describe(P)+Execute+Sync` pattern in the pgwire server and folds the held Describe into the trailing Execute, reducing the multipooler→postgres path from two backend round trips per prepared query to one.
- New `BindDescribeAndExecute` client method pipelines `Bind+Describe(P)+Execute+Sync` as a single flushed batch; portal `RowDescription` rides back on the streaming callback's first `Fields`-bearing chunk, so callers receive the same column metadata they would have gotten from a standalone `Describe('P')`.
- Plumbs an `includeDescribe` boolean down `Handler.HandleExecute` → `IExecute` → `Primitive` → `ScatterConn` → multipooler. The portal-only knob lives on a new `multipoolerservice.PortalExecuteOptions` message hung off `PortalStreamExecuteRequest`, keeping `query.ExecuteOptions` unchanged.
- sysbench `oltp_point_select` prepared @ 32c/300s on darwin/arm64: **17,337 → 25,736 TPS (+48%)**, multipooler CPU mean **272% → 227%** (-17%).

## Problem

For libpq's standard `PQexecPrepared` (`Bind+Describe(P)+Execute+Sync`), the gateway was unpacking the batch into two synchronous backend RPCs (`Describe`, then `PortalStreamExecute`). Each acquired a pooled connection, did its own `Bind+Sync` to postgres, and flushed the pgwire client buffer. On the multipooler→postgres link this doubled the syscall count per query and left a redundant second `Bind` (the first `Sync` had already dropped the portal). pprof showed `bufio.(*Writer).Flush` at ~33% cum on the prepared path with two distinct `BindAndDescribe` / `BindAndExecute` calls each contributing ~22% — exactly the shape of the doubled trip.

## Solution

The detection sits in the protocol layer (`pgprotocol/server/conn.go`):

- `handleDescribe('P')` stashes the portal name and writes nothing on the wire.
- `handleMessage` flushes any held describe via `resolveDeferredPortalDescribe` (which calls the synchronous `HandleDescribe` path) before any non-Execute message runs. This keeps wire ordering correct for describe-without-execute, two consecutive Describes of either type, and any other interleaving.
- `handleExecute` matches the held portal name and passes `includeDescribe=true` through `Handler.HandleExecute`. Mismatched names trigger the resolve before the unrelated Execute proceeds.

The handler stays opaque about the deferral — it just forwards the flag to the executor, which sets `PortalExecuteOptions.IncludeDescribe` on the gRPC call. The multipooler swaps `BindAndExecute` for `BindDescribeAndExecute`, which writes `Bind+Describe(P)+Execute+Sync` as one flushed batch. The portal `RowDescription` flows back through the existing streaming-Result `Fields` channel; pgwire-server's `handleExecute` callback writes it to the wire on the first `Fields`-bearing chunk, before any `DataRow`. DML (no fields) writes `NoData` before `CommandComplete` instead.

Wire ordering is preserved on every path: `BindComplete` → `RowDescription`/`NoData` → `DataRow*` → `CommandComplete` → `ReadyForQuery`. Verified by `TestExtendedQueryProtocol_DescribePortal` running 17 sub-scenarios against both direct PostgreSQL and multigateway with byte-equivalent expectations: two consecutive portal describes, statement-after-portal and portal-after-statement describe ordering, deferred-flush before unrelated Execute, DML `NoData`, empty result-set portal describe, repeated `Bind+Execute`, the fused path itself, repeated fused calls, and `ClosePortal` after describe.

## Performance

\`sysbench oltp_point_select\`, 32 clients, 300s/scenario, darwin/arm64:

| protocol | baseline TPS | new TPS | proxy CPU | pooler CPU | postgres CPU |
|---|---:|---:|---:|---:|---:|
| **prepared** | 17,337 | **25,736 (+48%)** | 267% | **227% (-17%)** | 48% |
| simple   | 25,386 | 22,836 | 278% | 205% | 74% |

The simple-protocol path is unchanged by this PR — its drop tracks against postgres-side / thermal variance (postgres CPU near saturation at 104% max), not the gateway code path. Multipooler-side \`bufio.Flush\` cum dropped from 33% to 30%, and the prepared path now shows a single \`BindDescribeAndExecute\` at 45% cum instead of two ~22% calls.

## Test plan

- [x] \`TestExtendedQueryProtocol_DescribePortal\` (17 sub-tests × postgres + multigateway = 34 cases) — all pass
- [x] Full \`go test ./go/test/endtoend/queryserving/...\` — passes
- [x] \`TestSysBench\` prepared+simple @ 32c/300s — passes; numbers above
- [ ] Cold-machine sysbench rerun to validate the simple-protocol drop is environmental


go/common/pgprotocol/server/conn.go
```diff
@@ -476,7 +488,33 @@ func (c *Conn) serve() error {
...
+// Describe(P)+Execute would produce.
 func (c *Conn) handleMessage(msgType byte) error {
+	switch msgType {
+	case protocol.MsgExecute:
+		// handleExecute consumes the deferred describe directly (it folds
+		// it into the backend call rather than flushing it separately).
+		return c.handleExecute()
+
+	case protocol.MsgFlush:
+		return c.flush()
```
haritabh17: When a client sends Describe('P') followed by Flush ('H'), the Flush handler here does not call resolveDeferredPortalDescribe() before flushing the write buffer. The Describe('P') response (RowDescription or NoData) has been deferred but not yet generated, so the flush sends nothing meaningful. The client, which sent Flush specifically to receive the describe response without a full Sync cycle, will hang waiting for the RowDescription/NoData that never arrives

Something like this works?

```
case protocol.MsgFlush:
		if err := c.resolveDeferredPortalDescribe(); err != nil {
			return err
		}
		return c.flush()
```

go/common/pgprotocol/server/conn.go
```diff
@@ -892,7 +948,19 @@ func (c *Conn) handleDescribe() error {
...
-	// Call the handler.
+	// Describe('P') is captured and held: if the next message is Execute on
+	// the same portal, the two fold into a single backend call. Otherwise
+	// the next non-Execute handler (including a follow-up Describe of any
+	// type) flushes the held state via resolveDeferredPortalDescribe in
+	// handleMessage before its own reply runs.
+	if typ == 'P' {
+		c.deferredPortalDescribe = true
+		c.deferredPortalDescribeName = name
+		return nil
```
haritabh17, line 985: When handleDescribe defers a Describe('P'), it calls startWriterBuffering() at entry and endWriterBuffering() via defer, but performs zero writes before returning. Can we save writer acquisition here??

haritabh17 (approved):
Rest looks good 🚀

go/common/pgprotocol/server/conn.go
GuptaManan100: Fixed

go/common/pgprotocol/server/conn.go
GuptaManan100: Fixed

PR #925, opened 2026-04-29
fix: bump recovery timeouts to avoid test flakiness

niconosenzo:
Root cause: DemoteStalePrimary executes synchronously within the recovery cycle — 5s drain + up to ~15s pg_rewind + ~5s postgres restart = ~25s worst case on slow CI. Using a 20s timeout makes the tests to fail eventually.

Fix: Bump timeout to 45s.


mkindahl (approved):
LGTM

PR #926, opened 2026-04-29
fix(tests): ensure cluster cleanup always runs and kills process groups

mkindahl:
Register test cluster cleanup via t.Cleanup inside setupTestCluster so it is guaranteed to run even when the function exits early via t.FailNow() / runtime.Goexit(). Previously, the caller's defer cleanup() was never registered in that case, leaving services running while their ports were returned to the pool.

Also send SIGKILL to the entire process group (kill -pid) in addition to the individual service PID, so child processes such as postgres under pgctld are terminated alongside the service.


rafael (approved):
LGTM

PR #928, opened 2026-04-29
feat(consensus): add coordinator proposal and revocation primitives

dweitzman:
- **`revocation.go`** — new file with two functions used by the coordinator to initiate and validate term revocations:
  - `NewTermRevocation`: derives the next safe term from observed node statuses and stamps the coordinator identity and initiation time
  - `ValidateRevocation`: checks whether a node should accept a given revocation (WAL position safety, coordinator idempotency, recruitment-round idempotency)

- **`proposals.go`** — three new public entry points for the coordinator's propose phase:
  - `CheckProposalPossible` / `CheckExternallyCertifiedProposalPossible`: pre-vote feasibility checks that run the full quorum and eligibility logic without requiring nodes to have accepted the revocation yet
  - `BuildExternallyCertifiedProposal`: bootstrap and stuck-quorum recovery path — the outgoing cohort's revocation is certified by an external agent (operator request via coordinator) rather than through normal Recruit RPCs; the incoming cohort is still recruited normally and its quorum is enforced

- **`ExternallyCertifiedRevocation` proto message** — bundles the outgoing rule number, frozen LSN, and term revocation as a self-contained certification. The coordinator names itself as initiator even when acting on an external operator request.

  Internal hardening of `buildProposalCore` to reject more broken or not-yet-supported cases.

- **Test coverage** — migrated standalone test functions to a table-driven `TestBuildProposalCore`


go/common/consensus/pooler_test.go
```diff
@@ -91,3 +91,65 @@ func TestIsLeader(t *testing.T) {
...
+		cs   *clustermetadatapb.ConsensusStatus
+		want int64
+	}{
+		{
+			name: "nil status",
+			cs:   nil,
+			want: 0,
+		},
+		{
+			name: "not primary",
```
rafael: nit - should we rename to leadeR?

go/common/consensus/proposals.go
```diff
@@ -28,102 +28,287 @@ import (
...
+type cohortQuorumMode int
+
+const (
+	// outgoingCohortMode requires that enough members of the current/outgoing
+	// cohort (identified from the highest committed rule across the recruited
+	// nodes) have accepted the term revocation. This is the standard safety
+	// check for normal failover: the existing cohort must consent to the
+	// leadership transition. validateProposal additionally verifies the
+	// proposed (incoming) cohort has sufficient recruited members.
+	outgoingCohortMode cohortQuorumMode = iota
```
rafael: maybe something like:
- requireOutgoingConsent
- requireIncomingConsent

go/common/consensus/proposals_test.go
```diff
@@ -77,203 +77,813 @@ func simpleProposal(proposedRule *clustermetadatapb.ShardRule) func(RecruitmentR
...
+			},
+			buildProposal: func(r RecruitmentResult) (*consensusdatapb.CoordinatorProposal, error) {
+				outsider := makeID("z1", "outsider")
+				return &consensusdatapb.CoordinatorProposal{
+					TermRevocation: r.TermRevocation,
+					ProposalLeader: &consensusdatapb.ProposalLeader{Id: a},
+					ProposedRule:   makeRule(5, []*clustermetadatapb.ID{a, b, outsider}),
+				}, nil
+			},
+		},
```
rafael: for completeness, it looks like we should add the wantLeader: `pooler-a` . This the one we were missing the check for either wantLeader or checkResult

go/common/consensus/proposals_test.go
```diff
@@ -77,203 +77,813 @@ func simpleProposal(proposedRule *clustermetadatapb.ShardRule) func(RecruitmentR
...
+		revocation        *clustermetadatapb.TermRevocation // nil → testRevocation
+		recruitedStatuses []*clustermetadatapb.ConsensusStatus
+		buildProposal     func(RecruitmentResult) (*consensusdatapb.CoordinatorProposal, error)
+		wantLeader        string
+		checkResult       func(*testing.T, RecruitmentResult)
+		wantErr           string
 	}
 
-	_, err := BuildSafeProposal(testRevocation, statuses, simpleProposal(rule))
+	tests := []tc{
```
rafael, line 182: A few suggestions to harden the test table:

1. Move cohort into the per-case struct so each row is self-contained — a couple of cases already use a different cohort inline, and lifting it makes the relationship between recruited statuses and the rule explicit.
2. Rename a, b, c to poolerA, poolerB, poolerC so fields like wantLeader: "pooler-b" read directly against the variables.
3. In the matrix runner, assert that when wantErr == "", at least one of wantLeader or checkResult is set. A few success-path cases currently assert nothing beyond "no error", which silently weakens coverage.

go/common/consensus/proposals.go
```diff
@@ -28,102 +28,287 @@ import (
...
+// at or above cert.frozen_lsn.
+func certLeaderFilter(
+	cert *clustermetadatapb.ExternallyCertifiedRevocation,
+	nodes []*clustermetadatapb.ConsensusStatus,
+) (func(*clustermetadatapb.ConsensusStatus) bool, error) {
+	if certRule := cert.GetOutgoingRuleNumber(); certRule != nil {
+		for _, cs := range nodes {
+			if rule := cs.GetCurrentPosition().GetRule(); rule != nil {
+				if CompareRuleNumbers(rule.GetRuleNumber(), certRule) > 0 {
+					return nil, fmt.Errorf("node %s is at rule term %d but certified outgoing rule is term %d",
```
rafael: nit - pooler, instead of node for consistency (these are all poolers)

proto/clustermetadata.proto
```diff
@@ -423,6 +423,36 @@ message TermRevocation {
...
+// The coordinator names itself as the initiator in term_revocation even though
+// it is acting on an external request.
+message ExternallyCertifiedRevocation {
+  // The rule number being superseded. Identifies which outgoing cohort is frozen.
+  RuleNumber outgoing_rule_number = 1;
+
+  // The LSN at which the outgoing cohort's progress is frozen. No durable writes
+  // occurred beyond this position under outgoing_rule_number. The chosen leader's
+  // WAL position must meet or exceed this LSN.
+  string frozen_lsn = 2;
```
rafael, line 461: we talked about adding invariant that enforces this is not empty

rafael (approved):
LGTM.

go/common/consensus/pooler_test.go
dweitzman: Done

go/common/consensus/proposals_test.go
dweitzman: Done. Lots of churn on the diff, but the test cases should all be the same and now they're all explicit about all the function inputs

go/common/consensus/proposals_test.go
dweitzman: I made it so all test cases are required to have wantLeader if they're not expecting an error

go/common/consensus/proposals.go
dweitzman: Changed for now to these:
* requireTransitionQuorum
* onlyRequireIncomingQuorum

I'm flexible about renaming them again in the future, but I think these names capture the intent of the transition types

proto/clustermetadata.proto
dweitzman: Done

go/common/consensus/proposals.go
dweitzman: I'm going to leave this as node for this PR and then follow up with another giant PR that changes node to pooler for the whole package. That'll be a mechanical change but it'll impact a lot of lines of code

go/common/consensus/proposals_test.go
rafael: Very cool. I think it makes it way clearer to read now

PR #929, opened 2026-04-29
feat(multipooler): implement Propose RPC for coordinator-led rule changes

dweitzman:
Propose() is the second step in the recruit→propose protocol. After Recruit() wins a term, the coordinator calls Propose() on each recruited node:

- Leader path: promotes postgres to primary (or no-ops if already primary), pre-configures sync replication from the proposed rule, writes the new ShardRule to the rule store, and updates topology.
- Replica path: configures primary_conninfo toward the new leader, updates pooler type to REPLICA, and resumes the postgres monitor.

Propose requires prior recruitment: the stored term revocation must match the proposal's term exactly. We can revisit implicit recruitment in a future PR.

Also extends the rule store to carry accepted_members and the durability policy.


go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -533,6 +536,191 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoAndReload(ctx context.Context, c
...
+	currentStatus, err := pm.getConsensusStatus(ctx)
+	if err != nil {
+		return nil, mterrors.New(mtrpcpb.Code_FAILED_PRECONDITION, err.Error())
+	}
+	if err := commonconsensus.ValidateRevocation(currentStatus, revocation); err != nil {
+		return nil, mterrors.New(mtrpcpb.Code_FAILED_PRECONDITION, err.Error())
+	}
+
+	// No implicit recruitment: require a prior Recruit() for this exact term.
+	storedTerm := currentStatus.GetTermRevocation().GetRevokedBelowTerm()
```
dweitzman, line 619: This is the meat of the PR: wire up replication so that a leader can write a proposal to WAL and followers can endorse it, making it durable.

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -533,6 +536,191 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoAndReload(ctx context.Context, c
...
+		state, err := pm.checkPromotionState(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+		// Pre-configure synchronous_standby_names before pg_promote() so the primary
+		// enforces sync replication from the moment it starts accepting connections.
+		syncCfg, err := syncConfigFromProposedRule(pm.logger, proposedRule, pm.serviceID)
+		if err != nil {
+			pm.logger.WarnContext(ctx, "Failed to compute sync config from proposal", "error", err)
+		} else if err := pm.applyGUCsForSyncReplication(ctx, syncCfg); err != nil {
```
dweitzman: In the future I'm imagining the application of sync standby names being more connected with the updateRule() call below. I have code for that but it's a bigger change so it'll be a follow-up PR

proto/consensusdata.proto
```diff
@@ -186,6 +186,13 @@ message RecruitResponse {
 // itself; all others configure replication toward the new primary.
 message ProposeRequest {
   CoordinatorProposal proposal = 1;
+
+  // Human-readable reason for this proposal (e.g. "dead_primary", "bootstrap").
```
dweitzman, line 190: Missed these .proto changes in the previous PR

go/services/multipooler/manager/rpc_consensus.go
rafael: Two comments around here:

1) Not a problem, but something subtle here is that we are coupling the read of the revocation that comes from disk with reading the position from postgres. I wonder if it would be clearer to be explicit and use `pm.consensusState.GetRevocation(ctx)`
2) nit - I would expand on the comment
```
// Strict equality, not >=: a Propose must be backed by a completed
// Recruit() at this exact term. Recruit is what revokes prior agents
// and performs discovery; letting Propose implicitly bump the term
// would skip both.
```

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -533,6 +536,208 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoAndReload(ctx context.Context, c
...
+			return nil, err
+		}
+		// Pre-configure synchronous_standby_names before pg_promote() so the primary
+		// enforces sync replication from the moment it starts accepting connections.
+		// TODO: Eventually updateRule() should own this GUC update so that the ordering
+		// of WAL writes and GUC changes is managed in one place, ensuring the durability
+		// configuration is always consistent with what is recorded in the rule history.
+		syncCfg, err := syncConfigFromProposedRule(pm.logger, proposedRule, pm.serviceID)
+		if err != nil {
+			pm.logger.WarnContext(ctx, "Failed to compute sync config from proposal", "error", err)
```
rafael: Don't we need to return error here?

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -533,6 +536,208 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoAndReload(ctx context.Context, c
...
+		if err := pm.promoteStandbyToPrimary(ctx, state); err != nil {
+			return nil, err
+		}
+		if err := pm.clearResignedLeaderAtTerm(ctx); err != nil {
+			return nil, mterrors.Wrap(err, "failed to clear resigned primary term")
+		}
+		// checkPromotionState already fetched the LSN when postgres was primary;
+		// only query again after a fresh pg_promote() call.
+		var finalLSN string
+		if state.isPrimaryInPostgres {
```
rafael: I don't get this check. If `promoteStandbyToPrimary` already happened, we know this will be a primary no?

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -533,6 +536,208 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoAndReload(ctx context.Context, c
...
+			return nil, err
+		}
+		// Ensure health state reflects REPLICA, even if this node was recently
+		// an emergency-demoted primary (which restarts postgres as standby but
+		// does not update the pooler type until a Propose arrives).
+		if err := pm.changeTypeLocked(ctx, clustermetadatapb.PoolerType_REPLICA); err != nil {
+			pm.logger.WarnContext(ctx, "Failed to update pooler type to REPLICA after propose", "error", err)
+		}
+		// Replication is now configured; allow the postgres monitor to resume.
+		pm.rewindPending.Store(false)
```
rafael: Why the `rewindPending` in the context of the followers/replicas? This should only apply for the stale primary

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -533,6 +536,208 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoAndReload(ctx context.Context, c
...
+	if proposalLeader == nil {
+		return nil, mterrors.New(mtrpcpb.Code_INVALID_ARGUMENT, "proposal.proposal_leader is required")
+	}
+	if proposalLeader.GetPostgresPort() == 0 {
+		return nil, mterrors.New(mtrpcpb.Code_INVALID_ARGUMENT, "proposal.proposal_leader.postgres_port is required")
+	}
+	proposedRule := proposal.GetProposedRule()
+	if proposedRule == nil {
+		return nil, mterrors.New(mtrpcpb.Code_INVALID_ARGUMENT, "proposal.proposed_rule is required")
+	}
```
rafael, line 590: maybe also validation that proposalLeader.GetID() is not nil ?

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -533,6 +536,208 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoAndReload(ctx context.Context, c
...
+
+	revokedBelowTerm := revocation.GetRevokedBelowTerm()
+	coordinatorID := revocation.GetAcceptedCoordinatorId()
```
rafael, line 593: Thinking that we should also have an invariant that a pooler with pgRewindPending set to true, can't participate in Propose

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -533,6 +536,208 @@ func (pm *MultiPoolerManager) setPrimaryConnInfoAndReload(ctx context.Context, c
...
+		// causing the gateway to buffer — so no client write transactions can be served.
+		//
+		// DO NOT call updateTopologyAfterPromotion() before updateRule() succeeds.
+		// The rule commit is the durability gate: it waits for sync-standby
+		// acknowledgment, proving the new primary position is replicated. Only after
+		// that gate closes is it safe to advertise PRIMARY + SERVING to the gateway.
+		if err := pm.promoteStandbyToPrimary(ctx, state); err != nil {
+			return nil, err
+		}
+		if err := pm.clearResignedLeaderAtTerm(ctx); err != nil {
```
rafael, line 685: Shouldn't this also be part of the DemoteStalePrimary?

go/services/multipooler/manager/rpc_consensus_test.go
```diff
@@ -1494,6 +1494,307 @@ func TestRecruit(t *testing.T) {
...
+		Name:      "other-pooler",
+	}
+
+	// recruitedTerm simulates a node that has been recruited for term 7 by coordinatorA.
+	recruitedTerm := &clustermetadatapb.TermRevocation{
+		RevokedBelowTerm:       7,
+		AcceptedCoordinatorId:  coordinatorA,
+		CoordinatorInitiatedAt: recruitTS,
+	}
+	validRevocation := &clustermetadatapb.TermRevocation{
```
rafael: why not use the same variable here?  (i.e recruitedTerm==validRevocation)

go/services/multipooler/manager/rpc_consensus_test.go
```diff
@@ -1494,6 +1494,307 @@ func TestRecruit(t *testing.T) {
...
+			expectErrContains: "proposal.proposed_rule is required",
+		},
+		{
+			name:        "TermNotRecruited",
+			initialTerm: &clustermetadatapb.TermRevocation{RevokedBelowTerm: 3},
+			ruleStore:   &fakeRuleStore{pos: makeRulePosition(0)},
+			req:         makeLeaderReq(),
+			setupMocks:  func(m *mock.QueryService) {},
+			expectError: true,
+			// stored term 3 != proposal term 7
```
rafael: maybe one extra test to check an intial_term that is greater than the proposed one, so we have coverage for exact equality

go/services/multipooler/manager/rpc_consensus.go
dweitzman: Here's what I propose:
- in a follow-up PR, I'll make Propose() do implicit recruitment
- Propose() will never change primary conn info or do a pg_promote() unless replication is fully stopped and, after being stopped, the term revocation is still valid (meaning it hasn't be invalidated by the appearance of a newer WAL entry)

dweitzman: Or even simpler: if the pooler is in primary mode or has replication turned on, the revocation is presumed to be consumed

go/services/multipooler/manager/rpc_consensus.go
dweitzman: Done

go/services/multipooler/manager/rpc_consensus.go
dweitzman: For now I've added code so that Recruit() fails if pgRewindPending is set. We can look at that closer later

go/services/multipooler/manager/rpc_consensus.go
dweitzman: Made the change to only allow Propose() if the pooler is a stopped replica

go/services/multipooler/manager/rpc_consensus.go
dweitzman: I think DemoteStalePrimary() doesn't necessarily need to clear a resignation

When it gets replaced by the Inform() RPC, that'll clear resignation unless the Inform() identifies the current node as the term leader

go/services/multipooler/manager/rpc_consensus.go
dweitzman: Removed this

go/services/multipooler/manager/rpc_consensus_test.go
dweitzman: Done

go/services/multipooler/manager/rpc_consensus_test.go
dweitzman: Done

rafael (approved):
LGTM

PR #931, opened 2026-04-30
feat(multiorch): extend /ready to check etcd connectivity

niconosenzo:
## Description
During startup, multiorch should be reported as = `Ready=True` only after confirmation that it can read the topology store.

## Changes
- Register a second ready check in multiorch that calls GetCellNames with a 4s timeout and confirm topo service reachability.


rafael (approved):
LGTM

PR #932, opened 2026-04-30
test(pgwire): regression coverage for async pusher / handler buffer race

GuptaManan100:
## Summary

- Adds a unit-level regression test in `go/common/pgprotocol/server/` that drives the synchronous reply path and the async notification pusher concurrently against the same `Conn` and asserts the wire bytes are well-framed.
- Adds an end-to-end `listen_with_concurrent_queries` subtest under `TestListenNotify` that exercises the same shape through multigateway.
- No production code change — the `bufMu` lock bracket in `startPacket` / `writePacket` already protects the race; what was missing was coverage that would have caught a regression.

## Problem

The pgwire server has two goroutines that write into the same `Conn`'s `bufferedWriter`: the synchronous request handler and the async notification pusher started by `EnableAsyncNotifications`. `bufio.Writer` is not goroutine-safe, so without serialization across body encoding the two writers can interleave byte-by-byte and produce torn packets on the wire (and `-race` data race reports on `bufio.Writer`'s internal cursor).

The lock bracket fix already shipped via the perf series (`startPacket` acquires `bufMu`, `writePacket` releases it, body encoding happens under the lock; the async pusher's `writeNotificationResponseMsg` goes through the same API). But every existing LISTEN/NOTIFY test used a dedicated idle listener connection — none of them exercised the same connection both listening and pipelining queries, so the race window was never opened in CI.

## Solution

Two tests, both designed to fail under `-race` if the lock bracket is removed:

- **Unit test** (`notification_race_test.go`): two goroutines on one `Conn` — one drives `writeRowDescription` / `writeDataRow` / `writeCommandComplete` / `writeReadyForQuery` in a tight loop, the other drives `writeNotificationResponseMsg`. After the soak, walk the captured wire bytes packet-by-packet and assert exact per-type counts. Verified to fail reliably under `-race` with the `bufMu.Lock` / `Unlock` in `startPacket` / `writePacket` commented out (race detector hit on `bufio.(*Writer).Write` vs `Flush`, plus `short write` errors). Passes with `-race -count=20` against the lock-bracketed code.
- **E2E test** (`listen_with_concurrent_queries`): one listener conn LISTENs and runs a tight extended-protocol prepared-query loop while a sibling notifier hammers `NOTIFY`. After the 5-second soak, drains a buffered `NotificationResponse` to confirm the async pusher path stayed well-formed end-to-end. The query loop alone exercises the server-side race because `pgx`'s `pgconn` buffers `NotificationResponse` messages it sees during normal reads — no separate `WaitForNotification` goroutine is needed (and pgx forbids concurrent use of a single `Conn`).

### DIFF
--- a/go/common/pgprotocol/server/notification_race_test.go
+++ b/go/common/pgprotocol/server/notification_race_test.go
@@ -1,0 +1,265 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/binary"
+	"errors"
+	"io"
+	"net"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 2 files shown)


haritabh17 (approved):
Looks good 🚀

PR #934, opened 2026-04-30
test(pgregresstest): persist per-test residual diffs to artifact

haritabh17:
## Summary

- `VerifyWithPatches` already computes the patched-vs-actual residual diff per failing test (`outcome.Diff` in `patch_verify.go`), but the value is dropped on the floor — `results.json`, `compatibility-report.md`, and pg_regress's raw `regression.diffs` all lack it, so investigating a failure required re-running the suite locally.
- Write each failure's residual diff to `<outputDir>/diffs/<name>.diff` and a concatenated `<outputDir>/failures.diffs`. Both ride the existing artifact upload — downloading the CI artifact now gives the full set of residual diffs.

## Problem

After a CI run, the only diff data shipped is `regression.diffs` produced by pg_regress itself, which:

- Compares unpatched upstream `expected/` against `results/` directly, so it includes diffs for tests we mark as **passing** under our `.patch`-based verification (whitespace-only changes, error-message wording, etc.).
- Lacks any signal for which diffs are *real* residuals after patching.

`VerifyWithPatches` already produces the right artifact internally — `outcome.Diff` is the patched-vs-actual residual — but never persists it.

## Solution

In the per-test verification loop, on `outcome.Status == "fail"` and non-empty `outcome.Diff`, write `<outputDir>/diffs/<test>.diff`. After the loop, write a concatenated `failures.diffs` with `=== <name> ===` separators for one-shot grep.

`VerifyWithPatches` gains an `outputDir` parameter; the regression-suite call site passes `builder.OutputDir`. No table / report format changes — diffs are only on disk, picked up automatically by the existing `actions/upload-artifact` glob (`/tmp/multigres_pg_cache/results/**/*`).

## Validation

- `go build ./go/test/endtoend/pgregresstest/...` — clean.
- `go test -short ./go/test/endtoend/pgregresstest/...` — pass.
- Existing `patch_verify_test.go` exercises `VerifyTest` (unaffected); no harness-level tests changed signature.

### DIFF
--- a/go/test/endtoend/pgregresstest/pgregress_test.go
+++ b/go/test/endtoend/pgregresstest/pgregress_test.go
@@ -139,7 +139,7 @@ func TestPostgreSQLRegression(t *testing.T) {
 			// verification. Operates on the regress build directory, where
 			// pg_regress wrote expected/ and results/ files.
 			regressDir := filepath.Join(builder.BuildDir, "src", "test", "regress")
-			if verr := builder.VerifyWithPatches(t, suiteCtx, results, regressDir); verr != nil {
+			if verr := builder.VerifyWithPatches(t, suiteCtx, results, regressDir, builder.OutputDir); verr != nil {
 				t.Logf("Warning: patch verification failed: %v", verr)
 			}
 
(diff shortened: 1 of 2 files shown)


mkindahl (approved):
LGTM

PR #935, opened 2026-04-30
fix(sqllogictest): skip per-file schema resets and bypass gateway pool

GuptaManan100:
## Summary

- Eliminates the deterministic ~30s `statement_timeout` that `test/random/select/slt_good_125.test` simple-protocol hits in CI by removing the catalog-invalidation backlog the harness was generating itself.
- Adds a `suiteutil.SchemaResetter` that holds one long-lived pgx connection per target and tracks a `dirty` bit so resets are skipped between sqllogictest parse-error files (which never reach the database).
- Points the multigateway target's resetter at the underlying primary multipooler PG (`setup.GetPrimary(t).Pgctld.PgPort`) instead of the gateway, so reset DDL pins to one backend instead of scattering across multipooler's connection pool.

## Problem

In nightly CI, `test/random/select/slt_good_125.test` deterministically failed on the multigateway target's simple protocol with `ERROR: canceling statement due to statement timeout` after exactly ~30s — the default `--statement-timeout`. The same file passed on extended protocol in <40 ms, and the byte-identical `slt_good_126.test` immediately afterwards passed both.

A captured `pg_stat_activity` snapshot mid-hang showed the offending PG backend running `CREATE TABLE tab0(col0 INTEGER, col1 INTEGER, col2 INTEGER)` on an empty schema in `state=active` with `wait_event=NULL` for 5+ seconds — i.e. burning CPU inside `AcceptInvalidationMessages`, not waiting on a lock or I/O.

The runner reset the public schema before every (file × protocol × target) invocation by opening a fresh pgx connection, running `DROP SCHEMA … CASCADE; CREATE SCHEMA public; …`, and disconnecting. Each cycle generates pg_namespace / pg_class shared-invalidation messages onto every other live PostgreSQL backend. Multipooler keeps a regular pool of ~80 idle backends per user, so all of them queue the SI traffic. Long stretches of the sqllogictest corpus (190+ files in `test/random/select/`) fail in sqllogictest's own parser before opening any connection — emitting no SQL — yet the runner still reset between them. By the time the first parseable simple-protocol file lands on a borrowed pooled backend, the SI queue is large enough that the first DDL pays for the whole drain and crosses the 30s wall.

## Solution

Two changes, both in test code only.

**1. Long-lived per-target resetter, with the multigateway one pointed at the underlying PG.** A new `suiteutil.SchemaResetter` holds one pgx connection open for the whole batch. Crucially, the multigateway-target resetter connects to the primary multipooler's PG directly (port `Pgctld.PgPort`, user `pgconstants.DefaultPostgresUser`, password `shardsetup.TestPostgresPassword`) rather than through the gateway. Multigres is itself a connection pooler, so a single gateway connection isn't pinned to one backend — bypassing the pool is the only way to keep all reset traffic on one PG session.

**2. `dirty`-bit-gated resets.** The resetter exposes `MarkDirty()` and `ResetIfDirty(ctx)`. The runner calls `ResetIfDirty` before each sqllogictest invocation and `MarkDirty()` afterwards only when the captured output does not contain `failed to parse` (the parse-error path provably never reached the database, so the schema is unchanged and the next file can reuse the prior reset).

Verified locally on a 162-file subset that previously reproduced an 11.45s simple-MG hang on `slt_good_125`: that file drops to 5.79s after the fix. In CI's heavier preamble (190+ parse-error files between the test/index workload and slt_good_125), the savings are larger and comfortably under the 30s wall.

The single-shot `suiteutil.ResetPublicSchema` API is unchanged, so the existing pgparity caller is unaffected.

## Test plan

- [ ] Re-run the SQLLogicTest Differential Suite nightly (or trigger via the `Run Extended Query Serving Tests` PR label) and confirm `slt_good_125.test` simple-MG passes.
- [x] Spot-check that no proxy divergences regress.

### DIFF
--- a/go/test/endtoend/queryserving/sqllogictest/runner.go
+++ b/go/test/endtoend/queryserving/sqllogictest/runner.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os/exec"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/multigres/multigres/go/test/endtoend/suiteutil"
@@ -58,6 +59,15 @@ type runResult struct {
 // CREATE TABLE statements and assumes no leftover state from the previous
 // file or previous protocol).
 //
+// resetter performs the per-file `DROP SCHEMA public CASCADE; CREATE SCHEMA
+// public; …` cycle. It is supplied by the caller so the same long-lived
+// connection is reused across all files. Without that reuse, each fresh pgx
+// connection-per-call generates a wave of pg_class / pg_namespace
+// invalidation messages that pile up against pooled multipooler backends
+// the simple-protocol path eventually lands on, producing a multi-second
+// catalog-cache rebuild on first use of those connections (most visibly:
+// `slt_good_125` simple hitting `statement_timeout` in CI).
+//
 // The runner always returns a *runResult; it never fails the test. Callers
 // decide what to do with the outcome (typically: aggregate into the
 // results.json report).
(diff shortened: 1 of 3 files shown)


haritabh17 (approved):
🚀

PR #936, opened 2026-04-30
refactor: replace types.ShardKey struct with *clustermetadatapb.ShardKey

mkindahl:
Remove the hand-rolled ShardKey struct from go/common/types and use the proto-generated *clustermetadatapb.ShardKey everywhere. Map keys that previously used the comparable struct directly now use commontypes.ShardKeyString (format: "db/tablegroup/shard") since proto pointers compare by address, not value.

### DIFF
--- a/go/services/multigateway/buffer/buffer.go
+++ b/go/services/multigateway/buffer/buffer.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/multigres/multigres/go/common/mterrors"
 	commontypes "github.com/multigres/multigres/go/common/types"
+	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 )
 
 // RetryDoneFunc must be called by the caller after the retry attempt completes.
@@ -52,7 +53,7 @@ type entry struct {
 	bufferCtx    context.Context
 	bufferCancel context.CancelFunc
 	// shardKey identifies which shard this entry belongs to.
-	shardKey commontypes.ShardKey
+	shardKey *clustermetadatapb.ShardKey
 	// createdAt records when the entry was enqueued for metrics.
 	createdAt time.Time
 }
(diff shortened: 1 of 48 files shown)


go/services/multigateway/buffer/buffer.go
```diff
@@ -88,8 +89,8 @@ type Buffer struct {
 	bufferSizeSema *semaphore.Weighted
 
 	mu      sync.Mutex
-	buffers map[commontypes.ShardKey]*shardBuffer
-	queue   []*entry // Global FIFO queue (all shards interleaved)
+	buffers map[string]*shardBuffer // keyed by ShardKeyString
```
dweitzman: Hypothetically if we wanted to we could keep the ShardKeyString type as the output of ShardKeyString() for stronger static typing. I'm not sure if that has other annoying side effects so I'll defer to your judgement

If the ShardKey type was working before, that may imply that using a custom type for this would be fine as a way of having more static type safety

dweitzman (approved):
Nice!

go/services/multigateway/buffer/buffer.go
mkindahl: > Hypothetically if we wanted to we could keep the ShardKeyString type as the output of ShardKeyString() for stronger static typing. I'm not sure if that has other annoying side effects so I'll defer to your judgement
> 
> If the ShardKey type was working before, that may imply that using a custom type for this would be fine as a way of having more static type safety

The old ShardKey worked because map accepts any structure type that has comparable members. A protobuf ShardKey cannot do that since it contains non-comparable members.

We could use the old ShardKey as key into the map, but this means we need to maintain both structures. Converting it to a string and use that in a map is a natural and it does require us to maintain two structures, so I would suggest that we just use strings.

dweitzman: We can have typed strings in Go that still work as map keys but are harder to misuse in other places, like this:

```
type ShardKeyString string
```

When a helper method takes a variety of string parameters, this would give us compile time protection against passing them in the wrong order, etc

mkindahl: Ah, I misunderstood what you meant. Will fix.

mkindahl: Done,

PR #937, opened 2026-04-30
ci: detect flaky tests, auto-retry on failure, weekly Slack report

GuptaManan100:
## Summary

- Wraps the unit / race / integration test workflows in `gotestsum --rerun-fails=3` so a single transient test failure no longer fails a PR. Coverage workflows use a bash retry loop instead — `gotestsum`'s per-package retry would truncate `-coverprofile` and corrupt `GOCOVERDIR`, so the whole suite is retried up to 3 times on failure.
- Adds `go/tools/flakyaggregator`, a weekly scheduled job that fetches gotestsum `--jsonfile` artifacts from the last 7 days of successful CI runs across `main` and PRs, identifies tests that failed-then-passed within a single run, and posts a Slack summary of any test that flaked in more than 3 distinct runs.
- Two-layer design separates _papering over_ flakes (rerun-fails) from _detecting_ them (the aggregator) — green CI doesn't hide flake history, because the original failure remains in the gotestsum JSONL even when the retry recovers.

### Layer 1: in-run retry (`test-short.yml`, `test-race.yml`, `test-integration.yml`)

`go test` is replaced with `gotestsum --format=testname --rerun-fails=3 --rerun-fails-max-failures=10 --jsonfile=...`. The circuit breaker (`--rerun-fails-max-failures=10`) prevents a genuinely broken commit from burning minutes on doomed retries. Artifacts upload only the JSONL stream now — the JUnit file we briefly emitted turned out to be useless for flake detection because gotestsum dedupes its JUnit output to last-result.

### Layer 1.5: whole-suite retry for coverage (`coverage.yml`)

Coverage runs cannot use `--rerun-fails` (a retry would overwrite the single `-coverprofile` file with only the rerun's package). Instead, a small bash retry loop wraps `go test`. On failure it logs `::warning::` and re-runs the whole suite up to 3 times. Costs more wall-clock on a flake but preserves coverage correctness; coverage isn't merge-blocking, so the duration cost is acceptable. Subprocess coverage also wipes `$GOCOVERDIR` between attempts.

### Layer 2: artifact retention

`*-test-results.jsonl` is uploaded with 14-day retention so the aggregator's 7-day rolling window has headroom for manual investigation.

### Layer 3: weekly aggregator (`flaky-test-report.yml` + `go/tools/flakyaggregator/`)

A weekly scheduled workflow (Mondays 14:00 UTC, also `workflow_dispatch`) runs a small Go program that:

- Lists successful workflow runs across `main` + all PR branches over the last `--days` (default 7), scoped per-workflow-file (`test-short.yml`, `test-race.yml`, `test-integration.yml`) via `Actions.ListWorkflowRunsByFileName`. The per-file scoping skips lint, docker-build, etc. and is the largest single API-rate-limit saver.
- Downloads each run's `*-test-logs` artifact zip (uses `*-test-results.jsonl` only — coverage artifacts are excluded since they can't contribute flake signal).
- Streams the gotestsum JSON event stream and flags any test that observed both a `fail` and a `pass` action — the canonical flake definition.
- Counts distinct **runs** in which each test flaked (so a test failing twice in one run only counts once).
- Filters strictly to tests with `>3` distinct flaking runs (configurable), sorts descending by count, and posts to Slack via `SLACK_WEBHOOK_URL`.

The conclusion=success API filter plus the fail-then-pass parser predicate is the load-bearing definition: a passing CI run with at least one observed failure means a retry recovered it. Tests that fail every retry make the run conclude failure and are correctly excluded as real-bug noise rather than reported as flakes — which is what makes including PR runs (and the larger sample size that comes with them) safe.

### Rate-limit hardening

- Per-workflow-file run listing (above).
- Generic `withRetry[T]` wrapper around every API call, detecting `*github.RateLimitError` and `*github.AbuseRateLimitError` via `errors.As`, sleeping until reset (capped at 5 min) for up to 3 retries.
- Rate-limit budget logged at start and end of the aggregator run.

### Dependencies

`go.mod` gains `github.com/google/go-github/v68` and indirect `github.com/google/go-querystring`. JUnit parsing was briefly handled by `github.com/joshdk/go-junit` and was then removed when we switched to gotestsum JSONL.

## Things to verify before merging

- [x] **Add `SLACK_WEBHOOK_URL` repo secret.** Aggregator fails fast with a clear error if it's missing.
- [x] Optional: trigger the new workflow manually via `workflow_dispatch` with `dry_run=true` once before the first scheduled run lands. This validates the GitHub API plumbing and lets you eyeball the Slack payload format without spamming the channel.
- [x] Confirm `Mondays 14:00 UTC` cron timing works for the team (~7am Pacific / 10am Eastern / 3pm UK). Edit the cron expression if not.
- [x] When a future test workflow with `gotestsum --rerun-fails` is added, append its filename to `testWorkflowFiles` in `go/tools/flakyaggregator/main.go` so the aggregator picks up its artifacts.

### DIFF
--- a/.github/workflows/flaky-test-report.yml
+++ b/.github/workflows/flaky-test-report.yml
@@ -43,6 +43,7 @@ jobs:
   report:
     name: Aggregate flaky tests and post to Slack
     runs-on: ubuntu-24.04
+    environment: ci
     permissions:
       actions: read
     steps:
--- a/.github/workflows/flaky-test-report.yml
+++ b/.github/workflows/flaky-test-report.yml
@@ -1,0 +1,67 @@
+# Copyright 2026 Supabase, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Flaky Test Report
+
+on:
+  schedule:
+    # Mondays at 14:00 UTC.
+    - cron: "0 14 * * 1"
+  workflow_dispatch:
+    inputs:
+      days:
+        description: "Lookback window in days"
+        required: false
+        default: "7"
+      threshold:
+        description: "Report tests with strictly more failures than this"
+        required: false
+        default: "3"
+      dry_run:
+        description: "Print Slack payload to logs instead of posting"
+        required: false
+        default: "false"
+
+permissions: {}
+
+jobs:
+  report:
+    name: Aggregate flaky tests and post to Slack
+    runs-on: ubuntu-24.04
+    permissions:
+      actions: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        with:
+          go-version-file: go.mod
+
+      - name: Run aggregator
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+          DAYS: ${{ github.event.inputs.days || '7' }}
+          THRESHOLD: ${{ github.event.inputs.threshold || '3' }}
+          DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}
+        run: |
+          args=(--days="$DAYS" --threshold="$THRESHOLD")
+          if [ "$DRY_RUN" = "true" ]; then
+            args+=(--dry-run)
+          fi
+          go run ./go/tools/flakyaggregator "${args[@]}"
--- a/.github/workflows/flaky-test-report.yml
+++ b/.github/workflows/flaky-test-report.yml
@@ -18,6 +18,10 @@ on:
   schedule:
     # Mondays at 14:00 UTC.
     - cron: "0 14 * * 1"
+  # checkov:skip=CKV_GHA_7: inputs are read-only flags that gate aggregator
+  # behavior (lookback window, threshold, dry-run); they do not influence any
+  # build artifact, so the "build output cannot be affected by user
+  # parameters" rule does not apply.
   workflow_dispatch:
     inputs:
       days:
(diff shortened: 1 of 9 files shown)


.github/workflows/flaky-test-report.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        with:
+          go-version-file: go.mod
+
+      - name: Run aggregator
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
```
rafael, line 63: We need to add the environment selector to the job. Otherwise I think the secret is not gonna be injected. We have been setting secrets per environment. Something like we did [here](https://github.com/multigres/multigres/pull/927/changes#diff-5477201dbb5cda254830c293a894fdc1b2788fce29c3fdc103102fcb635c663aR36)

rafael (approved):
Do you have an example of the report?  I'm eager to give this a try. I did not review in detail the flake aggregator as I'm seeing this as POC and we can iterate on the actual output. 

If we end up using `trunk.io` , I think we won't need most of the changes here. But we can cross that bridge later. This PR will have immediate impact.

GuptaManan100:
> Do you have an example of the report? I'm eager to give this a try. I did not review in detail the flake aggregator as I'm seeing this as POC and we can iterate on the actual output.
>
> If we end up using `trunk.io`, I think we won't need most of the changes here. But we can cross that bridge later. This PR will have immediate impact.

Yes, I ran the aggregator locally in `--dry-run` mode against `multigres/multigres` over the last 7 days. It scanned 640 successful runs across `test-short.yml`, `test-race.yml`, and `test-integration.yml`, handled two expired-artifact `410 Gone` responses gracefully, and produced this Slack payload:

```json
{
  "text": "*Flaky Test Report — week of 2026-04-30*\nTests that failed-then-recovered in more than 3 successful CI runs (`main` + PRs) over the last 7 days:\n\nNo flaky tests detected this week. :tada:"
}
```

That's the empty-state message, expected pre-merge, since most of those 640 runs predate the `gotestsum --rerun-fails` cutover this PR introduces, so their JSONL streams don't contain any fail-then-pass pairs to detect. Once a few weeks of post-merge runs accumulate, the format with real entries will look like:

```text
*Flaky Test Report — week of 2026-05-04*
Tests that failed-then-recovered in more than 3 successful CI runs (`main` + PRs) over the last 7 days:

• `pkg/a.TestFoo` — *7* failures (last seen 2026-05-03, <https://github.com/multigres/multigres/actions/runs/…|run>)
• `pkg/b.TestBar` — *4* failures (last seen 2026-05-02, <https://github.com/multigres/multigres/actions/runs/…|run>)
```

Agreed re: trunk.io, happy to rip Layer 3 out and lean on a managed service if the team prefers, but Layers 1+2 (gotestsum `--rerun-fails`, JSONL artifact retention) carry their own value regardless of where the analysis lives.

.github/workflows/flaky-test-report.yml
GuptaManan100: Done! Added `environment: ci` to the `report` job in `.github/workflows/flaky-test-report.yml`, matching the pattern from PR #927. `SLACK_WEBHOOK_URL` will now be resolved from the `ci` environment's secret store rather than the repo-level scope.

PR #949, opened 2026-05-01
feat(pgctld): enable operators to provide extra config to postgres

rafael:
# Description

The main change here is --extra-postgres-conf. There are few other changes that are mostly cleanup and bug fixes that the new feature surfaced.

## Changes
* Adds --extra-postgres-conf (and POSTGRES_EXTRA_CONF env var) to pgctld: a list of postgresql.conf files appended verbatim onto the generated config at init time. Mirrors vitess's EXTRA_MY_CNF pattern. Postgres applies last-write-wins, so extras override template defaults; the four pgctld-pinned settings (port, listen_addresses, unix_socket_directories, data_directory) remain safe because they're passed as -c args at start, which beat postgresql.conf.
* Removes Port, ListenAddresses, and UnixSocketDirectories from PostgresServerConfig: they were dead in-memory state since pgctld pins those via CLI args anyway. This also cleans up a read-back asymmetry where the in-memory struct could disagree with what postgres actually loaded.
* Teaches ReadPostgresServerConfig to follow postgres's include / include_if_exists / include_dir directives recursively, so the in-memory struct mirrors what postgres will actually load when extras pull in further files. Without this, an operator using include 'overrides.conf' in their extras would end up with a struct that disagreed with reality.

## Tests

* New unit tests cover the append, the include-following parser, the missing-file error, and the flag plumbing.
*  A new integration test in go/test/endtoend/pgctld exercises the whole stack end-to-end: extras → postgres include → live SHOW reflecting the override.

---
Fixes: MUL-380

### DIFF
--- a/go/cmd/pgctld/command/root.go
+++ b/go/cmd/pgctld/command/root.go
@@ -49,6 +49,7 @@ type PgCtlCommand struct {
 	postgresConfigTmpl viperutil.Value[string]
 	pgInitdbArgs       viperutil.Value[string]
 	initDbSQLFiles     viperutil.Value[[]string]
+	extraPostgresConf  viperutil.Value[[]string]
 
 	vc        *viperutil.ViperConfig
 	lg        *servenv.Logger
@@ -121,6 +122,12 @@ func GetRootCommand() (*cobra.Command, *PgCtlCommand) {
 			EnvVars:  []string{constants.PgInitDbSQLFilesEnvVar},
 			Dynamic:  false,
 		}),
+		extraPostgresConf: viperutil.Configure(reg, "extra-postgres-conf", viperutil.Options[[]string]{
+			Default:  []string{},
+			FlagName: "extra-postgres-conf",
+			EnvVars:  []string{constants.PgExtraConfFilesEnvVar},
+			Dynamic:  false,
+		}),
 		vc:        viperutil.NewViperConfig(reg),
 		lg:        servenv.NewLogger(reg, telemetry),
 		telemetry: telemetry,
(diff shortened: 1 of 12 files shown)


go/services/pgctld/postgresconfig_gen.go
```diff
@@ -118,28 +115,38 @@ func GeneratePostgresServerConfig(poolerDir string, port int, pgUser string) (*P
 	return ReadPostgresServerConfig(cnf, 0)
 }
 
-// LoadOrCreatePostgresServerConfig loads an existing PostgreSQL server configuration
-// from disk, or generates and writes a new one if it doesn't exist.
-// port is the port for the PostgreSQL server.
-// pgUser is the PostgreSQL user name for authentication.
-func LoadOrCreatePostgresServerConfig(poolerDir string, port int, pgUser string) (*PostgresServerConfig, error) {
```
rafael, line 125: This method was not being used, so I removed it.

go/cmd/pgctld/command/root.go
```diff
@@ -121,6 +122,12 @@ func GetRootCommand() (*cobra.Command, *PgCtlCommand) {
 			EnvVars:  []string{constants.PgInitDbSQLFilesEnvVar},
 			Dynamic:  false,
 		}),
+		extraPostgresConf: viperutil.Configure(reg, "extra-postgres-conf", viperutil.Options[[]string]{
```
mkindahl: Since the other options have the form "pg-initdb-something" it might make sense to follow this pattern and use, e.g., "pg-initdb-extra-conf".

go/common/constants/postgres.go
```diff
@@ -44,6 +44,12 @@ const (
 	// Multiple files are comma-separated.
 	PgInitDbSQLFilesEnvVar = "POSTGRES_INITDB_SQL_FILES"
 
+	// PgExtraConfFilesEnvVar is the environment variable for extra postgresql.conf
+	// files appended verbatim onto the generated config at init time.
+	// Multiple files are comma-separated. Postgres applies last-write-wins, so
+	// extras override values from the templated defaults.
+	PgExtraConfFilesEnvVar = "POSTGRES_EXTRA_CONF"
```
mkindahl: Might make sense to follow the pattern of the other files, e.g., something line `POSTGRES_INITDB_EXTRA_CONF`.

go/services/pgctld/postgresconfig.go
```diff
@@ -119,6 +129,99 @@ func (cnf *PostgresServerConfig) lookupFloat(key string) (float64, error) {
...
+		if e.IsDir() || strings.HasPrefix(e.Name(), ".") || !strings.HasSuffix(e.Name(), ".conf") {
+			continue
+		}
+		names = append(names, e.Name())
+	}
+	sort.Strings(names)
+	for _, n := range names {
+		if err := parseConfigInto(filepath.Join(dir, n), configMap, depth); err != nil {
+			return err
+		}
```
mkindahl, line 213: It seems this handles nested includes. Might be good to highlight in a comment for the main entry function.

mkindahl (approved):
LGTM. You might want to change the name to follow the existing convention.

PR #950, opened 2026-05-02
refactor(multipooler): add an internal/ package and move consensus code to a sub-package

dweitzman:
Move most of the multipooler service code under internal/ so violations are caught at compile time rather than only by golangci-lint depguard rules.

Also carve two focused sub-packages out of multipooler/manager/:
- manager/actionlock/: ActionLock and AssertActionLockHeld, used across the manager as a shared locking primitive
- manager/consensus/: ConsensusState, term storage, rule store, and the ReplicaID type (consolidating the duplicate replicaID definitions that previously existed in both pg_replication.go and rule_store.go)

grpcmanagerservice and grpcpoolerservice remain public because cmd/ blank-imports them for gRPC service registration.


go/services/multipooler/internal/manager/actionlock/action_lock.go
dweitzman, line 1: This moved to a standalone subpackage

go/services/multipooler/internal/manager/consensus/consensus_state.go
dweitzman, line 1: The whole consensus subpackage is new, splitting off things that were previously just in multipooler/manager

go/services/multipooler/internal/manager/consensus/replicaident.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+// MaxApplicationNameLength is the maximum length of a PostgreSQL application_name (NAMEDATALEN - 1).
+const MaxApplicationNameLength = 63
+
+// ReplicaID is a validated pooler identity that can be safely encoded as a
+// PostgreSQL application_name (format: {cell}_{name}). The validation rules
+// (no underscores, ≤63 chars) are driven by PostgreSQL's application_name
+// constraints. Construct via NewReplicaID or NewReplicaIDFromAppName.
+type ReplicaID struct {
```
dweitzman, line 34: This logic was previously in `pg_replication.go` and it was called `poolerID`, but the only purpose of `poolerID` was to help with validating what topo IDs are valid replication names in sync standby names + primary conninfo so renaming it to `ReplicaID` and putting it under consensus/ feels reasonable

go/services/multipooler/internal/manager/pg_paths.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/multigres/multigres/go/common/constants"
+)
+
+// postgresDataDir returns the PostgreSQL data directory path from the PGDATA env var.
+func postgresDataDir() string {
```
dweitzman, line 25: These were previously in consensus term storage code, but they're not consensus specific

mkindahl: It would be reasonable to have a separate package for the Postgres<->Multigres connection code, possibly some common code between multipooler and pgctld, but not necessary for this PR.

mkindahl (approved):
LGTM

PR #953, opened 2026-05-04
feat(multiorch): add Recruit/Propose rule_change workflow

dweitzman:
Introduces coordinatorLedRuleChange which orchestrates the full Recruit → pre-validate → Propose flow for coordinator-initiated failover. The implementation parallelizes both recruitment and proposal dispatch: each node is proposed to immediately once it has been recruited and a viable proposal is ready.

This PR doesn't handle appointing the initial leader -- that'll be a follow-up in https://github.com/multigres/multigres/pull/971

The Coordinator gains a useNewFlow flag (default false, configurable via use_new_consensus_flow) that gates the new path in AppointLeader. The old BeginTerm/EstablishLeadership path is completely unchanged.

The new flow requires DurabilityPolicy to always be present in committed rules, since BuildSafeProposal reads it for quorum validation. The rule_store sentinel row (written during createSidecarSchema at bootstrap) is now initialized with the bootstrap durability policy, making the durability_policy columns NOT NULL. Subsequent Promote operations carry the policy forward via COALESCE, so the new Recruit/Propose failover path always finds a non-nil policy in the committed rule.

Another small change: observePosition now always returns a non-nil position on success, which should have already been the case but now it's explicit in the code.


go/services/multiorch/consensus/rule_change.go
```diff
@@ -0,0 +1,305 @@
...
+		tryBuild:      tryBuild,
+		checkPossible: checkPossible,
+	}
+}
+
+// Run executes the rule change: derive term, pre-validate, recruit all nodes
+// concurrently, and propose as soon as a viable proposal can be assembled.
+// Each node receives its Propose immediately once it has been recruited and
+// the proposal is ready — there is no unnecessary waiting between the two.
+func (r *coordinatorLedRuleChange) Run(ctx context.Context, cohort []*multiorchdatapb.PoolerHealthState) error {
```
dweitzman, line 64: This is the highlight of the PR. It's the logic to:
- See if a rule change is even possible, given the reachable poolers
- Recruit as many poolers as possible
- As recruitment results come back, start Propose()ing immediately once it's safe to do so
- Send Propose() to all poolers but wait for a successful Recruit() on that pooler -- we don't have implicit recruitment at this time

Importantly, this file has essentially no understanding of the consensus algorithm. This code outsources all consensus knowledge to `go/common/consensus` and focuses as much as possible just on sending RPCs

go/services/multiorch/config/config.go
```diff
@@ -234,6 +235,12 @@ func NewConfig(reg *viperutil.Registry) *Config {
 			Dynamic:  true,
 			EnvVars:  []string{"MT_LEADER_POSTGRES_RESPONSE_THRESHOLD"},
 		}),
+		useNewConsensusFlow: viperutil.Configure(reg, "use-new-consensus-flow", viperutil.Options[bool]{
+			Default:  false,
```
dweitzman, line 239: All of this code is disabled by default, but I verified that dead primary test still passes if this default is changed to `true`

go/services/multipooler/manager/rule_store.go
```diff
@@ -198,19 +212,20 @@ func (rs *ruleStore) createRuleTables(ctx context.Context) error {
 		leader_id                 TEXT,
 		coordinator_id            TEXT,
 		cohort_members            TEXT[] NOT NULL,
-		durability_policy_name    TEXT,
-		durability_quorum_type    TEXT,
-		durability_required_count INT,
+		durability_policy_name    TEXT NOT NULL,
```
dweitzman, line 215: I made the durability policy non-nullable, since we know a bootstrap policy when `initdb` runs anyway

go/services/multipooler/manager/rule_store.go
```diff
@@ -271,14 +284,14 @@ func (rs *ruleStore) observePosition(ctx context.Context) (*clustermetadatapb.Po
 		return nil, mterrors.Wrap(err, "failed to query current position")
 	}
 	if len(result.Rows) == 0 {
-		return nil, nil
+		return nil, mterrors.Errorf(mtrpcpb.Code_INTERNAL, "current_rule sentinel row missing for shard 0: tables may not be initialized")
```
dweitzman, line 287: This scenario should not have been possible anyway, but it seems safest to return an error if no rule can be found instead of returning a nil rule

go/services/multiorch/consensus/coordinator.go
```diff
@@ -77,6 +85,22 @@ func (c *Coordinator) AppointLeader(ctx context.Context, shardID string, cohort
 		return mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT, "cohort is empty for shard %s", shardID)
 	}
 
+	if c.useNewFlow {
```
dweitzman, line 88: This is where we conditionally use the new logic, if the flag is enabled

go/services/multiorch/consensus/rule_change.go
```diff
@@ -0,0 +1,305 @@
...
+	healthByID map[string]*multiorchdatapb.PoolerHealthState,
+) (*consensusdatapb.CoordinatorProposal, error) {
+	if result.BestRule == nil {
+		return nil, errors.New("no committed rule found; use bootstrap path for fresh clusters")
+	}
+
+	var leader *clustermetadatapb.ConsensusStatus
+	for _, cs := range result.EligibleLeaders {
+		key := topoclient.ClusterIDString(cs.GetId())
+		if health, ok := healthByID[key]; ok && types.LeaderNeedsReplacement(health) {
```
dweitzman, line 286: `EligibleLeaders` is a list of all eligible leaders from a purely consensus point of view, and `LeaderNeedsReplacement()` is used to make sure we don't choose a leader if it's resigned

go/services/multiorch/consensus/rule_change.go
```diff
@@ -0,0 +1,305 @@
...
+			LeaderId:         leader.GetId(),
+		},
+	}, nil
+}
+
+// checkRecentAcceptance returns an error if any node in the cohort recently
+// accepted a term revocation, which may indicate another coordinator is making
+// a rule change.
+func checkRecentAcceptance(ctx context.Context, logger *slog.Logger, cohort []*multiorchdatapb.PoolerHealthState) error {
+	const backoffWindow = 4 * time.Second
```
dweitzman, line 323: The 4-second grace period for other coordinators to finish an intervention is ported from `preVote()`

A difference in behavior is that preVote() looked at when poolers accepted a BeginTerm() vs this logic is looking at when a coordinator created a `TermRevocation`

go/common/consensus/proposals.go
```diff
@@ -161,6 +161,31 @@ func BuildSafeProposal(
 	return proposal, nil
 }
 
+// CheckProposalPossible checks whether a safe leadership proposal is feasible
+// given current cached statuses, without requiring nodes to have accepted the
+// revocation. It filters to nodes that could accept (via ValidateRevocation),
+// simulates their acceptance so BuildSafeProposal can evaluate quorum and
+// eligible-leader constraints, and returns an error if no viable proposal exists.
+func CheckProposalPossible(
```
dweitzman: This is placeholder code. It'll go away once https://github.com/multigres/multigres/pull/928 is merged

go/common/consensus/revocation.go
```diff
@@ -19,13 +19,35 @@ import (
...
 	"github.com/multigres/multigres/go/common/topoclient"
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	"github.com/multigres/multigres/go/tools/pgutil"
 )
 
+// NewTermRevocation constructs a TermRevocation for a new coordinator-led rule
+// change. The revocation term is max(node accepted terms, committed rule terms)
+// across the provided statuses, incremented by one. Returns term 1 for a fresh
+// cluster where no prior terms are observed.
+func NewTermRevocation(statuses []*clustermetadatapb.ConsensusStatus, coordinatorID *clustermetadatapb.ID) *clustermetadatapb.TermRevocation {
```
dweitzman, line 36: This duplicates some things in https://github.com/multigres/multigres/pull/928 so it'll go away after rebasing

go/services/multiorch/consensus/coordinator.go
```diff
@@ -77,6 +85,22 @@ func (c *Coordinator) AppointLeader(ctx context.Context, shardID string, cohort
...
+		poolerByID, healthByID := buildCohortMaps(cohort)
+		buildProposalFn := func(result commonconsensus.RecruitmentResult) (*consensusdatapb.CoordinatorProposal, error) {
+			return buildFailoverProposal(ctx, c.logger, result, poolerByID, healthByID)
+		}
+		return c.newRuleChange(
```
rafael: I think for clarity, I would encapsulate all the details of the ruleChange in a helper function. Something like:
```
	poolerByID, healthByID := buildCohortMaps(cohort)
	buildProposal := func(r commonconsensus.RecruitmentResult) (*consensusdatapb.CoordinatorProposal, error) {
		return buildFailoverProposal(ctx, c.logger, r, poolerByID, healthByID)
	}

	tryBuild := func(rev *clustermetadatapb.TermRevocation, statuses []*clustermetadatapb.ConsensusStatus) (*consensusdatapb.CoordinatorProposal, error) {
		return commonconsensus.BuildSafeProposal(rev, statuses, buildProposal)
	}
	checkPossible := func(rev *clustermetadatapb.TermRevocation, statuses []*clustermetadatapb.ConsensusStatus) error {
		return commonconsensus.CheckProposalPossible(rev, statuses, buildProposal)
	}

	return c.newRuleChange(reason, tryBuild, checkPossible).Run(ctx, cohort)
}
```

Then from the context of this code, we just call: 
```
if c.useNewFlow {
	return c.runFailover(ctx, cohort, reason)
}
```

Maybe something like performFailover is even clearer than run?

go/services/multiorch/consensus/rule_change.go
```diff
@@ -0,0 +1,330 @@
...
+	multiorchdatapb "github.com/multigres/multigres/go/pb/multiorchdata"
+)
+
+// coordinatorLedRuleChange orchestrates the recruit → check → propose workflow
+// for a coordinator-initiated rule change. It is parameterized by action-specific
+// callbacks so the same workflow serves both normal failover and bootstrap.
+type coordinatorLedRuleChange struct {
+	coordinator   *Coordinator
+	reason        string
+	tryBuild      func(*clustermetadatapb.TermRevocation, []*clustermetadatapb.ConsensusStatus) (*consensusdatapb.CoordinatorProposal, error)
```
rafael: Two ideas here: 
 
* how about something proposalBuilder? I feel tryBuild is hard to understand what this is for. 
* Similar for checkPossible, could be preFlight, or a name that conveys more directly that the intent of this function is that it helps short-circuit when we know there is no chance of winning given the already known view of the world

go/services/multiorch/consensus/rule_change.go
```diff
@@ -0,0 +1,330 @@
...
+
+	sendPropose := func(p *multiorchdatapb.PoolerHealthState) {
+		isLeader := topoclient.ClusterIDString(p.MultiPooler.Id) == leaderKey
+		go func() {
+			err := r.propose(ctx, p, propReq)
+			proposeResults <- proposeResult{poolerName: p.MultiPooler.Id.Name, isLeader: isLeader, err: err}
+		}()
+	}
+
+	// Phase 1: collect recruits until tryBuild succeeds (or we run out).
```
rafael: The approach looks consensus-correct: once tryBuild succeeds, quorum overlap guarantees the most-advanced recruit carries every committed write. Worth flagging though that it's a behavioral change from our previous implementation, which waited for the full cohort to honor the most-advanced reachable replica even when those extra writes were uncommitted.

One thought: in practice this optimization may not buy us much. Because tryBuild enforces the transition quorum, we're already waiting for most of the nodes before it can succeed. I'm not sure streaming is buying us much here, and it adds complexity to the code.

go/services/multiorch/consensus/rule_change.go
```diff
@@ -0,0 +1,330 @@
...
+
+	// Phase 2: drain remaining recruits, proposing to each as it arrives.
+	for range remaining {
+		if rr := <-recruited; rr.cs != nil {
+			sendPropose(rr.pooler)
+			recruits = append(recruits, rr.pooler)
+		}
+	}
+
+	// Wait for every Propose to return, not just the leader's. The leader's
```
rafael: nits for the comment. Something like this is clearer to me:
```
// Wait for every Propose to return, not just the leader's. The rule
// change is committed when the leader's Propose succeeds — which can
// only happen because enough followers endorsed the proposal first to
// satisfy the cohort's durability quorum. Returning early here would
// let Run's context cancel the in-flight non-leader Proposes. The shard
// is already safe at that point, but any pooler whose Propose was
// cancelled wouldn't learn the new primary on this round and would
// need to be re-wired before it could route correctly. Waiting avoids
// that follow-up.
```

go/services/multiorch/consensus/rule_change.go
```diff
@@ -0,0 +1,330 @@
...
+	// success alone commits the rule change, but returning early would let
+	// Run's context cancel the in-flight non-leader Proposes. Letting them
+	// complete keeps the shard healthy: every pooler learns the new primary,
+	// even those whose endorsement wasn't required for the decision.
+	var leaderErr error
+	for range len(recruits) {
+		pr := <-proposeResults
+		if pr.err != nil {
+			if pr.isLeader {
+				leaderErr = pr.err
```
rafael, line 203: could we abort early if there is a leader error? This seems like the other side of the coin for building a quick proposal. If we get a leader failure, there is no point in waiting for the propose results. We should abort I think

go/services/multiorch/consensus/coordinator.go
dweitzman: Done

go/services/multiorch/consensus/rule_change.go
dweitzman: Done

go/services/multiorch/consensus/rule_change.go
dweitzman: Done

go/services/multiorch/consensus/rule_change.go
dweitzman: The tradeoff here seems to be whether we want the fastest possible failover or whether we want to reduce the chances that a pg_rewind is necessary

I think we could split the difference, but tune it in the future: I've put a TODO here that we could consider adding a configurable grace period after the proposal can be satisfied where we still allow slower Recruit() responses to arrive and influence the ultimate proposal.

I think we'll have to learn through experience how best to balance speed of failover vs risk of needing a pg_rewind. My inclination is to start by optimizing for speed of failover, but I wouldn't fight for it if you or others strongly prefer optimizing for low chance of rewind

go/services/multiorch/consensus/rule_change.go
dweitzman: What do you think of "tryBuildProposal" and "checkProposalPossible"? I've updated the code with those

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -255,7 +255,7 @@ func (pm *MultiPoolerManager) executeRevoke(ctx context.Context, term int64, res
 	if nodePosition, err := pm.rules.observePosition(ctx); err != nil {
 		pm.logger.WarnContext(ctx, "Failed to get rule history during revoke; candidate selection may be suboptimal",
 			"term", term, "error", err)
-	} else if nodePosition != nil {
```
rafael, line 258: to check my understanding. This is because we now return error if position is nil.

nit - should this be call `pos` to make it consistent with other places?

go/services/multipooler/manager/rule_store.go
rafael: ah makes sense.

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -324,7 +324,7 @@ func (pm *MultiPoolerManager) Status(ctx context.Context) (*multipoolermanagerda
 	// Get cohort members from the current rule (best-effort).
 	if pos, err := pm.rules.observePosition(ctx); err != nil {
 		pm.logger.WarnContext(ctx, "Failed to read current rule for status", "error", err)
-	} else if pos != nil && pos.Rule != nil {
```
rafael, line 327: this is because now we return error if the rule is nil?

rafael (approved):
LGTM

go/services/multipooler/manager/rpc_manager.go
dweitzman: Yes, exactly

go/services/multipooler/manager/rpc_consensus.go
dweitzman: Yes, exactly. Good catch on the naming. Renamed `nodePosition` -> `pos`

PR #954, opened 2026-05-05
feat(multipooler): client-side PG TLS with libpq sslmode

haritabh17:
Closes [MUL-370](https://linear.app/supabase/issue/MUL-370/client-side-pg-tls-multipooler-to-postgresql).

## Summary

- Adds libpq-style `sslmode` (`disable | allow | prefer | require | verify-ca | verify-full`) on the multipooler → PostgreSQL leg, exposed via `--pg-client-sslmode` and `--pg-client-sslrootcert`.
- Implements the TLS upgrade in `pgprotocol/client` so the SSLRequest path actually negotiates TLS instead of erroring with "TLS upgrade not yet implemented".
- `verify-full` matches the server certificate against SAN with a CN fallback when no SANs are present, matching libpq (Go's stdlib verifier removed CN matching in 1.17).
- Defaults to `prefer` to match libpq; Unix-socket dials always run plaintext, also matching libpq.
- New end-to-end test boots a shard with postgres provisioned for SSL and asserts the multipooler's connections show up in `pg_stat_ssl` as encrypted.

## Description

The multipooler's connection to PostgreSQL was plaintext. Multigateway already terminates TLS for incoming PG clients, but anything past the gateway was on the wire in the clear. This wires up the missing leg.

The work is split across two layers:

`go/common/pgprotocol/client/ssl.go` introduces an `SSLMode` enum with `ParseSSLMode`, `AttemptsTLS`, and `RequiresTLS` helpers, plus `BuildTLSConfig` which builds a `*tls.Config` per mode:

- `disable`/`allow` → returns `(nil, nil)` (allow's plaintext-then-TLS retry isn't a TLS-config concern).
- `prefer`/`require` → encryption only, no certificate verification, matching libpq.
- `verify-ca` → chain validated against `sslrootcert`, hostname not checked.
- `verify-full` → chain validated and `ServerName=host`. A custom `VerifyConnection` re-runs the chain check, then tries `Certificate.VerifyHostname` (SAN-based), and falls back to a case-insensitive CN compare only when the leaf has no SAN entries — RFC 6125 / libpq behavior.

`go/common/pgprotocol/client/startup.go` now sends the SSLRequest when `SSLMode.AttemptsTLS()` is true and the dial is TCP. On `'N'`, tolerant modes (`prefer`) continue plaintext while strict modes (`require`/`verify-*`) error. On `'S'`, the connection is upgraded in place via `tls.Client(...).HandshakeContext(ctx)` and the buffered reader/writer are rebuilt against the TLS conn. The pre-existing buffer is checked for stray data first; bytes after `'S'` would belong to the TLS record stream.

The multipooler side adds `--pg-client-sslmode` and `--pg-client-sslrootcert` to `connpoolmanager.Config`. `MultiPoolerManager.openConnectionsLocked` builds the `*tls.Config` once at startup (only on TCP — Unix-socket connections always run plaintext) and stores it on `ConnectionConfig`. `connpoolmanager.Manager.buildClientConfig` then propagates `SSLMode` and `TLSConfig` to every dial it builds, so admin and per-user pools share the same negotiation.

`MultiPoolerManager.openConnectionsLocked` also defaults `ConnectionConfig.Host` to the multipooler's own hostname when no Unix socket is configured, since postgres is colocated with pgctld on the same host. This is what enables a TCP dial in test fixtures and any deployment that omits the socket.

## Integration test

`shardsetup.WithMultipoolerPGTLS` provisions postgres with SSL on every pgctld in the setup and points each multipooler at the matching CA. It also installs a permissive `pg_hba` template that trusts 127.0.0.1 over the encrypted channel so per-user pools authenticate without SCRAM passthrough wiring, and clears the Unix socket on each multipooler so the dials go over TCP. `TestPGClientTLS_VerifyFull` boots a shard with the option, runs `SELECT 1` through the multipooler, and asserts `pg_stat_ssl` on the postgres side shows at least one TLS-encrypted backend that isn't the test's own connection.

## Out of scope

Client cert auth (`--pg-client-sslcert` / `--pg-client-sslkey`) and CRLs are deferred to MUL-383 under MUL-189. Supabase's managed PG doesn't expose `clientcert`-based auth.

## Test plan

- [x] Unit: `go test ./go/common/pgprotocol/client/...` — 12 tests covering `ParseSSLMode`, mode flags, `BuildTLSConfig` per mode, missing/empty rootcert, SAN match, hostname mismatch rejection under verify-full, CN fallback when no SANs, and CN suppression when SANs are present.
- [x] Integration: `go test -run TestPGClientTLS_VerifyFull ./go/test/endtoend/multipooler/` — passes consistently (`-count=2` clean), ~24s per run.
- [x] Regression: `go test ./go/services/multipooler/connpoolmanager/...` clean.


GuptaManan100 (approved):
LGTM!

PR #955, opened 2026-05-05
fix(multigateway): revert SET/RESET on ROLLBACK and ROLLBACK TO SAVEPOINT

haritabh17:
## Summary

- Adds a per-savepoint snapshot stack so SET/RESET issued inside a transaction or savepoint is reverted on ROLLBACK and ROLLBACK TO SAVEPOINT, matching PostgreSQL's GUC nested-stack semantics.
- Covers both `SessionSettings` (backend-managed GUCs) and `GatewayManagedVariable[T]` (e.g. `statement_timeout`) under one lockstep stack invariant.
- Closes MUL-392.

## Problem

Gateway-managed GUC values and `SessionSettings` did not revert on `ROLLBACK` or `ROLLBACK TO SAVEPOINT`. A `SET` issued inside a transaction or savepoint persisted after the transaction was undone, diverging from PostgreSQL's GUC stack semantics. The prior `SET LOCAL` fix (#911) cleared local overrides at the txn boundary but had no notion of per-savepoint snapshots, so non-LOCAL `SET` and savepoint-scoped reverts were still wrong.

## Solution

Introduce a `savepoints []savepointFrame` stack on `MultiGatewayConnectionState` and a parallel `snapshots []gmvSnapshot[T]` stack on every `GatewayManagedVariable`, kept lockstep by index. Each frame captures the `SessionSettings` map plus the full `(currentValue, isSet, localValue, isLocalSet)` tuple of every gateway-managed variable.

`TransactionPrimitive` drives the stack:

- **BEGIN** pushes a depth-0 frame (idempotent for nested BEGIN).
- **SAVEPOINT** pushes a named frame after the backend accepts.
- **RELEASE** drops the named frame and any nested ones, keeping current values (PG merges sub-transaction changes into parent scope).
- **ROLLBACK TO** restores from the named frame's snapshot but leaves the frame on the stack so the savepoint stays active (matches PG).
- **COMMIT** clears the stack and any `SET LOCAL` overrides.
- **ROLLBACK** restores from the depth-0 frame, then clears the stack.

State mutations only happen on backend success — if `SAVEPOINT`/`RELEASE`/`ROLLBACK TO` is rejected by the backend, the gateway stack is untouched.

## Tests

- 18 new unit tests in `handler/savepoint_state_test.go` cover the `GatewayManagedVariable` snapshot stack and the connection-state begin/commit/rollback/savepoint matrix (including nested savepoints, RELEASE merge semantics, ROLLBACK TO leaving the savepoint active, and lockstep depth invariant).
- Existing `TransactionPrimitive` tests updated for the new `NewTransactionPrimitive(kind, savepointName, sql, tableGroup, metrics)` signature.
- All multigateway unit packages green locally.

## Verified end-to-end

- `SET` inside `BEGIN` then `ROLLBACK` reverts to pre-BEGIN value
- `SET` inside `SAVEPOINT sp` then `ROLLBACK TO sp` reverts to pre-savepoint value; `sp` remains active
- `RELEASE sp` keeps current value (sub-transaction merged into parent)
- `SET LOCAL` interaction unchanged: still cleared at COMMIT/ROLLBACK
- Backend-managed GUCs (`DateStyle`, `search_path`, etc.) and gateway-managed (`statement_timeout`) follow the same semantics


GuptaManan100 (approved):
Just one thing to note - 

**Lockstep contract between `savepointFrame` and `GatewayManagedVariable[T].snapshots` is fragile**

The two-stack split is forced by generics — `GatewayManagedVariable[T comparable]` is parameterized, so its `gmvSnapshot[T]` can't live alongside the `map[string]string` snapshot in `savepointFrame` without erasing to `any` and losing type safety. That's fine, and the current design is the right call.

The cost is that every lifecycle method (`pushFrameLocked`, `RollbackToSavepoint`, `ReleaseSavepoint`, `CommitTransaction`, `RollbackTransaction`, `BeginTransaction`) has to explicitly name `m.statementTimeout.{Snapshot,RestoreFromDepth,PopFrom,ClearSnapshots}`. With one GMV today this is readable, but the next person adding a GMV has to remember to wire it into all six call sites or revert semantics will silently break for that variable.

When a second GMV lands, suggest extracting a non-generic `gmvLifecycle` interface (`Snapshot()`, `RestoreFromDepth(int)`, `PopFrom(int)`, `ClearSnapshots()`) and holding `[]gmvLifecycle` on the connection state, so lifecycle methods iterate a registry instead of naming each GMV. For now, a comment on the struct flagging the maintenance contract would be enough.

PR #956, opened 2026-05-05
ci: dispatch proto-updated event to operator on main push

niconosenzo:
Adds workflow that fires a repository_dispatch event to multigres-operator whenever a push to main touches proto/**.

Replaces the manual proto-bump step in the operator's release flow.


Verolop (approved):
LGTM! thank you 💖

PR #957, opened 2026-05-05
refactor(consensus): implement BuildPolicyTransition

dweitzman:
This is not-yet-used code with heuristics for implementing certain types of rule transitions.


mkindahl (approved):
LGTM. It seems it is almost entirely renames and no functional changes.

dweitzman:
> LGTM. It seems it is almost entirely renames and no functional changes.

BuildPolicyTransition is new logic, but it's not yet used. I'm working up towards a big change to the rules store and wanted to try to break it into smaller parts

go/common/consensus/durability.go
```diff
@@ -162,6 +161,111 @@ func cohortIntersect(cohort []*clustermetadatapb.ID, statuses []*clustermetadata
...
+// cohort changing at once, or cohorts with no subset relationship.
+//
+// For same-N transitions Both uses the smaller cohort (the subset), ensuring any
+// ack satisfies both the old and new policy. For same-cohort transitions Both uses
+// the larger N, ensuring the WAL record is acknowledged under the stricter policy.
+// When outgoing and incoming are identical Both matches incoming.
+//
+// Both AtLeastNPolicy and MultiCellPolicy are supported; mixing the two types
+// returns an error.
+func BuildPolicyTransition(outgoing, incoming PolicyWithCohort) (*PolicyTransition, error) {
```
mkindahl, line 218: As discussed: failing a rule change makes sense when build the proposal, but once a proposal has been made and supporters collected, failure of making the rule durable (which is what leader-led rule change does by writing to the database) means that the primary need to be marked as "necessary to rewind" in some manner.

The only reasonable way the rule change can fail at this stage is if it cannot reach enough standbys to get the rule change acknowledged, for example if it ended up in a network partition that is not progressing, and in that case the rule change need to be removed and the server need to rejoin as a standby on the timeline of the network partition that actually made progress.

dweitzman: Yes, stuck scenarios would either be due to a network partition (or other loss of nodes) or because a coordinator recruited those nodes and made them promise to stop acknowledging transactions from lower terms

PR #958, opened 2026-05-05
feat(multipooler): add SyncStandbyManager for centralized GUC writes

dweitzman:
SyncStandbyManager will be the sole writer of synchronous_standby_names and synchronous_commit. Centralizing writes will help ensure the ordering of GUC changes relative to WAL writes is always correct for the transition type.

- SetPolicy() computes and applies both GUC strings from a PolicyWithCohort;
  skips ALTER SYSTEM when the in-memory cache already reflects the desired
  state, avoiding spurious pg_reload_conf calls on every action
- NeedsApply() will be used in the postgres monitor. It probes whether a
  SetPolicy call would issue ALTER SYSTEM without acquiring any lock,
  so callers can skip the promotion hook when no GUC change is needed.
- Clear() resets synchronous_standby_names after demotion; refuses to run
  unless postgres reports pg_is_in_recovery() to prevent accidental
  durability loss on a primary
- Also adds priorRuleWritesDrainedKey context-token helpers to rule_store.go;
  these will be set by the rule store in a follow-up PR and checked by
  SetPolicy to ensure GUC writes only happen if we can confirm that any
  past attempts to update the rules table are finished.


go/services/multipooler/manager/rule_store.go
```diff
@@ -765,3 +765,34 @@ func scanRuleHistoryRow(rec *ruleHistoryRecord, leaderIDStr *string, cohortNames
...
 	return nil
 }
+
+// priorRuleWritesDrainedKey is a context key proving that any in-flight rule
+// writes from a previous action lock holder have been resolved before
+// SyncStandbyManager.SetPolicy is called. This is established in one of two ways:
+//
+//   - Primary path: a SELECT FOR UPDATE on current_rule blocks until any
+//     in-progress transaction from the prior holder commits or rolls back, after
+//     which our row lock prevents new writers from interposing.
```
mkindahl, line 783: This will only block other clients connected to the same server and only for the duration of the session.

From my understanding, sessions are ephemeral and will be wrapped in a BEGIN-COMMIT pair, so the `SELECT ... FOR UPDATE` will not take a long-lived lock on the row.

go/services/multipooler/manager/rule_store.go
```diff
@@ -765,3 +765,34 @@ func scanRuleHistoryRow(rec *ruleHistoryRecord, leaderIDStr *string, cohortNames
...
+//   - Recovery path (standby before pg_promote): the node is read-only, so no
+//     concurrent writes to current_rule are possible.
+//
+// The action lock (checked separately via AssertActionLockHeld) ensures no
+// concurrent goroutine in this process can also hold this proof.
```
mkindahl, line 788: If the server is waiting for acknowledgement from the standbys to meet the durability policy, the record is written to the database and to the WAL, so if the action lock is release while this wait is happening, the rule change will be there, but not acknowledged. This can lead to a split-brain situation since it might mean that the server is in a non-progressing network partition. At this point, it need to be rewound to match what exists in the progressing network partition.

go/services/multipooler/manager/sync_standby_manager.go
```diff
@@ -0,0 +1,278 @@
...
+	unchanged := s.lastSyncCommit == g.wantCommit && s.lastStandbyNames == g.wantStandby
+	s.mu.Unlock()
+	return !unchanged, nil
+}
+
+// SetPolicy computes the Postgres GUC configuration for the given durability policy and applies it.
+// Uses an in-memory cache of the last-written GUC strings to skip ALTER SYSTEM calls when the
+// desired values haven't changed. This is safe because postgresqlSyncStandbyManager is the sole
+// writer of synchronous_commit and synchronous_standby_names.
+func (s *postgresqlSyncStandbyManager) SetPolicy(ctx context.Context, pc commonconsensus.PolicyWithCohort) error {
```
mkindahl, line 181: Do you need to assert that the action lock is held? If not, this could be a race.

go/services/multipooler/manager/sync_standby_manager.go
```diff
@@ -0,0 +1,278 @@
...
+	select {
+	case <-time.After(100 * time.Millisecond):
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+
+	s.mu.Lock()
+	s.lastSyncCommit = g.wantCommit
+	s.lastStandbyNames = g.wantStandby
+	s.mu.Unlock()
```
mkindahl, line 218: Since the lock above where you check if the values are unchanged is released, you might _in theory_ at this point overwrite `lastSyncCommit` with an older value if there is no other protection than the mutex. However, since the action lock is supposed to be held at this point, it should work.

go/services/multipooler/manager/rule_store.go
dweitzman: Correct. The action lock is the long-lived lock, and the row lock is just for making sure there isn't lingering / leftover unfinished work in progress from a previous attempt at a rule change that no longer holds the action lock

go/services/multipooler/manager/rule_store.go
dweitzman: Yes, a rewind may be necessary if a failover happens when the former primary has WAL entries that haven't reached quorum.

Personally I wouldn't use the term split brain to describe a timeline containing transactions that haven't reached durability, since the shard as a whole has a single clear and consistent truth: there are no two timelines with conflicting durable transactions. All of this will become simpler once we have a replicated commit index, assuming the commit index never advanced without reaching quorum.

go/services/multipooler/manager/sync_standby_manager.go
dweitzman: Added the action lock assertion above. This method is not designed to be concurrency safe if the action lock holder calls it from multiple different gorountines, since that seems unlikely

The expected use cases are:
- postgres monitor gets the action lock and fixes GUC being inconsistent with the last-written rule table entry
- the updateRules() method is called either from Propose() or from the UpdateConsensusRule() RPC

All of those should be fairly straightforward from a concurrency perspective: 1) get action lock 2) try to change the rules and/or GUC 3) release lock

go/services/multipooler/manager/sync_standby_manager.go
dweitzman: Done

dweitzman:
Merged in the previous PR, so this diff is clean now

rafael (approved):
LGTM, stamping to unblock. But worth double checking with @mkindahl

go/services/multipooler/manager/sync_standby_manager.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	// pg_reload_conf sends SIGHUP to postgres backends asynchronously.
+	// Sleep briefly to give pool backends time to reload before the next WAL write.
+	// SHOW on a single connection is insufficient — it only confirms one backend
+	// processed the SIGHUP; other pooled connections may still use the old value.
+	select {
+	case <-time.After(100 * time.Millisecond):
+	case <-ctx.Done():
+		return ctx.Err()
+	}
```
mkindahl: Not sure if you want to put this inside `reloadConfig` function. However, it is also used below and is strictly speaking not necessary there, but isn't harmful either.

mkindahl (approved):
LGTM. Added a comment for your consideration.

go/services/multipooler/manager/sync_standby_manager.go
dweitzman: Good call, I'll move this into `reloadConfig()`

PR #959, opened 2026-05-06
refactor(proto): replace flat MultiPooler shard fields with nested ShardKey message

mkindahl:
Move database, table_group, and shard fields out of MultiPooler into a dedicated ShardKey sub-message (field 2), matching the ShardKey type already used elsewhere. Update all call sites across production code and tests to use the nested accessor pattern, and replace field-by-field shard equality checks with proto.Equal() where all three fields are compared.


mkindahl:
The changes to [‎proto/clustermetadata.proto‎](https://github.com/multigres/multigres/pull/959#diff-d6dcb29dd0cd8597a82402bf69dfa7724d2e6c00a580a5bf0b25996ba3e01937) is the main change. Everything else is just a consequence of this change.

haritabh17 (approved):
LGTM 🚀

PR #961, opened 2026-05-06
feat(multigateway): authenticate replication startup connections

haritabh17:
Closes [MUL-387](https://linear.app/supabase/issue/MUL-387/authenticate-replication-connections-replicationtrue-at-multigateway).

## Summary

- Parse the `replication` startup parameter (`true` / `false` / `database` plus libpq aliases).
- After SCRAM completes, require `pg_authid.rolreplication` on the role; on mismatch, emit PG's exact wording (`"must be superuser or replication role to start walsender"`, SQLSTATE 42501).
- Plumb `is_replication_role` through `GetAuthCredentials` so the rolreplication lookup adds no extra `pg_authid` query at the pooler.
- Walsender / COPY-mode command handling stays a follow-up; this PR delivers the auth gate.

## Description

Replication startup connections (physical and logical) flow through the same SCRAM path as normal sessions but additionally require `rolreplication = true` on the authenticated role — without that gate, multigateway would admit replication clients that PostgreSQL itself would reject. The check is sequenced between SASLFinal and AuthenticationOk so a rejected client sees a single FATAL message rather than a successful handshake followed by a closed connection, matching native PG.

The pooler-side change extends `admin.GetRolPassword` (now `GetRolAuthInfo`) to fetch `rolreplication` in the same `pg_authid` query and surfaces it on `GetAuthCredentialsResponse`. Multigateway uses a new `RoleAttributeVerifier` interface on the listener so the auth-flow refactor stays generic; `PoolerHashProvider` implements both `PasswordHashProvider` and `RoleAttributeVerifier`. Failure modes (no verifier configured, verifier error, verifier=false) all fail closed with the same 42501 FATAL so we never admit a possibly-unauthorized walsender session.

## Test plan

- [x] Unit: `parseReplicationMode` table covers libpq aliases plus rejection of unknown values (22023).
- [x] Unit: 6 startup tests covering accept (`replication=true`, `replication=database`), reject without rolreplication, no verifier configured, verifier returns error, normal connection skips verifier.
- [x] Unit: `PoolerHashProvider.IsReplicationRole` covering true/false, NotFound, login-disabled diagnostic, and transport-error propagation.
- [x] Unit: `admin.GetRolAuthInfo` extended with a rolreplication=true case; existing rolcanlogin / rolvaliduntil precedence tests preserved.
- [x] Integration: `TestMultiGateway_Authentication/replication=true_rejected_for_role_without_rolreplication` passes against a real shard.


go/services/multipooler/pools/admin/admin_conn.go
```diff
@@ -115,51 +128,55 @@ func (c *Conn) ResetAllSettings(_ context.Context) error {
...
 	// password-based auth. PG returns the same opaque "password authentication
 	// failed" message as wrong-password in this case.
 	if !parsePgBool(row[2]) {
-		return "", fmt.Errorf("%w: %q", ErrPasswordExpired, username)
+		return nil, fmt.Errorf("%w: %q", ErrPasswordExpired, username)
 	}
 
+	info := &RolAuthInfo{
+		IsReplicationRole: parsePgBool(row[3]),
+	}
```
GuptaManan100, line 183: Native PG's gate is `has_rolreplication()` (postgres/src/backend/utils/init/miscinit.c:734-750), which returns true if the role **is a superuser OR** has `rolreplication=true`:

```c
if (superuser_arg(roleid))
    return true;
...
result = ((Form_pg_authid) GETSTRUCT(utup))->rolreplication;
```

This SELECT only fetches `rolreplication`. A superuser whose `rolreplication=false` (the default for non-explicit-replication superusers) will be rejected by multigateway even though connecting directly to PG would succeed, and the FATAL message (`"must be superuser or replication role to start walsender"`) literally tells the user superuser should work.

The PR's own doc comments acknowledge the rule, `admin_conn.go:110-112` ("PostgreSQL requires this attribute (or rolsuper)…") and `startup.go` `ReplicationPhysical` ("The role must have rolreplication=true (or rolsuper=true)…"). So the intent is correct; the SQL just doesn't implement it.

GuptaManan100: EDIT: I read the Gateway side, and I see that this code is never reached when the rolreplication is false. Worth noting as a comment, but otherwise its fine 👍

go/services/multigateway/auth/pooler_hash_provider.go
```diff
@@ -106,5 +107,48 @@ func (p *PoolerHashProvider) GetPasswordHash(ctx context.Context, username, data
...
+// connection that requested replication=true / replication=database, to
+// match PostgreSQL's "must be superuser or replication role to start
+// walsender" check (SQLSTATE 42501).
+//
+// This makes a fresh GetAuthCredentials call. Replication startups are
+// rare relative to normal traffic, so the additional roundtrip is
+// acceptable; in exchange, we avoid sharing per-call state across
+// connections on the listener-shared provider.
+//
+// Auth-failure sentinels from GetPasswordHash (rolcanlogin=false,
```
GuptaManan100: The comment refers to *"Auth-failure sentinels from GetPasswordHash"* but the function it documents (`IsReplicationRole`) calls `GetAuthCredentials`, not `GetPasswordHash`. Update the reference.

GuptaManan100 (approved):
I don't like how the same struct is implementing 2 interfaces. And also the second round trip, I get is rare, but still feels unnecessary. I'm not super concerned by it, but it will add latency to connection time for replication connections. 

Why don't we collapse the two interfaces into one, that provides both PasswordHash, and if it's a replication user. That will be cleaner code won't it? I'm not sure what 2 interfaces is giving us.

Approved, so that you can merge once comments have been addressed

GuptaManan100:
@cuongdo This might be of interest to you too, it has the changes to parse and accept when a connection is coming for replication.

haritabh17:
> Why don't we collapse the two interfaces into one, that provides both PasswordHash, and if it's a replication user. That will be cleaner code won't it? I'm not sure what 2 interfaces is giving us.

Collapsed. Removed scram.PasswordHashProvider and server.RoleAttributeVerifier; introduced single server.CredentialProvider returning {Hash,IsReplicationRole}. ScramAuthenticator now takes a pre-fetched hash, so SCRAM does no I/O. authenticateSCRAM fetches once, caches creds on Conn, hands hash to SCRAM; verifyReplicationRole reads cached flag — no second pooler RPC. PoolerHashProvider → PoolerCredentialProvider, implements one interface.

One behavior change to flag: generic credential-lookup failure (pooler unreachable) on a replication startup now produces 28P01 before SCRAM instead of post-SCRAM 42501 but with the same outcome (fail closed), so seems okay.

go/services/multigateway/auth/pooler_hash_provider.go
haritabh17: Obsolete after the interface collapse — IsReplicationRole method removed entirely. Single GetCredentials returns hash + rolreplication in one shot.

go/services/multipooler/pools/admin/admin_conn.go
haritabh17: Ack, added a comment in GetRolAuthInfo documenting the assumption.

Issue #406, opened 2025-12-30
feat(telemetry): Add always-on Prometheus /metrics endpoint for debugging


ixchio:
## Summary

Add a dedicated Prometheus exporter endpoint at `/metrics` that is always available for debugging and observability purposes.

## Background

There is a TODO in the codebase at `go/tools/telemetry/telemetry.go:231`:
```go
// TODO(dweitzman): Add an additional prometheus exporter that's always at /metrics for debugging
```

## Motivation

Having a consistent `/metrics` endpoint across all Multigres components would:
1. **Improve Observability**: Enable operators to easily scrape metrics from Multigres components using standard Prometheus tooling
2. **Facilitate Debugging**: Provide real-time visibility into component health and performance
3. **Align with Industry Standards**: Most cloud-native infrastructure (like Vitess itself) exposes Prometheus metrics at a standard endpoint

## Proposed Implementation

1. Add a consistent `/metrics` HTTP endpoint to all Multigres server components (multigateway, multipooler, multiorch, pgctld)
2. Expose relevant metrics:
   - Connection pool stats (active connections, idle connections, wait time)
   - Query latencies (histogram)
   - gRPC call counts and latencies
   - Component health status
3. Make the endpoint configurable but enabled by default

## Related

This aligns with the existing Open Telemetry setup mentioned in the architecture docs.

cc @dweitzman (original TODO author)

dweitzman:
My thoughts on this have evolved since writing that TODO: in the spirit of prioritizing simplicity, for now I'd like to see how much mileage we can get out of deferring telemetry export problems to `opentelemetry-go-contrib`'s `autoexport`.

It's currently possible to set up a Prometheus `/metrics` endpoint with `OTLP_METRICS_EXPORTER=prometheus` environment variable (optionally with `OTEL_EXPORTER_PROMETHEUS_HOST` or `OTEL_EXPORTER_PROMETHEUS_PORT`). The catch is that it needs to be on a different HTTP port than the servenv debug info, and if you're using the local provisioner there's no graceful way to give different ports to different services and know what those ports are.

[...]

PR #965, opened 2026-05-06
feat(telemetry): expose shared Prometheus metrics endpoint

Verolop:
## Description

this PR adds shared Prometheus-compatible `/metrics` exposure for Multigres Go services through the common servenv runtime layer. Services that initialise through servenv, including multipooler, multigateway, multiorch, and pgctld, now expose the same scrape endpoint without each service needing its own metrics handler, so the operator can use it to scrape service, runtime, and process metrics for dashboards and alerts.

This is not a replacement for the existing OTel autoexport/OTLP configuration.

## Main Changes

- Add `/metrics` to the common servenv HTTP endpoints so multipooler, multigateway, multiorch, and pgctld inherit the same scrape path.
- Add an OTel Prometheus exporter as an additional metric reader, preserving the existing autoexport/OTLP metric path.
- Register Go runtime and process collectors in a service-local Prometheus registry so scrapes include go_*, process_*, and OTel service metrics.

## Testing
Adds coverage in the servenv telemetry integration test to exercise the full scrape path. The test records a counter through OpenTelemetry, scrapes /metrics, and verifies that the recorded metric appears in Prometheus output.

The test also verifies that the shared endpoint includes Go runtime and process collector metrics, proving the scrape output contains both Multigres service metrics and baseline process health metrics.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	github.com/lib/pq v1.10.9
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/pganalyze/pg_query_go/v6 v6.2.2
+	github.com/prometheus/client_golang v1.23.2
 	github.com/quasilyte/go-ruleguard/dsl v0.3.23
 	github.com/spf13/afero v1.15.0
 	github.com/spf13/cobra v1.10.2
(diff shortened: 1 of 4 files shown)


dweitzman:
It's possible to export prometheus metrics today with `OTLP_METRICS_EXPORTER=prometheus`

Does that not meet your needs? Some additional context: https://github.com/multigres/multigres/issues/406#issuecomment-3702722451

Verolop:
closing in favour of a new implementation since internal needs shifted from pulled to pushed metrics

PR #967, opened 2026-05-07
feat(multipooler): stamp multigres_vpid on PG backend application_name

haritabh17:
## Summary

- Plumb the multigateway client connection's virtual PID (vpid) to the multipooler over gRPC via a new `ExecuteOptions.client_connection_id` field.
- Multipooler stamps `multigres_vpid:<id>` on the PostgreSQL backend's `application_name` for every reserved and regular pool connection, so observability and lock-detection code can map a multigateway vpid back to the real backend PID via `pg_stat_activity`.
- Surface the real PostgreSQL backend PID on `ReservedState.backend_process_id` so the gateway can answer pid-bearing queries (`pg_stat_activity`, etc.) without an extra RPC.
- Switch the session-settings end-to-end test off `application_name`, which is now reserved for the vpid stamp; honest client-visible `application_name` round-tripping is tracked separately.

### Proto

`proto/query.proto` gains two pure additive fields:

- `ExecuteOptions.client_connection_id` (uint32, field 8) — the gateway's vpid for the client connection.
- `ReservedState.backend_process_id` (uint32, field 4) — the real PG backend PID for a reserved connection.

No behaviour change until callers wire them.

### Pools

`pools/regular` and `pools/reserved` each grow a `SetApplicationName(ctx, name)` method that issues a session-level `SET application_name = '<escaped>'` on the underlying connection, with correct single-quote escaping. This is the only place the rest of the multipooler should set `application_name` on a backend.

### Executor

Three coupled pieces:

- `sessionSettingsForPool` filters `application_name` (case-insensitive) out of any client-supplied `options.SessionSettings` before they reach the pool. Without this, the pool's connstate cache would track a client `application_name`; a later `ApplySettings` diff against the filtered desired map would emit `RESET application_name` and wipe the vpid stamp set out-of-band by `SetApplicationName`.
- `stampVpidOnReserved` / `stampVpidOnRegular` helpers issue `SET application_name = 'multigres_vpid:<id>'` on a freshly acquired pool conn. Called at every acquire point: `ExecuteQuery` / `StreamExecute` (reserved + regular), `reserveAndStreamExecute` for fresh reservations, and `portalExecuteWithReserved` (both new-reservation and existing-reservation branches, after `ApplySettingsToConn` so the stamp survives any `RESET ALL` it might issue).
- `buildReservedStateFromAPI` populates `BackendProcessId` from the underlying pgx conn via a new `ProcessID()` method on the `reservedConnAPI` interface; the test mock is updated to match.

### Scatterconn

Every `ExecuteOptions` constructed in `go/services/multigateway/scatterconn` (eleven sites covering read/write execute, portal execute, copy ready/data/finalize, etc.) now carries `ClientConnectionId: conn.ConnectionID()` so every shard request to the multipooler carries the vpid.

### Tests

`go/test/endtoend/queryserving/session_settings_test.go` previously used `application_name` as a generic GUC probe in three sub-tests (session-pool persistence, `RESET ALL` semantics, cross-connection isolation). Those sub-tests exercise generic GUC behaviour, not anything specific to `application_name`, so they switch to `search_path` and `client_min_messages`, which behave identically for these scenarios and are not shadowed by the multipooler.

### DIFF
--- a/go/services/multipooler/init.go
+++ b/go/services/multipooler/init.go
@@ -157,7 +157,7 @@ func NewMultiPooler(telemetry *telemetry.Telemetry) *MultiPooler {
 			Dynamic:  false,
 		}),
 		vpidStampEnabled: viperutil.Configure(reg, "vpid-stamp-enabled", viperutil.Options[bool]{
-			Default:  true,
+			Default:  false,
 			FlagName: "vpid-stamp-enabled",
 			Dynamic:  false,
 		}),
--- a/go/services/multipooler/init.go
+++ b/go/services/multipooler/init.go
@@ -56,6 +56,10 @@ type MultiPooler struct {
 	pgBackRestKeyFile  viperutil.Value[string]
 	pgBackRestCAFile   viperutil.Value[string]
 	pgBackRestPort     viperutil.Value[int]
+	// vpidStampEnabled controls stamping of multigres_vpid:<id> on PostgreSQL
+	// backends so lock-detection can map a multigateway virtual PID to the
+	// real backend PID via pg_stat_activity.application_name.
+	vpidStampEnabled viperutil.Value[bool]
 	// GrpcServer is the grpc server
 	grpcServer *servenv.GrpcServer
 	// Senv is the serving environment
@@ -152,6 +156,11 @@ func NewMultiPooler(telemetry *telemetry.Telemetry) *MultiPooler {
 			FlagName: "pgbackrest-port",
 			Dynamic:  false,
 		}),
+		vpidStampEnabled: viperutil.Configure(reg, "vpid-stamp-enabled", viperutil.Options[bool]{
+			Default:  true,
+			FlagName: "vpid-stamp-enabled",
+			Dynamic:  false,
+		}),
 		grpcServer:     servenv.NewGrpcServer(reg),
 		senv:           servenv.NewServEnvWithConfig(reg, servenv.NewLogger(reg, telemetry), viperutil.NewViperConfig(reg), telemetry),
 		telemetry:      telemetry,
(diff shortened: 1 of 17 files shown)


go/services/multipooler/pools/regular/regular_conn.go
```diff
@@ -208,6 +208,18 @@ func (c *Conn) State() *connstate.ConnectionState {
...
 // --- Query execution ---
 
+// SetApplicationName sets application_name on the underlying PostgreSQL
+// connection. Used to tag the backend with the client's virtual PID so
+// lock-detection functions (e.g. an override of
+// pg_isolation_test_session_is_blocked) can map vpid → real pid via
+// pg_stat_activity. Must be re-applied on every client hand-off for pooled
+// connections since the backend is shared across clients.
+func (c *Conn) SetApplicationName(ctx context.Context, name string) error {
+	escaped := strings.ReplaceAll(name, "'", "''")
```
GuptaManan100: We should use `ast.QuoteStringLiteral` instead. It pretty much does this, but let's use that function.

go/services/multipooler/pools/reserved/reserved_conn.go
```diff
@@ -311,6 +312,20 @@ func (c *Conn) SecretKey() uint32 {
 
 // --- Query execution ---
 
+// SetApplicationName sets the application_name on the underlying PostgreSQL
+// connection. Used to tag the backend with the client's virtual PID for
+// lock-detection mapping via pg_stat_activity.
+func (c *Conn) SetApplicationName(ctx context.Context, name string) error {
+	_, err := c.pooled.Conn.Query(ctx, "SET application_name = "+quoteStringLiteral(name))
```
GuptaManan100: This can just delegate the underlying conn's `SetApplicationName` so that it's implemented in 1 place. Pretty much like the other functions.

go/services/multipooler/init.go
```diff
@@ -152,6 +156,11 @@ func NewMultiPooler(telemetry *telemetry.Telemetry) *MultiPooler {
 			FlagName: "pgbackrest-port",
 			Dynamic:  false,
 		}),
+		vpidStampEnabled: viperutil.Configure(reg, "vpid-stamp-enabled", viperutil.Options[bool]{
+			Default:  true,
```
GuptaManan100: Don't we want this to default false and let this be opt-in?

GuptaManan100 (approved):
Can be merged once 👇 are fixed

go/services/multipooler/init.go
haritabh17: Ack, updated

PR #969, opened 2026-05-08
feat: graceful shutdown for multipooler on SIGTERM

mkindahl:
# feat(multipooler): graceful shutdown on SIGTERM

When a multipooler receives SIGTERM, it now runs a coordinated shutdown that lets the orchestrator elect a new leader cleanly instead of relying on stream-EOF detection alone. The pooler drains its own in-flight queries, closes its connection pool, stops postgres via pgctld, broadcasts a terminal `LIFECYCLE_SIGNAL_STOPPED` snapshot, marks itself `DRAINED` in topology, and only then exits. The orchestrator reacts to the STOPPED snapshot by triggering replacement immediately, without waiting for the usual stream-disconnect backoff.

## Sequence (multipooler)

The shutdown handler is registered as an `OnTermSync` hook in `manager.Start`. On SIGTERM the multipooler:

1. Arms a safety-net `time.AfterFunc` that force-exits the process by `gracefulShutdownAnnouncedDeadline` (60s) if shutdown overruns. Before the force-exit it runs a best-effort, bounded telemetry flush so OTel batchers drain. This bounds how long a hung shutdown can keep the pooler alive serving writes after the operator asked it to leave.
2. Transitions serving status to `NOT_SERVING` via the existing `OnStateChange` path. This rejects new gateway queries, drains in-flight queries (bounded by `--connpool-drain-grace-period`), and force-closes reserved connections on grace-period timeout.
3. Calls `connPoolMgr.Close()` to release idle backends to postgres. Idle connections are closed immediately; any non-reserved borrows still in flight from step 2 are disposed when they return, bounded by `PoolCloseTimeout`. In the common case step 2 has already brought `lentCount` to zero, so `Close` returns near-instantly.
4. Calls `pgctld.Stop` under the action lock, escalating fast → immediate. Smart shutdown is deliberately not in the chain: the pooler has already done what smart would wait for.
5. Broadcasts a terminal `LIFECYCLE_SIGNAL_STOPPED` snapshot via `healthStreamer.Shutdown`. The terminal broadcast uses **blocking sends** bounded by `--graceful-shutdown-final-snapshot-timeout` (default 5s) so the STOPPED snapshot is not silently dropped under buffer pressure (unlike ordinary heartbeats, which use non-blocking sends and drop on a full buffer). After the snapshot lands, subscriber channels are closed, which lets servenv's `grpcServer.GracefulStop` release without waiting for the lameduck-period timeout.
6. Marks its own topology entry as `Type=DRAINED, ServingStatus=NOT_SERVING` via `UpdateMultiPoolerFields` on a fresh bounded context. Best-effort: a topology-write failure is logged but does not block shutdown.
7. Returns from the hook so the process can exit.

## Reaction (multiorch)

On the orchestrator side, `health_stream.go` reacts to two events:

1. A `STOPPED` snapshot. `handleLifecycleSnapshot` fires `synthesizeRequestingDemotion`, which writes `LeadershipSignal_REQUESTING_DEMOTION` onto the cached `LeadershipStatus` for the topology primary at the current consensus `primary_term`. `LeaderNeedsReplacement` then reads that signal with its usual staleness check, and the existing voluntary-handoff machinery (the `LEADERSHIP_SIGNAL_REQUESTING_DEMOTION` path already exercised by `EmergencyDemote`) takes over. Synthesis is idempotent and is a no-op for non-primary poolers.
2. A stream EOF on a stream whose cached `LifecycleSignal` is `STOPPED`. The orchestrator skips the usual reconnect-with-backoff (the pooler is intentionally gone) and re-fires `synthesizeRequestingDemotion` as a no-op safety net.

There is no on-wire shutdown deadline. The pooler's local safety-net force-exit bounds how long a hung shutdown can stay alive; the orchestrator never receives or interprets that value. The 60s deadline is set in code as `gracefulShutdownAnnouncedDeadline` and used only locally to size the safety-net.

## Proto additions

`AvailabilityStatus` gains an optional `LifecycleStatus` sub-message with a single field:

- `LifecycleSignal` — enum with values `UNKNOWN`, `RUNNING`, `STOPPED`. `UNKNOWN` is reserved for "not set"; `RUNNING` is the explicit healthy state.

## Shared constant

`go/common/constants/lifecycle.go` defines:

- `MinShutdownDeadline = 10 * time.Second` — the floor on the multipooler's graceful-shutdown deadline. A compile-time assertion in the multipooler manager rejects any value below this floor so misconfiguration surfaces at build time rather than at runtime.

## Configuration

Three timeout flags on multipooler bound the post-drain phases:

| Flag | Default | Purpose |
| --- | --- | --- |
| `--graceful-shutdown-fast-timeout` | `5s` | How long fast-mode `pgctld.Stop` runs before escalating to immediate. |
| `--graceful-shutdown-immediate-timeout` | `5s` | How long immediate-mode `pgctld.Stop` runs before giving up. |
| `--graceful-shutdown-final-snapshot-timeout` | `5s` | Time budget for the terminal STOPPED broadcast to land on each subscriber's gRPC stream before its channel is force-closed. |

Drain is bounded by the pre-existing `--connpool-drain-grace-period` (default `3s`), shared with operational `NOT_SERVING` transitions like demotion. The pooler's shutdown deadline is a fixed `60s` (matching `pg_ctl stop`'s default), set in code as `gracefulShutdownAnnouncedDeadline` and used only locally to size the safety-net force-exit. Validation rejects misconfiguration where `fast + immediate + final-snapshot` exceeds this deadline.

Servenv's `--onterm-timeout` must be at least the shutdown deadline (60s) or `OnTermSync` will be cut off mid-flight. Integration tests pass `--onterm-timeout 80s` in `shardsetup/process.go`, and the local provisioner (`go/provisioner/local/local.go`) passes the same value when launching multipoolers, so `multigres cluster start` followed by SIGTERM exercises the full graceful-shutdown sequence without the servenv default truncating it.

`PoolerType_DRAINED` is set from two pooler-side writes: phase 6 of `GracefulShutdown` (described above) and the existing `OnRun` unregister hook in `init.go`, which runs from servenv's `OnClose` phase. Both writes target the same fields with the same values; the second one is a harmless no-op overwrite. DRAINED is purely cosmetic so `multigres getpoolers` reflects reality; it is not a functional input to failover, which keys off `LeadershipStatus`. Bare unreachability (SIGKILL, kernel panic, partition that severs the pooler-to-topo connection) leaves the pooler at its last live type — that case is "crashed without intent" rather than "drained" and is honestly represented as such.

## Refactor

`LeaderIsDeadAnalyzer` and related symbols are renamed to `LeaderIsDown` across the `multiorch/recovery` package. "Dead" was misleading because the analyzer also fires when the leader is reachable but unhealthy.

## CI

`.github/workflows/coverage.yml` passes `-timeout=30m` to both the direct and subprocess coverage jobs. Coverage instrumentation adds 30–50% overhead and the default per-package 10m timeout was tipping the integration packages (multiorch, multipooler, queryserving) into "panic: test timed out after 10m0s" under instrumentation. 30m matches what `test-integration.yml` already uses for uninstrumented runs.

## Tests

Unit tests:

- `manager/graceful_shutdown_test.go` — eleven shutdown-sequence paths: happy path, fast→immediate escalation, all-modes-fail, never-calls-smart, closes-pool-before-pgctld, lifecycle signal is `UNKNOWN` during the slow phases and `STOPPED` only at the end, safety-net exit, safety-net flushes telemetry before exit, nil pgctld client, terminal broadcast closes subscribers, validator rejects post-announce budget that exceeds the shutdown deadline.
- `recovery/health_stream_lifecycle_test.go` — stream EOF after STOPPED skips reconnect-with-backoff, EOF without lifecycle signal keeps backoff, plus four `REQUESTING_DEMOTION` synthesis tests (sets signal for primary, no-op for non-primary, no-op for zero term, idempotent on repeated calls at same term).
- `recovery/types/pooler_signals_test.go` — subtests exercising every signal combination.

Integration tests in `go/test/endtoend/multiorch/graceful_shutdown_test.go`:

- Primary graceful shutdown triggers failover and the old primary's topology entry shows `Type=DRAINED` after failover.
- Standby graceful shutdown does not trigger failover.
- Multi-replica continuity verifies remaining standbys keep streaming after one shuts down.
- Sequential graceful shutdowns of two replicas both complete cleanly.
- `TestPrimaryGracefulShutdownNoSmartEscalation` bounds SIGTERM-to-exit at 8s, locking in that drain-before-pgctld means fast doesn't have to wait. A regression that reintroduces smart-style waiting would balloon past that bound.

## Out of scope

A `Shutdown` RPC on `MultiPoolerManagerService` for triggering the same sequence remotely is intentionally deferred. Whether `--connpool-drain-grace-period` should be raised from its current 3s default for graceful-shutdown use is a separate review.

Fixes MUL-36


go/services/multipooler/manager/graceful_shutdown.go
```diff
@@ -0,0 +1,309 @@
...
+//     failover proceeds against a leader that has cleanly resigned.
+func (pm *MultiPoolerManager) GracefulShutdown(ctx context.Context) {
+	deadline := gracefulShutdownAnnouncedDeadline
+	deadlineAt := time.Now().Add(deadline)
+
+	// Safety-net force-exit. Stops if GracefulShutdown returns first.
+	safetyNet := time.AfterFunc(deadline, func() {
+		pm.logger.ErrorContext(ctx, "graceful shutdown overran announced deadline; forcing process exit",
+			"deadline", deadline)
+		gracefulShutdownExitFunc(1)
```
dweitzman: A risk of os.Exit() here is failing to export any final telemetry data (flushing logs, pushing metrics, etc.). I feels like existing without at least trying to flush logs to disks to disk could cause problems debugging a shutdown that doesn't behave as expected

ServEnv already has `onCloseTimeout` to try to limit the amount of time spent on a graceful shutdown to 10 seconds. Could we use the existing graceful shutdown capabilities of ServEnv by making all shutdown behavior an onCloseHook and raising that setting from 10 seconds to 60 seconds?

go/services/multipooler/manager/health_provider.go
```diff
@@ -124,12 +124,41 @@ func (hs *healthStreamer) OnStateChange(ctx context.Context, poolerType clusterm
...
+// sent. Subscribers' gRPC stream handlers detect the channel close as EOF and
+// return cleanly, releasing the gRPC server's GracefulStop wait. Subsequent
+// state changes will not reach the closed subscribers; new subscribers
+// (unlikely during a shutdown window, but possible) get an immediate channel
+// that they will never receive from.
+//
+// Safe to call on a nil receiver: tests and other constructors that build a
+// MultiPoolerManager without a healthStreamer can call broadcast paths
+// unconditionally without nil-checking at each call site.
+func (hs *healthStreamer) Broadcast(terminal bool) {
```
dweitzman: Something feels a little off to me about having `terminal` be a parameter here

Something like having a shutdown Context in the healthStreamer constructor or having a separate Shutdown() method would feel more idiomatic

go/services/multiorch/recovery/types/pooler_signals.go
```diff
@@ -29,17 +29,52 @@ import (
...
 // coordinator-synthesized signals (e.g. TEMPORARILY_UNAVAILABLE for unreachable
 // nodes) are handled here too.
 func LeaderNeedsReplacement(p *multiorchdatapb.PoolerHealthState) bool {
-	leadershipStatus := p.GetAvailabilityStatus().GetLeadershipStatus()
+	avail := p.GetAvailabilityStatus()
+
+	// Lifecycle path: pooler has completed graceful shutdown. Only STOPPED
+	// triggers replacement; SHUTTING_DOWN is an announcement and does not
+	// fire failover on its own (see comment above).
+	switch avail.GetLifecycleStatus().GetSignal() {
```
dweitzman: Seems like this existing LEADERSHIP_SIGNAL_REQUESTING_DEMOTION signal can already handle this use case?

The thing missing at the moment is cohort resignation, which may be a little awkward to fix because the `FixReplication` action does a lot of things including adds but nothing does removes. It seems like cohort management shouldn't be the responsibility of `FixReplication` and maybe we should have a separate analyzer and action for keeping the cohort list in good shape (both adds and removes, as needed)?

dweitzman: I put together a PR at https://github.com/multigres/multigres/pull/973 for the add/remove cohort member analysis & action so this PR can stay focused on the tracking a shutdown request, leadership resignation, and the pooler shutdown process

go/services/multipooler/manager/graceful_shutdown.go
```diff
@@ -0,0 +1,309 @@
...
+
+	postAnnounce := cfg.GracefulShutdownFastTimeout +
+		cfg.GracefulShutdownImmediateTimeout +
+		cfg.GracefulShutdownFinalSnapshotTimeout
+	if postAnnounce > gracefulShutdownAnnouncedDeadline {
+		return mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT,
+			"graceful shutdown post-announce budget %s exceeds announced deadline %s; lower the --graceful-shutdown-* timeouts",
+			postAnnounce, gracefulShutdownAnnouncedDeadline)
+	}
+	if gracefulShutdownAnnouncedDeadline < constants.MinShutdownDeadline {
```
GuptaManan100: Shouldn't `gracefulShutdownAnnouncedDeadline` also live in the constants package like `constants.MinShutdownDeadline`.

GuptaManan100: This check seems wrong, its comparing 2 constants 😆, maybe it should be `if postAnnounce < constants.MinShutdownDeadline`

go/services/multipooler/manager/graceful_shutdown.go
```diff
@@ -0,0 +1,309 @@
...
+	)
+
+	// Phase 3: close the connection pool. Idle backends are released
+	// immediately so postgres has no lingering pooler clients; borrows that
+	// returned naturally during phase 1 are already gone, and any remaining
+	// non-reserved borrows are disposed when they return (bounded by
+	// PoolCloseTimeout). Action lock not needed for pool teardown.
+	if pm.connPoolMgr != nil {
+		pm.connPoolMgr.Close()
+	}
```
GuptaManan100: `MultiPoolerManager` already has a `Shutdown` method that takes care of state transition and closing the connPoolMgr. It does even more than that, and closes other components like `pubsubListener` and `replTracker` which this function doesn't close. 

Anyways having 2 `Shutdown` and `GracefulShutdown` isn't a very good idea. We should reconcile these 2 functions.

go/services/multipooler/manager/graceful_shutdown.go
mkindahl: That seems to be remains from a previous attempt and having a dynamically announced deadline. Removed it, even though we might in the future make it dynamic, and added a compile-time check above instead (that will complain if we start to make it dynamic).

mkindahl: Intention was to keep it local since we might decide to dynamically compute it in the future. constants.MinShutdownDeadline is an absolute minimum below we will never set a shutdown, but gracefulShutdownAnnouncedDeadline is a value that pooler can decide either dynamically or statically when it starts the shutdown.

go/services/multipooler/manager/graceful_shutdown.go
mkindahl: > `MultiPoolerManager` already has a `Shutdown` method that takes care of state transition and closing the connPoolMgr. It does even more than that, and closes other components like `pubsubListener` and `replTracker` which this function doesn't close.

This just calls closeLocked() with a message, so I did that instead of modifying the Shutdown() message to accept a message.

> Anyways having 2 `Shutdown` and `GracefulShutdown` isn't a very good idea. We should reconcile these 2 functions.

They are doing slightly different things: `Shutdown()` is direct and abrupt and can leave the server is a strange state, but `GracefulShutdown()` is more careful and gives the server a chance to cleanup. Are your suggestion that we replace `Shutdown()` with `GracefulShutdown()`?

go/services/multipooler/manager/health_provider.go
mkindahl: I removed the "terminal" and am now calling `closeLocked()` inside `GracefulShutdown()` instead (this was the only place that used "true" for `Broadcast()`), so now we have a `Shutdown()` and a `GracefulShutdown()`, where the latter is doing an orderly shutdown while the former is just closing everything and can leave the system in a strange state.

go/services/multiorch/recovery/types/pooler_signals.go
mkindahl: > Seems like this existing LEADERSHIP_SIGNAL_REQUESTING_DEMOTION signal can already handle this use case?

Functionally, they are similar, but the lifecycle events work for any server (including a standby), while leadership events are only used for leaders.

> The thing missing at the moment is cohort resignation, which may be a little awkward to fix because the `FixReplication` action does a lot of things including adds but nothing does removes. It seems like cohort management shouldn't be the responsibility of `FixReplication` and maybe we should have a separate analyzer and action for keeping the cohort list in good shape (both adds and removes, as needed)?

For this functionality, the pooler is just notifying that it is going away, but it is not really doing that until it has actually left, and as a result `LeaderIsDownAnalyzer` will be triggered and promote a standby. The remaining servers need to negotiate a new cohort excluding this server.

mkindahl: I modified the code to instead of having LeaderNeedsReplacement check if the LifecycleStatus is STOPPED, this is updated when the snapshot is processed in the orchestrator and it will synthesize a REQUESTING_DEMOTION for poolers that are primaries but not for anything else.

go/services/multipooler/manager/graceful_shutdown.go
mkindahl: I replaced this with a "best-effort" shutdown to flush logs. It is running inside a separate go-routine, and I place a hard deadline on this as well to avoid a "weird" event causing it to not finish in time. If the deadline is exceeded, it is a hard stop using os.Exit.

go/services/multiorch/recovery/engine.go
```diff
@@ -239,7 +239,23 @@ func NewEngine(
...
 	poolerStore := store.NewPoolerStore(rpcClient, logger)
-	healthStream := NewHealthStream(ctx, rpcClient, poolerStore, logger)
+
+	// markDrained writes Type=DRAINED + ServingStatus=NOT_SERVING to the pooler's
+	// topology entry. The orchestrator drives this from HealthStream's three
+	// shutdown-intent trigger paths (STOPPED snapshot, EOF after SHUTTING_DOWN,
+	// deadline expiry) so DRAINED reflects the pooler's intentional departure
+	// even when the pooler crashed or was partitioned and could not update
+	// topology itself.
+	markDrained := func(ctx context.Context, id *clustermetadatapb.ID) error {
```
dweitzman: My short-term preference would be to have the pooler mark itself as drained rather than going through orch

My longer-term preference (out of scope for this PR) would be:
- eliminate the concept of pooler type entirely
- have poolers publish their lifecycle stage to their etcd record
- have poolers first advertise in lifecycle that they're shutting down and then advertise that they're shut down. These can serve as tombstones so that orch knows the etcd entry can be safely deleted

mkindahl: > My short-term preference would be to have the pooler mark itself as drained rather than going through orch

That is reasonable. Changed.

> My longer-term preference (out of scope for this PR) would be:
> 
> * eliminate the concept of pooler type entirely
> * have poolers publish their lifecycle stage to their etcd record
> * have poolers first advertise in lifecycle that they're shutting down and then advertise that they're shut down. These can serve as tombstones so that orch knows the etcd entry can be safely deleted

No objection, but as you say, this is out of scope for this PR.

go/services/multipooler/manager/graceful_shutdown.go
mkindahl: Note: the timeout for the best-effort log flush is subtracted from the announced deadline to make sure that we stop before.

go/services/multiorch/recovery/health_stream.go
```diff
@@ -111,6 +126,54 @@ func WithStalenessTimeout(d time.Duration) Option {
...
+// LIFECYCLE_SIGNAL_STOPPED into the cached PoolerHealthState so the analyzer
+// can fail over without contact with the pooler.
+//
+// On duplicate SHUTTING_DOWN announcements (timer already armed) the original
+// deadline is kept; if the duplicate announces a different value a warning is
+// logged (a buggy or misconfigured pooler signal). Identical re-announcements
+// are silent.
+//
+// Caller is expected to have already clamped deadline to MinShutdownDeadline.
+func (entry *streamEntry) armTimer(ctx context.Context, hs *HealthStream, poolerID string, deadline time.Duration) {
```
dweitzman: Maybe we can skip the armTimer() behavior for now and consider adding it in the future if we notice any situations it'd help during hardening testing

go/services/multiorch/recovery/health_stream.go
```diff
@@ -449,6 +484,99 @@ func (hs *HealthStream) applySnapshot(ctx context.Context, poolerID string, pool
...
+// Called from the two convergent shutdown-intent paths: STOPPED snapshot
+// received, and stream EOF after SHUTTING_DOWN. For non-primary poolers it
+// is a no-op: only a leader's departure should request demotion. Idempotent
+// across the two paths (skips if the signal is already set for the current
+// term).
+//
+// Captures the leader_term at observation time so LeaderNeedsReplacement's
+// staleness check (leader_term == consensus primary_term) passes when the
+// analyzer reads the cached state on its next tick.
+func (hs *HealthStream) synthesizeRequestingDemotion(poolerID, reason string) {
```
dweitzman: This logic seems like it belongs in the pooler instead of orch. It seems like orch can only get three useful pieces of information from a pooler's `Status` 
- eligible for cohort? yes/no
- eligible to continue being leader? yes/no
- eligible to become leader? yes/no

Any reason not to make `Status` a deep module that encapsulates all the complexity behind answering those questions and let's orch focus on rule changes without concerning itself with pooler internals?

go/services/multiorch/recovery/health_stream.go
mkindahl: The timer is completely removed as well as the shutdown deadline, but it still sends a shutdown announcement.

I still think we might run into issues if we do not use a timer and pre-announce the shutdown (mainly as a result of unexpected behavior c.f. bugs or mistakes, not a specific design issue), but since I have no concrete scenario yet I'll remove it and keep the branch around so that we can add it later when we have a concrete case.

go/services/multiorch/recovery/health_stream.go
mkindahl: > This logic seems like it belongs in the pooler instead of orch. It seems like orch can only get three useful pieces of information from a pooler's `Status`
> 
> * eligible for cohort? yes/no
> * eligible to continue being leader? yes/no
> * eligible to become leader? yes/no
> 
> Any reason not to make `Status` a deep module that encapsulates all the complexity behind answering those questions and let's orch focus on rule changes without concerning itself with pooler internals?

To get a good separation of concerns, I would argue that the pooler should only concern itself with the availability of the components that make up a pod and just focus on aspects that are relevant to the pod. As a result, it should declare the status of the pod components (on the level that is relevant) and let any other cluster-wide decision (such as who is the leader, the concept of a leader and follower, and what is required to be a leader) stay the orchestrator.

The flow right now is very simple: the pooler declare that it shut down and let the orchestrator make more advanced decisions.

go/services/multiorch/recovery/health_stream.go
mkindahl: I removed the shutdown announcement as well, per your request.

go/services/multipooler/manager/manager.go
```diff
@@ -287,6 +297,14 @@ func NewMultiPoolerManagerWithTimeout(logger *slog.Logger, multiPooler *clusterm
 		cancel: cancel,
 	}
 
+	// shutdownCtx is independent of ctx: ctx is recreated on every Open(),
+	// while shutdownCtx exists for the lifetime of the manager and is
+	// cancelled exactly once, by GracefulShutdown. Background root is
+	// intentional: this ctx must outlive any Open()/Close() cycle so
+	// long-lived stream subscribers can keep watching it.
+	//nolint:gocritic // intentional; see comment above
+	pm.shutdownCtx, pm.shutdownCancel = context.WithCancel(context.Background())
```
dweitzman:
```suggestion
	pm.shutdownCtx, pm.shutdownCancel = context.WithCancel(ctxutil.Detach(ctx))
```

go/services/multipooler/init.go
```diff
@@ -332,7 +321,6 @@ func (mp *MultiPooler) Init(startCtx context.Context) error {
 		PgBackRestCertFile: mp.pgBackRestCertFile.Get(),
 		PgBackRestKeyFile:  mp.pgBackRestKeyFile.Get(),
 		PgBackRestCAFile:   mp.pgBackRestCAFile.Get(),
-		VpidStampEnabled:   mp.vpidStampEnabled.Get(),
```
dweitzman, line 335: VpidStampEnabled getting removed looks like a merge accident?

dweitzman (approved):

go/services/multipooler/init.go
mkindahl: Yes, it was a merge problem.

PR #971, opened 2026-05-08
feat(multiorch): support bootstrap via Recruit/Propose flow

dweitzman:
Wires AppointInitialLeader to the new flow when useNewFlow is set. Bootstrap has no outgoing cohort to recruit consent from, so it uses BuildExternallyCertifiedProposal / CheckExternallyCertifiedProposalPossible with a cert derived from the cohort's cached state — discovery, but on the snapshot rather than recruit responses.

Discovery surfaces both the most-advanced rule number and the most-advanced LSN. The rule number caps how far the outgoing cohort could have progressed; the LSN is the frozen point any new leader must match. This handles partial bootstraps too — if any cohort member already carries a rule, we surface its rule number rather than falsely claiming term 0 (which certLeaderFilter would then reject).

Adds a MostAdvancedPosition helper to go/common/consensus, a buildBootstrapProposal builder analogous to buildFailoverProposal, and TestAppointInitialLeader_NewFlow exercising the wiring end-to-end.

### DIFF
--- a/go/services/multiorch/config/config.go
+++ b/go/services/multiorch/config/config.go
@@ -236,7 +236,7 @@ func NewConfig(reg *viperutil.Registry) *Config {
 			EnvVars:  []string{"MT_LEADER_POSTGRES_RESPONSE_THRESHOLD"},
 		}),
 		useNewConsensusFlow: viperutil.Configure(reg, "use-new-consensus-flow", viperutil.Options[bool]{
-			Default:  false,
+			Default:  true,
 			FlagName: "use-new-consensus-flow",
 			Dynamic:  false,
 			EnvVars:  []string{"MT_USE_NEW_CONSENSUS_FLOW"},
--- a/go/services/multiorch/consensus/coordinator.go
+++ b/go/services/multiorch/consensus/coordinator.go
@@ -222,6 +222,58 @@ func (c *Coordinator) AppointInitialLeader(ctx context.Context, shardID string, 
 		return mterrors.Wrap(err, "failed to load durability policy from topology")
 	}
 
+	if c.useNewFlow {
+		// Bootstrap has no outgoing cohort to recruit consent from, so we use
+		// the externally-certified path. The "external" certification is the
+		// most-advanced timeline observed across the cohort's cached statuses:
+		// its rule number caps how far the outgoing cohort could have
+		// progressed, and its LSN is the frozen point any new leader must
+		// match. This handles partial bootstraps too — if any cohort member
+		// already carries a rule, we surface its rule number rather than
+		// falsely claiming term 0.
+		var cohortStatuses []*clustermetadatapb.ConsensusStatus
+		for _, p := range cohort {
+			if cs := p.GetConsensusStatus(); cs != nil {
+				cohortStatuses = append(cohortStatuses, cs)
+			}
+		}
+
+		// This is the discovery phase of coordinator-led rule changes. For externallly-certified
+		// rule changes, the client (this method) is responsible for discovery.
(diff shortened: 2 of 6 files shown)


go/services/multiorch/config/config.go
```diff
@@ -236,7 +236,7 @@ func NewConfig(reg *viperutil.Registry) *Config {
 			EnvVars:  []string{"MT_LEADER_POSTGRES_RESPONSE_THRESHOLD"},
 		}),
 		useNewConsensusFlow: viperutil.Configure(reg, "use-new-consensus-flow", viperutil.Options[bool]{
-			Default:  false,
+			Default:  true,
```
dweitzman, line 239: This PR is also flipping on the new consensus flow as the default behavior

go/services/multiorch/consensus/coordinator.go
```diff
@@ -117,21 +118,39 @@ func (c *Coordinator) AppointLeader(ctx context.Context, shardID string, cohort
...
+// and runs it. Poolers that have signaled REQUESTING_DEMOTION are excluded from
+// the cohort entirely: a writing leader's local LSN is always at least as
+// advanced as any replica's (async replication), so including a resigned
+// leader would let it dominate discovery and dead-end the proposal when health
+// check rejects it. The full cohort identity is preserved via OutgoingRule's
+// CohortMembers (replicas carry the same rule), so the consensus layer's
+// outgoing-quorum check still runs against the original cohort size.
 func (c *Coordinator) runFailover(ctx context.Context, cohort []*multiorchdatapb.PoolerHealthState, reason string) error {
-	poolerByID, healthByID := buildCohortMaps(cohort)
+	liveCohort := make([]*multiorchdatapb.PoolerHealthState, 0, len(cohort))
```
dweitzman, line 129: A change here is removing resigned poolers from the list of recruited nodes before trying to validate or construct a rule change proposal

The trouble was that a resigning leader could be stuck at a higher LSN than all the other nodes. It might not be a durable LSN, but if we considered it during discovery it might cause the same outgoing primary to be discovered as the most advanced timeline and listed as the only eligible leader. If that happened, we'd see that the outgoing leader is resigning and ineligible to stay at leader and we'd get stuck: the most advanced timeline is not eligible to be the leader

Dropping a resigned leader from the recruitment list is a simple workaround. A more flexible alternative could be using consensus rules to prove that the leader's timeline has non-durable transactions and have the build proposal logic ignore it regardless of whether it's included as an input.

go/services/multiorch/consensus/coordinator.go
```diff
@@ -222,6 +241,58 @@ func (c *Coordinator) AppointInitialLeader(ctx context.Context, shardID string,
 		return mterrors.Wrap(err, "failed to load durability policy from topology")
 	}
 
+	if c.useNewFlow {
```
dweitzman, line 244: This is the meat of the PR

go/services/multiorch/consensus/rule_change.go
```diff
@@ -267,33 +266,21 @@ func (r *coordinatorLedRuleChange) propose(
...
-			logger.InfoContext(ctx, "Skipping resigned primary during leader selection",
-				"pooler", cs.GetId().GetName())
-			continue
-		}
-		leader = cs
-		break
-	}
-	if leader == nil {
-		return nil, errors.New("all eligible leaders have resigned")
+	if len(result.EligibleLeaders) == 0 {
```
dweitzman, line 280: A resigned leader is excluded before trying to make a proposal, so there's no need to filter out resigned leaders here anymore

rafael (approved):

PR #972, opened 2026-05-08
refactor(pgctld): align initdb SQL files flag/field naming

rafael:
## Summary 

- Rename pgctld flag `--init-db-sql-file` → `--pg-initdb-sql-files` so it matches the sibling flags `--pg-initdb-args` and `--pg-initdb-extra-conf`.
   - Rename associated Go identifiers from two-word `InitDb` to one-word `Initdb` (e.g. `InitdbSQLFiles`, `pgInitdbSQLFiles`, `runInitdbSQLFiles`, `WithInitdbSQLFiles`,
   `PgInitdbSQLFilesEnvVar`), matching `InitdbArgs` / `InitdbExtraConfFiles` in the same struct.
   - Env var `POSTGRES_INITDB_SQL_FILES` is unchanged — it already follows the `POSTGRES_INITDB_*` convention.
   - Add a backwards-compat alias via cobra's `SetGlobalNormalizationFunc` so `--init-db-sql-file` keeps working transparently during the migration window. The alias is meant to be removed once downstream callers have updated.

# Tests
- Adds relevant unit tests and update existing ones

### DIFF
--- a/go/cmd/pgctld/command/flags_test.go
+++ b/go/cmd/pgctld/command/flags_test.go
@@ -64,40 +64,40 @@ func TestPgInitdbArgsEnvVar(t *testing.T) {
 	})
 }
 
-func TestInitDbSQLFilesFlag(t *testing.T) {
+func TestPgInitdbSQLFilesFlag(t *testing.T) {
 	t.Run("defaults to empty slice", func(t *testing.T) {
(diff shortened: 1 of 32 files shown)


dweitzman (approved):

mkindahl (approved):
LGTM

PR #973, opened 2026-05-08
feat(multiorch): split cohort management into dedicated analyzer/action

dweitzman:
## Summary

`FixReplicationAction` was doing two unrelated jobs: (1) repairing a replica's
postgres-side replication setup (`primary_conninfo`, pg_rewind, streaming
verification) and (2) mutating the primary's synchronous standby list as a
proxy for cohort membership. This PR pulls the cohort-membership work into
its own analyzer/action and tightens the RPC contract so cohort changes
race-safely.

### What's new

- **`CohortMismatchAnalyzer`** — detects two flavors of cohort drift:
  - `ProblemPoolerNotInCohort`: a healthy, replicating, eligible REPLICA is
    absent from the leader's cohort
  - `ProblemCohortMemberIneligible`: a current cohort member has self-reported
    `INELIGIBLE` and should be removed/replaced

- **`ReconcileCohortAction`** — applies a single cohort change per execution
  (one ADD or REMOVE) via `UpdateConsensusRule` on the leader. Runs at
  `PriorityNormal` so replication repair (`PriorityHigh`) finishes first.

- **`CohortEligibilityStatus`** signal on `AvailabilityStatus`. Every pooler
  publishes whether it is willing to serve in the cohort (`ELIGIBLE` /
  `INELIGIBLE`) — symmetric, regardless of current membership. The
  multipooler emits an immediate health broadcast when the signal changes
  so coordinators don't wait a heartbeat to react. No operator-facing
  setter yet (test-only); that lands when there's a use case.

### What's gone

- `ReplicaNotInStandbyListAnalyzer` (and its `ProblemReplicaNotInStandbyList`
  code) — folded into the cohort analyzer.
- `FixReplicationAction` no longer mutates the synchronous standby list.

### Tightening the `UpdateConsensusRule` contract

`UpdateConsensusRule`'s old request type was
`UpdateSynchronousStandbyListRequest`/`Response` with operation enum
`StandbyUpdateOperation` — names that framed the operation as a postgres-GUC
mutation. Renamed to:


go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -357,14 +357,39 @@ func (pm *MultiPoolerManager) getInconsistentConsensusStatus(ctx context.Context
...
+	return &clustermetadatapb.CohortEligibilityStatus{Signal: signal}
+}
+
+// setCohortEligibility records this pooler's cohort eligibility. If the
+// signal actually changed, an immediate health broadcast is pushed so the
+// coordinator sees the new value without waiting for the next heartbeat —
+// otherwise a transition INELIGIBLE → cohort removal could be delayed by up
+// to a heartbeat interval. Currently test-only — there is no operator/admin
+// RPC to flip it yet.
+func (pm *MultiPoolerManager) setCohortEligibility(signal clustermetadatapb.CohortEligibilitySignal) {
```
dweitzman, line 385: A SIGTERM or similar would end up calling this method to request resigning from the cohort

proto/multipoolermanagerdata.proto
```diff
@@ -556,29 +556,31 @@ message ConfigureSynchronousReplicationRequest {
...
   repeated clustermetadata.ID standby_ids = 2;
 
-  // Whether to reload configuration immediately
-  bool reload_config = 3;
-
-  // Consensus term (used by MultiOrch for term validation)
-  int64 consensus_term = 4;
-
-  // Force the operation even if the term doesn't match
-  bool force = 5;
```
dweitzman, line 575: Removed some unnecessary options like "force" and "reload_config" from the request to update the cohort members, but also make a compare-and-swap calling style required so there aren't weird race conditions with multiple coordinators undoing each other's work

go/services/multiorch/recovery/analysis/cohort_mismatch_analyzer.go
```diff
@@ -0,0 +1,157 @@
...
+//   - ProblemPoolerNotInCohort: a healthy, replicating, eligible pooler exists
+//     in the shard but is absent from the leader's recorded cohort.
+//   - ProblemCohortMemberIneligible: a current cohort member has self-reported
+//     INELIGIBLE via AvailabilityStatus.CohortEligibilityStatus and should be
+//     removed.
+//
+// Both surface a single ReconcileCohortAction; the action interprets the
+// problem code and applies the appropriate ADD/REMOVE.
+//
+// TODO: this currently includes every healthy, eligible pooler in the cohort.
```
dweitzman, line 41: Just calling attention to this TODO: as written now, all poolers will be added to the cohort. I think in the future we'll want the cohort to only be as large as "necessary", but we'll have to have a way to configure what that means and a way to decide which poolers should be in the cohort vs being non-participating observers.

rafael (approved):
LGTM

PR #974, opened 2026-05-08
ci(coverage): add Codecov upload alongside Coveralls

dweitzman:
Coveralls hasn't been sending PR comments and it's not obvious to me why. Codecov seems to work fine, so we can enable that and either figure out how to fix or remove coveralls later


Verolop (approved):
💖

PR #976, opened 2026-05-11
feat(security): require postgres password and remove trust authentication

mkindahl:
The pg_hba.conf template previously granted the admin user `trust` authentication on the local socket as a bootstrap shortcut for multipooler's admin pool. A misconfiguration that exposed the socket or relaxed a network rule would have given an attacker passwordless superuser access.

This change closes the bootstrap path and makes scram-sha-256 mandatory for every connection. pgctld init and pgctld server now refuse to start without POSTGRES_PASSWORD, and multipooler refuses to start without CONNPOOL_ADMIN_PASSWORD / POSTGRES_PASSWORD. pgbackrest receives the password via PGBACKREST_PG1_PASSWORD (requires pgbackrest >= 2.53, which is checked at startup before any pgbackrest config is written). pg_rewind already used PGPASSWORD, so the env var requirement is enough there. The transient psql invocations used during post-initdb setup inherit the password through PGPASSWORD and fail loudly at construction time if it is missing.

The local provisioner is also fixed to source the password from pgctldConfig (same place pgctld reads it) and to stop clobbering the multipooler env on the line after setting POSTGRES_PASSWORD.

Integration test fixes for scram-sha-256 auth:
- All direct postgres socket connections in tests updated to include password=TestPostgresPassword (backup_test_helpers, backupfaults, multiorch helpers, pgctld tests).
- pgctld_test: TestPostgreSQLAuthentication socket psql now sets PGPASSWORD; TestCustomDatabaseCreation does the same.
- WaitForManagerReady per-check context raised from 1s to 15s; total wait extended from 30s to 60s.
- checkBootstrapStatus per-call context raised from 5s to 15s; waitForShardBootstrap timeout extended from 60s to 120s.
- TestConsensus_BeginTerm/BeginTerm_NewTerm_Accepted deadline raised from 2s to 15s (postgres restart with scram auth takes ~4-5s).
- waitForReceiverDisconnect timeout raised from 10s to 30s; switched from stale pg_stat_wal_receiver to pg_stat_activity; combined terminate+count into single query to avoid pool contention.
- resetPrimaryConnInfo: adds best-effort pg_terminate_backend for any walreceiver blocked in TCP recv during concurrent primary demotion.
- breakReplication: switched to pg_stat_activity, uses per-query fresh context (60s polling, decoupled from 30s SetupTest ctx).
- emergencyDemote: returns to SERVING state on pre-restart failure so subsequent BeginTerm attempts are not stuck.
- BeginTerm emergency demotion test: fixed state leakage between runs.

Fixes MUL-431


go/cmd/pgctld/command/init.go
```diff
@@ -240,32 +241,35 @@ func initializeDataDir(logger *slog.Logger, cfg PgCtldServiceConfig) error {
...
-		}
-		if err := tmpFile.Close(); err != nil {
-			return fmt.Errorf("failed to close temporary password file: %w", err)
-		}
+	// Write the password to a temporary file for initdb --pwfile. The filename
+	// uses a random pattern with no prefix so the file's purpose isn't
+	// advertised in /tmp listings. This is the only way to set the password
+	// during initdb without exposing it in process arguments or environment
+	// variables, and the file is removed immediately after initdb completes.
+	tmpFile, err := os.CreateTemp("", "*")
```
dweitzman, line 255: FYI, you don't have to change anything here but I'll probably make a separate PR to pass this file to the child process more securely using a pipe

go/cmd/pgctld/command/server.go
```diff
@@ -306,6 +306,15 @@ func NewPgCtldService(
 	if listenAddresses == "" {
 		return nil, errors.New("listen-addresses needs to be set")
 	}
+	if cfg.Password == "" {
+		return nil, errors.New("POSTGRES_PASSWORD must be set")
+	}
+
+	// Set PGPASSWORD so pgbackrest (archive-push runs as a postgres subprocess and
+	// inherits this process's environment) can authenticate against PostgreSQL.
+	if err := os.Setenv("PGPASSWORD", cfg.Password); err != nil {
```
dweitzman: It's probably worth looping in more people to discuss how we feel about secrets in environment variables

Personally secrets in env variables make me nervous, so I'd prefer to only support `PGPASSFILE` over `PGPASSWORD`

If we were using a PGPASSFILE, we could use either a pipe with a file descriptor or an already-deleted file's file descriptor to securely pass secrets to a sub process.

mkindahl: It is worth looking into, but right now we are using environment variables for passwords already so better take this as a separate PR.

go/cmd/pgctld/command/init.go
mkindahl: A better idea is probably to use a dedicated volume for this and clamp down the access to it.

go/cmd/pgctld/command/server.go
mkindahl: Since this was a modification regarding pgbackrest, I switched to using a file for this particular case.

go/test/endtoend/multipooler/rpc_consensus_test.go
```diff
@@ -325,7 +325,10 @@ func TestConsensus_BeginTerm(t *testing.T) {
 			Action:  consensusdatapb.BeginTermAction_BEGIN_TERM_ACTION_REVOKE,
 		}
 
-		resp, err := primaryConsensusClient.BeginTerm(utils.WithShortDeadline(t), req)
+		// BeginTerm REVOKE on a primary triggers synchronous emergency demotion
+		// (postgres stop + restart as standby). Use a timeout consistent with
+		// multiorch's RemoteOperationTimeout (15s) rather than WithShortDeadline (2s).
+		resp, err := primaryConsensusClient.BeginTerm(utils.WithTimeout(t, 15*time.Second), req)
```
mkindahl: The timeout was extended since scram-sha-256 takes longer to connect.

cuongdo (approved):
I'm approving, but I think you should consult with our resident security experts to see if further hardening of environment variable use is recommended.

cuongdo (approved):

mkindahl:
> I'm approving, but I think you should consult with our resident security experts to see if further hardening of environment variable use is recommended.

The most sensible approach would be to use a mounted secret and skip the environment variables, but then we need to modify other components to honor this before we can stop using it. I'll merge this for now and continue with wiring that into the existing infrastructure.

PR #977, opened 2026-05-11
refactor(multipooler): remove dead transaction-status reconciliation

GuptaManan100:
## Summary

- Deletes the post-`StreamExecute` reconciliation block in `streamExecuteOnReservedConn` that synced in-memory `ReasonTransaction` against PG's `ReadyForQuery` transaction status indicator.
- That block existed to catch simple-query payloads like `"BEGIN; SELECT 1; COMMIT;"` where transaction control was bundled inline and bypassed `ReservationOptions`. The gateway now parses multi-statement payloads into individual `ast.Stmt`s and dispatches one statement per RPC (`executeWithImplicitTransaction` in `transaction_helpers.go`), so the pooler only ever sees a single statement and `ReservationOptions` is always authoritative.
- Also drops `TxnStatus()` from the `reservedConnAPI` interface (no remaining caller in the executor) and removes the two reconciliation-specific unit tests.

## Problem

`executor.go` reconciled `rc.TxnStatus()` against `rc.IsInTransaction()` after every `StreamExecute` on a reserved connection. The accompanying comment explained the case it guarded: bundled transaction control in a simple-query payload, where the SQL itself changed PG's transaction status without going through `ReservationOptions`. That case can no longer reach the pooler — the gateway's per-statement dispatch makes the reconciliation a no-op on all current traffic.

## Solution

Delete the reconciliation block and the now-unused `TxnStatus()` member of `reservedConnAPI`. No defense-in-depth assertion was added: per the repo guidance we avoid validation for scenarios that can't reach the code, and a gateway regression would surface as a stuck reserved connection — visible without a custom log path.

## Test plan

- [x] `go test -short ./go/services/multipooler/executor/...` — passes.
- [x] `go test ./go/test/endtoend/queryserving/...` — all four packages pass (`queryserving`, `benchmarking`, `pgparity`, `sqllogictest`).

### DIFF
--- a/go/services/multipooler/executor/executor.go
+++ b/go/services/multipooler/executor/executor.go
@@ -357,8 +357,7 @@ func (e *Executor) reserveAndStreamExecute(
 // streamExecuteOnReservedConn executes a query on an existing reserved
 // connection. It optionally promotes the reservation by adding new reasons
 // (e.g., starting a transaction on a temp-table-reserved connection) before
-// running the query, and reconciles the in-memory transaction tracking against
-// PG's actual transaction status afterwards.
+// running the query.
 //
 // Defined over reservedConnAPI rather than *reserved.Conn so that unit tests
 // can substitute a mock and exercise this path without a live PG connection.
--- a/go/services/multipooler/executor/executor.go
+++ b/go/services/multipooler/executor/executor.go
@@ -24,7 +24,6 @@ import (
 	"log/slog"
 
 	"github.com/multigres/multigres/go/common/mterrors"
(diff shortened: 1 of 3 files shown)


go/services/multipooler/executor/executor.go
haritabh17, line 461: Still describes removed reconciliation behavior

haritabh17 (approved):
Looks good, only one doc update needed

go/services/multipooler/executor/executor.go
GuptaManan100: Fixed!

PR #978, opened 2026-05-11
perf(multigateway): demote per-query log to DEBUG with sampling

GuptaManan100:
## Summary

- Per-query structured logging was ~5% of multigateway CPU under load (CPU pprof: `emitQueryLog` + `JSONHandler.Handle` + slog bridge). The hot path built a `[]slog.Attr` and stringified the error for every query even though most operators only care about errors and slow queries.
- `emitQueryLog` now short-circuits before allocating any attrs when the normal path is disabled or sampled out. Normal queries log at `DEBUG`, so the default `INFO`-level slog handler drops them entirely without ever entering `Handle`. Errors and slow queries still log unconditionally at `WARN`.
- New `--query-log-sample-rate` flag (default `0`, i.e. sampling disabled) emits 1/N of normal queries when an operator wants to peek at the stream. Visibility also requires `--log-level=debug`, reusing the existing global level flag rather than introducing a parallel knob.
- Emission counts are exported via a new `mg.gateway.query.log.emits` OTel counter, labeled by level, so operators can size log volume and verify sampling is engaged.

## Description

The CPU win comes from skipping `[]slog.Attr` construction in the steady state. In the default config (no sampling, `INFO`-level handler, normal query), `emitQueryLog` now does a couple of comparisons and a `logger.Enabled` check, then returns — no allocations, no attr building, no error stringification. With sampling enabled, the 99-of-100 sampled-out case adds only a single `atomic.Uint64.Add` for the modulo cursor.

Sampling is deterministic via `counter % rate == 0` on an `atomic.Uint64`; no `math/rand`, no allocation. The cursor stays internal — the observable signal is `mg.gateway.query.log.emits{level=debug|warn}`, which is incremented exactly once per record actually handed to slog.

To re-enable full per-query logging on demand:

```
multigateway --log-level=debug --query-log-sample-rate=1 ...
```

Or with sampling at 1%:

```
multigateway --log-level=debug --query-log-sample-rate=100 ...
```

Untouched on the hot path: metrics recording (`queryDuration`, `rowsReturned`, `tableQueries`, `queryErrors`), query-registry recording, span enrichment, and the slow-query threshold logic. The shape of `queryLogEntry` is unchanged.

## Test plan

- [x] `go test ./go/services/multigateway/handler/...` — 8 tests pass, including a new `TestEmitQueryLog_EmitsMetric` that wires up `sdkmetric.NewManualReader` and asserts the OTel counter values for both `debug` (sampled emissions) and `warn` (errored/slow) data points.
- [x] `go build ./...` clean.
- [x] `go vet ./go/services/multigateway/...` clean.
- [x] Manual smoke: run multigateway with `--log-level=info` (default) and verify no per-query log spam; rerun with `--log-level=debug --query-log-sample-rate=1` and verify per-query records appear.

### DIFF
--- a/go/services/multigateway/handler/handler.go
+++ b/go/services/multigateway/handler/handler.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"sync/atomic"
 	"time"
 
 	"github.com/multigres/multigres/go/common/constants"
@@ -99,6 +100,16 @@ type MultiGatewayHandler struct {
 	// targetReplica is set to true for the replica-port listener. When true,
 	// new connection states target replicas so the planner routes queries there.
 	targetReplica bool
+
+	// normalQueryLogSampleRate controls 1/N sampling for normal-path query
+	// logs. 0 disables sampling (handler level alone governs emission); 1
+	// emits every normal query; N>1 emits every Nth. Normal queries always
+	// log at DEBUG, so the default INFO-level handler drops them entirely.
+	normalQueryLogSampleRate uint64
+	// normalQueryLogSamplingCursor is the 1/N modulo cursor. It is only
+	// touched when sampleRate > 1; emission counts are exposed via the
+	// queryLogEmits OTel counter on h.metrics.
+	normalQueryLogSamplingCursor atomic.Uint64
 }
 
 // NewMultiGatewayHandler creates a new PostgreSQL protocol handler.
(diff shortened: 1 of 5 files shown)


go/services/multigateway/handler/handler.go
```diff
@@ -99,6 +100,16 @@ type MultiGatewayHandler struct {
...
+
+	// normalQueryLogSampleRate controls 1/N sampling for normal-path query
+	// logs. 0 disables sampling (handler level alone governs emission); 1
+	// emits every normal query; N>1 emits every Nth. Normal queries always
+	// log at DEBUG, so the default INFO-level handler drops them entirely.
+	normalQueryLogSampleRate uint64
+	// normalQueryLogSamplingCursor is the 1/N modulo cursor. It is only
+	// touched when sampleRate > 1; emission counts are exposed via the
+	// queryLogEmits OTel counter on h.metrics.
+	normalQueryLogSamplingCursor atomic.Uint64
```
haritabh17, line 112: Not atomic -- think its fine as I dont think we plan to make them dynamic

haritabh17 (approved):

go/services/multigateway/handler/handler.go
GuptaManan100: Right — `normalQueryLogSampleRate` is set once during `Init` (from the `--query-log-sample-rate` flag) before either listener starts accepting connections, then read-only for the lifetime of the handler. The flag is registered with `Dynamic: false` and there's no setter that mutates it post-start, so a plain `uint64` is sufficient. If we ever want to make sampling tunable at runtime, we'd lift this to `atomic.Uint64` (and probably do the same for the slow-query threshold and normal log level) as one change.

PR #979, opened 2026-05-11
feat(buffer,pooler): add failover capacity-planning metrics

GuptaManan100:
## Summary

- Add the metrics operators need to size `--buffer-size`, `--buffer-window`, and `--connpool-drain-grace-period` against real failover behavior instead of guessing from defaults.
- Gateway side: `multigateway.buffer.queue.depth` gauge and `multigateway.buffer.failover.duration` histogram, plus explicit seconds-scale buckets on the existing `wait.duration` histogram (which had been silently using the OTel ms-default boundaries while recording in seconds).
- Pooler side: `mg.pooler.drain.duration` histogram, `mg.pooler.drain.outcome` counter (graceful / force_close), and `mg.pooler.drain.force_closed` counter.
- New operator guide at `docs/query_serving/failover_sizing.md` walking through the sizing procedure with PromQL examples, a worked example, alert queries, and common pitfalls.

## Description

The previous instrumentation surfaced rate-of-buffer-events (counters) and per-request wait time, but operators had no way to answer the questions that actually matter for tuning: how full does the buffer get at peak, how long does a failover take end-to-end, and how often does the pooler drain hit its grace period.

These metrics fill those gaps. The gateway-side `failover.duration` measures the gateway-perceived failover window per shard; the pooler-side `drain.duration` measures the pooler's contribution to that window. Subtracting one from the other (at matching quantiles) tells you whether the drain or the new-primary election is the bottleneck.

While wiring up the depth gauge I caught a latent race: the depth decrement was happening after `close(e.done)`, which wakes the waiting goroutine. A caller observing the gauge immediately after `WaitForFailoverEnd` returns could see a stale value. Reordered all three release sites (drain, timeout eviction, full-buffer eviction) so the gauge is updated before the wake-up. The new test covering this scenario caught the race on first run.

All new histograms use explicit `WithExplicitBucketBoundaries(...)` sized for their actual data range — both `wait.duration` and `failover.duration` had to be fixed (they're seconds-scale, the OTel default is tuned for HTTP milliseconds).

Tests cover: queue depth tracking across enqueue / drain / eviction / cancel / shutdown; failover-duration recording with simulated clock advance; drain outcome on both graceful and force-close paths; explicit bucket distribution for all three histograms; and OTel boundary-inclusivity (`sample <= Bound[i]` → `Bucket[i]`).

### DIFF
--- a/docs/query_serving/failover_sizing.md
+++ b/docs/query_serving/failover_sizing.md
@@ -155,12 +155,21 @@ After deploying the new values, watch these queries for a week:
 # Should stay flat at zero for healthy operation.
 rate(multigateway_buffer_requests_evicted_total{reason="window_exceeded"}[5m])
 rate(multigateway_buffer_requests_evicted_total{reason="buffer_full"}[5m])
-rate(multigateway_buffer_requests_evicted_total{reason="max_duration"}[5m])
 ```
 
 Any non-zero rate means at least one query failed because of buffer
-limits — either size, per-request window, or shard-level max
-duration. Each maps to a specific flag to raise.
+limits — per-request window timeout or buffer at capacity.
+
+Exceeding `--buffer-max-failover-duration` does not produce an
+eviction. The timer drains buffered entries early, and those entries
+then fail on retry against the still-unavailable primary. Detect
+this by watching `failover.duration` p99 approach the configured
+cap:
+
+```promql
+histogram_quantile(0.99,
+  sum(rate(multigateway_buffer_failover_duration_bucket[1h])) by (le))
+```
 
 ## Worked Example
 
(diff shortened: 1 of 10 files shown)


docs/query_serving/failover_sizing.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Step 4: Validate
+
+After deploying the new values, watch these queries for a week:
+
+```promql
+# Should stay flat at zero for healthy operation.
+rate(multigateway_buffer_requests_evicted_total{reason="window_exceeded"}[5m])
+rate(multigateway_buffer_requests_evicted_total{reason="buffer_full"}[5m])
+rate(multigateway_buffer_requests_evicted_total{reason="max_duration"}[5m])
```
haritabh17: Dont see this being emitted 🤔

haritabh17 (approved):

mkindahl (approved):

docs/query_serving/failover_sizing.md
GuptaManan100: Good catch! you're right, `reason="max_duration"` is not emitted anywhere. The only eviction reasons the buffer produces are `buffer_full` (`buffer.go:256`), `window_exceeded` (`timeout_thread.go:141`), and `context_canceled` (`shard_buffer.go:172`).

`--buffer-max-failover-duration` exceedance doesn't produce an eviction, when the timer fires, `stopBuffering` drains the entries (they show up under `requests.drained`) and they fail on retry against the still-unavailable primary.

Fixed in `docs/query_serving/failover_sizing.md`:

- Removed the bogus `reason="max_duration"` query from the Step 4 validation section.
- Added a note explaining how `--buffer-max-failover-duration` exceedance actually manifests, with a query against `failover.duration` p99 vs. the configured cap.
- Removed `max_duration` from the regex in the Continuous Validation alert, leaving `window_exceeded` as the standalone signal.

PR #980, opened 2026-05-11
test(pgregresstest): install vpid-aware lock-detection shim for isolation harness

haritabh17:
## Summary

- Installs a PL/pgSQL shim `public.multigres_test_session_is_blocked` in the `postgres` DB. Resolves a multigateway vpid to one or more PostgreSQL backend PIDs by LIKE-matching the `multigres_vpid:<id>` stamp on `pg_stat_activity.application_name` (stamp produced by #967), then aggregates `pg_blocking_pids` and `pg_safe_snapshot_blocking_pids` across every stamped backend currently associated with the vpid.
- Aggregation across multiple backends is required: a vpid can map to several backends in flight (a leftover stamp on a pool conn plus the live reserved conn), and SSI safe-snapshot waits (e.g. `read-only-anomaly-3`) need both blocking views.
- Logs every invocation to `public.isolation_debug_log` with a full `pg_stat_activity` snapshot. `dumpIsolationDebugLog` prints the last 30 entries to test output after each isolation run for post-mortem debugging of hung specs.
- Plumbs `directPgPort` through `RunIsolationTests` so the shim is installed directly on PostgreSQL, bypassing the multigateway. Multipooler is configured with `--database=postgres` and routes every query to that DB regardless of the dbname in the client startup packet, so installing in `isolation_regression` would land the function in a DB the harness never touches.
- `ensureIsolationHarnessDatabase` recreates `isolation_regression` so `pg_isolation_regress`'s `--use-existing` invocation finds the database.

## Stack

This PR is part of a 3-PR series:

- #967 — vpid stamping pipeline on PostgreSQL backends. **Must merge first.** The shim relies on the `multigres_vpid:<id>` stamp; without it, the shim returns no backends and every blocking spec reports not-blocked.
- **This PR** — vpid-aware shim function + diagnostic plumbing.
- Follow-up PR — patches `isolationtester.c` to call the shim and caps `PG_TEST_TIMEOUT_DEFAULT`. Depends on this PR.

Until #967 merges, this PR's diff against `main` also contains the B commits; once #967 lands the branch will be rebased onto `main` and the diff will reduce to only the harness changes.

## Test plan

- [ ] Confirm shim installs successfully against a primary's `postgres` DB.
- [ ] Run isolation suite once #967 + the follow-up patch PR are stacked; verify blocking specs report blocked/unblocked correctly via the shim.
- [ ] Verify `isolation_debug_log` accumulates entries and `dumpIsolationDebugLog` prints them on test failure.

### DIFF
--- a/go/test/endtoend/pgregresstest/postgres_builder.go
+++ b/go/test/endtoend/pgregresstest/postgres_builder.go
@@ -666,9 +666,9 @@ func (pb *PostgresBuilder) BuildIsolation(t *testing.T, ctx context.Context) err
 	return nil
 }
 
-// truncate clips s to at most n characters (with an ellipsis suffix when
+// truncateForLog clips s to at most n characters (with an ellipsis suffix when
 // truncation occurs). Used for compact log/error messages.
-func truncate(s string, n int) string {
+func truncateForLog(s string, n int) string {
 	if len(s) <= n {
 		return s
 	}
@@ -838,7 +838,7 @@ $$`,
 	}
 	for _, stmt := range stmts {
 		if _, err := db.Exec(stmt); err != nil {
-			return fmt.Errorf("failed to execute statement [%s]: %w", truncate(stmt, 80), err)
+			return fmt.Errorf("failed to execute statement [%s]: %w", truncateForLog(stmt, 80), err)
 		}
 	}
 
--- a/go/test/endtoend/pgregresstest/postgres_builder.go
+++ b/go/test/endtoend/pgregresstest/postgres_builder.go
@@ -17,6 +17,7 @@ package pgregresstest
 import (
 	"bytes"
 	"context"
+	"database/sql"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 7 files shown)


go/test/endtoend/pgregresstest/postgres_builder.go
```diff
@@ -662,16 +666,276 @@ func (pb *PostgresBuilder) BuildIsolation(t *testing.T, ctx context.Context) err
 	return nil
 }
 
+// truncate clips s to at most n characters (with an ellipsis suffix when
+// truncation occurs). Used for compact log/error messages.
+func truncate(s string, n int) string {
```
GuptaManan100: This is a very generic name, could cause incorrect usage or confusion later. Consider renaming to something more specific like `truncateForLog`

GuptaManan100 (approved):
LGTM! I will admit I did not read the whole SQL that is being passed through 😆, I'll just wait for them time when it works!

PR #981, opened 2026-05-11
test(pgregresstest): patch isolationtester for multigres + cap step timeout

haritabh17:
## Summary

- Patches the cached upstream PG `isolationtester.c` at build time (via a new `patchIsolationtester` helper hooked into `BuildIsolation`) to fix two patterns that don't work through multigateway:
  - Per-session `set_config('application_name', current_setting() || '/' || \$1, false)` uses a bound parameter; multigateway's planner rejects non-literal `set_config` because the pooler tracks the literal value for `is_local=false`. Rewritten to build the literal client-side via `PQescapeLiteral` and send via `PQexec`. Without this, every per-session setup exits 1 and the entire suite reports 117 instant failures.
  - The wait-query prepares `pg_catalog.pg_isolation_test_session_is_blocked(\$1, '{...}')`. The C builtin is not vpid-aware (operates on real backend PIDs), and `CREATE OR REPLACE` on a `pg_catalog` C builtin was unreliable in practice (fresh backends bound the C entry, returning false for every probe and hanging every blocking spec at max_step_wait). Rewritten to call `public.multigres_test_session_is_blocked` with explicit `::int4` / `::int4[]` casts so PG resolves the function under extended-protocol Parse with `paramTypes=NULL`.
- Caps `PG_TEST_TIMEOUT_DEFAULT=5` in `runTestSuite`. isolationtester's default 180s per-step timeout means `max_step_wait=360s` and a hard-exit at 720s, so a single compat-incompatible spec can burn ~12 min of suite context. The cap brings each step to ~10s cancel / ~20s hard-exit — still ~500× headroom over isolationtester's 10ms poll interval — so the schedule actually finishes.
- `patchIsolationtester` runs `git checkout --` on the source before each patch, so repeat invocations against the cached PG checkout are idempotent. Regression suite is unaffected (regression tests don't read `PG_TEST_TIMEOUT_DEFAULT`).

## Stack

This PR is the last of a 3-PR series:

- #967 — vpid stamping pipeline. Provides the `multigres_vpid:<id>` tag the shim resolves.
- #980 — installs `public.multigres_test_session_is_blocked`. The patched wait-query in this PR calls that function. **Must merge before this PR**, otherwise every blocking spec fails with `function does not exist`.
- **This PR** — isolationtester patches + step-timeout cap.

Until the dependencies land, this PR's diff against `main` also contains commits from #967 and #980. After they merge, this branch will be rebased onto `main` and the diff will reduce to the two harness commits.

## Test plan

- [x] Confirm `patchIsolationtester` produces a clean diff against a fresh PG checkout and is idempotent on a previously patched checkout.
- [x] After #967 and #980 merge, run full isolation suite end-to-end and verify blocking / SSI specs pass.
- [x] Confirm 117-spec schedule completes within reasonable time (no spec stuck at 720s hard-exit).

### DIFF
--- a/go/test/endtoend/pgregresstest/postgres_builder.go
+++ b/go/test/endtoend/pgregresstest/postgres_builder.go
@@ -146,6 +146,16 @@ func (pb *PostgresBuilder) runTestSuite(t *testing.T, ctx context.Context, cmd *
 		"PGDATABASE=postgres",
 		"PGCONNECT_TIMEOUT=10",
 		"PGOPTIONS="+pgOptions,
+		// PG_TEST_TIMEOUT_DEFAULT (in seconds) caps how long isolationtester
+		// waits per step before cancelling. Upstream default is 180 →
+		// max_step_wait = 360 s with hard-exit at 720 s, which lets a
+		// single compat-incompatible spec burn ~12 minutes of the suite
+		// ctx and starve the rest of the schedule. The cap applies per
+		// step and a multi-permutation spec where every permutation
+		// hangs scales linearly. 5 s gives 10 s cancel / 20 s hard-exit
+		// per step, ~500x headroom over the 10 ms poll interval used to
+		// detect legitimate blocking.
+		"PG_TEST_TIMEOUT_DEFAULT=5",
 	)
 
 	// Capture stdout for result parsing while still printing to the terminal.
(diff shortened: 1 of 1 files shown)


GuptaManan100 (approved):
Merge once the checklists in the PR description have been checked off

PR #982, opened 2026-05-11
refactor(consensus): make externally certified discovery more crisp

dweitzman:
This is a pure refactoring -- no intended behavior change.

The leaderFilter parameter on buildProposalCore feels like a potential mistake waiting to happen, so this PR reframes it as a discovery function. For externally certified proposals, discovery is based on the frozen LSN. For safe proposals, it's the most advanced timeline among recruited poolers.

In the future we could do something like say that it's possible to recruit a resigning leader but choose to ignore it for discovery if it has transactions that are provably not durable.


mkindahl (approved):
LGTM

PR #986, opened 2026-05-12
feat(scram): advertise SCRAM-SHA-256-PLUS over TLS

haritabh17:
## Summary

- Multigateway now advertises `SCRAM-SHA-256-PLUS` alongside `SCRAM-SHA-256` over TLS, mirroring native PostgreSQL.
- Implements `tls-server-end-point` channel binding (RFC 5929) with PG's MD5/SHA-1 → SHA-256 quirk; Ed25519 rejected to match PG.
- Enforces GS2 flag ↔ mechanism agreement, rejects `y`-flag downgrade attempts, rejects SASL authzid.
- SQLSTATEs (28000 / 08P01 / 0A000) and error messages match PG17 `auth-scram.c` verbatim — libpq-compatible clients see identical diagnostics.

Closes [MUL-371](https://linear.app/supabase/issue/MUL-371).

## Test plan

- [x] Unit tests in `scram/`: cbind hash compute (RSA/ECDSA at SHA-256/384/512, MD5/SHA-1 quirk, Ed25519 reject), PLUS handshake happy path, tampered cbind, downgrade, authzid, protocol violations, state guards.
- [x] Wire-level tests in `server/`: TLS+PLUS handshake end-to-end via real `tls.Conn`, tampered cbind FATAL 28000, downgrade FATAL 28000, authzid FATAL 0A000, plaintext never advertises PLUS.
- [x] `go test -race -count=5 ./go/common/pgprotocol/...` — clean.
- [x] Downstream `go/services/...` race tests — no regressions.
- [x] Manual smoke with real `psql --set=channel_binding=require` against a local cluster — relaunched zone1 `multigateway` with `--pg-tls-cert-file`/`--pg-tls-key-file` (self-signed RSA cert), then verified four scenarios over `sslmode=require`: `channel_binding=require` (PLUS selected, query OK), `channel_binding=prefer` (PLUS picked, query OK), `channel_binding=disable` (base SCRAM-SHA-256 used, query OK), and `channel_binding=require` against a plaintext port (libpq aborts with `channel binding required, but SSL not in use`).

### DIFF
--- a/go/common/pgprotocol/server/startup.go
+++ b/go/common/pgprotocol/server/startup.go
@@ -16,6 +16,7 @@ package server
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"maps"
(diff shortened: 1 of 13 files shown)


go/common/pgprotocol/server/startup.go
```diff
@@ -182,6 +183,33 @@ func (c *Conn) handleSSLRequest() error {
...
+	// fall back to parsing Certificate[0] for the common case where Leaf was
+	// left nil. A failure here is non-fatal: we just won't advertise PLUS and
+	// auth will fall back to SCRAM-SHA-256.
+	//
+	// LIMITATION: tlsConfig.Certificates[0] is the canonical cert in this
+	// config slot. If a deployment uses tlsConfig.GetCertificate or
+	// GetConfigForClient (dynamic SNI), the cert actually negotiated for
+	// this handshake may differ — PLUS will then fall back to base SCRAM
+	// because Go's tls.Conn.ConnectionState() does not expose the server-
+	// presented cert. See MUL-434 for the wrapper-based fix that captures
```
GuptaManan100: Let's not have linear links in our docs

GuptaManan100 (approved):
Rest LGTM! We should complete the manual `psql --set=channel_binding=require` smoke test in the PR plan before merging!

PR #987, opened 2026-05-12
fix(server): update ssl_test.go for HashProvider→CredentialProvider rename

mkindahl:
The RequireTLS tests in ssl_test.go were authored against the old HashProvider API. Commit f23e6fb4 renamed the field and constructor to CredentialProvider/newMockCredentialProvider before this test file was merged, leaving three stale references that broke the build.


haritabh17 (approved):
✅

PR #988, opened 2026-05-12
fix(multipooler): wait for pg_reload_conf to be processed before returning

dweitzman:
pg_reload_conf() returns immediately after sending SIGHUP to postmaster. We can't determine if the config was reloaded on every backend process, but we can at least determine if it's been loaded on more than zero processes.

This PR also centralizes config reloading into a single helper.


mkindahl (approved):
LGTM

PR #990, opened 2026-05-13
fix(replication): actively terminate walreceiver during pause; use pg_stat_activity

mkindahl:
pg_reload_conf() sends SIGHUP asynchronously, so the startup process may restart the walreceiver before it has read the new (empty) primary_conninfo. Actively terminate the walreceiver after clearing primary_conninfo and re-terminate on every poll tick until it is confirmed gone.

Switch from pg_stat_wal_receiver to pg_stat_activity throughout: the former reads from WalRcvCtlBlock shared memory which can hold stale entries after a walreceiver exits, causing spurious "still running" results.

Combine the count check and pg_terminate_backend into a single query per tick so only one pool connection is consumed and the terminate cannot fail silently due to pool saturation.

Use ctxutil.Detach for per-query contexts in waitForReceiverDisconnect so that expiry of the outer wait context does not cause pool.Get to return ErrCtxTimeout instead of propagating the proper DEADLINE_EXCEEDED error.

Update test mocks: pattern "SELECT COUNT.*pg_stat_wal_receiver" → "terminated_count"; result columns {"count"} → {"total_count", "terminated_count"}.

### DIFF
--- a/go/services/multipooler/manager/pg_replication.go
+++ b/go/services/multipooler/manager/pg_replication.go
@@ -27,6 +27,7 @@ import (
 	"github.com/multigres/multigres/go/common/mterrors"
 	"github.com/multigres/multigres/go/common/parser/ast"
 	"github.com/multigres/multigres/go/services/multipooler/executor"
+	"github.com/multigres/multigres/go/tools/ctxutil"
 
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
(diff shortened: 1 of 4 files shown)


go/services/multipooler/manager/pg_replication.go
```diff
@@ -549,10 +568,19 @@ func (pm *MultiPoolerManager) queryReplayState(ctx context.Context) (replayLsn s
...
+// table and accurately reflects whether a walreceiver backend is still running.
 func (pm *MultiPoolerManager) waitForReceiverDisconnect(ctx context.Context) (*multipoolermanagerdatapb.StandbyReplicationStatus, error) {
-	// Create a context with timeout for the polling loop
-	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	// Create a context with timeout for the polling loop.
+	// 30 s gives enough margin for the case where the walreceiver's upstream
+	// (the primary's postgres) is itself restarting: the walreceiver will not
+	// exit until the upstream TCP connection is torn down, which typically
+	// happens within ~7 s but can take longer under load.
+	waitCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
dweitzman, line 583: 30s here seems high. For an RPC like Recruit(), I wouldn't expect orch to wait around that long.

Is all that waiting necessary? Can we prove earlier that no more WAL will be received even if the WAL receiver process isn't in an explicitly disconnected/stopped state?

mkindahl:
Closing this one for now. Killing the process repeatedly to ensure that is dead is not a good solution.

PR #992, opened 2026-05-13
feat(multipooler): rule store owns GUC writes and controls promotion timing

dweitzman:
## Summary

Makes `ruleStore` the sole orchestrator of both the WAL write and the
`synchronous_standby_names`/`synchronous_commit` GUC writes, so their
ordering is always correct.

`updateRule` now sequences GUC changes around the WAL write: a Both
policy (satisfying old and new durability simultaneously) is applied
before the write; the Incoming policy narrows it after the write
commits. For promotions, the pre-write GUC fires while the node is still
a standby — `pg_promote()` only runs after replication is in place.

A `SELECT FOR UPDATE NOWAIT` at the start of every `updateRule` call
drains any in-flight writes from the previous action lock holder before
`SetPolicy` touches the GUC, stamped on the context as a proof token.

`hasInconsistentGUC` / `reconcileGUC` let the postgres monitor repair a
stale GUC without waiting for the next rule change.


go/services/multipooler/manager/rule_store.go
```diff
@@ -344,6 +473,10 @@ func (rs *ruleStore) observePosition(ctx context.Context) (*clustermetadatapb.Po
 // complete within that time. A timeout typically indicates that synchronous
 // replication is not functioning.
 func (rs *ruleStore) updateRule(ctx context.Context, update *ruleUpdateBuilder) (*clustermetadatapb.PoolerPosition, error) {
```
dweitzman, line 494: This is the meat of the PR. `updateRule()` now handles both the GUC change and the WAL change

go/services/multipooler/manager/rule_store.go
```diff
@@ -355,24 +488,110 @@ func (rs *ruleStore) updateRule(ctx context.Context, update *ruleUpdateBuilder)
 		return nil, nil
 	}
 
-	// Convert optional leader ID; empty string causes NULLIF→COALESCE to keep existing.
-	var leaderStr string
+	isPromotion := update.promotionHook != nil
+
+	// Read the current rule to establish the CAS baseline and drain any in-flight
```
dweitzman, line 528: Previously reading the rule row and updating it was multiple statements within a single common table expression query, but now reading and writing are two separate queries

The action + row lock ensure that no previous rule change is lingering and still in progress if we make it beyond `readCurrentRuleLocked()`

There is an unhandled edge case if a previous write went into the WAL and committed without achieving quorum, due to a process crash or something like that. We'll handle that edge case when we deal with propagation, since it's essentially the same problem: recognizing and completing and unfinished proposal.

go/services/multipooler/manager/rule_store.go
```diff
@@ -385,120 +604,94 @@ func (rs *ruleStore) updateRule(ctx context.Context, update *ruleUpdateBuilder)
...
+	// the caller omits withDurabilityPolicy(), so these values are always present.
+	dpName := newDP.PolicyName
+	dpQuorumType := newDP.QuorumType.String()
+	dpRequiredCount := int64(newDP.RequiredCount)
+
+	// Apply the transition GUC before writing the rule. The transition (Both) policy
+	// satisfies both old and new durability requirements simultaneously.
+	// Promotion path: set GUC while still a standby, then call pg_promote().
+	// Primary path: set GUC immediately before the write CTE.
+	if isPromotion {
```
dweitzman, line 652: This is new: we can do promotion to leadership within the updateRule() call, but we set the GUC for sync standby names _before_ promoting so that all writes as primary will need to satisfy the "both" quorum even if something like autovacuum or a checkpoint sneaks in before the rule table update.

go/services/multipooler/manager/rule_store.go
```diff
@@ -578,6 +765,12 @@ func (rs *ruleStore) updateRule(ctx context.Context, update *ruleUpdateBuilder)
 	if err != nil {
 		return nil, mterrors.Wrap(err, "failed to parse written rule position")
 	}
+
+	// Apply the incoming (new) GUC after the write commits.
+	if err := rs.syncStandby.SetPolicy(lockedCtx, transition.Incoming); err != nil {
+		return nil, fmt.Errorf("post-write GUC: %w", err)
+	}
+
```
dweitzman, line 813: In the future for propagation & dealing with failed rule changes, at this point we can record success so that coordinators and the postgres monitor understand the transition is complete.

go/services/multipooler/manager/rpc_backup.go
```diff
@@ -348,25 +348,13 @@ func (pm *MultiPoolerManager) restoreFromBackupLocked(ctx context.Context, backu
...
-		pm.logger.InfoContext(ctx, "Deleting consensus term file after restore; node will re-join consensus from term 0")
-		if err := pm.consensusState.DeleteTermFile(); err != nil {
-			return mterrors.Wrap(err, "failed to delete consensus term file after restore")
-		}
-		pm.healthStreamer.UpdateLeaderObservation(nil)
-		return nil
-	}); err != nil {
-		return err
-	}
+	// Clear the in-memory leader observation. The restored PGDATA may have been
```
dweitzman, line 351: Removed a prior behavior that deleted the term revocation when a backup is restored. Deleting consensus promises feels risky

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -728,51 +727,14 @@ func (pm *MultiPoolerManager) Propose(ctx context.Context, req *consensusdatapb.
...
-	}
 	// TODO: If this proposal already exists, we're being asked to propagate
 	// rather than make a new entry. We can make the rule store understand
 	// propagation for that case.
 	reason := req.GetReason()
 	if reason == "" {
 		reason = "propose"
 	}
-	if _, err = pm.rules.updateRule(ctx, newRuleUpdate(
+	ruleUpdate := newRuleUpdate(
```
dweitzman, line 752: The Propose() endpoint becomes simpler because the GUC change complexity is now pushed down into updateRule() within the rule store

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -539,27 +539,6 @@ func (pm *MultiPoolerManager) configureSynchronousReplicationLocked(ctx context.
 	return nil
 }
 
-// applyGUCsForSyncReplication sets the synchronous_commit and synchronous_standby_names
-// GUCs without the primary guardrail or rule-history write. Safe to call on a standby
-// (e.g., before pg_promote) so the promoted primary takes effect with the new config.
-// The caller must hold the action lock.
-func (pm *MultiPoolerManager) applyGUCsForSyncReplication(
```
dweitzman, line 546: Some code is disappearing from rpc_manager.go because `updateRule()` handles the GUC changes now

go/services/multipooler/manager/sync_standby_manager.go
```diff
@@ -162,16 +163,56 @@ func (s *postgresqlSyncStandbyManager) computeGUC(pc commonconsensus.PolicyWithC
...
 // NeedsApply returns true if the given policy would produce GUC strings that
-// differ from the in-memory cache. Safe to call without the action lock.
-func (s *postgresqlSyncStandbyManager) NeedsApply(pc commonconsensus.PolicyWithCohort) (bool, error) {
+// differ from what postgres currently has. Safe to call without the action lock.
+//
+// It queries postgres directly so it can detect GUC changes made outside this
+// manager (e.g. manual ALTER SYSTEM). On any query failure it falls back to the
+// in-memory cache, which is reliable because this manager is the sole writer of
+// these GUCs.
+func (s *postgresqlSyncStandbyManager) NeedsApply(ctx context.Context, pc commonconsensus.PolicyWithCohort) (bool, error) {
```
dweitzman, line 172: A change here vs the previous PR: when the postgres monitor checks if the GUC for sync standby names is out of date, it always checks against Postgres directly rather than assuming the cached values are correct. This allows it to recover after a rogue person or agent manually sets the wrong value.

go/services/multipooler/manager/rule_store.go
```diff
@@ -90,6 +105,52 @@ func (rs *ruleStore) cachedPosition() *clustermetadatapb.PoolerPosition {
 	return rs.lastPos
 }
 
+// hasInconsistentGUC returns true if the cached rule's policy would produce
+// different GUC strings than what postgres currently has. Safe to call
+// without the action lock.
+func (rs *ruleStore) hasInconsistentGUC(ctx context.Context) bool {
```
dweitzman, line 115: This is used without holding the action lock to estimate whether it's likely that sync standby names will need updating

If so, reconcileGUC() must be called while holding the action lock to actually do the update

rafael (approved):

PR #994, opened 2026-05-14
perf(multigateway): cache OTel metric attribute sets per query dimension

GuptaManan100:
## Summary

- Per-query metric instrumentation in `recordQueryCompletion` showed up at ~7.5% of gateway CPU in pprof, dominated by attribute-set construction (sort + dedupe) inside the OTel SDK on every `Record`/`Add` call.
- Cache a `metric.MeasurementOption` per stable dimension tuple (db, op, fingerprint, …) so that the OTel attribute-set work happens once per tuple instead of once per query.
- Dedupe `mterrors.ClassifyErrorSource`: skip it entirely on the success path and compute it exactly once (then reuse) on the error path. Previously called twice on errors and once on success.
- Adds unit tests for cache reuse + emitted attribute shape, and benchmarks comparing the cached vs. pre-cache path.

## Description

The hot path for every query funnels into three OTel calls — `queryDuration.Record`, `rowsReturned.Record`, and one `tableQueries.Add` per table touched. Each call previously built a fresh `[]attribute.KeyValue` and handed it to `metric.WithAttributes`, which the SDK then sorted and deduplicated into an `attribute.Set`. The attribute *values* for a given (db, op, protocol, fingerprint) tuple are stable across millions of queries, so re-sorting them every time was wasted work.

This PR caches the `MeasurementOption` (which wraps a pre-built `attribute.Set`) on a per-instrument `sync.Map`, keyed by a small struct of the dimensions. The hot path becomes a `sync.Map.Load` followed by a single `Record`/`Add`. Cache misses (cold start, new fingerprint) take a write through `LoadOrStore` and then proceed normally.

Cardinality is bounded by external invariants — the query registry caps fingerprint labels into `__other__` / `__utility__`, SQLSTATE / status / protocol are enum-shaped, and `db.namespace` / `db.operation.name` / `db.collection.name` come from authenticated PG protocol fields rather than arbitrary user input. A short comment in `metrics.go` notes this so no one reaches for an explicit cap reflexively.

### Benchmarks (Apple M4, single-tuple warm cache)

```
BenchmarkQueryDuration_Cached            ~115 ns/op    16 B/op   1 allocs/op
BenchmarkQueryDuration_Uncached          ~430 ns/op   808 B/op   4 allocs/op   (3.7×)

BenchmarkRowsReturned_Cached              ~84 ns/op    16 B/op   1 allocs/op
BenchmarkRowsReturned_Uncached           ~253 ns/op   424 B/op   4 allocs/op   (3.0×)

BenchmarkTableQueries_Cached              ~75 ns/op    16 B/op   1 allocs/op
BenchmarkTableQueries_Uncached           ~246 ns/op   424 B/op   4 allocs/op   (3.3×)

BenchmarkRecordQueryCompletion_Mixed_Cached    ~320 ns/op    48 B/op    3 allocs/op
BenchmarkRecordQueryCompletion_Mixed_Uncached ~1035 ns/op  1656 B/op   12 allocs/op   (3.2×)
```

The Mixed benchmark rotates across 16 distinct (op, fingerprint) tuples per iteration, so the speedup isn't artifact of best-case single-tuple replay.

### Notes for reviewers

- Metric names, descriptions, units, bucket boundaries, and attribute *keys* are unchanged — dashboards continue to see the same shape. `TestHandlerMetrics_EmittedAttributes` asserts this end-to-end via a `sdkmetric.ManualReader`.
- Wrapper types (`QueryDuration`, `QueryErrors`, `RowsReturned`, `TableQueries`) moved to pointer receivers because they now hold a `sync.Map`. `HandlerMetrics` correspondingly holds pointers. No external callers were affected.

## Test plan

- [x] `go test ./go/services/multigateway/handler/...` passes
- [x] `go build ./...` passes
- [x] `golangci-lint run ./go/services/multigateway/handler/...` clean
- [x] New cache-hit tests verify identical dimensions reuse the entry and distinct dimensions add separate entries
- [x] End-to-end attribute-shape test via `sdkmetric.ManualReader` confirms emitted metric data points carry the same attributes as before
- [ ] Re-run the original pprof capture (select-only + tpcb-like, c=100) against this branch to confirm the projected ~1.5% gateway CPU drop

### DIFF
--- a/go/services/multigateway/handler/handler.go
+++ b/go/services/multigateway/handler/handler.go
@@ -543,6 +543,16 @@ func (h *MultiGatewayHandler) ConnectionClosed(conn *server.Conn) {
 	h.psc.RemoveConnection(conn.ConnectionID())
 }
 
+// classifyErrorSource calls mterrors.ClassifyErrorSource only when err is
+// non-nil. The success path is overwhelmingly common, so short-circuiting
+// here avoids a per-query function call + interface check.
+func classifyErrorSource(err error) string {
+	if err == nil {
+		return ""
+	}
+	return mterrors.ClassifyErrorSource(err)
+}
+
 // recordQueryCompletion records all three metrics and emits a structured
 // query log entry. Centralises the instrumentation logic shared by
 // HandleQuery and HandleExecute.
@@ -585,10 +595,12 @@ func (h *MultiGatewayHandler) recordQueryCompletion(
 
 	status := QueryStatusOK
 	errorType := ""
+	errorSource := ""
 	if err != nil {
 		status = QueryStatusError
 		errorType = mterrors.ExtractSQLSTATE(err)
-		h.metrics.queryErrors.Add(ctx, errorType, mterrors.ClassifyErrorSource(err), dbNamespace, operationName, fingerprintLabel)
+		errorSource = classifyErrorSource(err)
+		h.metrics.queryErrors.Add(ctx, errorType, errorSource, dbNamespace, operationName, fingerprintLabel)
 	}
 
 	h.metrics.queryDuration.Record(ctx, totalDuration.Seconds(), dbNamespace, operationName, queryProtocol, errorType, status, fingerprintLabel)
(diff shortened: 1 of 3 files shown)


haritabh17 (approved):
🚀

PR #995, opened 2026-05-14
ci(sqllogictest): run weekly, alert on any pg/mg divergence

GuptaManan100:
## Summary

- Drop the nightly cron and switch the sqllogictest differential suite to a single weekly run (Mondays, 09:30 UTC).
- Remove the cached-baseline + `detect-regressions.sh` machinery. The weekly run now Slack-alerts whenever any file passes on postgres but fails on multigateway, with no diff against prior runs.
- Bump the per-file timeout from 5m to 10m to absorb the occasional slow file.

## Description

Now that the suite is stable, daily runs were overkill — the signal we actually want is "did any file regress relative to postgres?", which is fully derivable from a single run's `results.json`. Removing the baseline cache also eliminates a class of confusion (stale baselines, PR-labeled runs reading old data) and makes the workflow easier to reason about. The Slack alert wording was updated from "regression" to "divergence" to match the new semantics. The `detect-regressions.sh` script is still used by `test-pgregress.yml`, so it's left in place.

### DIFF
--- a/.github/workflows/test-sqllogictest.yml
+++ b/.github/workflows/test-sqllogictest.yml
@@ -16,8 +16,7 @@ name: SQLLogicTest Differential Suite
 
 on: # yamllint disable-line rule:truthy
   schedule:
-    - cron: "30 3 * * *" # Nightly at 3:30 AM UTC (offset from pgregress at 3:00)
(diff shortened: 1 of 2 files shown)


haritabh17 (approved):
🚀

PR #996, opened 2026-05-14
fix(multigateway): error on duplicate PREPARE name in simple protocol

GuptaManan100:
## Summary

- SQL-level `PREPARE name AS ...` now errors with `prepared statement "name" already exists` (SQLSTATE 42P05) when the name is already in use on the session, matching upstream PostgreSQL.
- Previously the consolidator silently replaced the existing statement, so a buggy client never saw the conflict and pgregress diffs included ~50 hunks for this case.

## Problem

Multigateway's prepared-statement consolidator replaces any existing entry when a Parse comes in for the same name on the same connection. That behavior is intentional for the extended protocol — clients (notably pq) re-Parse with the same name after a failed Describe, and we need to tolerate that — but it also leaks through to simple-protocol `PREPARE`, where upstream PostgreSQL is required to raise `duplicate_prepared_statement` (42P05). See MUL-388 / MUL-436.

## Solution

Add the duplicate check at the simple-PREPARE boundary instead of inside the consolidator: `PreparedStatementPrimitive.executePrepare` looks up the user-visible name via `Handler.GetPreparedStatementInfo` and returns a new `NewDuplicatePreparedStatementError` if it already exists. `HandleParse`'s replace-on-reparse behavior is left intact so the extended-protocol retry path keeps working.

Also adds:

- `PgSSDuplicatePreparedStmt = "42P05"` and `NewDuplicatePreparedStatementError` in `mterrors`.
- A unit test in the planner package covering the duplicate path.
- A subtest in the existing `TestSimpleProtocolPreparedStatements` comparison harness so the behavior is verified against both direct PostgreSQL and multigateway, and the first statement's definition is confirmed to survive the rejected re-PREPARE.

This addresses defect 3 from MUL-388 (silent re-PREPARE). Defects 1 (user name not surfaced in `pg_prepared_statements`) and 2 (state leak across pooled backend sessions) are intentionally out of scope here.

## Test plan

- [x] `go test ./go/services/multigateway/planner -run TestPlanPrepareStmt -v`
- [x] `go test ./go/test/endtoend/queryserving -run TestSimpleProtocolPreparedStatements -v` (postgres + multigateway, all 16 subtests pass)
- [x] `go test -short ./go/services/multigateway/... ./go/common/mterrors/... ./go/common/preparedstatement/...`

### DIFF
--- a/go/common/mterrors/code.go
+++ b/go/common/mterrors/code.go
@@ -34,6 +34,7 @@ const (
 	PgSSAuthFailed              = "28P01" // invalid_password
 	PgSSInvalidAuthSpec         = "28000" // invalid_authorization_specification
 	PgSSInvalidCursorName       = "34000" // invalid_cursor_name
+	PgSSDuplicatePreparedStmt   = "42P05" // duplicate_prepared_statement
 	PgSSInsufficientPrivilege   = "42501" // insufficient_privilege
 	PgSSSyntaxError             = "42601" // syntax_error
 	PgSSUndefinedObject         = "42704" // undefined_object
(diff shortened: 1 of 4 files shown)


haritabh17 (approved):
🚀

PR #998, opened 2026-05-14
feat(scram): capture per-conn TLS cert for SCRAM-SHA-256-PLUS under SNI

haritabh17:
## Summary

- Capture the leaf certificate that crypto/tls actually presents during the handshake so SCRAM-SHA-256-PLUS channel binding works under dynamic-cert deployments (SNI multi-tenant edges).
- Wrap the listener's `tls.Config` to instrument `GetCertificate` and `GetConfigForClient`; static `Certificates[0]` deployments keep the existing post-handshake recovery and skip the wrapper.
- Warn at listener init when a non-nil `TLSConfig` yields no cert via any path, so the silent PLUS-off state is visible to operators.
- Cover the new paths end-to-end: GetCertificate, GetConfigForClient (including the per-handshake inner config), SNI dispatch with distinct leaves, predicate, static bypass, nil guards.

## Problem

SCRAM-SHA-256-PLUS (RFC 5929 `tls-server-end-point`) requires the server leaf certificate that was actually presented to the client. Prior to this change the server recovered it from `tlsConfig.Certificates[0]` after the handshake, which only works for static-cert deployments. Any deployment using `tls.Config.GetCertificate` or `GetConfigForClient` — the typical pattern for SNI-based multi-tenant TLS — left `tlsServerCert` nil and silently fell back to base SCRAM-SHA-256 even though TLS was negotiated. crypto/tls does not expose the server-presented leaf via `(*tls.Conn).ConnectionState()`, so there is no post-handshake way to recover it.

## Solution

Wrap the listener's `tls.Config` before calling `tls.Server` so the dynamic cert-selection paths invoke a capture hook with the `*tls.Certificate` crypto/tls is about to present. The captured leaf is stored on the `Conn` for the SCRAM advertisement that follows on the same goroutine, so no synchronization is required.

`GetConfigForClient` is handled by also wrapping the per-handshake inner `*tls.Config` it returns, since crypto/tls runs cert selection against that inner config. When the inner config carries only static `Certificates`, a synthetic `GetCertificate` is installed that mirrors crypto/tls's `SupportsCertificate`-based SNI matching so the chosen leaf is captured.

The static-only path is left untouched — the wrapper short-circuits and returns the original config without cloning, and the post-handshake `Certificates[0]` recovery still runs (now gated on `tlsServerCert == nil` so the dynamic path takes precedence when it has already captured).

Listener init validates that a non-nil `TLSConfig` will yield a server certificate via at least one supported path; otherwise it logs a warning, because the handshake would fail at runtime and PLUS would be silently disabled without one.

### DIFF
--- a/go/common/pgprotocol/server/tls_cert_capture.go
+++ b/go/common/pgprotocol/server/tls_cert_capture.go
@@ -1,0 +1,182 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+)
+
+// captureTLSServerCert records the leaf certificate that crypto/tls
+// selected for this connection's handshake. Invoked by the GetCertificate
+// / GetConfigForClient wrappers installed by wrapTLSConfigForCertCapture.
+//
+// Prefers the pre-parsed Leaf to avoid re-parsing on hot paths; falls back
+// to parsing Certificate[0] for callers that don't populate Leaf. A parse
+// failure is logged at warn and leaves tlsServerCert nil — auth then falls
+// back to SCRAM-SHA-256 without channel binding rather than failing the
+// handshake.
+//
+// Called synchronously from the handshake goroutine before tls.Handshake
+// returns; the subsequent reader (handleSSLRequest / SCRAM advertisement)
+// runs on the same goroutine, so no synchronization is needed.
+func (c *Conn) captureTLSServerCert(tlsCert *tls.Certificate) {
+	if tlsCert == nil {
+		return
+	}
+	if tlsCert.Leaf != nil {
+		c.tlsServerCert = tlsCert.Leaf
+		return
+	}
+	if len(tlsCert.Certificate) == 0 {
+		return
+	}
+	parsed, err := x509.ParseCertificate(tlsCert.Certificate[0])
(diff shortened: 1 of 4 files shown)


go/common/pgprotocol/server/tls_cert_capture.go
```diff
@@ -0,0 +1,182 @@
...
+//
+// The static Certificates[0] path is intentionally NOT routed through
+// capture here — the caller continues to handle it after the handshake
+// to keep behavior unchanged for the common single-cert deployment.
+//
+// capture is invoked synchronously on the handshake goroutine; the caller
+// must serialize subsequent reads (in this package, capture writes to the
+// owning *Conn which is only read after Handshake returns on the same
+// goroutine).
+func wrapTLSConfigForCertCapture(base *tls.Config, capture func(*tls.Certificate)) *tls.Config {
```
GuptaManan100, line 82: This function always clones `inner` even when it has neither `GetCertificate` nor any `Certificates` - the clone is then returned unmodified. Consider a guard returning `inner` unchanged in that branch, parallel to the static short-circuit in `wrapTLSConfigForCertCapture`

go/common/pgprotocol/server/tls_cert_capture.go
```diff
@@ -0,0 +1,182 @@
+// Copyright 2026 Supabase, Inc.
```
GuptaManan100, line 1: This whole file is structurally a workaround for a missing Go stdlib API: `(*tls.Conn).ConnectionState()` doesn't expose the server-presented leaf (the asymmetric counterpart of `PeerCertificates`). Other stacks call into OpenSSL's `SSL_get_certificate(port->ssl)` directly. The wrapper-of-getters pattern this PR uses is the canonical Go-community workaround, but it's load-bearing only because of that stdlib gap. Suggest adding a `// TODO(go-stdlib): replace with tls.ConnectionState.LocalCertificate (or equivalent) once Go exposes the server-side leaf — see golang/go issue tracker for the open ask.` comment near the top of `wrapTLSConfigForCertCapture` so a future maintainer knows the whole file can be deleted when the stdlib fills the gap. Also worth a one-liner pointing at this TODO from the `selectCertificate` doc-comment, since that helper is the most fragile piece (it re-implements stdlib cert-selection rules) and the one most likely to drift.

GuptaManan100 (approved):
Couple minor nits, but largely 🚀

PR #999, opened 2026-05-14
refactor(multiorch): drop useNewConsensusFlow flag and legacy appointment RPCs

dweitzman:
Now that the Recruit/Propose flow is the default and there is no need to fall back, remove the transitional `useNewConsensusFlow` flag along with the BeginTerm/Promote/EmergencyDemote RPCs that only the old appointment path called. AppointLeader and AppointInitialLeader now take a ShardKey so log/error context retains database, tablegroup, and shard.


go/services/multipooler/manager/rpc_backup.go
```diff
@@ -348,25 +348,11 @@ func (pm *MultiPoolerManager) restoreFromBackupLocked(ctx context.Context, backu
...
-	// The term file lives outside PGDATA and is not included in the backup, so
-	// its on-disk value may be ahead of what the restored PGDATA participated in.
-	// Deleting it resets the node to term 0; multiorch will advance the term to
-	// the current cluster value on first contact via BeginTerm.
-	//
-	// TODO: Revisit this when we restore backups for other reasons than to
-	// bootstrap a new node, e.g. point-in-time recovery for an existing node.
-	if err := telemetry.WithSpan(ctx, "restore/reset-consensus-term", func(ctx context.Context) error {
-		pm.logger.InfoContext(ctx, "Deleting consensus term file after restore; node will re-join consensus from term 0")
-		if err := pm.consensusState.DeleteTermFile(); err != nil {
```
dweitzman: This is a change in behavior. Previously a restore from backup deleted the term revocation. With the change, that's no longer the case

Preserving the revocation makes sense to me: it's a promise about revoking previous agents, not a statement about what is or isn't in the WAL. Rolling back the WAL shouldn't invalidate any promises made to coordinators

The reason for changing this now in the middle of a mostly-delete PR is just to avoid a test failure since this PR replaces some "force" settings in tests that bypasses term revocation behavior in some tests with real term numbers that only go up.

go/test/endtoend/queryserving/buffer_test.go
```diff
@@ -383,25 +384,30 @@ func triggerFailover(t *testing.T, setup *shardsetup.ShardSetup) {
...
 	require.NoError(t, err)
 	oldTerm := statusResp.ConsensusStatus.GetTermRevocation().GetRevokedBelowTerm()
+	outgoingRule := statusResp.ConsensusStatus.GetCurrentPosition().GetRule().GetRuleNumber()
+	require.NotNil(t, outgoingRule, "primary should have a recorded rule before recruit")
 
-	t.Logf("Triggering failover: BeginTerm on %s (term %d → %d)", currentPrimaryName, oldTerm, oldTerm+1)
+	t.Logf("Triggering failover: Recruit on %s (term %d → %d)", currentPrimaryName, oldTerm, oldTerm+1)
 
-	beginTermResp, err := primaryClient.Consensus.BeginTerm(
+	recruitResp, err := primaryClient.Consensus.Recruit(
```
dweitzman, line 392: Some tests have changed to call Recruit() instead of BeginTerm()

mkindahl (approved):
LGTM

PR #1000, opened 2026-05-15
feat(consensus): split Propose (leader) and SetTermPrimary (followers)

dweitzman:
Reshape the leader-appointment RPC surface: each RPC now has a single role.

**SetTermPrimary** (formerly Inform): tells a pooler the current leader.
The pooler compares the supplied `RuleNumber` against its own and either
ignores (revoked or not-newer), updates `primary_conninfo` (standby), or
demotes itself (stale primary). No prior Recruit required — rule numbers
are globally unique because a coordinator only earns the authority to
propose at term N by successfully recruiting term N. The coordinator's
choice of primary is not up for debate, so the RuleNumber compare is sufficient
under retries and out-of-order delivery.

**Propose is now leader-only.** It rejects when `proposal_leader != self`
and directs the coordinator to use SetTermPrimary for followers. The
coordinator dispatches accordingly: leader → Propose, followers →
SetTermPrimary.

**PoolerAddress** replaces `ProposalLeader` and the `MultiPooler` shape
inside `ReplicationPrimary` / `SetTermPrimary.leader` — one type
(id + host + postgres_port) for all consensus contact-info uses.

**MonitorPostgres self-heals primary_conninfo.** The monitor
compares the live `primary_conninfo` against the recorded
`ReplicationPrimary` and retries `ALTER SYSTEM SET primary_conninfo` if
they diverge (with revoked-rule guards). A SetTermPrimary that lands
while postgres is unavailable, fails mid-apply, or gets clobbered by an
external restart converges on its own — no follow-up RPC needed.

Other things folded in: `ReplicationPrimary` published in
`ConsensusStatus` so coordinators can skip redundant calls;
`TermRevocation.outgoing_rule` for stranded-recruit self-heal;
SetTermPrimary's standby branch publishes `LeaderObservation` and clears
any resignation signal.

`DemoteStalePrimary` and `SetPrimaryConnInfo` are now redundant —
SetTermPrimary covers their work. They remain for the deprecation window
but new code shouldn't call them.


go/services/multiorch/recovery/actions/fix_replication.go
```diff
@@ -215,22 +216,31 @@ func (a *FixReplicationAction) fixNotReplicating(
...
-	// Configure primary_conninfo on the replica
-	req := &multipoolermanagerdatapb.SetPrimaryConnInfoRequest{
-		Primary:               primary.MultiPooler,
-		StopReplicationBefore: true,
-		StartReplicationAfter: true,
-		CurrentTerm:           consensusTerm,
-		Force:                 false,
-	}
+	// Configure primary_conninfo on the replica.
+	if a.config.GetUseNewConsensusFlow() {
```
dweitzman, line 220: This is the multiorch change to use the new RPC if the new flow is enabled

go/services/multipooler/manager/manager.go
```diff
@@ -1848,6 +1924,25 @@ func (pm *MultiPoolerManager) takeRemedialAction(ctx context.Context, action rem
 			pm.logger.ErrorContext(ctx, "MonitorPostgres: failed to change pooler type to replica", "error", err)
 		}
 
+	case remedialActionFixPrimaryConnInfo:
```
dweitzman, line 2001: This is a postgres monitor behavior to keep trying to fix the primaryconninfo GUC if it's in a weird state, like if a previous Inform() RPC timed out or failed because of postgres crashing or something like that

If we've recorded in memory what the leader is for the highest known rule, we can always keep retrying to set the primary conn info to match until it works

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -789,6 +833,185 @@ func (pm *MultiPoolerManager) Propose(ctx context.Context, req *consensusdatapb.
...
+		return nil, mterrors.Wrap(err, "failed to check recovery status")
+	}
+
+	// A standby with rewindPending=true was emergency-demoted earlier and still
+	// has divergent WAL relative to the new primary. Routing through
+	// demoteStalePrimaryLocked runs pg_rewind, which clears rewindPending and
+	// makes the node recruitable again. Without this, the lightweight standby
+	// branch sets primary_conninfo but leaves the WAL divergent, and the next
+	// Recruit refuses with "rewind pending after emergency demotion".
+	needsRewind := pm.rewindPending.Load()
```
dweitzman, line 948: The `needsRewind` here is because anything that looks like an emergency / unplanned demotion might need a rewind. This wasn't perfect before and still isn't -- `rewindPending` doesn't survive a multipooler crash and restart

In the future it'd be nice to revisit this and maybe say that a `Recruit` or `Inform` that causes an unplanned demotion could record to disk that a rewind is likely necessary until the last known rule in the WAL advances beyond some point. We may need/want to record some extra information before rewinding anyway to avoid some theoretical edge cases.

go/common/consensus/revocation.go
```diff
@@ -31,23 +31,95 @@ import (
...
+// The second clause is the runaway-recruit override: if durable WAL exists
+// for a rule strictly newer than outgoing_rule, the cohort has demonstrably
+// moved past the rule the revocation was authored to transition away from,
+// so the promise is moot. See the TermRevocation proto for the full safety
+// argument.
+//
+// revocation may be nil (treated as no revocation). outgoing_rule may be nil
+// on revocations written by older code that predates the field; nil means
+// "no override available" and the override branch cannot fire.
+func IsRuleRevoked(rule *clustermetadatapb.ShardRule, revocation *clustermetadatapb.TermRevocation) bool {
```
dweitzman, line 106: This is a new helper to so `Inform()` can't avoid rejoining replication if the pooler in question has been recruited for a higher term

go/common/consensus/revocation.go
```diff
@@ -31,23 +31,95 @@ import (
...
+	revokedBelow := revocation.GetRevokedBelowTerm()
+	if revokedBelow == 0 {
+		return false
 	}
+	if rule.GetRuleNumber().GetCoordinatorTerm() >= revokedBelow {
+		return false
+	}
+	outgoing := revocation.GetOutgoingRule()
+	if outgoing != nil && CompareRuleNumbers(rule.GetRuleNumber(), outgoing) > 0 {
```
dweitzman, line 115: The idea is that:
- multiorch will pick a specific outgoing rule before recruiting
- a recruitment is automatically invalidated if any proposal or decision is discovered that proves the recruitment was unsuccessful, allowing partitioned poolers to rejoin replication

proto/clustermetadata.proto
```diff
@@ -379,23 +379,29 @@ message PoolerPosition {
...
-//
-// HighestKnownRule does not need to be persisted. It's a best-effort attempt to fix GUC
-// settings so that nodes can participate in replication. Nodes with out-of-date information
-// will be re-informed of the latest HighestKnownRule by any coordinator that notices the
-// pooler has incorrect replication settings.
-message HighestKnownRule {
+// Not persisted across pooler restarts — best-effort, populated by Inform and
+// Propose RPCs. Coordinators that notice a pooler has incorrect replication
+// settings will re-inform it.
+message ReplicationPrimary {
```
dweitzman, line 409: Renamed `HighestKnownRule` in the ConsensusStatus to `ReplicationPrimary`, which feels like a more direct for its purpose and also explains why it needs the contact info (host/port) for the primary

go/services/multipooler/manager/rpc_manager.go
```diff
@@ -1065,18 +1065,73 @@ func (pm *MultiPoolerManager) DemoteStalePrimary(
...
+// The helper does not touch term_revocation: revocations are authored by
+// coordinators via Recruit/AcceptRevocation, not by side effects of demotion.
+// Callers that want to record the new term (the explicit DemoteStalePrimary
+// RPC) call updateTermIfNewer themselves after the demotion succeeds; SetTermPrimary
+// deliberately does not, because an SetTermPrimary is a notification, not a revoke.
+//
+// Sequence: stop postgres -> pg_rewind -> fix pgbackrest paths -> restart as
+// standby -> reset sync replication -> set primary_conninfo -> report leader
+// observation -> read final LSN -> flip topology type to REPLICA.
+func (pm *MultiPoolerManager) demoteStalePrimaryLocked(
```
dweitzman, line 1174: This is factored out so it can be reused in the old and new RPCs

rafael (approved):
LGTM

PR #1002, opened 2026-05-18
feat(consensus): record outgoing rule in TermRevocation

dweitzman:
This should be a mostly pure refactoring, it's for two purposes:

- Future recruit-time staleness rejection. A pooler should not accept a Recruit() and revoke replication if the request is from a coordinator operating on an out-of-date view of the shard state. I've added a TODO about this in revocation.go since we may need to have a clearer distinction between proposals vs decisions before enforcing that decisions override impossible proposals.
- Future SetTermLeader RPC. A higher rule version should override a revocation with a lower expected starting point.

This is all based around the idea that a coordinator doesn't begin recruitment until it has a clear sense of the outgoing rule and never changes its mind midway through a rule change.

### DIFF
--- a/proto/clustermetadata.proto
+++ b/proto/clustermetadata.proto
@@ -379,23 +379,29 @@ message PoolerPosition {
   string lsn = 2;
 }
 
-// HighestKnownRule is the most recent ShardRule this pooler is aware of,
-// which could be beyond the end of this pooler's WAL / PoolerPosition.
+// ReplicationPrimary advertises the primary this pooler believes it should be
+// pointed at for replication, along with the rule under which that primary
+// holds leadership. The primary contact info is the main payload — it's how a
+// pooler in a degraded state (partition, mid-bootstrap, Inform-while-postgres-
+// was-down) figures out who to reconnect replication to. The rule is supporting
+// evidence and has secondary uses such as coordinators learning of rules newer
+// than what's in any replica's WAL.
 //
-// If a pooler is disconnected from replication due to network partitions
-// or stale coordinators, the HighestKnownRule helps them fix GUC and reconnect
-// to learn the latest shard state.
-//
-// If a coordinator is still establishing a new term's first rule, the HighestKnownRule
-// may represent a rule that the coordinator is _about_ to write that doesn't exist in
-// any WAL entries yet.
-//
-// HighestKnownRule does not need to be persisted. It's a best-effort attempt to fix GUC
-// settings so that nodes can participate in replication. Nodes with out-of-date information
-// will be re-informed of the latest HighestKnownRule by any coordinator that notices the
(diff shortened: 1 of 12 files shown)


proto/clustermetadata.proto
```diff
@@ -379,23 +379,29 @@ message PoolerPosition {
...
-//
-// HighestKnownRule does not need to be persisted. It's a best-effort attempt to fix GUC
-// settings so that nodes can participate in replication. Nodes with out-of-date information
-// will be re-informed of the latest HighestKnownRule by any coordinator that notices the
-// pooler has incorrect replication settings.
-message HighestKnownRule {
+// Not persisted across pooler restarts — best-effort, populated by Inform and
+// Propose RPCs. Coordinators that notice a pooler has incorrect replication
+// settings will re-inform it.
+message ReplicationPrimary {
```
dweitzman, line 393: Renamed HighestKnownRule -> ReplicationPrimary and made it include the host/port. This isn't strictly necessary in this PR, but it feels like the right thing to do for making subsequent PRs simpler

proto/clustermetadata.proto
```diff
@@ -431,6 +437,13 @@ message TermRevocation {
   // recruiting. All poolers that accept the same recruitment store the same value.
   // TODO: populate once BeginTermRequest carries this timestamp.
   google.protobuf.Timestamp coordinator_initiated_at = 3;
+
+  // The rule the coordinator observed across the cohort at recruit time —
+  // the "from" side of the transition this recruit was authoring. Used as
+  // an override key during Inform: if a subsequent Inform carries a rule
+  // strictly greater than this, the cluster has demonstrably moved past
+  // the runaway recruit's premise and the revocation is moot.
+  RuleNumber outgoing_rule = 4;
```
dweitzman, line 446: This is the change to TermRevocation

proto/clustermetadata.proto
```diff
@@ -440,27 +453,25 @@ message TermRevocation {
...
+//     time of revocation are captured at or below that position.
 //
 //  2. The term in term_revocation is globally unique; no other coordinator has
 //     used or will use it.
 //
 // The coordinator names itself as the initiator in term_revocation even though
 // it is acting on an external request.
 message ExternallyCertifiedRevocation {
-  // The rule number being superseded. Identifies which outgoing cohort is frozen.
-  RuleNumber outgoing_rule_number = 1;
```
dweitzman, line 453: ExternallyCertifiedRevocation doesn't need an outgoing rule here anymore since the TermRevocation already carries it

dweitzman:
Broke this off of https://github.com/multigres/multigres/pull/1000 for easier review

rafael (approved):
LGTM

PR #1004, opened 2026-05-19
feat(multigateway): pin backend for DECLARE … WITH HOLD cursors

haritabh17:
## Summary

- Detect `DECLARE … WITH HOLD` in the gateway planner and pin the cursor name on the reserved backend via the existing `ReasonPortal` reservation bit so the cursor survives COMMIT.
- Unpin on `CLOSE <name>` / `CLOSE ALL`; clear every HOLD pin on ROLLBACK to match PostgreSQL (which closes all open cursors at rollback).
- Extends `ReservationOptions` with `pin_portal_names` / `release_portal_names` to carry the per-statement pin/release set to the multipooler.

## Problem

Through multigateway, a `DECLARE … WITH HOLD` cursor was lost on COMMIT because the multipooler released the reserved backend to the pool when the transaction ended, and the next `FETCH` landed on a different backend with `ERROR: cursor "x" does not exist`. WITH HOLD explicitly promotes the cursor to session-level state that must outlive the transaction. Fixes MUL-389.

## Solution

`HoldCursorRoute` (planner primitive) tags `DECLARE … WITH HOLD` for portal pinning. ScatterConn OR's `ReasonPortal` into the reservation and forwards the cursor name in `pin_portal_names`. The multipooler calls `ReserveForPortal(name)` so the bit survives `COMMIT` (which only clears `ReasonTransaction`). `CloseCursorRoute` does the inverse on `CLOSE` / `CLOSE ALL`, and `ConcludeTransaction(ROLLBACK)` drops every portal pin since PostgreSQL closes all open cursors at rollback.

## Test plan

- [x] `go build ./...` clean
- [x] Unit tests pass: planner, engine, handler, scatterconn, multipooler executor (mock extended for pin/release/Release).
- [x] New e2e in `multigateway_pg_test.go` exercises the MUL-389 repro: DECLARE WITH HOLD → FETCH 1 → COMMIT → FETCH (continues at row 2) → CLOSE → backend reclaimed.
- [x] Existing skipped `cursor via DECLARE and FETCH` unskipped and passing.
- [x] Upstream pgregress `portals` test: regression.diffs collapsed from ~700 lines to 34, with zero remaining `WITH HOLD` / `cursor "x" does not exist` lines (residual diffs are unrelated pre-existing items: pg_cursors extra row and a Pico-GUC `EXPLAIN` plan diff).


mkindahl (approved):
You seem to handle this case correctly, but it is not clear if it is tested:

1. Create a cursor WITH HOLD inside a transaction
2. Start executing something causing an error. This will close the cursor(s), but the connection should stay pinned to deal with any subsequent statements.
3. Rollback (or commit) the transaction.
4. The connection should now be unpinned.

haritabh17:
> You seem to handle this case correctly, but it is not clear if it is tested:
> 
> 1. Create a cursor WITH HOLD inside a transaction
> 2. Start executing something causing an error. This will close the cursor(s), but the connection should stay pinned to deal with any subsequent statements.
> 3. Rollback (or commit) the transaction.
> 4. The connection should now be unpinned.

Ack, added test with explicit rollback, previously `WITH_HOLD_released_on_COMMIT_of_failed_transaction` did commit, now `WITH_HOLD_released_on_explicit_ROLLBACK_after_failure` does rollback, rest all same.

PR #1005, opened 2026-05-19
feat(security): support POSTGRES_PASSWORD_FILE

mkindahl:
Operators can now deliver the postgres superuser password via a mounted  Kubernetes Secret instead of the plaintext POSTGRES_PASSWORD env var. Using  POSTGRES_PASSWORD can potentially leak, for example when using commands like  "kubectl get pod -o yaml and /proc/PID/environ". The env path stays as a transitional fallback for the published images and the endtoend harness, but the local provisioner now uses the file-based path natively so it dogfoods what production should be doing.

Resolution order in pgctld and the multipooler admin pool:
  1. --pg-password-file / --connpool-admin-password-file flag
  2. POSTGRES_PASSWORD_FILE env (PGDG / docker-library convention)
  3. POSTGRES_PASSWORD env (fallback)

The new pgsecret helper centralizes the file-reading + trim logic so every consumer applies the same trim semantics. pgctld hands the operator's file directly to initdb --pwfile when available, falling back to staging a temp file only for the env-var path. The demo StatefulSet mounts the Secret at /var/run/secrets/multigres/ (tmpfs, FHS-appropriate for ephemeral runtime data) and per-pod passwords are supported via $(POD_NAME) substitution.

Local provisioner: writes <poolerDir>/postgres-password from the YAML pg-password value and passes POSTGRES_PASSWORD_FILE to both pgctld and multipooler instead of POSTGRES_PASSWORD. The file's path is threaded through PgctldProvisionResult so both children agree on the source. The write is idempotent so config rotation propagates on re-provision.

Fixes MUL-462

### DIFF
--- a/demo/k8s/k8s-multipooler-statefulset.yaml
+++ b/demo/k8s/k8s-multipooler-statefulset.yaml
@@ -155,6 +155,9 @@ spec:
             - name: sampling-config
               mountPath: /etc/otel
               readOnly: true
+            - name: postgres-password
+              mountPath: /var/run/secrets/multigres
+              readOnly: true
           env:
             - name: POD_NAME
               valueFrom:
@@ -170,8 +173,11 @@ spec:
                   fieldPath: metadata.labels['cell']
             - name: OTEL_SERVICE_NAME
               value: "multipooler"
-            - name: POSTGRES_PASSWORD
-              value: postgres
+            # Read the postgres password from the file mounted from the
+            # multigres-postgres-password Secret. Swap to
+            # /var/run/secrets/multigres/$(POD_NAME) for per-pod passwords.
+            - name: POSTGRES_PASSWORD_FILE
+              value: /var/run/secrets/multigres/password
             - name: PGDATA
               value: /data/pg_data
           envFrom:
@@ -210,6 +216,9 @@ spec:
             - name: pg-hba-template
               mountPath: /etc/pgctld
               readOnly: true
+            - name: postgres-password
+              mountPath: /var/run/secrets/multigres
+              readOnly: true
           env:
             - name: POD_NAME
               valueFrom:
@@ -217,8 +226,11 @@ spec:
                   fieldPath: metadata.name
             - name: PGDATA
               value: /data/pg_data
-            - name: POSTGRES_PASSWORD
-              value: postgres
+            # initdb / start / stop all read the postgres password from this
+            # file. Swap to /var/run/secrets/multigres/$(POD_NAME) for per-pod
+            # passwords (POD_NAME is already in scope above).
+            - name: POSTGRES_PASSWORD_FILE
+              value: /var/run/secrets/multigres/password
 
       volumes:
         # We have to give different mount points.
@@ -246,3 +258,12 @@ spec:
         - name: sampling-config
           configMap:
             name: sampling-config
+        # The postgres password is mounted from a Secret. The mode is set to
+        # 0444 to ensure the file is readable by the non-root user in the
+        # container. This is just for the demo. In production, you may want to
+        # use a more secure mode and ensure the Secret is only readable by the
+        # appropriate users.
+        - name: postgres-password
+          secret:
+            secretName: multigres-postgres-password
+            defaultMode: 0444
--- a/go/provisioner/local/pgctld.go
+++ b/go/provisioner/local/pgctld.go
@@ -147,6 +147,27 @@ func (p *localProvisioner) provisionPgctld(ctx context.Context, dbName, tableGro
 	// Create unique pgctld service ID using multipooler's service ID
 	pgctldServiceID := "pgctld-" + serviceID
 
+	// Resolve pgctld config up-front so we can materialize the postgres
+	// password file before either branch (already-running or fresh start)
+	// returns. The file is the wire format both pgctld and the multipooler
+	// in this cell consume via POSTGRES_PASSWORD_FILE.
+	pgctldConfig, err := p.getCellServiceConfig(cell, "pgctld")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get pgctld config for cell %s: %w", cell, err)
+	}
+	poolerDir, ok := pgctldConfig["pooler_dir"].(string)
+	if !ok || poolerDir == "" {
+		return nil, errors.New("pooler_dir not found in config")
+	}
+	pgPassword, ok := pgctldConfig["password"].(string)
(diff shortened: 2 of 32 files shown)


go/services/multipooler/connpoolmanager/config.go
```diff
@@ -339,12 +352,35 @@ func (c *Config) PgUser() string {
...
 }
 
-// PgPassword returns the configured PostgreSQL superuser password.
-// Set via POSTGRES_PASSWORD environment variable.
+// PgPassword returns the resolved PostgreSQL superuser password. ResolvePgPassword
+// must be called once during startup before this is read; until then it returns
+// the env/flag value without reading the file (so call-site invariants stay
+// simple and the hot path stays lock-free).
 func (c *Config) PgPassword() string {
+	if c.resolvedPgPassword != "" {
```
Verolop: little edge case: an empty password file can still end up using the env/flag password later, which could mask a bad or mis-mounted kube Secret, e.g. the pod could keep running with an older/env-sourced credential while pgctld/postgres are using the file-source path. Could you make the chosen source sticky so file-source selection never falls through to the env fallback?

demo/k8s/k8s-multipooler-statefulset.yaml
```diff
@@ -246,3 +258,7 @@ spec:
         - name: sampling-config
           configMap:
             name: sampling-config
+        - name: postgres-password
+          secret:
+            secretName: multigres-postgres-password
+            defaultMode: 0400
```
Verolop: since the multipooler image runs as 65532:65532 and the pgctld image runs as postgres, both containers can fail to read POSTGRES_PASSWORD_FILE. Could we make this non-root-readable, maybe 0440 plus an appropriate pod fsGroup, or add a note that the operator should use a different pattern?

Verolop (commented):
looking good! This will need a companion PR in the multigres-operator to actually mount the shard password Secret and wire POSTGRES_PASSWORD_FILE into the pgctld and multipooler containers; I left a couple of comments to make the file-source behaviour fail fast on malformed or empty files, so that the follow-up doesn’t inherit silent fallback behaviour

demo/k8s/k8s-multipooler-statefulset.yaml
mkindahl: Since this is the demo, the most sensible approach is to use world-readable (otherwise we need to create a group for both users and use that) and add a note that operator (or other provisioners) need to think about this.

go/services/multipooler/connpoolmanager/config.go
mkindahl: Good point. The empty password was used to check if a password was configured, but I switched to using a password source and check that instead of the password to decide if one was configured. This means that an empty password is allowed in the password file and will not fall through to the environment variable.

I also check if the environment variable is empty. The logic is this:
1. File path: --connpool-admin-password-file flag (Changed), or     CONNPOOL_ADMIN_PASSWORD_FILE / POSTGRES_PASSWORD_FILE env (LookupEnv).
     - Path explicitly empty                  → error.
     - Path set, file content empty           → error.
     - Path set, file content non-empty       → use it, source=File.
2. Flag option: --connpool-admin-password (Changed).
     - Flag set to empty                      → error.
     - Flag set to non-empty                  → use it, source=Option.
3. Env var: CONNPOOL_ADMIN_PASSWORD or POSTGRES_PASSWORD (LookupEnv).
     - Env set to empty                       → error.
     - Env set to non-empty                   → use it, source=Env.

Verolop (approved):
thank you, LGTM!

go/provisioner/local/pgctld.go
```diff
@@ -382,3 +391,20 @@ func (p *localProvisioner) deprovisionPgctld(ctx context.Context, service *Local
...
 }
+
+// writePostgresPasswordFile materializes the postgres password into a 0600
+// file at <poolerDir>/postgres-password and returns its path. The file is the
+// wire format both pgctld and the multipooler in this cell consume via
+// POSTGRES_PASSWORD_FILE; writing it idempotently (overwriting on each
+// provision) keeps the file in sync with the YAML config when operators rotate
+// the value. Cleanup is handled by the provisioner tearing down poolerDir.
+func writePostgresPasswordFile(poolerDir, password string) (string, error) {
+	if err := os.MkdirAll(poolerDir, 0o755); err != nil {
```
cuongdo: Does the pooler directory have to be accessible to other users/groups?

go/services/multipooler/connpoolmanager/config.go
```diff
@@ -339,10 +373,107 @@ func (c *Config) PgUser() string {
...
+}
+
+// ResolvePgPassword chooses the password source and caches the result so
+// subsequent PgPassword() calls return it. Three independent inputs are
+// considered, in strict precedence order — once a higher-precedence input is
+// "explicitly set" it is authoritative and lower-precedence inputs are NOT
+// consulted, even when the higher-precedence value turns out to be empty
+// (those empty cases become errors instead of fallthroughs):
+//
+//  1. File path: --connpool-admin-password-file flag (Changed), or
```
cuongdo: what does "(changed)" mean?

go/services/multipooler/connpoolmanager/config.go
```diff
@@ -339,10 +373,107 @@ func (c *Config) PgUser() string {
...
+//     CONNPOOL_ADMIN_PASSWORD_FILE / POSTGRES_PASSWORD_FILE env (LookupEnv).
+//     - Path explicitly empty                  → error.
+//     - Path set, file content empty           → error.
+//     - Path set, file content non-empty       → use it, source=File.
+//  2. Flag option: --connpool-admin-password (Changed).
```
cuongdo: same question here

cuongdo (approved):
lgtm, with some minor comments

go/provisioner/local/pgctld.go
mkindahl: Nope, updated it. (Kept scan permissions, so `0o700`.)

go/services/multipooler/connpoolmanager/config.go
mkindahl: It's the state value from the `Flag` type.  Implementation details, so removed.

go/services/multipooler/connpoolmanager/config.go
mkindahl: Same as above. Implementation details that does not need to be there.

PR #1006, opened 2026-05-19
feat(multiadmin,multiorch): add ApplyCertifiedRuleChange API and CLI

dweitzman:
## Summary

Adds an operator-facing path for installing a shard rule via an externally
certified revocation. Used whenever the normal Recruit/Propose path can't
run on its own — either because there is no prior rule to derive a
revocation from, or because the existing cohort can't reach the
durability policy's recruitment threshold. Concrete scenarios:

- **Fresh-shard bootstrap.**
- **Unpause / scale-from-zero**, where the most recent rule is stale from
  a previous provisioning and the cohort can't form a quorum to revoke it.
- **Stuck-quorum recovery** after losing too many cohort members.

A new `multigres cluster apply-rule-change` CLI submits to a new
`multiadmin.ApplyCertifiedRuleChange` RPC, which proxies to
`multiorch.ApplyCertifiedRuleChange`.

## Architecture

- **multiorch** is a pure executor: validates the request, then runs the
  existing Recruit + Propose fan-out against the proposed cohort with the
  cert's revocation in place of the locally-derived one.
- **multiadmin** picks a multiorch (preferring the leader's cell), fills
  in any identity/timing fields the operator omitted, and optionally
  derives `outgoing_rule` + `frozen_lsn` by probing the proposed cohort
  when `unsafe_derive_cert` is set. The derive path requires the
  durability policy's recruitment threshold to be satisfied by the
  reachable subset.
- **CLI** supports both an empty-outgoing-rule cert (any reported LSN is
  acceptable; useful for fresh-shard and unpause flows) and an explicit
  cert with a real `outgoing_rule_term` + `frozen_lsn`. Prompts for
  shard-name confirmation by default; `--yes` skips.

## Notable consensus changes that came along for the ride

- `Propose` now writes the rule using `proposed_rule.coordinator_id` and
  `proposed_rule.creation_time` instead of inventing them locally from the
  revocation's coordinator + `time.Now()`. Identity and timing flow
  end-to-end from the caller.
- Proposal builders take `PoolerAddress` (not `MultiPooler`) since
  consensus only needs contact info; `poolerAddressFor` was promoted to
  shared `topoclient.PoolerAddressFor`.


rafael (approved):
LGTM - Linter is not happy though

PR #1008, opened 2026-05-20
feat(multiorch, multipooler): retire shutting-down poolers

mkindahl:
## Summary

Introduces a `PoolerLifecycle` field on the pooler topology entry that records where a pooler is in its process lifecycle, and wires it through `GracefulShutdown` and the orchestrator's pooler watcher so a gracefully-shutdown pooler is no longer dialled for ~4 hours until the bookkeeping eviction tears down the cache entry.

`PoolerLifecycle` is a deliberate replacement for an earlier (unmerged) attempt to overload `COHORT_ELIGIBILITY_SIGNAL_INELIGIBLE` for the "I'm leaving" purpose. Cohort eligibility and lifecycle are now cleanly separated: `INELIGIBLE` keeps its current meaning ("don't put me in the cohort right now"); lifecycle answers the orthogonal question ("what phase of its life is this pooler in?").

## Problem

When a multipooler exits cleanly (rolling upgrade, ConfigMap rotation, node drain, manual restart), the orchestrator currently keeps dialling the dead address for ~4 hours until `forgetLongUnseenInstances` tears down the cache entry:

| What runs                                                     | For how long                  |
| ------------------------------------------------------------- | ----------------------------- |
| `runStream` reconnect loop (30 s backoff cap)                 | ~4 h until eviction           |
| Per-attempt: gRPC dial → connection refused → backoff → retry | ~120-180 attempts over 4 h    |
| Goroutines per dead pooler                                    | 1 reconnect loop + 1 watchdog |

For a _primary_, `REQUESTING_DEMOTION` on `LeadershipStatus` already drives the failover path and the cluster recovers in seconds — but the reconnect loop runs anyway. For a _standby_, no analyzer fires at all (correctly, given today's signals), so the dead-address dialling is pure waste plus log/metric noise that drowns out real connectivity problems. The pooler _knows_ it's leaving; discarding that information at the gRPC layer and forcing the orchestrator to re-derive it is fragile.

## Solution

### Proto

A new `PoolerLifecycle` message — `{status, reason, updated}` — and a `PoolerLifecycleStatus` enum with five values (`UNKNOWN`, `STARTING`, `ACTIVE`, `STOPPING`, `SHUTDOWN`):

- `MultiPooler.lifecycle_status` (field 12) — persisted in topology so the orchestrator observes lifecycle states even on cold start. Every writer sets all three sub-fields together: `status` for the state, `reason` for an operator-readable hint, and `updated` for "how long has this pooler been here" diagnostics.

### Pooler-side lifecycle state machine

| Transition              | Trigger                                                                                            | Write                                                                                                                                                                                                                                                                                  |
| ----------------------- | -------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `(none)` → `STARTING`   | Process boot. `topoclient.NewMultiPooler` initialises `LifecycleStatus` to `STARTING` + `"process starting"` + `timestamppb.Now()`. | Persisted on the first `RegisterMultiPooler` call.                                                                                                                                                                                                                                     |
| `STARTING` → `ACTIVE`   | `pgMonitor` observes `postgresRunning` (pgctld reports `RUNNING`).                                  | `markPoolerActive` writes `{ACTIVE, "pooler active", now}` to topology, then flips the in-memory record and publishes via `topoPublisher.Notify`. Idempotent — every subsequent tick short-circuits. Monotonic per boot: postgres bouncing afterwards is communicated via `Status.PostgresReady` / `PostgresStatus`, not by flipping lifecycle back. |
| `ACTIVE` → `STOPPING`   | Top of `GracefulShutdown`, before any blocking work.                                                 | Best-effort topology write of `{STOPPING, "shutting down", now}`. Operators see the state immediately, before pgctld stop runs (~5–15 s). Survives action-lock-acquire failures by design.                                                                                                                                |
| `STOPPING` → `SHUTDOWN` | `OnClose → tr.Unregister → unregisterFunc`, after `GracefulShutdown` returns.                       | Single `UpdateMultiPoolerFields` call writes `ServingStatus=NOT_SERVING`, `Type=DRAINED`, and `LifecycleStatus={SHUTDOWN, "pooler shutdown", now}` together. The topology entry is intentionally **not** deleted — the 4 h unseen-instance bookkeeping handles eventual eviction.                                                                          |

`PoolerType=DRAINED` is preserved as-is for the other consumers that observe it (e.g. `multigres getpoolers` display). SIGKILL-style exits skip `unregisterFunc` entirely and leave the entry in its last-known state — correct, because a crash is indistinguishable from a network partition.

### Orchestrator-side reaction

A single signal: the `PoolerWatcher` observes the lifecycle transition on the durable topology entry. `handlePoolerEvent` switches on `(existing, newLifecycle, prevLifecycle)`:

| Case                                          | Action                                                                                                                                                                                                                                                                                  |
| --------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Cold-start, already `SHUTDOWN`                | Ignore. The entry is not added to the in-memory `PoolerStore`; no callback fires. Nothing live to monitor.                                                                                                                                                                              |
| Transition into `SHUTDOWN`                    | Evict the in-memory `PoolerStore` entry (analyzers stop seeing a frozen-timestamp entry) and invoke `onPoolerStopped(id)`. The durable topology record is **not** touched.                                                                                                              |
| Metadata update on a tracked pooler            | Refresh `MultiPooler` in place via `DoUpdate`, preserving the health-worker-owned fields (`IsUpToDate`, `IsLastCheckValid`, timestamps).                                                                                                                                                |
| Fresh discovery (not in store, not `SHUTDOWN`) | `Set` the new state and invoke `onNewPooler(id)`, queuing the first health check.                                                                                                                                                                                                       |

The watcher exposes a second callback, `onPoolerDeleted`, that fires on `NoNode` topology events. It is **reserved**: today the watcher logs the event and invokes the callback if a caller wired one, but it does **not** evict the cache and does **not** stop per-pooler resources. The current engine deliberately passes `nil`. The intent is that an accidental external deletion (operator typo, automation bug, manual `etcdctl del`) of a still-running pooler's entry should not tear down its health stream — that would compound the operator mistake.

Engine wiring: `healthStream.Stop` is wired as `onPoolerStopped`, `nil` as `onPoolerDeleted`. `SHUTDOWN → ACTIVE` re-entry works for free: the in-memory cache is empty after the eviction, so the next watcher event for the same pooler falls into the fresh-discovery branch and `onNewPooler` fires, restarting the health stream.

Stream closure is driven **exclusively** by the lifecycle transition into `SHUTDOWN`. `INELIGIBLE` on `AvailabilityStatus.CohortEligibilityStatus` is a preference about consensus participation (e.g. the `StopReplication` admin path) and is **not** a terminal lifecycle signal — the two are deliberately decoupled.

## Operational consequences

| Scenario                                                | Behaviour                                                                                                                                                                                                                                                                            |
| ------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| Standby graceful shutdown                                | `STOPPING` is announced first; on `OnClose` the entry transitions to `{SHUTDOWN, DRAINED, NOT_SERVING}`. The orchestrator's watcher evicts the cache and calls `healthStream.Stop`; the reconnect loop exits cleanly. No 4 h reconnect storm.                                       |
| Standby crash / SIGKILL                                  | `unregisterFunc` does not run; no `SHUTDOWN` written. The entry stays in its last-known state and the reconnect loop runs as today. Correct: a crash is indistinguishable from a network partition, and retrying is the right default.                                              |
| Primary graceful shutdown                                | `SHUTDOWN` published alongside the existing `REQUESTING_DEMOTION` path. Failover is unchanged (`LeaderResignedAnalyzer` drives it). The reconnect loop on the standby orchestrators exits the same way as for any other pooler.                                                       |
| Pooler restart within 4 h of `SHUTDOWN`                  | The new process boots with `STARTING` and `RegisterMultiPooler(allowUpdate=true)` overwrites the topology entry. The watcher sees a non-`SHUTDOWN` entry with no cache record and treats it as a fresh discovery → `onNewPooler` fires → `Start` resumes dialling.                  |
| Orchestrator restart while a pooler is `SHUTDOWN`        | The cold-start branch in `handlePoolerEvent` ignores the already-`SHUTDOWN` entry: it is not added to the cache and no callback fires. The orchestrator does not dial it.                                                                                                            |
| Topology write of `SHUTDOWN` fails                        | `unregisterFunc` returns the error and the entry stays in its previous state (`STOPPING` or earlier). The orchestrator's reconnect loop will continue until the 4 h bookkeeping eviction fires. Same recovery as a SIGKILL crash.                                                     |
| `StopReplication` admin path on a live standby            | Already publishes `INELIGIBLE` today via `walReceiverManuallyStopped`. Unchanged — the stream stays open, the orchestrator just removes the node from the cohort.                                                                                                                    |
| External `etcdctl del` of a live pooler's entry          | `NoNode` event fires. The watcher logs at WARN ("cache and stream left intact"); the cache and health stream are unaffected. Recovery story (a pooler self-watch that re-creates its own entry) is captured in `FOLLOWUPS.md`.                                                       |

## Test plan

- [x] **Pooler unit:**
  - `TestMarkPoolerActive_TransitionsLifecycle` — `STARTING → ACTIVE` writes in-memory and topology; the factory default of `STARTING` + reason/timestamp is pinned here.
  - `TestMarkPoolerActive_Idempotent` — second call on an already-`ACTIVE` pooler does not retrigger a topology write.
  - `TestGracefulShutdown_AnnouncesStopping` — in-memory `STOPPING` transition fires.
  - `TestGracefulShutdown_AnnouncesStoppingInTopology` — `STOPPING` persists to a real `memorytopo` backend with reason `"shutting down"` and a recent timestamp.
- [x] **Orchestrator unit:**
  - `TestPoolerWatcher_PoolerEntersShutdownLifecycle` — `ACTIVE → SHUTDOWN` transition evicts the cache, fires `onPoolerStopped` once, and does not fire `onPoolerDeleted`.
  - `TestPoolerWatcher_RestartAfterShutdownFiresOnNewPooler` — `SHUTDOWN → ACTIVE` re-fires `onNewPooler` so the health stream restarts.
  - `TestPoolerWatcher_ColdStartShutdownIgnored` — pre-existing `SHUTDOWN` entry on watcher boot is not cached and triggers no callbacks.
  - `TestPoolerWatcher_PoolerDeletedFromTopology` — `NoNode` fires `onPoolerDeleted` only; the cache is **not** evicted and `onPoolerStopped` does **not** fire.
  - `TestHealthStream_StreamEOFWithoutSpecialSignal_KeepsBackoff` — bare EOFs still enter the reconnect-with-backoff loop; the only thing that tears down the stream is the watcher's `onPoolerStopped`.
- [x] **Integration:**
  - `TestStandbyGracefulShutdownLifecycleShutdown` — SIGTERM a standby and assert the topology entry's `LifecycleStatus.Status == LIFECYCLE_SHUTDOWN` and `Type == DRAINED` within 30 s, with the entry still present (no `NoNode`) and no spurious failover. Passes locally in ~27 s.
- [x] **Regression:** all existing graceful-shutdown integration tests (`TestPrimaryGracefulShutdownTriggersFailover`, `TestStandbyGracefulShutdownDoesNotTriggerFailover`, `TestMultiReplicaContinuityAfterStandbyShutdown`, `TestSequentialGracefulShutdowns`) continue to pass.

## Out of scope

- **Removing `PoolerType=DRAINED`.** The `unregisterFunc` continues to write `Type=DRAINED` alongside `LifecycleStatus=LIFECYCLE_SHUTDOWN`. `DRAINED` is used for other purposes (e.g. `multigres getpoolers` display); whether to retire it later is a separate decision.
- **Aggressive cache eviction on `SHUTDOWN` in topology.** The `unregisterFunc` deliberately does **not** delete the topology record. The 4 h `forgetLongUnseenInstances` threshold handles eventual cleanup; the entry survives for that window so operators and post-mortem tooling can inspect the durable end-state. The in-memory `PoolerStore` is evicted immediately by the watcher.
- **Wiring `onPoolerDeleted` to behaviour.** The callback exists and is invoked on `NoNode`, but the engine passes `nil`. Hooking it up to anything (re-create the entry, page on-call, telemetry) is a follow-up.
- **Pooler self-watch for accidental deletions.** A pooler whose topology entry is externally removed does not currently notice or recover. Captured in `FOLLOWUPS.md`.
- **Lifecycle-driven cohort changes.** Cohort membership remains governed by `CohortEligibilityStatus`. A pooler going `SHUTDOWN` does not itself trigger cohort removal in this PR.
- **Wiring an `UNUSABLE` state.** The enum has `UNKNOWN`, `STARTING`, `ACTIVE`, `STOPPING`, `SHUTDOWN`. Adding an explicit non-recoverable state can be done later if needed.

Closing MUL-478

### DIFF
--- a/go/services/multiorch/recovery/health_stream.go
+++ b/go/services/multiorch/recovery/health_stream.go
@@ -148,6 +148,12 @@ func (hs *HealthStream) Shutdown() {
 // If a stream is already running for this pooler the call is a no-op.
 // The pooler's MultiPooler metadata is read from the store on each
 // reconnect attempt so topology updates are automatically picked up.
+//
+// Dialing is unconditional: poolers whose lifecycle reads STOPPING in
+// topology are dialled the same as any other. If the pooler subsequently
+// goes away (its OnClose unregisterFunc deletes the topology entry),
+// the pooler watcher's onDeletedPooler callback invokes Stop, which
+// cancels the per-pooler context and unwinds the reconnect loop.
 func (hs *HealthStream) Start(id *clustermetadatapb.ID) {
 	poolerID := topoclient.MultiPoolerIDString(id)
 	hs.mu.Lock()
--- a/go/services/multiorch/recovery/pooler_watcher.go
+++ b/go/services/multiorch/recovery/pooler_watcher.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"log/slog"
+	"path"
 	"strings"
 	"sync"
 	"time"
@@ -44,11 +45,13 @@ import (
 // are reported via the onNewPooler callback so the caller can start monitoring
 // them (e.g. open a ManagerHealthStream stream).
 type PoolerWatcher struct {
-	topoStore   topoclient.Store
-	targets     func() []config.WatchTarget // live accessor, same as Engine.shardWatchTargets
-	store       *store.PoolerStore
-	onNewPooler func(id *clustermetadatapb.ID) // called when a new pooler is first discovered
-	logger      *slog.Logger
+	topoStore       topoclient.Store
+	targets         func() []config.WatchTarget // live accessor, same as Engine.shardWatchTargets
+	store           *store.PoolerStore
+	onNewPooler     func(id *clustermetadatapb.ID) // called when a new pooler is first discovered
+	onPoolerStopped func(id *clustermetadatapb.ID) // called on lifecycle transition into LIFECYCLE_SHUTDOWN
+	onPoolerDeleted func(id *clustermetadatapb.ID) // called when a tracked pooler's topology entry is removed (NoNode); reserved for future use
+	logger          *slog.Logger
 
 	// Control
 	ctx    context.Context
@@ -62,26 +65,35 @@ type PoolerWatcher struct {
 
 // NewPoolerWatcher creates a new PoolerWatcher.
 // targets is a function that returns the current WatchTargets (consulted on every event).
-// onNewPooler is called when a new pooler is discovered; the pooler is already present
-// in the store when the callback fires.
+// onNewPooler is called when a new pooler is discovered; the pooler is already
+// present in the store when the callback fires.
+// onPoolerStopped is called when a tracked pooler's lifecycle transitions to
+// LIFECYCLE_SHUTDOWN; the store entry has already been evicted when it fires.
+// onPoolerDeleted is called when a tracked pooler's topology entry is removed
+// (NoNode). Reserved for future use — pass nil to leave NoNode events as a
+// log-only no-op that does not evict the cache or stop per-pooler resources.
 func NewPoolerWatcher(
 	ctx context.Context,
 	topoStore topoclient.Store,
 	targets func() []config.WatchTarget,
 	poolerStore *store.PoolerStore,
 	onNewPooler func(id *clustermetadatapb.ID),
+	onPoolerStopped func(id *clustermetadatapb.ID),
+	onPoolerDeleted func(id *clustermetadatapb.ID),
 	logger *slog.Logger,
 ) *PoolerWatcher {
 	watchCtx, cancel := context.WithCancel(ctx)
 	return &PoolerWatcher{
-		topoStore:    topoStore,
-		targets:      targets,
-		store:        poolerStore,
-		onNewPooler:  onNewPooler,
-		logger:       logger,
-		ctx:          watchCtx,
-		cancel:       cancel,
-		cellWatchers: make(map[string]*cellPoolerWatcher),
+		topoStore:       topoStore,
+		targets:         targets,
+		store:           poolerStore,
+		onNewPooler:     onNewPooler,
+		onPoolerStopped: onPoolerStopped,
+		onPoolerDeleted: onPoolerDeleted,
+		logger:          logger,
+		ctx:             watchCtx,
+		cancel:          cancel,
+		cellWatchers:    make(map[string]*cellPoolerWatcher),
 	}
 }
 
@@ -227,7 +239,7 @@ func (pw *PoolerWatcher) processCellEvent(event *topoclient.WatchDataRecursive) 
 
 // startCellWatcher starts a per-cell pooler watcher. Caller must hold pw.mu.
 func (pw *PoolerWatcher) startCellWatcher(cell string) {
-	w := newCellPoolerWatcher(pw.ctx, pw.topoStore, cell, pw.targets, pw.store, pw.onNewPooler, pw.logger)
+	w := newCellPoolerWatcher(pw.ctx, pw.topoStore, cell, pw.targets, pw.store, pw.onNewPooler, pw.onPoolerStopped, pw.onPoolerDeleted, pw.logger)
 	pw.cellWatchers[cell] = w
 	w.start()
 }
@@ -249,12 +261,14 @@ func extractCellNameFromPath(watchPath string) string {
 // ---------------------------------------------------------------------------
 
 type cellPoolerWatcher struct {
-	topoStore   topoclient.Store
-	cell        string
-	targets     func() []config.WatchTarget
-	store       *store.PoolerStore
-	onNewPooler func(id *clustermetadatapb.ID)
-	logger      *slog.Logger
+	topoStore       topoclient.Store
+	cell            string
+	targets         func() []config.WatchTarget
+	store           *store.PoolerStore
+	onNewPooler     func(id *clustermetadatapb.ID)
+	onPoolerStopped func(id *clustermetadatapb.ID)
+	onPoolerDeleted func(id *clustermetadatapb.ID)
+	logger          *slog.Logger
 
 	ctx      context.Context
 	cancel   context.CancelFunc
@@ -269,19 +283,23 @@ func newCellPoolerWatcher(
 	targets func() []config.WatchTarget,
 	poolerStore *store.PoolerStore,
 	onNewPooler func(id *clustermetadatapb.ID),
+	onPoolerStopped func(id *clustermetadatapb.ID),
+	onPoolerDeleted func(id *clustermetadatapb.ID),
 	logger *slog.Logger,
 ) *cellPoolerWatcher {
 	watchCtx, cancel := context.WithCancel(ctx)
 	return &cellPoolerWatcher{
-		topoStore:   topoStore,
-		cell:        cell,
-		targets:     targets,
-		store:       poolerStore,
-		onNewPooler: onNewPooler,
-		logger:      logger.With("cell", cell),
-		ctx:         watchCtx,
-		cancel:      cancel,
-		syncChan:    make(chan chan struct{}, 1),
+		topoStore:       topoStore,
+		cell:            cell,
+		targets:         targets,
+		store:           poolerStore,
+		onNewPooler:     onNewPooler,
+		onPoolerStopped: onPoolerStopped,
+		onPoolerDeleted: onPoolerDeleted,
+		logger:          logger.With("cell", cell),
+		ctx:             watchCtx,
+		cancel:          cancel,
+		syncChan:        make(chan chan struct{}, 1),
 	}
 }
 
@@ -368,13 +386,28 @@ func (cw *cellPoolerWatcher) sync(ctx context.Context) error {
 // handlePoolerEvent processes a single watch event for a pooler file.
 func (cw *cellPoolerWatcher) handlePoolerEvent(wd *topoclient.WatchDataRecursive) {
 	if wd.Err != nil {
-		// Deletions (NoNode) are intentionally ignored here: the pooler store
-		// entry is left in place so that ongoing health checks can continue until
-		// bookkeeping removes it after the configured unseen threshold.
-		if !errors.Is(wd.Err, &topoclient.TopoError{Code: topoclient.NoNode}) {
+		// NoNode = the pooler's topology entry was removed outright. On the
+		// happy-path graceful-shutdown flow the pooler does NOT delete its
+		// entry; it writes LifecycleStatus=LIFECYCLE_SHUTDOWN and lets
+		// bookkeeping eviction clean up. So a NoNode here is unexpected —
+		// an operator typo, an automation bug, or a manual etcd delete.
+		// We log the event and invoke onPoolerDeleted if a caller wired
+		// it, but we deliberately do NOT evict the cache or stop per-pooler
+		// resources: the running pooler is unaffected, and tearing down
+		// the health stream over an accidental deletion would compound
+		// the operator mistake.
+		if errors.Is(wd.Err, &topoclient.TopoError{Code: topoclient.NoNode}) {
+			key := path.Base(path.Dir(wd.Path))
+			if entry, ok := cw.store.Get(key); ok {
+				cw.logger.Warn("pooler topology entry removed; cache and stream left intact",
+					"pooler_id", key)
+				if cw.onPoolerDeleted != nil {
+					cw.onPoolerDeleted(entry.MultiPooler.Id)
+				}
+			}
+		} else {
 			cw.logger.Warn("watch error on pooler path", "error", wd.Err, "path", wd.Path)
 		}
-		return
 	}
 
 	// Only handle files named "Pooler"
@@ -400,21 +433,66 @@ func (cw *cellPoolerWatcher) handlePoolerEvent(wd *topoclient.WatchDataRecursive
 	}
 
 	poolerID := topoclient.MultiPoolerIDString(pooler.Id)
+	newLifecycle := pooler.GetLifecycleStatus().GetStatus()
 
-	// Use DoUpdate to atomically update only the topology metadata, preserving all
-	// health-check fields (timestamps, IsUpToDate, IsLastCheckValid, etc.) that may
-	// be written concurrently by the health check worker. A closure variable tracks
-	// whether the key existed so we know whether to treat this as a new pooler.
-	existing := false
+	// Atomic read-modify-write: capture whether the entry already existed and
+	// what its previous lifecycle was, then refresh MultiPooler in place.
+	// Health-check fields written concurrently by the health worker are
+	// preserved.
+	var (
+		existing      bool
+		prevLifecycle clustermetadatapb.PoolerLifecycleStatus
+	)
 	cw.store.DoUpdate(poolerID, func(state *multiorchdatapb.PoolerHealthState) *multiorchdatapb.PoolerHealthState {
 		existing = true
+		prevLifecycle = state.MultiPooler.GetLifecycleStatus().GetStatus()
 		state.MultiPooler = pooler
 		return state
 	})
 
-	if existing {
+	switch {
+	case !existing && newLifecycle == clustermetadatapb.PoolerLifecycleStatus_LIFECYCLE_SHUTDOWN:
+		// Cold-start discovery of an already-SHUTDOWN pooler. Cache the entry
+		// so that if the pooler restarts later (within the 4 h bookkeeping
+		// window) we'll detect the SHUTDOWN→non-SHUTDOWN transition and
+		// re-fire onNewPooler. Don't open a health stream now — there's
+		// nothing live to monitor.
+		cw.store.Set(poolerID, &multiorchdatapb.PoolerHealthState{
+			MultiPooler: pooler,
+			IsUpToDate:  false,
+		})
+		cw.logger.Debug("cached already-SHUTDOWN pooler without opening stream", "pooler_id", poolerID)
+
+	case existing && newLifecycle == clustermetadatapb.PoolerLifecycleStatus_LIFECYCLE_SHUTDOWN && prevLifecycle != clustermetadatapb.PoolerLifecycleStatus_LIFECYCLE_SHUTDOWN:
+		// Transition into SHUTDOWN: notify the caller so per-pooler resources
+		// (health stream) can be torn down. We deliberately do NOT evict the
+		// in-memory PoolerStore entry — analyzers still need to see the entry
+		// to drive failover (e.g. LeaderIsDeadAnalyzer needs the cached
+		// PoolerHealthState to fire after stream EOF), and the topology
+		// record itself stays in place until the 4 h bookkeeping eviction
+		// window. The cached MultiPooler proto carries
+		// LifecycleStatus=LIFECYCLE_SHUTDOWN so analyzers and operator
+		// tooling can distinguish "just stopped" from "stalled".
+		if cw.onPoolerStopped != nil {
+			cw.onPoolerStopped(pooler.Id)
+		}
+		cw.logger.Info("pooler entered SHUTDOWN lifecycle", "pooler_id", poolerID)
+
+	case existing && prevLifecycle == clustermetadatapb.PoolerLifecycleStatus_LIFECYCLE_SHUTDOWN && newLifecycle != clustermetadatapb.PoolerLifecycleStatus_LIFECYCLE_SHUTDOWN:
+		// Restart from SHUTDOWN: re-fire onNewPooler so per-pooler resources
+		// (health stream) get re-established. The cache entry was retained
+		// on the original SHUTDOWN transition so analyzers had consistent
+		// state; here we re-trigger the discovery callback.
+		cw.onNewPooler(pooler.Id)
+		cw.logger.Info("pooler restarted from SHUTDOWN lifecycle",
+			"pooler_id", poolerID,
+			"new_lifecycle", newLifecycle.String(),
+		)
+
+	case existing:
 		cw.logger.Debug("pooler metadata updated from topology", "pooler_id", poolerID)
-	} else {
+
+	default:
 		// New pooler — add to store and queue for immediate health check.
 		cw.store.Set(poolerID, &multiorchdatapb.PoolerHealthState{
 			MultiPooler: pooler,
--- a/go/services/multipooler/manager/graceful_shutdown.go
+++ b/go/services/multipooler/manager/graceful_shutdown.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"google.golang.org/protobuf/types/known/durationpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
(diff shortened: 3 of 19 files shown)


go/services/multipooler/manager/graceful_shutdown.go
```diff
@@ -65,6 +65,20 @@ var pgctldStopModes = []struct {
...
+	// so the snapshot reaches the orchestrator before the stream closes in
+	// shutdownCancel. This lets the orchestrator's runStream short-circuit its
+	// reconnect-backoff loop on the EOF that follows, instead of dialing the
+	// dead address for the 4h bookkeeping eviction window.
+	//
+	// This is uniform for primaries and standbys: for a primary the later
+	// REQUESTING_DEMOTION publish drives the failover path independently, and a
+	// current leader is structurally excluded from the cohort-mismatch removal
+	// branch.
+	pm.setCohortEligibility(clustermetadatapb.CohortEligibilitySignal_COHORT_ELIGIBILITY_SIGNAL_INELIGIBLE)
```
dweitzman: Publishing this here seems like it ought to cause an immediate failover without waiting for query serving to drain

(I don't think we've implemented cohort resignation implying leader resignation yes, but it feels like we should)

I would propose that we replace the `setResignedLeaderAtTerm()` call below with `setCohortEligibility(INELIGIBLE)` and say that leader resignation is implied by cohort ineligiblility.

go/services/multiorch/recovery/health_stream.go
```diff
@@ -227,6 +228,20 @@ func (hs *HealthStream) runStream(ctx context.Context, poolerID string, entry *s
...
+
+		// If the pooler's last snapshot declared it INELIGIBLE before the
+		// stream closed, it left on purpose (graceful shutdown calls
+		// setCohortEligibility(INELIGIBLE) above the action lock so the
+		// snapshot leaves before the stream EOFs). Exit the reconnect
+		// loop instead of dialing the dead address until the bookkeeping
+		// eviction window (4h) tears us down. The cached snapshot is
+		// preserved by markDisconnected; the pooler watcher will start a
+		// fresh stream if the pooler reappears.
+		if current, ok := hs.store.Get(poolerID); ok && types.PoolerIsCohortIneligible(current.AvailabilityStatus) {
```
dweitzman: A non-cohort member can still be a participant in replication, so `PoolerIsCohortIneligible()` doesn't feel like the right signal that a pooler is deprovisioned.

I'm imagining we'll want a new signal for deprovisioned poolers that lives in etcd, where the provisioner can also write to it and where it can serve as a tombstone even after the pooler is fully stopped and deprovisioned.

go/services/multiorch/recovery/health_stream.go
```diff
@@ -148,6 +148,12 @@ func (hs *HealthStream) Shutdown() {
 // If a stream is already running for this pooler the call is a no-op.
 // The pooler's MultiPooler metadata is read from the store on each
 // reconnect attempt so topology updates are automatically picked up.
+//
+// Cold-start guard: poolers whose topology entry already records
+// POOLER_LIFECYCLE_SHUTDOWN are not dialled. The pooler is on its way out
+// (or already gone); reconnecting would burn cycles against a dead
+// address. The pooler watcher will call Start again with a fresh
+// MultiPooler if the lifecycle is later set back to ACTIVE.
 func (hs *HealthStream) Start(id *clustermetadatapb.ID) {
```
dweitzman, line 157: What do you think of pushing this down lower into the discovery / etcd monitoring instead of the health stream? We could treat "SHUTDOWN" as a reason to delete orch's knowledge of this pooler, so the impact would be more in the spirit of "you can forget this pooler exists, which implies disconnecting" instead of "you know abou this pooler you can disconnect from it"

go/services/multiorch/recovery/pooler_watcher.go
```diff
@@ -403,17 +403,38 @@ func (cw *cellPoolerWatcher) handlePoolerEvent(wd *topoclient.WatchDataRecursive
...
-	// be written concurrently by the health check worker. A closure variable tracks
-	// whether the key existed so we know whether to treat this as a new pooler.
+	// be written concurrently by the health check worker. Closure variables track
+	// whether the key existed (so we know whether to treat this as a new pooler)
+	// and whether the lifecycle just transitioned out of SHUTDOWN (so we can
+	// restart the health stream — runStream stops dialling on SHUTDOWN and the
+	// stream goroutine exits, so subsequent re-activation needs an explicit
+	// Start).
 	existing := false
+	lifecycleLeftShutdown := false
```
dweitzman: This is the flip side of my other comment: what if we just treat "SHUTDOWN" as meaning "this pooler doesn't exist"?

Discovery's behavior would be:
- If a new pooler is found and it's not SHUTDOWN, start tracking it
- If a pooler is marked as SHUTDOWN, stop tracking it
- If a pooler is _missing_ (we know about it but the entry is deleted, not marked as shutdown) remember it and try to stay connected. This could be because of operator error (accidentally `rm -rf`ing etcd, etc.). If the etcd entry is missing and we can't connect for a long enough period of time (hours, etc.) maybe then it's safe to forget about it

go/services/multipooler/manager/graceful_shutdown.go
```diff
@@ -107,6 +107,40 @@ func (pm *MultiPoolerManager) GracefulShutdown(ctx context.Context) {
...
+	// a prediction. The trade-off is that an action-lock-acquire failure
+	// or a stuck stopPostgresLocked leaves lifecycle unchanged; that's
+	// fine — those are abnormal exits and the topology Type=DRAINED set
+	// by the post-exit OnClose hook still acts as the floor signal.
+	if pm.topoClient != nil && pm.serviceID != nil {
+		updateFn := func(mp *clustermetadatapb.MultiPooler) error {
+			mp.Lifecycle = clustermetadatapb.PoolerLifecycle_POOLER_LIFECYCLE_SHUTDOWN
+			return nil
+		}
+		_, err := pm.topoClient.UpdateMultiPoolerFields(lockCtx, pm.serviceID, updateFn)
```
dweitzman: This is maybe not super intuitive, but the way to update the pooler's record is probably actually to modify `pm.multipooler` and then call `pm.topoPublisher.Notify(ctx, pm.multipooler)`

I feel like this is error-prone, so it doesn't have to be in this PR but I would propose for the future that we make a change like, for example, having `topoPublisher` own the etcd record, any locks necessary to modify it, and all publishing of the record to etcd. Expecting people to know they need to modify `multipooler` and then publish it feels error-prone

go/services/multipooler/manager/manager.go
```diff
@@ -164,6 +164,15 @@ type MultiPoolerManager struct {
...
 	cohortEligibility clustermetadatapb.CohortEligibilitySignal
 
+	// lifecycle is this pooler's self-reported lifecycle stage. Starts at
+	// STARTING; transitioned to ACTIVE by markPoolerActive once the pgMonitor
+	// observes postgres responding (pgctld reports RUNNING), and to SHUTDOWN
+	// by GracefulShutdown. Published in every health snapshot via
+	// Status.PoolerLifecycle and mirrored to topology (MultiPooler.Lifecycle)
+	// so the orchestrator can observe lifecycle states even on cold start.
+	// Protected by mu.
+	lifecycle clustermetadatapb.PoolerLifecycle
```
dweitzman: I think this shouldn't be necessary if `pm.multipooler.Lifecycle` is modified directly (while holding appropriate locks)

proto/multipoolermanagerdata.proto
```diff
@@ -235,6 +235,11 @@ message Status {
   // Use postgres_running to check only if the process exists.
   bool postgres_ready = 14;
 
+  // Lifecycle stage of this pooler. Mirrors MultiPooler.lifecycle in
+  // topology so the orchestrator receives transitions in band; the
+  // orchestrator uses POOLER_LIFECYCLE_SHUTDOWN to terminate its
+  // reconnect loop cleanly on the EOF that follows.
+  clustermetadata.PoolerLifecycle pooler_lifecycle = 15;
```
dweitzman: I don't have strong feelings here since proto3 messages are always changeable, but I wonder if it might be useful in the longer term to make this a message that contains an enum instead of just an enum. For example, maybe we want to record both that the pooler is stopped and a reason like that it was stopped due to a SIGTERM, etc.

There was also the possibility of recording at what time the lifecycle stage last changed

dweitzman (commented):
The overall idea makes sense to me. Comments are about some of the details

go/services/multipooler/manager/graceful_shutdown.go
mkindahl: This was removed, so not using that signal any more.

go/services/multiorch/recovery/health_stream.go
mkindahl: Same as above.

proto/multipoolermanagerdata.proto
mkindahl: Adding a reason is actually a good idea, and it makes sense to not use a "bare" enum here.

go/services/multipooler/manager/graceful_shutdown.go
dweitzman: FYI, just sent out https://github.com/multigres/multigres/pull/1020 to try to make the management of the pooler record less error prone

go/services/multiorch/recovery/health_stream.go
mkindahl: Moving the status into the topology and deleting the entry on a graceful shutdown as well as moving the health stream "management" into the pooler watcher as we discussed has the following consequences:

- The graceful shutdown will mark the entry with SHUTDOWN and DRAINED before terminating.
- The orchestrator will cancel the health stream when the lifecycle transitions to SHUTDOWN. This is done inside the pooler watcher.
- If the entry is accidentally deleted, nothing will happen. The health stream will still be connected.
- I added a `STOPPING` and used it as an announcement that the pooler is stopping so that orchestrator (and users) can see that the pooler is in the process of stopping.
- An orchestrator that starts and finds a SHUTDOWN pooler will just ignore it.
- If the pooler crashes or is killed during shutdown, it will leave the life-cycle in the current state.
- I added a timestamp to the life-cycle message tracking the last update. This allow us to clean up very old entries regardless of what state they are in.

go/services/multiorch/recovery/pooler_watcher.go
mkindahl: > This is the flip side of my other comment: what if we just treat "SHUTDOWN" as meaning "this pooler doesn't exist"?
> 
> Discovery's behavior would be:
> 
> * If a new pooler is found and it's not SHUTDOWN, start tracking it

Added. This also means that if it switches from SHUTDOWN to STARTED (for example), a health stream will be set up (again).

> * If a pooler is marked as SHUTDOWN, stop tracking it

This is done by checking the transition. It is also the case that if the orchestrator starts and finds a SHUTDOWN pooler, it is ignored.

> * If a pooler is _missing_ (we know about it but the entry is deleted, not marked as shutdown) remember it and try to stay connected. This could be because of operator error (accidentally `rm -rf`ing etcd, etc.). If the etcd entry is missing and we can't connect for a long enough period of time (hours, etc.) maybe then it's safe to forget about it

This is just left as it is. A warning is logged if a record is deleted, but no other action is taken.

go/services/multipooler/manager/manager.go
mkindahl: It is necessary to avoid writing to the topology store every 5 seconds, but I write the `pm.multipooler.Lifecycle` directly here and just call `Notify` if it changed.

proto/multipoolermanagerdata.proto
mkindahl: Added a field to record the last updated timestamp as well. Good for debugging.

go/services/multipooler/manager/topo_publisher.go
```diff
@@ -67,7 +67,26 @@ func (tp *topoPublisher) Notify(ctx context.Context, mp *clustermetadatapb.Multi
...
+// immediate publish, without requiring the action lock to be held. Lifecycle
+// transitions (STARTING → ACTIVE via markPoolerActive, etc.) are orthogonal
+// to the consensus state machine the action lock guards, and are driven from
+// callers (e.g. the pgMonitor tick) that have no reason to take that lock.
+//
+// Concurrency: scheduleUpdate's tp.mu.Lock plus the caller's pm.mu (held while
+// mutating pm.multipooler.LifecycleStatus prior to calling here) is sufficient
+// to serialise lifecycle writers among themselves.
+func (tp *topoPublisher) NotifyLifecycle(mp *clustermetadatapb.MultiPooler) {
+	tp.scheduleUpdate(mp)
```
dweitzman: I'm a little nervous about making updates here without holding the action lock

Once https://github.com/multigres/multigres/pull/1020 it'll maybe be less of a concern since there won't be any race conditions, but prior to that the action lock protects two goroutines from clobbering each other's writes if they try to notify simultaneously

go/services/multipooler/manager/graceful_shutdown.go
```diff
@@ -65,6 +66,35 @@ var pgctldStopModes = []struct {
...
+	}
+	pm.mu.Lock()
+	pm.multipooler.LifecycleStatus = stopping
+	pm.mu.Unlock()
+	if pm.topoClient != nil && pm.serviceID != nil {
+		updateFn := func(mp *clustermetadatapb.MultiPooler) error {
+			mp.LifecycleStatus = stopping
+			return nil
+		}
+		if _, err := pm.topoClient.UpdateMultiPoolerFields(ctx, pm.serviceID, updateFn); err != nil {
```
dweitzman: I'm a little nervous about battling other code for the pooler record here since something like `SetState` below might override these changes, but I think https://github.com/multigres/multigres/pull/1020 would eliminate those concerns

go/test/endtoend/multiorch/graceful_shutdown_test.go
```diff
@@ -349,6 +349,95 @@ func TestMultiReplicaContinuityAfterStandbyShutdown(t *testing.T) {
...
+
+	t.Logf("Terminating standby %s", terminatedStandby)
+	terminateMultipoolerGracefully(t, setup.Multipoolers[terminatedStandby], 90*time.Second)
+
+	// The unregisterFunc runs after GracefulShutdown returns. Within a few
+	// seconds the topology entry must read LIFECYCLE_SHUTDOWN + DRAINED;
+	// 30 s is a generous bound that catches regressions in the unregister
+	// path while still leaving headroom for slower-responsive GitHub Actions
+	// runners.
+	require.Eventually(t, func() bool {
```
dweitzman, line 422: Not a blocker, just a thought: from an e2e test performance perspective it might not be unreasonable to put this assertion into an existing test instead of making a new one since bootstrapping for tests takes a long time

dweitzman:
The overall shape seems reasonable to me. I'm still a little nervous about all the different ways of modifying the pooler record, so I would propose merging https://github.com/multigres/multigres/pull/1020 first and then modifying the pooler record will be (famous last words) safe and easy.

I'm also not sure about the need to modify the pooler record without holding the action lock, so maybe we can discuss more what that specific scenario is and whether there would be any harm in holding the action lock.

go/services/multipooler/manager/topo_publisher.go
mkindahl: The main reason for this is that the action lock is taken for many "heavy" operations but the lifecycle is updated independently of anything else.

My worry was that if we gate this on the action lock, it could mean that the lifecycle update is blocked on a more long-running operation and hence the state is not updated until a lot later, perhaps after a failure c.f. a result of the lifecycle state. This would make debugging quite difficult, hence it is better to promptly update the lifecycle state.

With #1020 this will be less of a problem, but we can deal with the PRs in either order.

go/test/endtoend/multiorch/graceful_shutdown_test.go
mkindahl: No objections, but I think it might be better to optimize the test performance when we note that they are a problem rather than speculatively.

mkindahl:
> The overall shape seems reasonable to me. I'm still a little nervous about all the different ways of modifying the pooler record, so I would propose merging #1020 first and then modifying the pooler record will be (famous last words) safe and easy.

It can be done in either order, so no opinion there, especially since I just approved #1020 😉 

> I'm also not sure about the need to modify the pooler record without holding the action lock, so maybe we can discuss more what that specific scenario is and whether there would be any harm in holding the action lock.

I have a suspicion that the action lock is too coarse and we will have problem because we are not updating the "status" promptly, e.g., the system is shutting down but we do not update the state because we are holding the action lock because we are taking a backup.

I reverted the `NotifyLifecycle()` so that we can discuss this separately.

dweitzman:
> I have a suspicion that the action lock is too coarse and we will have problem because we are not updating the "status" promptly, e.g., the system is shutting down but we do not update the state because we are holding the action lock because we are taking a backup.
> 
> I reverted the `NotifyLifecycle()` so that we can discuss this separately.

Two possibilities that might work:
- shutdown preempts something like backup, causing it to fail quickly and release the action lock 
- the pooler record "unregister/mark as stopped" function is a special case that doesn't require the action lock like normal mutation. Unregister would get the last word on what the published record should be, and any future update attempts by an unfinished action would effectively be a no-op

mkindahl:
> > I have a suspicion that the action lock is too coarse and we will have problem because we are not updating the "status" promptly, e.g., the system is shutting down but we do not update the state because we are holding the action lock because we are taking a backup.
> > I reverted the `NotifyLifecycle()` so that we can discuss this separately.
> 
> Two possibilities that might work:
> 
> * shutdown preempts something like backup, causing it to fail quickly and release the action lock

There is already an issue for this that I will pick up once my queue is smaller. It would be particular actions that can pre-empt the action lock, such as pooler shutdown.

> * the pooler record "unregister/mark as stopped" function is a special case that doesn't require the action lock like normal mutation. Unregister would get the last word on what the published record should be, and any future update attempts by an unfinished action would effectively be a no-op

I think it might make sense to exclude particular fields if we know that there is just a single updater, e.g., such as the lifecycle field, which is what I excluded here.

go/services/multipooler/manager/manager.go
```diff
@@ -440,6 +441,17 @@ func (pm *MultiPoolerManager) openLocked(ctx context.Context, targetServingStatu
...
+			// Transition lifecycle STARTING → ACTIVE once postgres is up
+			// and responding. markPoolerActive is idempotent (short-circuits
+			// when already ACTIVE) so calling it every tick is cheap, and
+			// it retries the topology write on the next tick if it fails.
+			// The transition is monotonic per boot: once ACTIVE, postgres
+			// going down doesn't flip the lifecycle back to STARTING —
+			// runtime health is communicated via Status.PostgresReady /
+			// PostgresStatus, not via Lifecycle.
+			if newState.postgresRunning {
+				pm.markPoolerActive(ctx)
```
dweitzman, line 453: Seems like it should be in the pooler manager's Start() method rather than openLocked() as a direct poolerRecord.Mutate() call -- the action lock is already held there

go/services/multipooler/manager/rpc_consensus.go
```diff
@@ -184,6 +185,52 @@ func (pm *MultiPoolerManager) setCohortEligibility(signal clustermetadatapb.Coho
...
+// LifecycleStatus.Updated from "first time ACTIVE" to "last seen ACTIVE".
+// Keep the guard.
+//
+// pm.multipooler is the single in-memory source of truth: the guard reads it,
+// the mutation writes it, and topoPublisher.Notify hands the same pointer to
+// the publisher goroutine which clones-then-writes. The action lock is held
+// across the mutate-then-Notify sequence so lifecycle writes serialise
+// against the consensus state machine (Promote/Demote/BeginTerm) the same
+// way every other Notify caller does.
+func (pm *MultiPoolerManager) markPoolerActive(ctx context.Context) {
```
dweitzman, line 203: I don't think this method will be needed if MultiPoolerManager.Start() directly updates the lifecycle after Open()

dweitzman (approved):

go/services/multipooler/manager/manager.go
mkindahl: This code deliberately waits for the Postgres server to be ready before switching to ACTIVE, which I think is important from a user perspective to not signal a server recovering as active when it does not accept connections and is still in recovery. Especially since recovery can take some time.

PR #1009, opened 2026-05-20
fix(parser): quote identifiers in deparser + property-test coverage

thebengeu:
## Summary

- Fixes a deparser round-trip bug surfaced by a GoTrue SAML migration through multigateway: constraint names like `"resource_id not empty"` were emitted unquoted, causing Postgres to fail with `syntax error at or near "not"`. Multi-statement Simple Query batches go AST → `SqlString()` → multipooler, so any missed `QuoteIdentifier` in the deparser becomes a wire-protocol failure.
- Audits and fixes ~80 missed `QuoteIdentifier` sites across the AST deparser: constraints, tablespaces, roles, servers, FDW, publications/subscriptions, event triggers, triggers, policies, languages, access methods, cursors, portals, prepared statements, savepoints, NOTIFY/LISTEN channels, function parameters, windows, named args, stats expressions, `CURRENT OF cursor`, `IntoClause.AccessMethod`, `ColumnDef.StorageName`, `CreateStmt.TableSpaceName`, `ReplicaIdentityStmt.Name`, `AlterTableCmd.Name`, `MergeWhenClause` columns, and the `TRUNCATE`/`LOCK`/`SET CONSTRAINTS`/`DROP ROLE`/`REINDEX` container walks that were bypassing the `RangeVar`/`RoleSpec` `SqlString` helpers.
- Adds two layers of test coverage:
  - **`identifier_quoting_roundtrip_test.go`** — curated regression-anchor tests (constraint names with spaces, tablespace/role/index/access-method quoting, the original SAML migration excerpt as a multi-statement batch).
  - **`identifier_quoting_fuzz_test.go`** — property-based check that walks every statement in `testdata/` (full PostgreSQL regression corpus included unless `-short`), reflection-mutates each identifier-typed string field to a value requiring quoting, deparses, and re-parses. Any parse failure is a missed `QuoteIdentifier` site. The `nonIdentifierFields` map is the durable audit artifact — adding a new identifier-typed string field to the AST forces explicit classification.
- Updates 7 PostgreSQL regression expected outputs that previously baked in deparser bugs: schema qualifiers, `ONLY`, and quoted identifiers are now preserved end-to-end instead of being silently dropped.

## Test plan

- [x] `go test ./go/common/parser/...` — all green, including the new fuzz and round-trip tests
- [x] `go test ./go/common/parser/ -run TestIdentifierQuotingFuzz` over the full PostgreSQL regression corpus — passes
- [x] `go test ./go/services/multigateway/...` — handler, executor, planner, etc. all green
- [x] Manual round-trip of the GoTrue `add_saml` migration excerpt — comment-string semicolon survives, every CREATE TABLE deparse re-parses cleanly


thebengeu:
@claude review once

go/common/parser/ast/ddl_additional_statements.go
haritabh17: While at it, lets also update this

haritabh17: Lets quote this.

go/common/parser/ast/parse_infrastructure.go
```diff
@@ -1185,7 +1185,7 @@ func (c *ColumnDef) SqlString() string {
 
 	// Add storage clause if specified
 	if c.StorageName != "" {
-		parts = append(parts, "STORAGE", c.StorageName)
+		parts = append(parts, "STORAGE", QuoteIdentifier(c.StorageName))
```
haritabh17: Dont think this is needed

haritabh17 (approved):
🚀

PR #1011, opened 2026-05-20
feat(pgctld): add --pg-initdb-sql-dir flag for directory-based SQL init

mkindahl:
Adds --pg-initdb-sql-dirs (repeatable StringSlice, env var POSTGRES_INITDB_SQL_DIRS) accepting role:path entries. pgctld reads all .sql files from the directory in lexicographic order (os.ReadDir sort), skips non-.sql files, and runs them in a single psql session wrapped in SET SESSION AUTHORIZATION "<role>" / RESET SESSION AUTHORIZATION.

Execution order within postInitdbSetup:
  1. Create target database (if non-default)
  2. --pg-initdb-sql-dirs: bulk schema/migration directories under role
  3. --pg-initdb-sql-files: targeted overrides applied on top

Implementation details:
- parseSQLDirEntry splits role:path on the first colon; rejects empty role
- sqlFilesInDir returns sorted .sql paths; os.ReadDir order is guaranteed
- quoteIdentifier wraps identifiers in double quotes (ReplaceAll approach, used for both SET SESSION AUTHORIZATION and CREATE DATABASE)
- postInitdbSetup condition extended to include InitdbSQLDirs

Tests:
- flags_test.go: TestPgInitdbSQLDirsFlag (repeated flag, comma-separated, env var, flag overrides env var)
- start_test.go: TestRunInitdbSQLDirs (happy path, empty dir, invalid entry, unreadable dir, psql failure) and TestInitdbSQLExecutionOrder (dirs run before files)
- shardsetup_test.go: TestShardSetup_InitdbSQLDirsExecuted (E2E)

Closes MUL-484

### DIFF
--- a/go/cmd/pgctld/command/flags_test.go
+++ b/go/cmd/pgctld/command/flags_test.go
@@ -147,6 +147,43 @@ func TestPgInitdbExtraConfFlag(t *testing.T) {
 	})
 }
 
+func TestPgInitdbSQLDirsFlag(t *testing.T) {
+	t.Run("defaults to empty slice", func(t *testing.T) {
+		_, pc := GetRootCommand()
+		assert.Empty(t, pc.pgInitdbSQLDirs.Get())
+	})
+
+	t.Run("accepts repeated flag", func(t *testing.T) {
+		root, pc := GetRootCommand()
+		require.NoError(t, root.ParseFlags([]string{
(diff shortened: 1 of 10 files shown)


rafael (approved):
Makes sense. Straightforward change.

PR #1012, opened 2026-05-20
fix(multipooler): WAL receiver disconnect handling, backup-restore term, skip_outgoing_quorum proto

dweitzman:
Four small, independent changes split out of the rule store PR for easier review.

### 1. Better diagnostics on WAL receiver disconnect timeout
`waitForReceiverDisconnect`'s 10s budget expiring at a tick boundary was surfacing as
`connection pool context already expired` (an opaque pool error) rather than the real
"timeout waiting for WAL receiver to disconnect". Short-circuits on `waitCtx.Err()`
before issuing the next query, and folds `primary_conninfo` + `pg_stat_wal_receiver.status`
into the polling query so the timeout log shows the last-observed state — next failure
self-diagnoses whether the receiver was stuck or `primary_conninfo` was never cleared.

### 2. Accept WALRCV_WAITING + empty primary_conninfo as disconnected
Diagnostics from #1 revealed a bimodal pattern: walreceiver exits within 500ms (count=0)
or stays "waiting" for the full 10s budget. The stuck case happens when the walreceiver
is mid-`PQconnectdb` when SIGHUP arrives — the cooperative shutdown signal isn't
processed until libpq returns or `connect_timeout` expires (tens of seconds under load).
`pg_terminate_backend` can't help: it doesn't work on auxiliary processes like the
walreceiver. Treat WALRCV_WAITING with empty `primary_conninfo` as effectively-
disconnected; safe because we hold the action lock and WAITING→STREAMING requires
non-empty `primary_conninfo`.

### 3. Preserve term revocation across backup restores
The term revocation file lives outside PGDATA intentionally so restores don't affect it.
A prior change was deleting it on restore, probably left over from the previous bootstrap
design where orch needed to begin terms before `initdb` ran.

### 4. `skip_outgoing_quorum` proto field on `CoordinatorProposal`
Adds the field, validation (`validateProposal` restricts it to externally-certified
proposals), and the coordinator-side setter (`buildBootstrapProposal`). The multipooler-side
reader lands separately with the rule store work. Until then, multipoolers ignore the
flag — which matches today's behavior since `Propose()` already operates under
incoming-cohort-quorum semantics.

### DIFF
--- a/go/services/multipooler/manager/pg_replication.go
+++ b/go/services/multipooler/manager/pg_replication.go
@@ -550,32 +550,57 @@ func (pm *MultiPoolerManager) waitForReceiverDisconnect(ctx context.Context) (*m
 	ticker := time.NewTicker(500 * time.Millisecond)
 	defer ticker.Stop()
 
+	// Track the latest poll snapshot so a timeout has diagnostic detail without
+	// needing another query after the ctx has already expired.
+	var (
+		lastCount    int64 = -1 // -1 = not yet polled
+		lastStatus   string
+		lastConnInfo string
+	)
+
+	timedOut := func(cause error) (*multipoolermanagerdatapb.StandbyReplicationStatus, error) {
+		pm.logger.ErrorContext(ctx, "WAL receiver did not disconnect",
+			"cause", cause,
+			"last_receiver_count", lastCount,
+			"last_walreceiver_status", lastStatus,
+			"last_primary_conninfo", lastConnInfo)
+		if errors.Is(cause, context.DeadlineExceeded) {
+			return nil, mterrors.New(mtrpcpb.Code_DEADLINE_EXCEEDED, "timeout waiting for WAL receiver to disconnect")
+		}
+		return nil, mterrors.Wrap(cause, "context cancelled while waiting for WAL receiver to disconnect")
+	}
+
 	for {
 		select {
 		case <-waitCtx.Done():
-			if waitCtx.Err() == context.DeadlineExceeded {
-				pm.logger.ErrorContext(ctx, "Timeout waiting for WAL receiver to disconnect")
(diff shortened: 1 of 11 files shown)


go/services/multipooler/manager/pg_replication.go
```diff
@@ -550,33 +550,82 @@ func (pm *MultiPoolerManager) waitForReceiverDisconnect(ctx context.Context) (*m
...
+			//     RequestXLogStreaming, which only fires when primary_conninfo
+			//     is non-empty.
+			//   - We can't actively terminate a walreceiver from SQL —
+			//     pg_terminate_backend only works on regular backends, not
+			//     auxiliary processes like the walreceiver. The walreceiver
+			//     only exits when its in-flight libpq call returns, which is
+			//     bounded by connect_timeout (potentially tens of seconds).
+			//     Treating WAITING+empty as done lets us proceed without
+			//     waiting out that timeout.
+			done := lastCount == 0 || (lastStatus == "waiting" && lastConnInfo == "")
```
dweitzman, line 624: @mkindahl FYI, this is a change to the logic for waiting until WAL receiver is disconnected. It seems like there are cases where it gets stuck in `waiting` state until the connect timeout before disconnecting, so this is saying that if the primary conninfo is empty and we're in waiting state it's probably fine and we shouldn't wait for the full connect timeout

rafael (approved):

go/services/multipooler/manager/pg_replication.go
mkindahl: Is there a connection with #990 ?

dweitzman: Not intentionally. The rule store PR was hitting a timeout because waiting for a fully stopped wal receiver could take as long as connection timeout 

This was one way to avoid waiting too long, but killing wal receiver might be as good or better

PR #1014, opened 2026-05-20
feat(servenv): expose binary build info via gRPC, HTTP /version, and OTel

dweitzman:
Every multigres process now reports its embedded VCS identity through three surfaces, populated from runtime/debug.BuildInfo:
  - gRPC ServiceInfo.GetBuildInfo (auto-registered by servenv on every gRPC server, alongside reflection and health)
  - HTTP /version (JSON) with a "build info" link in the shared footer of every index page
  - OTel resource attribute service.version, attached automatically to every emitted metric, span, and log

The bottom of status pages:
image

The output at `/version`:
image


rafael (approved):
LGTM - but codecoverage is not happy

PR #1016, opened 2026-05-20
fix(parser+pgregress): LIKE deparse + C locale parity (+6 net)

haritabh17:
## Summary

- Fix `NOT LIKE` / `NOT ILIKE` / `... ESCAPE` deparse: `A_Expr.SqlString()` was always emitting `LIKE`/`ILIKE` regardless of `a.Name`, and dropping ESCAPE clauses on the floor. `NOT LIKE` round-tripped as `LIKE`, inverting results; `SIMILAR TO ... ESCAPE '#'` lost the escape character.
- Plumb `--pg-initdb-args` from the pgctld root flag (and `POSTGRES_INITDB_ARGS` env) through `PgCtldServiceConfig.InitdbArgs` — the `server` and `init` subcommands were silently dropping it.
- Run pg_regress against an initdb forced to C locale + UTF8 encoding, with an `lc_*='C'` snippet appended via `--pg-initdb-extra-conf` so the template's hard-coded `en_US.UTF-8` GUCs don't override initdb. Drop `select_having.patch` (it was masking en_US case-insensitive sort order; under C locale upstream's expected output already matches).

## Result

`pg_regress` against `pg17`: **147/222 → 153/222** (+6 net wins, zero regressions). Wins: `char`, `varchar`, `int8`, `numeric`, `opr_sanity`, `type_sanity`. `opr_sanity` / `type_sanity` were collateral — the broken `LIKE '_%'` deparse turned their pg_type antijoins into match-all.


GuptaManan100 (approved):
Love it!

PR #1018, opened 2026-05-21
fix(parser): deparser drops OVERRIDING, mangles SHOW ALL, doubles space

thebengeu:
InsertStmt.SqlString dropped the OVERRIDING { SYSTEM | USER } VALUE
clause, which is required to insert into GENERATED AS IDENTITY columns
and would cause the deparsed SQL to fail at runtime.

VariableShowStmt.SqlString ran the grammar's "all" sentinel through
QuoteQualifiedIdentifier, producing SHOW "all" (a lookup of a parameter
named "all", which doesn't exist) instead of the SHOW ALL keyword form.

CreateOpClassItem.SqlString unconditionally inserted a space before "("
for operator items, producing a double space in the DROP form where the
operator name is absent (e.g. DROP OPERATOR 1  (int4, int4)).


haritabh17 (approved):
LGTM 🚀

PR #1019, opened 2026-05-21
fix(copy): preserve reserved connection on PG-level COPY errors

haritabh17:
## Summary

- Stop tearing down reserved connections when PostgreSQL rejects a COPY at initiation or finalization. A PG ErrorResponse on these paths means the socket is back at `ReadyForQuery`, not broken, so the reservation must survive when other reasons (transaction, temp tables) still hold it.
- Forward the surviving `ReservedState` through the multipooler ↔ gateway bidi stream's `ERROR` phase so the gateway keeps tracking the connection instead of clearing it.
- Drain trailing `ReadyForQuery` in `ReadCopyDoneResponse` on the client side so subsequent operations on the same socket are protocol-clean.
- Guard the gateway-side deferred `CopyAbort` so it no-ops once `CopyFinalize` has already cleaned up the COPY reservation reason.

## Problem

A single bad COPY statement was poisoning the entire session. In the multipooler, both `CopyReady` and `CopyFinalize` treated _any_ error from `InitiateCopyFromStdin` / `ReadCopyDoneResponse` as a fatal connection failure and called `Release(ReleaseError)`. That tore down the reserved connection even when PostgreSQL had cleanly answered with `ErrorResponse` + `ReadyForQuery`. Any session that was holding that connection for an unrelated reason — a wrapping transaction or a temp table from earlier in the script — would then see every subsequent statement fail with:

```
reserved connection N not found for user postgres
```

Because no `CopyInResponse` was sent on the rejected COPY, psql never entered copy-in mode and shipped the script's `COPY` data lines (and `\.`) as ordinary `Query` messages, producing the cascading `parse error at position N: syntax error` and `invalid command \.` noise we were seeing across `copy`, `copy2`, `copyselect`, `copydml`, and many other pgregress suites.

## Solution

Split connection-level failures from PG-level rejections on both the initiate and finalize paths in `go/services/multipooler/executor/executor.go`. `mterrors.IsConnectionError(err)` keeps the existing teardown path for genuinely broken sockets. Otherwise the COPY reservation reason is removed and the connection is either released (no other reasons) or kept alive with the surviving state returned alongside the error. `ReadCopyDoneResponse` in `go/common/pgprotocol/client/conn.go` now drains the trailing `ReadyForQuery` on the error path (mirroring `InitiateCopyFromStdin`) so the conn is genuinely clean before that decision is made. `CopyAbort` short-circuits when the conn no longer holds a COPY reason, so the gateway-side deferred abort can't write a stray `CopyFail` onto a backend already back at RFQ.

The grpc service (`go/services/multipooler/grpcpoolerservice/service.go`) forwards the surviving `ReservedState` through the bidi stream's `ERROR` phase on init-failure. The gateway-side gRPC client (`grpc_query_service.go`) and `scatter_conn.go` propagate that state instead of nil, and `applyReservedState` either keeps tracking the live conn or clears tracking + marks the transaction failed depending on what came back. `copy_statement.go` marks `completed = true` after `CopyFinalize` returns (success or error) so the deferred `CopyAbort` cannot run on top of already-cleaned-up state.

## Test plan

- [x] `go test -count=1 -short ./go/common/pgprotocol/client/... ./go/services/multipooler/executor/... ./go/services/multipooler/grpcpoolerservice/... ./go/services/multigateway/scatterconn/... ./go/services/multigateway/poolergateway/... ./go/services/multigateway/engine/...` — all 6 packages green.
- [x] `RUN_PGREGRESS=1 go test -timeout 30m -count=1 -run TestPostgreSQLRegression ./go/test/endtoend/pgregresstest/` — 147 / 222 pass; **0 cascade markers** in `copy.diff`, `copy2.diff`, `copyselect.diff`, `copydml.diff` (down from 46 in baseline `copy2.diff` alone). 5 residual `reserved connection ... not found` markers remain only in `plancache.diff` (3) and `transactions.diff` (2) — both originate from prepared-statement/DESCRIBE paths and are unrelated to the COPY code paths this PR touches.
- [x] Manual: `CREATE TEMP TABLE x(a int, b text);` followed by `COPY x (xyz) FROM stdin;` (PG error: invalid column) then `INSERT INTO x VALUES (1, 'alice'); SELECT * FROM x;` — all succeed on the same backend connection. Pre-fix the SELECT/INSERT would have failed with `reserved connection N not found for user postgres`.


GuptaManan100 (approved):
Love it!

PR #1020, opened 2026-05-21
refactor(multipooler): centralize pooler topology record management

dweitzman:
Introduce poolerRecord — the single owner of the local MultiPooler proto and all its writes to etcd. Other multipooler code now reads or changes the record without thinking about when or how the change reaches etcd; the publisher reflects every Mutate to etcd asynchronously.

Collapses three publish paths (toporeg.Register on the live proto, topoPublisher.Notify callsites, and UpdateMultiPoolerFields on shutdown) into a single Mutate → publisher pipeline. A ruleguard rule forbids RegisterMultiPooler outside poolerRecord to keep this property over time.


go/services/multipooler/manager/pooler_record.go
```diff
@@ -0,0 +1,348 @@
...
+		return errors.New("poolerRecord.Mutate: PgDataDir is immutable")
+	}
+	if before.Hostname != after.Hostname {
+		return errors.New("poolerRecord.Mutate: Hostname is immutable")
+	}
+	if !portMapsEqual(before.PortMap, after.PortMap) {
+		return errors.New("poolerRecord.Mutate: PortMap is immutable")
+	}
+	return nil
+}
```
mkindahl: Instead of having this check, you could split the fields into two different structures: one with immutable fields and one with mutable fields. The function would then accept the mutable fields and allow them to be modified.

By doing it this way you use structural constraints to enforce the policy and prevent immutable fields from being mutated rather than having to add a check like this (which potentially can generate errors, if we add buggy code).

mkindahl (commented):
This is a significant improvement of how to handle Multipooler topology fields, but I would suggest that you split the fields into mutable and immutable fields in separate structures and only pass down the mutable fields to the callback.

This completely eliminates the risk of bugs modifying the immutable fields and generating errors, removes the need for the check, and also the need for the "transactional" update that you have.

go/services/multipooler/manager/pooler_record.go
dweitzman: Good call, pulled out a mutable struct. Feels cleaner now, and the mutate function doesn't need to have an error return value

go/services/multipooler/manager/pooler_record.go
```diff
@@ -0,0 +1,328 @@
...
+	current := r.desired.Load()
+	state := MutablePoolerRecordState{
+		Type:          current.Type,
+		ServingStatus: current.ServingStatus,
+	}
+	fn(&state)
+	next := proto.Clone(current).(*clustermetadatapb.MultiPooler)
+	next.Type = state.Type
+	next.ServingStatus = state.ServingStatus
+	r.desired.Store(next)
```
mkindahl, line 187: This works, but since you're holding the action lock, there should be no problems mutating the record in place rather than using a Read-Modify-Write pattern.

In the presence of bugs (which I find unlikely for this case, but anyway) where we assume that some actor modifies the state without holding the action lock, this code would actually overwrite the modification made in the buggy path. Not sure if that is good or bad, but it will make some changes "disappear".

mkindahl (approved):
LGTM. Not sure you need to use the Read-Modify-Write pattern inside `Modify`, but there is no harm in doing it, so approving.

go/services/multipooler/manager/pooler_record.go
dweitzman: The action lock isn't required for reads, so some kind of protection for reads against concurrent writes is probably needed

The protection could be using an atomic field and immutable copies like this, or reads and writes could alternatively be guarded by a mutex

I don't have strong feelings either way

go/services/multipooler/manager/manager.go
```diff
@@ -450,36 +464,72 @@ func (pm *MultiPoolerManager) Open() {
...
+func (pm *MultiPoolerManager) Pause(ctx context.Context) (resume func(context.Context)) {
+	preServingStatus := pm.record.ServingStatus()
+	if !pm.closeLocked(ctx, "paused") {
+		pm.logger.ErrorContext(ctx, "MultiPoolerManager: Pause() called on already-closed manager")
 	}
 
-	return func() {
-		pm.Open()
+	return func(resumeCtx context.Context) {
+		pm.openLocked(resumeCtx, preServingStatus)
```
dweitzman, line 493: Made a change here so that Pause()/resume() for the multipooler manager restores the previous serving state instead of always switching to SERVING

The previous codepaths must have had some situations where we got into SERVING state but didn't publish it. Since this PR makes the publishing more aggressive, the unintended switch from NOT_SERVING to SERVING during pause/unpause led to problems

PR #1022, opened 2026-05-22
feat(multigateway): COPY TO STDOUT, notice forwarding, un-wrapped PG errors

haritabh17:
## Summary

Three coupled fixes that unblock five failing pgregress tests (`copy`, `copyselect`, `copydml`, `insert`, `generated`) and clean up COPY error reporting end-to-end.

### 1. COPY ... TO STDOUT streaming

- `planCopyStmt` now plans `CopyStatement` for the TO STDOUT direction (previously rejected with `PgSSFeatureNotSupported`).
- New pgprotocol/client helpers: `InitiateCopyToStdout`, `ReadCopyOutMessage`, `FinishCopyToStdout` — read `CopyOutResponse` + interleaved `CopyData` / `NoticeResponse` + trailing `CommandComplete` / `ReadyForQuery`.
- New pgprotocol/server helpers: `WriteCopyOutResponse`, `WriteCopyData`, `WriteCopyDone`.
- Multipooler executor: `CopyOutReady` reserves a backend conn (reusing existing reservations / deferred BEGIN like `CopyReady` FROM STDIN); `CopyOutStream` pumps each PG frame back via callback.
- `CopyBidiExecuteRequest` grows a `Direction` field; the multipooler bidi handler routes `TO_STDOUT` requests through `copyBidiExecuteToStdout` while `FROM_STDIN` keeps the existing flow.
- Multigateway `grpcQueryService` / `PoolerGateway` / `ScatterConn` / `IExecute` all gain `CopyOutReady` + `CopyOutStream` symmetric with the FROM STDIN plumbing. `CopyStatement.streamCopyOut` writes `CopyOutResponse` → forwards each pumped chunk as `CopyData` → writes `CopyDone` → hands the final Result to the per-statement callback (which writes `CommandComplete`).

### 2. NoticeResponse forwarding during COPY

Previously dropped by `ReadCopyDoneResponse` / `ReadCopyFailResponse` (`// Ignore notices`). Now:

- Capture `NoticeResponse` diagnostics that PG emits between `CopyDone` and `ReadyForQuery` (trigger `RAISE NOTICE`, COPY progress reporting, etc.) and return them alongside the command tag.
- `InitiateCopyFromStdin` / `InitiateCopyToStdout` also capture pre-response notices.
- `executor.CopyFinalize` stores the captured slice on `sqltypes.Result.Notices`; the multipooler grpc service serializes them onto `CopyBidiExecuteResponse.notices`; the gateway decodes them back and the existing per-result Conn writer emits `NoticeResponse` frames to the client in order before `CommandComplete`.

### 3. Un-wrapped PG errors from the COPY path

`CopyBidiExecuteResponse` grows an `error_diagnostic` field. The multipooler attaches the full `PgDiagnostic` (severity/SQLSTATE/DETAIL/HINT/position) for every ERROR phase; the gateway's `copyErrorFromResp` prefers it over the legacy text-only `error` field. PG `ErrorResponse` text now round-trips verbatim.

Error wraps that previously produced

```
ERROR: failed to initiate COPY: failed to initiate COPY: COPY initiation failed:
       failed to initiate COPY FROM STDIN: ERROR: column "xyz" of relation "x" does not exist
```

are gone; the underlying `PgError` is returned as-is to the client.

## Stack

Standalone fix branched off `upstream/main`.

## Test plan

- [x] `go test ./...` (multigateway / multipooler / pgprotocol packages all green; new unit tests cover notice capture + un-wrapped error path).
- [x] Full pgregress run in verify mode against upstream/main baseline:
  - **4 fewer failures** than baseline (`copy`, `copydml`, `insert`, `generated`).
  - **No new regressions.**
  - `merge.patch` updated to drop now-stale "COPY TO STDOUT not yet supported" hunks; `merge` continues to pass.
  - Residual `copyselect` / `insert_conflict` failures left for follow-up (parser syntax-error format + `MultiAssignRef` deparser — separate issues, not in scope).

### DIFF
--- a/go/common/pgprotocol/client/conn.go
+++ b/go/common/pgprotocol/client/conn.go
@@ -28,6 +28,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/multigres/multigres/go/common/mterrors"
 	"github.com/multigres/multigres/go/common/pgprotocol/protocol"
 )
 
@@ -398,40 +399,48 @@ func (c *Conn) WriteCopyFail(errorMsg string) error {
 	return c.flush()
 }
 
-// ReadCopyDoneResponse reads the CommandComplete response after WriteCopyDone()
-// Returns the command tag (e.g., "COPY 100") and rows affected
-// Note: In simple query protocol, PostgreSQL sends CommandComplete followed by ReadyForQuery
-// We need to consume both messages to clear the buffer
-func (c *Conn) ReadCopyDoneResponse(ctx context.Context) (string, uint64, error) {
+// ReadCopyDoneResponse reads the CommandComplete response after WriteCopyDone().
+// Returns the command tag (e.g., "COPY 100"), rows affected, and any
+// NoticeResponse / InfoResponse diagnostics received between CopyDone and
+// ReadyForQuery. Notices come from BEFORE/AFTER row triggers that fire during
+// COPY FROM STDIN (e.g. RAISE NOTICE) and from COPY progress reporting; the
+// gateway forwards them to the client as NoticeResponse frames before the
+// final CommandComplete so client-side test fixtures see the expected
+// NOTICE / INFO lines.
+//
+// Note: In simple query protocol, PostgreSQL sends CommandComplete followed
+// by ReadyForQuery — both messages must be consumed to clear the buffer.
+func (c *Conn) ReadCopyDoneResponse(ctx context.Context) (string, uint64, []*mterrors.PgDiagnostic, error) {
 	c.bufMu.Lock()
 	defer c.bufMu.Unlock()
 
 	var commandTag string
 	var rowsAffected uint64
+	var notices []*mterrors.PgDiagnostic
 	gotCommandComplete := false
 
 	// Read messages until we get both CommandComplete and ReadyForQuery
 	for {
 		msgType, err := c.readMessageType()
 		if err != nil {
-			return "", 0, fmt.Errorf("failed to read message type: %w", err)
+			return "", 0, notices, fmt.Errorf("failed to read message type: %w", err)
 		}
 
 		length, err := c.readMessageLength()
 		if err != nil {
-			return "", 0, fmt.Errorf("failed to read message length: %w", err)
+			return "", 0, notices, fmt.Errorf("failed to read message length: %w", err)
 		}
 
 		body, err := c.readMessageBody(length)
 		if err != nil {
-			return "", 0, fmt.Errorf("failed to read message body: %w", err)
+			return "", 0, notices, fmt.Errorf("failed to read message body: %w", err)
 		}
 
 		switch msgType {
 		case protocol.MsgCommandComplete:
 			// Parse command tag
 			if len(body) == 0 {
-				return "", 0, errors.New("empty CommandComplete body")
+				return "", 0, notices, errors.New("empty CommandComplete body")
 			}
 			// Command tag is null-terminated string
 			tag := string(body)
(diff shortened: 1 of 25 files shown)


go/common/pgprotocol/client/conn.go
```diff
@@ -659,10 +674,226 @@ func (c *Conn) InitiateCopyFromStdin(ctx context.Context, copyQuery string) (for
...
+// InitiateCopyToStdout sends a COPY ... TO STDOUT query and reads the
+// CopyOutResponse. The caller is then expected to drive the response stream
+// with ReadCopyOutMessage until io.EOF (CopyDone), followed by
+// FinishCopyToStdout to consume the trailing CommandComplete + ReadyForQuery.
+// Returns the COPY format, per-column formats from the CopyOutResponse, and
+// any NoticeResponse diagnostics seen before the CopyOutResponse arrived.
+//
+// Modeled after InitiateCopyFromStdin — same Q-then-flush, same notice /
+// ParameterStatus / ErrorResponse handling.
+func (c *Conn) InitiateCopyToStdout(ctx context.Context, copyQuery string) (format int16, columnFormats []int16, notices []*mterrors.PgDiagnostic, err error) {
```
GuptaManan100, line 676: CopyOutResponse/CopyInResponse body parsing (`int16(body[0])` + column-format loop) is now duplicated across `InitiateCopyToStdout` / `InitiateCopyFromStdin`. A tiny shared parse helper would prevent drift.

go/services/multipooler/grpcpoolerservice/service.go
```diff
@@ -537,6 +557,129 @@ func (s *poolerService) CopyBidiExecute(stream multipoolerpb.MultiPoolerService_
...
+
+// copyBidiExecuteToStdout drives the COPY ... TO STDOUT half of the bidi
+// stream. Unlike COPY FROM STDIN, the client doesn't send DATA messages;
+// PG pushes CopyData frames (interleaved with NoticeResponse) until
+// CopyDone, then CommandComplete + ReadyForQuery. We translate that into
+// the bidi protocol as READY → DATA* → RESULT.
+//
+// A FAIL message from the gateway during the data phase still aborts the
+// COPY (e.g. client disconnect mid-stream), but we don't expect DATA from
+// the gateway — receiving one is a protocol violation.
```
GuptaManan100: This happens via context cancelation? Asking cause `copyBidiExecuteToStdout` never calls `stream.Recv()` during the data phase. So I want to understand how the FAIL works in cancelling a running stream. I would expect a context cancelation.

go/services/multigateway/poolergateway/grpc_query_service.go
```diff
@@ -731,5 +747,190 @@ func (g *grpcQueryService) ReleaseReservedConnection(
...
+				for _, pd := range resp.Notices {
+					if err := onMessage(client.CopyOutMessage{Notice: mterrors.PgDiagnosticFromProto(pd)}); err != nil {
+						g.copyStreamsMu.Lock()
+						delete(g.copyStreams, options.ReservedConnectionId)
+						g.copyStreamsMu.Unlock()
+						// Close the send half so the multipooler's
+						// stream ctx cancels and its CopyOutStream
+						// pump exits via the abortCopyOut path. Mirrors
+						// the RESULT / ERROR terminal branches below.
+						_ = stream.CloseSend()
```
GuptaManan100: Ties with ☝️, comment says `CloseSend()` makes "the multipooler's stream ctx cancel". A half-close does **not** cancel the server-side stream context; cleanup actually happens when the query ctx is cancelled as `streamCopyOut` returns the error upward. Reword so a future reader doesn't rely on CloseSend for cancellation.

GuptaManan100 (approved):
Overall LGTM! Just a few nits

go/common/pgprotocol/client/conn.go
haritabh17: Done — extracted shared COPY response parsing into a helper

go/services/multipooler/grpcpoolerservice/service.go
haritabh17: Yup, COPY TO STDOUT data phase doesn’t Recv FAIL; cleanup/abort is driven by context cancellation and send failures. Updated comment accordingly.

go/services/multigateway/poolergateway/grpc_query_service.go
haritabh17: Agree. CloseSend only half-closes client send. Server cleanup happens via returned error/context-cancel path and pooler send failures. Reworded comment.

PR #1027, opened 2026-05-24
ci(coverage): only upload coverage on successful runs

dweitzman:
Coverage was previously uploaded even when the coverage jobs failed, which could surface misleading partial numbers in Codecov and Coveralls. Test reliability has improved enough that gating uploads on success is the better default. Also drops the now-redundant `if: always()` on coveralls-finish so it no longer finalizes builds that were never uploaded.


mkindahl (approved):
LGTM

PR #1028, opened 2026-05-24
refactor(consensus): remove non-determinism from NewTermRevocation

dweitzman:
## Summary

`NewTermRevocation` no longer reads the wall clock — callers pass an
explicit `initiatedAt` timestamp.

This PR also adds ruleguard guardrails to reduce the odds of `go/common/consensus`
having non-deterministic going forward:

- Direct map range loops and stdlib `maps.Keys/Values/All` — use the new
  `go/tools/sortedmaps` helper instead
- `go` statements (goroutine scheduling is non-deterministic)
- Wall-clock reads: `time.Now/Since/Until/After/Tick/Sleep/NewTimer/NewTicker`
  and `timestamppb.Now()`


mkindahl (approved):
LGTM

PR #1030, opened 2026-05-25
test(parser): add parse→deparse→parse AST round-trip checks and fix the bugs they surface

GuptaManan100:
## Summary

- Adds `TestParseDeparseRoundtrip`, a new check that parses a query, deparses the AST back to SQL, parses that SQL again, and asserts the two ASTs are structurally identical — catching deparser bugs where the emitted SQL is valid but reconstructs a different tree.
- Fixes ~30 distinct deparser/parser bugs surfaced by the check (information loss or wrong-tree reconstruction across DDL, DML, GRANT, publications, XML, aggregates, and more).
- Re-bakes the affected `testdata/postgres/*.json` expectations to the corrected deparse output.

## How the check works

The round-trip check deparses the **pristine** parse and only canonicalizes both trees for the final comparison (deparsing a canonicalized tree produces fake mismatches). A small set of strictly **loss-free** canonicalization rules (in the test only) collapse spellings PostgreSQL treats as identical — bare boolean flags vs `= true`, `on`/`off` vs `true`/`false`, `CROSS JOIN` vs `INNER JOIN ON TRUE`, the implicit `pg_catalog` schema, the default `IN` parameter mode, order-independent `ALTER DEFAULT PRIVILEGES` options, and `LIMIT NULL`/`LIMIT ALL`. Each rule is narrow enough that a genuine dropped clause still fails.

Also added `SetLocation` to the `Node` interface so the comparison can zero source locations in a single `Rewrite` pass.

## Parser / grammar bugs fixed (`postgres.y`)

- **Bare `Integer` typmods** — `CHAR(n)`, `CHARACTER(n)`, `TIMESTAMP/TIME(n)`, and the `SUBSTRING(x FOR n)` length were built as bare `Integer` instead of `A_Const`, so they re-parsed to a different node than the canonical path produces.
- **`NodeList` tag** — eight sites built `&ast.NodeList{...}` directly, leaving `Tag` at the `T_Invalid` zero value instead of `T_List`. All now use the `ast.NewNodeList(...)` constructor. This was the root cause of ~980 raw mismatches and let the corresponding test-side normalization be deleted entirely.

## Deparser bugs fixed (`SqlString`)

- **Constraints** — dropped `INCLUDE (...)` on `PRIMARY KEY`/`UNIQUE`/`EXCLUDE`; dropped `UNIQUE NULLS NOT DISTINCT`; dropped the name on `CONSTRAINT <name> NOT NULL`; dropped `INCLUDE`/deferrability on `EXCLUDE`; standalone constraint-attribute nodes (`DEFERRABLE`, `INITIALLY DEFERRED`, …) printed nothing; `GENERATED BY DEFAULT` collapsed to `GENERATED ALWAYS`.
- **`DefElem` namespace** — `toast.<reloption>` namespace dropped, colliding with the same-named non-toast option.
- **`ALTER TABLE` identity** — `SET INCREMENT BY n` corrupted by an injected `SET GENERATED BY DEFAULT`; `DROP IDENTITY IF EXISTS` dropped `IF EXISTS`; `ALTER CONSTRAINT ... NOT DEFERRABLE INITIALLY DEFERRED` dropped the `INITIALLY DEFERRED`.
- **`%TYPE`** — `col%TYPE` references were alias-normalized and lost the `%TYPE` suffix.
- **Typed tables** — `OF type (...)` dropped its column/constraint list.
- **Multi-column `SET`** — `SET (a, b) = <source>` was split into duplicated per-column assignments in UPDATE, ON CONFLICT, **and** MERGE.
- **`DROP TRIGGER`/`RULE`/`POLICY`** — mangled a schema-qualified table (`DROP TRIGGER t ON s.tbl` → `... ON s`).
- **Set-op selects** — `FOR UPDATE`/`FOR NO KEY UPDATE` locking clause dropped.
- **`GROUPING SETS`** — single-column members wrapped in parens, re-parsing as `ParenExpr`.
- **FDW column `OPTIONS (...)`** — dropped on foreign-table columns.
- **Column `COLLATE`** — bare `COLLATE` dropped on partition/typed-table columns.
- **`btrim`** — incorrectly rewritten to `trim` (not a real rename).
- **`CREATE FUNCTION`** — C-language `AS 'objfile', 'symbol'` dropped the second element; `BEGIN ATOMIC` body injected a spurious `LANGUAGE sql`.
- **`CREATE TABLE AS`** — dropped `ON COMMIT` and `TABLESPACE`.
- **`CREATE`/`ALTER OPERATOR`** — `COMMUTATOR`/`NEGATOR` operator names quoted as string literals.
- **Ordered-set aggregate** — trailing `VARIADIC` dropped the `ORDER BY` arg, re-parsing as a plain aggregate.
- **XML** — `XMLELEMENT` dropped `XMLATTRIBUTES`; `XMLROOT` dropped `STANDALONE` and mis-rendered `VERSION NO VALUE`; `XMLTABLE` dropped `XMLNAMESPACES`.
- **RECORD-returning functions** — dropped the column-definition list (`... AS x(a int, b text)`).
- **JOIN aliases** — `(a JOIN b) AS j` and `JOIN ... USING (c) AS x` lost the alias.
- **`GRANT`/`REVOKE`** — dropped the object-type keyword (`FOREIGN DATA WRAPPER`, `DOMAIN`, `ROUTINE`, …); membership options rendered as `OPTION` instead of `TRUE`/`FALSE`.
- **`COPY ... WHERE`** — dropped the WHERE clause.
- **`ALTER SUBSCRIPTION`** — `DISABLE` deparsed as `ENABLE`; `SET PUBLICATION p WITH (...)` dropped the options.
- **Publications** — mixed `TABLE`/`TABLES IN SCHEMA` lists reordered/dropped; schema entries carrying a column list/`WHERE` dropped it; bare `FOR CURRENT_SCHEMA` dropped.

All deparser fixes re-bake their affected `testdata/postgres/*.json` expectations.


haritabh17 (approved):
🚀

PR #1031, opened 2026-05-25
fix(multipooler): retry pool ops on transient close instead of generation bump

haritabh17:
## Summary

- Stabilizes `TestWrappedPreparedStatementExecution/multigateway` (top queryserving flake — 13 failures in the past week per the flaky-test report).
- Closes a race in `withReopenRetry` that surfaced `ErrPoolClosed` to clients during a reopen.

## Problem

`reopenConnections()` does `Close → Open` under `pm.mu`, and `Manager.Open()` bumps its generation counter only after clearing the snapshot and resetting `closed`. A query whose `op()` resolves to `ErrPoolClosed` in the Close-but-not-yet-Open window observes an unchanged generation, and the retry guard treats it as a terminal shutdown and surfaces the transient error. The eviction path (`evictUserPool`) hit the same hole — it closes the pool without bumping generation at all.

Client-visible symptom against `multigateway`: `pq: query execution failed: Code: UNKNOWN — failed to get connection for user postgres: connection pool is closed`.

## Solution

Use `m.closed` as the source of truth for "is this shutdown terminal." On `ErrPoolClosed`, retry against the current snapshot unless `m.closed` is set — in which case `getOrCreateUserPool` returns `"manager is closed"` and the retry exits cleanly without re-invoking `op`. Drop the now-dead generation field.

Tests:
- Existing `TestWithReopenRetry_RetriesOnReopen` / `_OnlyRetriesOnce` no longer need a generation bump.
- `_SurfacesGenuineClose` rewritten to flip `closed` inside `op`, simulating the Close-races-op race the new guard must catch.
- New `_FailsFastOnPreClosedManager` covers the pre-closed path where `op` never fires.
- `TestWrappedPreparedStatementExecution` run under `-race -count=20`: 20/20 pass.

### DIFF
--- a/go/services/multipooler/connpoolmanager/config.go
+++ b/go/services/multipooler/connpoolmanager/config.go
@@ -606,6 +606,6 @@ func (c *Config) NewManager(logger *slog.Logger) *Manager {
 		logger:  logger,
 		metrics: metrics,
 	}
-	mgr.closed.Store(true) // Manager is closed until Open() is called
+	mgr.lifecycle.Store(uint32(managerLifecycleClosed)) // Manager is closed until Open() is called
 	return mgr
 }
--- a/go/services/multipooler/connpoolmanager/manager.go
+++ b/go/services/multipooler/connpoolmanager/manager.go
@@ -79,15 +79,10 @@ type Manager struct {
 	// This mutex is only acquired when a new user pool needs to be created.
 	createMu sync.Mutex
 
-	// closed indicates whether the manager has been closed.
+	// closed indicates whether the manager has been closed. withReopenRetry
+	// reads this on ErrPoolClosed to distinguish a transient pool swap
+	// (reopen/eviction) from a genuine terminal shutdown.
 	closed atomic.Bool
-
-	// generation is incremented on every Open(). Callers that get
-	// ErrPoolClosed from an in-flight pool operation can compare the
-	// pre-call generation against the current one: if it advanced, a
-	// reopen swapped the pools underneath them and the call is safe to
-	// retry against the fresh snapshot.
-	generation atomic.Uint64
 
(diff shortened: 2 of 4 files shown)


GuptaManan100 (commented):
I'm not sure if this PR fixes the bug it's trying to. For the reopen race, the thing `TestWrappedPreparedStatementExecution` actually exercises, the new `m.closed` guard makes the **identical** retry decision as the old `generation` guard in every sub-case. It's logically a no-op on that path, if I'm not mistaken.

### Old vs. new, every case the guard fires

| Scenario | Old (`gen != startGen`) | New (`!m.closed`) | Differ? |
|---|---|---|---|
| Reopen: Close **and** Open both done | gen advanced → **retry** | closed=false → **retry** | no |
| Reopen window: Close done, Open not yet | gen same → **surface** | closed=true → **surface** | no |
| Eviction: pool evicted, no reopen | gen same → **surface** | closed=false → **retry** | **yes** |
| Terminal shutdown | gen same → **surface** | closed=true → **surface** | no |

For the reopen race the decision is identical in both sub-cases. The PR changes nothing there. It is still a good change cause we get rid of an unnecessary field, and we do improve the eviction part. 

I think the correct way to fix this would be to make a new variable that is set when we are reopening the connections, so that we can distinguish the case where a close is going to be followed by a open soon, vs the case where the close is permanent.

haritabh17:
> I think the correct way to fix this would be to make a new variable

Makes sense, updated by adding explicit reopen state variable plus reopen window markers . So now it distinguishes:
  - close + reopen soon (reopening) -> wait, retry
  - permanent close (closed) -> surface error

GuptaManan100:
Closing in favour of https://github.com/multigres/multigres/pull/1098

PR #1032, opened 2026-05-26
feat(multigateway): support bound params in set_config()

haritabh17:
## Summary

- Storage migrations against multigres hit `set_config value argument must be a literal, not a bound parameter` for `SELECT set_config('search_path', $1, false)` over the extended query protocol. Substitute the portal's bind values into a cloned AST at portal-execute time so the planner sees literals and mints the SessionSettings tracker entry it needs.
- Substituted plans are single-use (never cached). `Route` reissues the original portal + binds to PG, so the gateway tracker and PG decode the same bytes — no desync, no plan-cache poisoning.
- Position rules and the security blocklist are unchanged: set_config in WHERE/subquery/CTE/DEFAULT is still rejected, and `dblink`/`pg_read_file`/etc. with bound params are still rejected at the walker.

Fixes [MUL-531](https://linear.app/supabase/issue/MUL-531).

## Test plan

- [x] Unit tests in `planner/set_config_bind_test.go` — 11 cases (value/name/is_local slots, text+binary bool, NULL bind rejected, unsupported OID rejected, unspecified OID accepted, passthrough, schema-qualified, original AST unmodified, out-of-range param)
- [x] Integration test `TestExtendedQueryProtocol_SetConfigBoundParam` — 3 subtests × 2 targets (direct PG + multigateway): text value (Storage repro), unspecified param type, repeated bind with different values + SHOW mirror
- [x] Existing `TestMultiGateway_UnsafeFuncCallRejection`, `TestMultiGateway_SetConfigRoutedAsSET`, `TestMultiGateway_SessionSettings`, `TestExtendedQueryProtocol_*`, `TestPreparedStatement*` — no regression
- [x] Unit + race detector across `go/services/multigateway/...`, `go/common/parser/...`, `go/common/preparedstatement`
- [x] Manual psql against local cluster: Slack repro, Storage value, bound name/is_local/all-three slots, `pg_catalog.set_config(...)` qualified, repeated bind diff values, session isolation across clients, plus all rejection paths (WHERE, nested, dblink+bind, pg_read_file+bind, out-of-range param)

### DIFF
--- a/go/services/multigateway/planner/unsafe_funccall.go
+++ b/go/services/multigateway/planner/unsafe_funccall.go
@@ -75,13 +75,13 @@ var funcBlocklist = map[string]string{
 //
 // Two shapes:
 //   - all-literal: Name and Value carry the parsed strings, *Bind fields
-//     are nil, Bound is false. is_local was literal false (literal true
-//     short-circuits below and never produces a setConfigCall).
+//     are nil. is_local was literal false (literal true short-circuits
+//     below and never produces a setConfigCall).
 //   - any-bound: at least one of NameBind/ValueBind/IsLocalBind is non-nil
 //     and points at the parser-produced ParamRef for that slot. Name/Value
-//     hold the parsed string for any slot that was NOT bound. Bound is
-//     true — the planner emits the deferred-resolution primitive that
-//     decodes the bound slots from the portal at execute time.
+//     hold the parsed string for any slot that was NOT bound. The planner
+//     emits the deferred-resolution primitive that decodes the bound slots
+//     from the portal at execute time.
 //
 // `is_local=true` literals never produce a setConfigCall — validation
 // returns (nil, nil) early so the call goes to PG via Route only, with no
@@ -95,11 +95,10 @@ type setConfigCall struct {
 	NameBind    *ast.ParamRef
 	ValueBind   *ast.ParamRef
 	IsLocalBind *ast.ParamRef
+}
 
-	// Bound is true when at least one of *Bind is non-nil. Callers branch
-	// on this to choose ApplySessionStateSilent (literal) vs
-	// ApplySessionStateFromBind (deferred resolution).
-	Bound bool
+func (sc setConfigCall) hasBoundParams() bool {
+	return sc.NameBind != nil || sc.ValueBind != nil || sc.IsLocalBind != nil
 }
 
 // expressionCheckResult carries the output of inspectExpressionFuncCalls.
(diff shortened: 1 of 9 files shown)


go/test/endtoend/queryserving/extended_query_protocol_test.go
```diff
@@ -644,3 +644,135 @@ func TestExtendedProtocol_TransactionIsolation(t *testing.T) {
 		})
 	}
 }
+
+// TestExtendedQueryProtocol_SetConfigBoundParam pins the Supabase Storage
+// migration shape: SELECT set_config('search_path', $1, false) over the
```
GuptaManan100: Let's not say Supabase storage shap, its just an additional query pattern that we want to support.

go/test/endtoend/queryserving/extended_query_protocol_test.go
```diff
@@ -644,3 +644,135 @@ func TestExtendedProtocol_TransactionIsolation(t *testing.T) {
...
+// extended protocol. Without the bind-materialization path the multigateway
+// planner rejects the bound parameter with
+// "set_config value argument must be a literal, not a bound parameter"
+// because the AST validator runs at Plan() time, when the AST still carries
+// a ParamRef. With the fix, MaterializeBoundSetConfigs substitutes the
+// portal's bind value before planning, and the gateway's session tracker is
```
GuptaManan100: I think `MaterializeBoundSetConfigs` is a stale reference. Let's just remove this and not talk about a fix. The test stands alone

go/services/multigateway/planner/unsafe_funccall.go
```diff
@@ -69,15 +69,37 @@ var funcBlocklist = map[string]string{
...
 	Value string
+
+	NameBind    *ast.ParamRef
+	ValueBind   *ast.ParamRef
+	IsLocalBind *ast.ParamRef
+
+	// Bound is true when at least one of *Bind is non-nil. Callers branch
+	// on this to choose ApplySessionStateSilent (literal) vs
+	// ApplySessionStateFromBind (deferred resolution).
+	Bound bool
```
GuptaManan100: Do we need this field? Why don't we make it a function that just checks if any of them are non-nil?

GuptaManan100 (approved):
Amazing! This is awesome 🚀

go/test/endtoend/queryserving/extended_query_protocol_test.go
haritabh17: Ack, updated

go/test/endtoend/queryserving/extended_query_protocol_test.go
haritabh17: Okay

go/services/multigateway/planner/unsafe_funccall.go
haritabh17: Good point, updating

PR #1033, opened 2026-05-26
feat(release): add dry-run workflow and enhance release artifacts

Verolop:
## Description

this PR expands the Multigres release setup so GitHub Releases produce conventional downloadable artifacts, while keeping the real release flow separate from manual release validation.

The real release workflow remains tag-driven and is responsible for publishing versioned images and creating the GitHub Release. The new dry-run workflow is manual-only and runs GoReleaser in snapshot mode, which lets maintainers inspect generated archives, checksums, and metadata without creating a tag, publishing images, or creating a GitHub Release. Later on, this will evolve into a nightly build. 

This follows the standard shape used by multi-component Go/database projects: release one distribution archive per platform containing the project binaries, with checksums, alongside container images for Kubernetes/runtime consumption.

## Main Changes

- Adds GoReleaser builds for the Multigres release binaries and packages them into platform distribution archives.
- Adds a manual GoReleaser dry-run workflow that uploads generated `dist/` output for inspection.
- Keeps the production release workflow focused on real tag-based releases only.
- Tracks the Pico CSS file in `go/common/web/templates/css/`, which is the directory embedded into the Go binaries and served by the built-in web handlers. Previously the file lived under `external/pico/` and the Makefile copied it into the embedded template tree before builds, which meant release tooling also had to know about that hidden preparation step. Keeping the asset in the embedded tree preserves the same served `/css/pico.classless.jade.min.css` path while making normal Go builds, Makefile builds, Docker builds, and GoReleaser builds use the same source layout.

## Tests

Vaalidated that the edited workflow and GoReleaser YAML parse correctly, and `goreleaser check` passes with the updated artifact configuration. I also verified a direct Go build of `go/cmd/multigres` after removing the CSS copy step, which confirms the embedded web asset is now available from the tracked source tree rather than relying on Makefile preparation.

A full GoReleaser snapshot was run before the final archive rename and produced the expected platform archives, each containing the six release binaries plus checksums.


rafael (approved):
LGTM -

Good idea cleaning up the css part of it

PR #1035, opened 2026-05-26
fix(multiorch,multipooler): allow re-promoting previously demoted leader

dweitzman:
A 2-pooler AT_LEAST_2 cohort wedges after a Recruit fan-out emergency-demotes the primary: runFailover excluded the resigning leader from the cohort, leaving 1 of 2 members recruitable.

Three small changes:

- coordinator.runFailover stops excluding leaders that signaled REQUESTING_DEMOTION. The signal means "please replace me if you can," not "remove me from the cohort," so a resigning leader stays a recruitment candidate and may be re-promoted at the new term. To avoid wasting Recruit RPCs on members we believe are dead, cohort membership now requires LastSeen freshness within 10s.

- Recruit no longer hard-rejects when rewindPending=true. The proper rewind happens at SetTermPrimary before the node serves writes or replicates from a new primary; keeping the gate at Recruit was unnecessarily strict and could block the only path back to quorum.

- promoteStandbyToPrimary clears rewindPending. The consensus protocol picked this node as the new leader at a higher term, so its WAL is by definition the rule going forward; the postgres monitor can resume.

TestSpuriousFailoverRecovery reproduces the wedge. TestDeadPrimaryRecovery iteration 4 is updated to wait on cluster recovery rather than assume the old primary is excluded; a new WaitForHigherTermPrimary helper waits for any pooler to advance the term with enough connected followers.

### DIFF
--- a/go/services/multiorch/consensus/coordinator.go
+++ b/go/services/multiorch/consensus/coordinator.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"log/slog"
 	"sync"
+	"time"
 
 	"google.golang.org/protobuf/types/known/timestamppb"
 
@@ -29,7 +30,6 @@ import (
 	consensusdatapb "github.com/multigres/multigres/go/pb/consensusdata"
 	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
 	multiorchdatapb "github.com/multigres/multigres/go/pb/multiorchdata"
-	"github.com/multigres/multigres/go/services/multiorch/recovery/types"
 )
 
 // Coordinator orchestrates consensus-based leader election for shards.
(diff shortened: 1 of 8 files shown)


rafael (approved):
LGTM

go/services/multiorch/consensus/coordinator.go
```diff
@@ -75,27 +75,43 @@ func (c *Coordinator) AppointLeader(ctx context.Context, shardKey *clustermetada
...
 	liveCohort := make([]*multiorchdatapb.PoolerHealthState, 0, len(cohort))
 	for _, p := range cohort {
-		if types.LeaderNeedsReplacement(p) {
-			c.logger.InfoContext(ctx, "Excluding resigned pooler from failover cohort",
-				"pooler", p.GetMultiPooler().GetId().GetName())
+		last := p.GetLastSeen()
+		if last == nil || time.Since(last.AsTime()) > failoverCohortReachableThreshold {
+			c.logger.InfoContext(ctx, "Excluding unreachable pooler from failover cohort",
+				"pooler", p.GetMultiPooler().GetId().GetName(),
+				"last_seen", last.AsTime())
```
Verolop, line 107: Can you add a nil-safe log value here, or log the never-seen case separately? During clean cluster startup, a pooler can be present in topo before multiorch has received its first successful health snapshot, so `LastSeen` may still be nil during leader appointment. It would be useful for that case to be excluded cleanly instead of risking a multiorch panic while recovery is already in progress

dweitzman: `(*timestamppb.Timestamp)(nil).AsTime()` returns the epoch of `1970-01-01 00:00:00 UTC`, which is maybe a little odd looking in logs but isn't a panic

PR #1036, opened 2026-05-26
fix(multipooler): use Pause/resume around stale-primary stopPostgresIfRunning

dweitzman:
When multiorch's DemoteStalePrimary routed through demoteStalePrimaryLocked and pg_rewind failed (e.g. rewind source temporarily unreachable), the connection pool was left permanently closed: stopPostgresIfRunning closed the pool before pg_ctl, and only the success-path restartPostgresAsStandby reopened it. Every subsequent RPC on the demoted pooler surfaced "manager is closed" until the postgres monitor's next startPostgres cycle reopened the pool — observed in a k8s scale-up incident.

stopPostgresIfRunning now wraps the pgctld.Stop call with pm.Pause() and defer resume(), mirroring RewindToSource's pattern for "temporarily stop postgres, then bring it back." Pause sets NOT_SERVING and tears down the heartbeat/monitor for the stop window; the deferred resume() reopens on every exit path (including pg_rewind failure), so no caller is left holding a permanently closed pool.

pgctld.Stop additionally escalates fast → immediate (matching graceful shutdown). fast SIGTERMs postgres backends directly so a held client pool never blocks stop; immediate is the safety escalation if fast hangs.

### DIFF
--- a/go/services/multipooler/manager/rpc_manager.go
+++ b/go/services/multipooler/manager/rpc_manager.go
@@ -1008,29 +1008,37 @@ func (pm *MultiPoolerManager) stopPostgresIfRunning(ctx context.Context) error {
 
 	pm.logger.InfoContext(ctx, "Stopping postgres if running")
 
-	// Close ONLY connection pools to release database connections.
-	// This allows postgres to stop cleanly without waiting for connections,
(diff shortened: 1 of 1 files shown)


rafael (approved):
LGTM

PR #1037, opened 2026-05-26
refactor(multipooler): unify rewind-then-restart-as-standby paths

dweitzman:
`RewindToSource` (orch-driven) and the stale-primary branch of `SetTermPrimary`
(after an unplanned demotion) both went stop → pg_rewind → restart-as-standby,
with their own copies of every step. This collapses them into a single
`restartAsStandbyLocked` helper, gated on `rewindPending` for whether to run
pg_rewind at all.

`demoteStalePrimaryLocked` is gone — the remainder inlines into
`SetTermPrimary`. `fixPgBackRestPaths` moves into the shared helper, which
fixes a latent bug where `RewindToSource` was missing it.


mkindahl (approved):
LGTM. A quite straightforward refactor.

PR #1038, opened 2026-05-26
test(multiorch): fix TestSequentialGracefulShutdowns flake via TryFindPrimary

dweitzman:
assert.Never callbacks were calling RefreshPrimary, which fatals the test when no pooler currently reports IsInitialized + PoolerType=PRIMARY. A single Status RPC timeout (e.g. while multiorch reconciles sync_standby_names after a standby exits) inside any one of the 10 polls was enough to fail the test — the actual condition "primary stays the same" was perfectly happy.

Split TryFindPrimary out of RefreshPrimary as the non-fatal probe. The assert.Never callbacks use it and treat (nil, false) as "can't tell right now, keep polling" instead of fatal. A transient miss no longer fails the test; an observed *different* primary still does.


mkindahl (approved):
LGTM

PR #1039, opened 2026-05-26
chore(release): consolidate GoReleaser dry run

Verolop:
## Description

this PR updates the release workflow now that the GoReleaser set of produced artifacts has been validated.

The first dry-run workflow was intentionally kept separate from the publishing workflow so we could inspect the generated Multigres archives and checksums without changing the release path itself. After running that workflow and confirming the archive contents/checksums, this PR folds the dry-run path into `release.yml` so release validation and release publishing live in one place. This check was important so we are certain that release paths won't interfere with the integrity of each other's artifacts, especially as we add branch-specific behaviours and automations.

Manual runs now use GoReleaser snapshot mode. Tag pushes remain the only path that publish GHCR images and create GitHub Releases.

## Main Changes

- Moves the GoReleaser snapshot dry run into `release.yml`.
- Removes the standalone `release-dry-run.yml` workflow.
- Keeps publishing steps restricted to semver tag pushes.
- Selects dry-run upload contents from GoReleaser’s `dist/artifacts.json` manifest instead of uploading the full `dist/` directory.
- Adds checks that the snapshot archives exist, contain the expected files, and match `checksums.txt`.

## Tests

The standalone dry-run workflow was run successfully before this consolidation. The generated archives were inspected locally, and `checksums.txt` verified successfully.

For this PR, `release.yml` was parsed locally. The remaining validation is to run the manual `Release` workflow and confirm the snapshot artifact contains only the GoReleaser archive and checksum artifacts.


haritabh17 (approved):
🚀

PR #1041, opened 2026-05-27
feat(parser): add identifier-rewrite deparse validity test and fix the quoting bugs it surfaces

GuptaManan100:
## Summary

- Adds `TestParseIdentifierRewriteRoundtrip`, which rewrites every identifier in a parsed query to a form that *requires* double-quoting, deparses, and re-parses — asserting the result is still valid SQL. This stress-tests identifier quoting in deparse paths that the existing round-trip test never reaches, because the real-SQL corpus only uses plain lowercase names that never need quoting.
- Fixes ~3,100 identifier-quoting bugs the test surfaced across roughly two dozen deparse paths — function/trigger/aggregate/operator-class/domain/type/collation/statistics names, qualified function names, field indirection, access-method names, CTE SEARCH/CYCLE columns, GRANT/COPY column lists, and more — by routing each through `QuoteIdentifier`.
- Adds an e2e test (`keyword_funcname_test.go`) verifying keyword-named functions (`left`, `right`, `overlaps`, `current_schema`) behave identically quoted vs unquoted, against both Postgres and multigateway.

## Problem

The deparser emitted many identifiers via raw string concatenation instead of `QuoteIdentifier`. This was invisible to the existing `TestParseDeparseRoundtrip` because that test's inputs are real PostgreSQL regression queries whose identifiers are all simple lowercase names — `QuoteIdentifier` is a no-op on them, so a missing quote call produces identical output. The bug only manifests when an identifier actually needs quoting (a keyword, mixed case, a space, an embedded quote), which the corpus never exercises in those positions.

## Solution

The new test walks each parsed statement and mangles every identifier to a form that must be quoted (appending a suffix containing a space, an uppercase letter, and a double quote), then requires the deparsed SQL to re-parse. It deliberately does **not** compare ASTs — equivalent re-spellings are fine here; AST fidelity remains the job of `TestParseDeparseRoundtrip`. Identifiers are captured generically: every simple-identifier `*String` node (the parser's identifier representation) plus the structural Go-string fields (`RangeVar`, `ResTarget`, `ColumnDef`, …). A small skip-set excludes things that look like identifiers but are not — option values (`DefElem`), built-in/pg_catalog type keywords, and `SET XML OPTION` keywords — and a shape guard leaves operator symbols, function bodies, and already-complex names untouched.

Each failure it reported was a genuine deparser bug, fixed by quoting the offending name. A new helper, `QuoteIdentifierOrOperator`, quotes identifiers while passing operator symbols (`+`, `~~`) through verbatim, so shared paths like `ObjectWithArgs` and `DefineStmt` (which carry both function names and operator symbols) stay correct and `DROP OPERATOR +` does not regress.

## Notes for reviewers

- The only golden-output changes are 3 `collate.*` files: `CREATE COLLATION ... FROM nonsense` was previously *over*-quoted as `FROM "nonsense"`; it now matches Postgres's `quote_identifier`, which leaves simple names bare. The other touched golden files (`horology`, `inherit`, `text`, etc.) reflect keyword-named functions (`"left"`, `"overlaps"`, …) now being quoted as Postgres does.
- All parser suites (`TestParsing`, `TestPostgresTestsParsing`, `TestParseDeparseRoundtrip`) and the `ast` package remain green; the new round-trip test passes (0 failures), so it now serves as a regression guard — a new failure means a deparse path emits an identifier raw instead of through `QuoteIdentifier`.


haritabh17 (approved):
🚀

PR #1042, opened 2026-05-27
test(shardsetup): report startup timing stats to aid timeout tuning

mkindahl:
Add a TimingCollector that records elapsed/limit for each timeout-bounded setup operation and prints a percentage table at test teardown. Values approaching 80–90% flag operations at risk on slow CI runners.

Collected operations:
- manager ready: <pooler> (via WaitForManagerReady)
- shard bootstrap (via waitForShardBootstrap)

Process startup functions (waitForStartup, WaitForPortReady) log elapsed time inline on their success lines rather than feeding the collector.

The integration test workflow now passes -v so the table always appears in CI logs, and a post-run step writes a Startup Timing code block to the GitHub step summary via GITHUB_STEP_SUMMARY.

### DIFF
--- a/.github/scripts/startup-timing-summary.py
+++ b/.github/scripts/startup-timing-summary.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python3
+# Copyright 2026 Supabase, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


GuptaManan100 (approved):
I like the idea!

PR #1044, opened 2026-05-27
test: run integration tests against the Supabase Postgres image

mkindahl:
## Run integration tests against the Supabase Postgres image

This PR adds the ability to run the Multigres integration test suite against the Supabase-flavored Postgres image (from `github.com/supabase/postgres`) and fixes the test failures that surface when doing so.

### Background

The existing integration tests run against vanilla Postgres. Testing against the Supabase image is important because it ships with a different default user (`supabase_admin` instead of `postgres`), a patched socket directory, and Supabase-specific init SQL — all of which exposed real incompatibilities in the test infrastructure.

### What breaks and why

The Supabase container exports `POSTGRES_USER=supabase_admin` into its environment. Test processes inherited this via `os.Environ()`, causing `initdb -U supabase_admin`. This gave `supabase_admin` the SCRAM verifier and left `postgres` with none. Since `pg_hba_template.conf` requires `scram-sha-256` for all connections, every test connection (authenticating as `postgres`) failed with *password authentication failed*.

The container also exports `POSTGRES_INITDB_SQL_FILES` pointing at Supabase migration SQL that assumes a pre-existing `supabase_admin` role — running that during tests would break initdb post-setup.

### Fixes (commit 1)

- **`go/test/utils/process.go`**: new `BaseTestEnv()` helper that strips `POSTGRES_USER`, `POSTGRES_INITDB_SQL_FILES`, and `POSTGRES_INITDB_SQL_DIRS` from `os.Environ()` and appends `POSTGRES_USER=postgres`, ensuring `initdb` always creates `postgres` as the superuser.
- **`test_helpers.go`, `pgctld_test.go`, `cluster.go`, `cluster_test.go`**: replace all `os.Environ()` calls in pgctld, shardsetup, and localprovisioner test processes with `BaseTestEnv()`.
- **`go/cmd/pgctld/command/init.go`**: change `--auth-local` from `trust` to `scram-sha-256` to match `--auth-host` and the `pg_hba` template.
- **`go/test/endtoend/shardsetup/setup.go`**: raise `waitForShardBootstrap` timeout from 60 s to 240 s for Docker and slow CI runners.

After these fixes the full integration suite passes against the Supabase image: **947 tests, 0 failures**.

### Infrastructure (commit 2)

- **`Dockerfile.integration-test`**: extends the supabase-postgres base image with Go, etcd, and gotestsum; the multigres source tree is volume-mounted at runtime.
- **`.github/scripts/`**: three scripts (`build-supabase-postgres.sh`, `build-supabase-postgres-test-image.sh`, `run-integration-tests-supabase.sh`) that encapsulate the docker build/run steps and are shared between the Makefile and the workflow.
- **`.github/workflows/test-integration-supabase.yml`**: CI workflow that checks out both repos, builds the images, runs the suite with `RERUN_FAILS=3`, and uploads JUnit/JSON artifacts. Runs nightly at 02:00 UTC and on demand via `workflow_dispatch`.
- **`Makefile`**: `docker-supabase-postgres`, `docker-supabase-postgres-test`, and `test-integration-supabase` targets for local use.
- **`docs/supabase-postgres-testing.md`**: local usage guide.

Fixes MUL-534

### DIFF
--- a/.github/workflows/test-integration-supabase.yml
+++ b/.github/workflows/test-integration-supabase.yml
@@ -1,0 +1,88 @@
+# Copyright 2026 Supabase, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Multigres Integration Tests (Supabase Postgres)
+
+on:
+  schedule:
+    - cron: "0 2 * * *" # 02:00 UTC nightly
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  test-integration-supabase:
+    name: Run integration tests against Supabase Postgres
+    runs-on: ubuntu-24.04
+    environment:
+      name: ci
+      deployment: false
+
+    steps:
+      - name: Checkout multigres
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Checkout supabase/postgres
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: supabase/postgres
+          path: supabase-postgres
+          persist-credentials: false
+
+      - name: Build supabase Postgres base image
+        env:
+          SUPABASE_POSTGRES_DIR: ${{ github.workspace }}/supabase-postgres
+        run: .github/scripts/build-supabase-postgres.sh
+
+      - name: Build integration test runner image
+        run: .github/scripts/build-supabase-postgres-test-image.sh
+
+      - name: Run integration tests
+        env:
+          TEST_PRINT_LOGS: "1"
+          AWS_ACCESS_KEY_ID: test-access-key
+          AWS_SECRET_ACCESS_KEY: test-secret-key
+          RERUN_FAILS: "3"
+        run: .github/scripts/run-integration-tests-supabase.sh
+
+      - name: Upload test logs
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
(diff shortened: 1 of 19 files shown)


.github/workflows/test-integration-supabase.yml
```diff
@@ -0,0 +1,90 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Multigres Integration Tests (Supabase Postgres)
+
+on:
+  push:
+    branches: [chore/supabase-integration-tests]
```
rafael: would we change this in a follow up PR?

misc/hooks/shfmt
```diff
@@ -12,10 +12,34 @@ fi
 FILES=$(git diff --cached --name-only --diff-filter=ACMR)
 [ -z "$FILES" ] && exit 0
 
-# Format all selected files
-echo "$FILES" | xargs shfmt --apply-ignore -w
+# Filter to shell files only: shfmt errors on files that start with '#' but
```
rafael, line 12: why this change in the context of this PR?

rafael (approved):

.github/workflows/test-integration-supabase.yml
mkindahl: This is a separate commit, so I will drop this before I merge. It is only there to test the workflow.

misc/hooks/shfmt
mkindahl: There are changes to the `Makefile` in this PR, and since `git diff --cached --name-only --diff-filter=ACMR` extract all files this cause `shfmt` to break since it treats any file starting with `#` as a shell file, for example:

```bash
$ shfmt Makefile 
Makefile:47:9: a command can only contain words and redirects; encountered `(`
```

This means that it is necessary to filter out anything that is not a shell script from the list of files being processed.

This probably did not cause problems before because we did not make modifications to files that `shfmt` bail on, like the `Makefile`.

I could probably create something more streamlined than this, which Claude generated.

mkindahl: Done

.github/workflows/test-integration-supabase.yml
mkindahl: Dropped from the PR now.

mkindahl:
>go/test/endtoend/shardsetup/setup.go: raise waitForShardBootstrap timeout from 60 s to 240 s for Docker and slow CI runners.

This was necessary to make it pass when using Docker locally, but I added #1042 so that we should be able to tune the timeouts in CI as well.

PR #1046, opened 2026-05-27
fix(multigateway): route by LeaderObservation, not topology PoolerType

dweitzman:
## What

Replaces the gateway's `cachedPrimaries map[shardKey]{*PoolerConnection,term}` with `leaders map[shardKey]*LeaderObservation`. Leader identity is now sourced exclusively from the highest-term `LeaderObservation` received on health streams. Topology `pooler.Type` is reduced to a cold-start hint and to replica filtering.

## Why

The connection-keyed cache coupled leader identity to a resource lifecycle: removing a pooler erased identity, and observations arriving before the named pooler was connected were silently dropped. Combined with `evictConflictingPrimary` — which deleted the real leader's connection whenever a stale pooler re-asserted `Type=PRIMARY` in etcd — this could pin the gateway to a demoted pooler for the rest of its process lifetime.

## Bugs fixed

- **Eviction-induced primary loss.** A demoted-then-restarted pooler re-asserting `Type=PRIMARY` no longer causes the gateway to drop its connection to the actual leader. `evictConflictingPrimary` is deleted; two `Type=PRIMARY` entries in topology during a failover are now normal and harmless.
- **Silent-drop on out-of-order discovery.** Higher-term observations naming a not-yet-connected pooler are now recorded and applied as soon as the connection arrives, instead of being dropped.
- **Identity erased on connection removal.** `leaders[shard]` survives `RemovePooler` so a brief connection drop no longer reverts the gateway to "no leader observed yet."

## Notes

- Cold-start: when no observation has arrived yet, `AddPooler` seeds a term-0 synthetic `LeaderObservation` from `pooler.Type == PRIMARY`. Any real observation (term ≥ 1) wins on term comparison.
- Buffer drain still requires the leader to self-report `PoolerType_PRIMARY` on its health stream, matching pre-refactor invariants.
- Follow-up: have multipooler publish its current `LeaderObservation` into its etcd record so the synthetic seed can be dropped.

### DIFF
--- a/go/services/multigateway/poolergateway/load_balancer.go
+++ b/go/services/multigateway/poolergateway/load_balancer.go
@@ -125,10 +125,19 @@ func (lb *LoadBalancer) SetReplicationLagThresholds(lowLag, highTolerance time.D
 // AddPooler creates a new PoolerConnection for the given pooler.
 // If a connection already exists for this pooler, it updates the pooler info.
 //
-// AddPooler does not touch leader state — that comes exclusively from
-// LeaderObservation. When a newly-added connection happens to be the leader
-// already-known via prior observations, the buffer is drained promptly via
-// notifyIfLeaderServingLocked.
+// AddPooler does not establish leader identity from real consensus state —
+// that comes exclusively from LeaderObservation on health streams. The one
+// concession is a cold-start fallback: if no LeaderObservation has been
+// received for the shard yet and topology says this pooler is PRIMARY, we
+// seed `leaders[shard]` with a synthetic LeaderObservation at term 0. Any
+// real observation (term >= 1) beats this seed unconditionally, so a stale
+// Type=PRIMARY assertion from a demoted-then-restarted pooler cannot
+// override consensus. The seed only matters during the gateway's startup
+// window before health streams have produced any observations.
+//
+// TODO: once multipooler publishes its current LeaderObservation into its
+// etcd record (the proto would need a `last_observed_leader` field), drop
+// the synthetic seed and use that real observation instead.
 func (lb *LoadBalancer) AddPooler(pooler *clustermetadatapb.MultiPooler) error {
 	poolerID := poolerIDString(pooler.Id)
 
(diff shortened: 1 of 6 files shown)


go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -187,6 +179,34 @@ func (lb *LoadBalancer) AddPooler(pooler *clustermetadatapb.MultiPooler) error {
...
+// observation has been received yet for the shard. Any real observation
+// (term >= 1) overrides this seed via the standard term-reconciliation in
+// onPoolerHealthUpdate, so a stale Type=PRIMARY assertion from a
+// demoted-then-restarted pooler cannot override consensus. Caller must hold
+// lb.mu.
+//
+// TODO: once multipooler publishes its current LeaderObservation into its
+// etcd record (the proto would need a `last_observed_leader` field), drop
+// this synthetic seed and use the real observation instead.
+func (lb *LoadBalancer) maybeSeedColdStartLeaderLocked(key shardKey, pooler *clustermetadatapb.MultiPooler) {
```
rafael, line 193: wouldn't it be even safer to not seed the primary and always wait for an actual observation ?

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -229,28 +248,50 @@ func (lb *LoadBalancer) GetConnection(target *query.Target) (*PoolerConnection,
...
+	leaderObs := lb.leaders[key]
 
-	if targetType == clustermetadatapb.PoolerType_PRIMARY {
-		// Use cached primary (populated by health stream callbacks).
-		// The health stream's LeaderObservation is the authoritative source
-		// for leader identity — no type-based fallback.
-		key := shardKey{tableGroup: target.TableGroup, shard: target.Shard}
-		if cached, ok := lb.cachedPrimaries[key]; ok {
-			return cached.conn, nil
+	if target.PoolerType == clustermetadatapb.PoolerType_PRIMARY {
```
rafael, line 254: or do we want to bail here if the observation is the seed? Just as extra guardrail

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -229,28 +248,50 @@ func (lb *LoadBalancer) GetConnection(target *query.Target) (*PoolerConnection,
...
-			candidates = append(candidates, conn)
+	for id, conn := range lb.connections {
+		if !matchesShardTarget(conn, target) {
+			continue
+		}
+		if confirmedLeaderID != "" {
+			if id == confirmedLeaderID {
+				continue
+			}
+		} else if conn.Type() != clustermetadatapb.PoolerType_REPLICA {
```
rafael: This check is not obvious to me. Why we are doing this?

rafael (approved):
Few comments small comments.

go/services/multigateway/poolergateway/load_balancer.go
dweitzman: Conceptually yes, but it'd make the tests flaky. I'll do a follow-up to put LeaderObservation into etcd to solve the problem "correctly>

go/services/multigateway/poolergateway/load_balancer.go
dweitzman: If the observation is still the seed that means we have no observations from the poolers themselves, so the etcd seed would still be worth respecting as the only information we have.

go/services/multigateway/poolergateway/load_balancer.go
dweitzman: Pulled out a helper which I think makes this more clear

PR #1048, opened 2026-05-27
feat(release): add version metadata to multigres binaries

niconosenzo:
## Description:

Expose build metadata (version, commit, build date) in every core binary via a new `go/common/version` package, populated at link time through `-ldflags -X`. All six binaries (multigateway, multipooler, pgctld, multiorch, multiadmin, multigres) now respond to `--version` with the binary name, version, short commit, build timestamp, Go version, and target platform.

## Main changes:

- CLI --version: all six core binaries (multigateway, multipooler, pgctld, multiorch, multiadmin, multigres) wire cmd.Version =  version.String(name) on their root cobra command, with a template that prints `<binary>\n  version: ...\n  commit: ...\n  committed:    ...\n  go: ...\n  platform: ...`.
- HTTP /version (servenv): JSON payload gains the release tag alongside the existing VCS fields.
- gRPC ServiceInfo.GetBuildInfo: BuildInfo proto gains a `version` field, serviceInfoServer populates it from version.Version.
- OTel service.version resource attribute: emits the release tag (falling back to the VCS revision for unversioned dev builds) so metrics/traces/logs tag builds by semver, not commit SHA.
- OCI image labels: Dockerfile and Dockerfile.pgctld expose org.opencontainers.image.version/.revision/.created, fed by VERSION/COMMIT/DATE Docker build args, for offline `docker inspect`.

## Validation:

Makefile target can be validated building a test version and passing random values:
`make build VERSION=v0.1.0-test 2>&1 | tail -5 && echo "---" && bin/multigateway --version`


Makefile
```diff
@@ -30,6 +30,15 @@ export SQLLOGICTEST_VER
 CMDS = multigateway multipooler pgctld multiorch multigres multiadmin portpoolserver
 BIN_DIR = bin
 
+# Version metadata injected at build time via -ldflags -X. Overridable from
+# the environment (e.g. Docker build args) so release pipelines can supply
+# their own values.
+VERSION ?= $(shell git describe --tags --always --dirty 2>/dev/null || echo dev)
+COMMIT  ?= $(shell git rev-parse --short HEAD 2>/dev/null || echo unknown)
```
dweitzman: I think Go natively already compiles in the commit and the commit's time (not the current/build time): https://pkg.go.dev/runtime/debug#BuildSetting

If we want the exact time of build rather that commit and the git tags then we may need to include some of that information manually here

Verolop:
thank you for putting together!

having the `--version` surface is definitely useful, especially for local debugging and release verification purposes.

Given the related work in #1014, I think the cleanest path is use that for the shared build info model, and then add the cli/release-facing pieces from your proposal on top.

The main thing I’d like us to avoid is having separate version mechanisms that can drift. Ideally we get one consistent story across:

- `--version`
- `/version`
- gRPC `ServiceInfo`
- Docker image labels
- GitHub release artifacts

For the release pipeline, we may still want to inject the release tag/version as an additional field, while continuing to use Go’s native BuildInfo for commit/revision metadata. That way we keep #1014 as the foundation but still expose the semver users expect to see.

niconosenzo:
Made some assumptions: 
- Left the `go/common/version/version.go` as the single source of truth, it actually uses both the release tag injected by the ci surface and the metadata embedded by go. So moved David's buildSnapshot logic (`buildinfo.go `) into this package. This way we don't have to import the whole `servenv` package for each binary entry point just to print a version string
- OTel `service.version` resource attribute emits the release tag and fall back to the VCS revision for unversioned dev builds, so metrics/traces/logs tag builds by semver, not commit SHA.

Verolop (commented):
Thanks for the context on the assumptions. I like the direction of using `go/common/version` as the single source of truth 😎 

The one place I think the model still breaks is Docker image builds. The release workflow sets OCI image labels, but it does not pass the release metadata as Docker build args, and `.dockerignore` excludes `.git`, so the Go build inside Docker cannot populate `vcs.revision` / `vcs.time` from BuildInfo. That means the image labels can say the right revision/version while `multigres --version` or `pgctld --version` inside the same image reports `version: dev`, `commit: unknown`, and `committed: unknown`.

Could you make the Docker image path preserve the same #1014  + release-tag model, either by passing explicit ldflags for the binary metadata or by making VCS metadata available to the Go build?

niconosenzo:
Good catch, confirmed and fixed!

Dockerfile.cluster
```diff
@@ -27,8 +27,13 @@ RUN apk add --no-cache git=2.52.0-r0 make=4.4.1-r3 bash=5.3.3-r1
...
 RUN go mod download
 
+# Release metadata injected into binaries via the Makefile's -ldflags.
+ARG VERSION=dev
+ARG COMMIT=unknown
+ARG DATE=unknown
+
 COPY . .
-RUN make build-release
+RUN make build-release VERSION=$VERSION COMMIT=$COMMIT DATE=$DATE
```
Verolop, line 41: looks like docker-build-cluster just needs the same treatment as the others, otherwise that image may still report dev/unknown

docs/building.md
```diff
@@ -69,6 +69,43 @@ This starts a multigres cluster. To stop:
 multigres cluster stop
 ```
 
+## Version Metadata
+
+Every binary reports build metadata via `--version`:
```
Verolop: non-blocking wording nit: since portpoolserver is still built too, “every binary” might be a bit broad unless we add --version there as well. Maybe “the main service/CLI binaries” or something like that?

Verolop (commented):
sweet! Left one comment about Dockerfile.cluster now expects VERSION/COMMIT/DATE to be passed in ( which can be addressed in a follow up pr), and a non-blocking wording nit. 
Apologies for letting this become stale to the point of getting a little merge conflict.

mkindahl (approved):

PR #1050, opened 2026-05-28
test(pgproto): add wire-protocol conformance differential suite

GuptaManan100:
## Summary

- Adds a `pgproto` wire-protocol conformance suite (`go/test/endtoend/queryserving/pgproto/`) that drives [Tatsuo Ishii's `pgproto`](https://github.com/tatsuo-ishii/pgproto) tool against both a standalone PostgreSQL and the Multigres multigateway, then diffs the frontend/backend message traces. It exercises raw V3 protocol sequences (simple query, extended Parse/Bind/Describe/Execute/Sync, COPY, error-then-Sync recovery, fast-path FunctionCall) that our Go-native protocol tests can only cover piecemeal.
- PostgreSQL is the oracle; known, accepted divergences are recorded as **patches** (same pattern as `pgregresstest`), so the suite gates on _zero unpatched divergences_ rather than tracking a baseline. The Go test fails on any unpatched divergence.
- Fixes **two real multigateway bugs** the suite surfaced (details below).
- `make tools` builds `pgproto` from a pinned source commit (SHA-verified). CI runs on the `Run Extended Query Serving Tests` PR label plus a daily cron that Slack-alerts on failure.

## Bugs fixed

**1. Syntax errors returned the wrong SQLSTATE.** The multigateway parses queries locally, and a parse failure surfaced as `MTD04` on the extended-query path and `XX000` (internal_error) on the simple-query path, instead of PostgreSQL's `42601` (syntax_error). Drivers branch on `42601`, and `XX000` actively signals "server bug" to clients. Fixed by introducing `mterrors.NewParseError` as the single source of truth for the parse-error SQLSTATE, applied at the parser call sites (the parser package stays `mterrors`-free), with `conn.go` preserving the structured diagnostic through the extended-query Parse path instead of blanket-wrapping it as `MTD04`.

**2. COPY FROM STDIN `CopyFail` returned the wrong SQLSTATE.** A client `CopyFail` produced `08P01` (protocol_violation, "COPY failed") instead of PostgreSQL's `57014` (query_canceled, "COPY from stdin failed: <msg>"). Now matches PostgreSQL's `copyfromparse.c` exactly.

## How the suite works

The trace is normalized before comparison: `Error`/`NoticeResponse` lines are reduced to their SQLSTATE, dropping fields a proxy can't reproduce (PostgreSQL's internal source file/line/routine) or that legitimately differ for gateway-raised errors (message wording, position). A SQLSTATE mismatch is still a genuine divergence and is caught. Known divergences are absorbed by per-file patches under `testdata/patches/<name>.patch`; each patch carries a comment preamble explaining _why_ the divergence is accepted, and the comment is preserved across `make pgproto-update-patches`. One patch ships today — `error_recovery.patch` — recording an extended-query `BindComplete`-before-error sequencing difference on a plan-time error (`SELECT 1/0`), which stems from the gateway planning lazily at Execute vs. PostgreSQL folding constants at Bind; matching it would need an eager bind-time backend round-trip on the hot path for negligible client benefit.

Run with `make pgproto` (verify) or `make pgproto-update-patches` (regenerate patches). See the package README for the data-file format and how to expand the corpus.

## Follow-ups

- [ ] Expand the corpus with more adversarial sequences (Bind before Parse, Execute against a closed portal, pipelined batches that error mid-batch, multi-statement simple query, Describe of unknown statements, Flush behavior).
- [ ] Fix the bug we found in `pgproto` itself: its `process_bind` is missing a `SKIP_TABS` between a bound parameter's length and value, so parameter _values_ can't be sent from a data file (only zero-parameter Binds work). The corpus currently embeds literals instead of `$1`. We should fix it and send an upstream PR.
- [ ] Consider extracting the shared parts of the `pgregresstest` and `pgproto` harnesses (patch verify/generate, diff helpers, normalization) into a common package — both implement the same patch pattern independently today.

## Test plan

- [x] `make tools` builds `bin/pgproto` from pinned source.
- [x] `make pgproto` → 4/4 files match (3 exact, 1 via patch), 0 unpatched divergences, test passes.
- [x] Introducing an unpatched divergence fails the test (red check); `make pgproto-update-patches` records it (preserving patch comments).
- [x] Unit tests cover trace normalization and patch split/apply.


haritabh17 (approved):
🚀

PR #1051, opened 2026-05-28
docs: clean up stale references and gaps

niconosenzo:
Reviewed README.md files and docs/:

- demo/local/README.md: add Prerequisites, prepend `cluster init` to Quick Start so first-time users do not hit a missing-config error.
- demo/k8s/README.md: add `make images` command to Step 1, flag the supafirehose step as external/optional since the directory is not in this repo.
- web/multiadmin/README.md: document `MULTIADMIN_API_URL` (defaults to `http://localhost:15000`), correct dev server port to 3000 and explain how to override.
- docs/architecture.md: add MultiAdmin to the Components section.
- go/tools/asthelpergen/README.md: fix module path examples (`github.com/supabase/multigres` -> `github.com/multigres/multigres`) and restore the missing `common/` segment in the iface path.
- docs/ha/decision-log/2026-01-31-separate-voting-term-from-primary-term.md: flip Status from `In Progress` to `Implemented` (shipped via `primary_term` in proto + consensus code).
- docs/contributing.md: note that `make tools` installs the pre-commit hooks from `misc/git/`, add DCO sign-off (`git commit -s`) requirement.
- docs/README.md: add Design Documents section linking to `query_serving/`, `ha/`, and `general/` subtrees (previously unreachable from any index).
- proto/README.md: document `make proto` for regenerating bindings.
- config/README.md: document `POSTGRES_PASSWORD_FILE` (used by the k8s demo).
- go/test/endtoend/pgregresstest/README.md: list `patch_verify.go` / `patch_verify_test.go` in the File Structure section.
- go/common/parser/testdata/postgres/README.md: clarify that `../postgres-tests/` is a local scratch directory and not checked in.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -64,18 +64,23 @@ The following environment variables are recognized by `pgctld` (and to
 some extent, Multipooler) and follow the Docker `postgres` image
 convention.
 
-| Variable               | CLI flag               | Default    | Description                           |
-| ---------------------- | ---------------------- | ---------- | ------------------------------------- |
-| `POSTGRES_USER`        | `--pg-user` / `-U`     | `postgres` | PostgreSQL user name                  |
-| `POSTGRES_PASSWORD`    | _(none)_               | _(empty)_  | PostgreSQL password                   |
-| `POSTGRES_DB`          | `--pg-database` / `-D` | `postgres` | PostgreSQL database name              |
-| `POSTGRES_INITDB_ARGS` | `--pg-initdb-args`     | _(empty)_  | Extra arguments forwarded to `initdb` |
+| Variable                 | CLI flag               | Default    | Description                                                |
+| ------------------------ | ---------------------- | ---------- | ---------------------------------------------------------- |
+| `POSTGRES_USER`          | `--pg-user` / `-U`     | `postgres` | PostgreSQL user name                                       |
+| `POSTGRES_PASSWORD`      | _(none)_               | _(empty)_  | PostgreSQL password (inline)                               |
+| `POSTGRES_PASSWORD_FILE` | _(none)_               | _(empty)_  | Path to a file containing the password (preferred for k8s) |
+| `POSTGRES_DB`            | `--pg-database` / `-D` | `postgres` | PostgreSQL database name                                   |
+| `POSTGRES_INITDB_ARGS`   | `--pg-initdb-args`     | _(empty)_  | Extra arguments forwarded to `initdb`                      |
 
 `POSTGRES_USER`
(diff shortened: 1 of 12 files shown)


config/README.md
```diff
@@ -64,18 +64,22 @@ The following environment variables are recognized by `pgctld` (and to
...
+| Variable                 | CLI flag               | Default    | Description                                                |
+| ------------------------ | ---------------------- | ---------- | ---------------------------------------------------------- |
+| `POSTGRES_USER`          | `--pg-user` / `-U`     | `postgres` | PostgreSQL user name                                       |
+| `POSTGRES_PASSWORD`      | _(none)_               | _(empty)_  | PostgreSQL password (inline)                               |
+| `POSTGRES_PASSWORD_FILE` | _(none)_               | _(empty)_  | Path to a file containing the password (preferred for k8s) |
+| `POSTGRES_DB`            | `--pg-database` / `-D` | `postgres` | PostgreSQL database name                                   |
+| `POSTGRES_INITDB_ARGS`   | `--pg-initdb-args`     | _(empty)_  | Extra arguments forwarded to `initdb`                      |
 
 `POSTGRES_USER`
 : PostgreSQL user for connecting to PostgreSQL server and perform administrative actions. This user requires `SUPERUSER` privileges and is expected to use SCRAM-256 password authentication over a local socket connection, but this is not enforced.
```
mkindahl: I think it might actually be enforced now: it depends a little on what we mean by "enforced".

demo/local/README.md
```diff
@@ -2,11 +2,32 @@
...
+
+- `multigres` binary built (`make build` from repo root) and on `PATH`, e.g.
+  `export PATH="$PWD/bin:$PATH"`
+- PostgreSQL 17.x installed locally (multigres invokes the system `postgres`)
+- Docker (only for the optional observability stack)
+- `psql` client (for connecting to the cluster)
+- `pgbench` (only for the traffic-generation example)
+
+`/path/to/multigres_local` below is a local directory you choose — pick any
+path; every subsequent command must pass the same `--config-path`.
```
mkindahl: There is a default path for this, which is `multigres_local` in the same directory as where you run the command, so you could simplify the commands below and add a note that the `--config-path` option exists.

I usually do something like this:

```bash
$ cd /tmp
$ multigres cluster init
```

mkindahl (approved):
Some minor suggestions for improvement, but it looks good.

demo/local/README.md
niconosenzo: `MTDATAROOT` envvar should also be mentioned as an alternative.

PR #1052, opened 2026-05-28
feat(docker): add docker-compose cluster setup for CI compatibility testing

GuptaManan100:
## Summary

- Adds an all-in-one cluster image (`Dockerfile.cluster`) plus a `docker-compose.yml` that boot a complete Multigres cluster inside a single container and expose a PostgreSQL endpoint, so any external tool, application, or test suite can run against Multigres with one `docker compose up` — no manual multi-process setup.
- Makes the cluster configurable for different testing shapes: number of cells via `MULTIGRES_NUM_CELLS` and the gateway's PostgreSQL port via `MULTIGRES_GATEWAY_PG_PORT` (so it can act as a drop-in PostgreSQL on the standard `5432` for suites that hardcode it).
- Adds CI: a smoke-test workflow that builds the image, boots the cluster, and runs a query round-trip (pinging Slack on `main` failures), plus a workflow that publishes the cluster image to GHCR on `main` and release tags.
- Builds the existing `multigres` image on release tags too, so tagged releases produce published images.

## Description

Multigres already ships component images, but running it normally means standing up etcd, multiadmin, and a per-cell stack of pgctld/PostgreSQL, multipooler, multiorch, and multigateway. That setup friction makes it hard to point external test suites or compatibility checks at Multigres. This PR packages the whole thing into a single container that behaves like an ordinary PostgreSQL server, so the common testing pattern — "bring up a database and run my suite against it" — works against Multigres unchanged.

`Dockerfile.cluster` bundles PostgreSQL, etcd, pgBackRest, and the multigres binaries. The entrypoint (`docker/cluster-entrypoint.sh`) runs `multigres cluster init` + `multigres cluster start` (the local provisioner) to bootstrap the cluster, then supervises it and tears it down cleanly on signal. Clients connect through the zone1 multigateway, which speaks the PostgreSQL wire protocol.

Knobs, both settable from compose or any orchestrator that consumes the image:

- `MULTIGRES_NUM_CELLS` — number of full stacks to run. The provisioner bootstraps the shard with an `AtLeastN(2)` durability policy, so at least two poolers are required to elect a leader; the value is therefore clamped to the supported `2`–`3` range, defaulting to `2` (the minimum that serves queries). `3` tolerates losing a node.
- `MULTIGRES_GATEWAY_PG_PORT` — the port the gateway serves PostgreSQL on (default `15432`). Setting it to `5432` lets the container drop straight into setups that expect a standard PostgreSQL.

CI:

- `docker-compose-ci.yml` builds the image, waits for health, runs a DDL/DML round-trip as a compatibility smoke test, dumps logs on failure, and notifies Slack when it fails on `main`.
- `docker-build-cluster.yml` publishes `ghcr.io/<owner>/multigres-cluster` on `main` (`:latest`) and on `v*` release tags, mirroring the existing image-publish workflows.
- `docker-build.yml` now also triggers on `v*` tags so the `multigres` image is built for releases.

Verified locally end to end: the default 2-cell cluster boots healthy and passes the smoke test, and the `5432` gateway-port path serves queries.

## Test plan

- [x] `docker compose up --build` brings the cluster up healthy
- [x] `PGPASSWORD=postgres psql -h 127.0.0.1 -p 15432 -U postgres -d postgres -c 'SELECT 1'` succeeds
- [x] `docker-compose-ci.yml` passes on the PR


.github/workflows/docker-build.yml
```diff
@@ -19,6 +19,8 @@ on:
   push:
     branches:
       - main
+    tags:
```
Verolop: `release.yml` already publishes images with SBOM, provenance, and signing. Please remove the `tags: ["v*"]` trigger from this workflow so the two workflows do not race to publish the same semver tags

Dockerfile.cluster
```diff
@@ -0,0 +1,124 @@
...
+
+# Ports: zone1 multigateway PostgreSQL, multigateway HTTP, multiadmin HTTP.
+EXPOSE 15432 15100 15000
+
+# initdb and the postgres server both refuse to run as root, so run as the
+# postgres user that the base image provides.
+USER postgres
+WORKDIR /multigres/cluster
+
+HEALTHCHECK --interval=10s --timeout=5s --start-period=60s --retries=12 \
```
Verolop, line 123: the entrypoint lets callers change the gateway port with `MULTIGRES_GATEWAY_PG_PORT`, but this healthcheck always probes `15432`. Please make the healthcheck use the configured port so `MULTIGRES_GATEWAY_PG_PORT=5432` does not mark a healthy container as unhealthy.

docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      # Storage's tenant_db). Keep ports/healthcheck below in sync.
+      MULTIGRES_GATEWAY_PG_PORT: "15432"
+    ports:
+      - "15432:15432" # zone1 multigateway — PostgreSQL wire protocol
+      - "15100:15100" # multigateway HTTP (status/metrics)
+      - "15000:15000" # multiadmin HTTP (status/metrics)
+    healthcheck:
+      test:
+        - "CMD-SHELL"
+        - "PGPASSWORD=postgres psql -h 127.0.0.1 -p 15432 -U postgres -d postgres -tAc 'SELECT 1' || exit 1"
```
Verolop: please keep this healthcheck in sync with `MULTIGRES_GATEWAY_PG_PORT`; otherwise changing the gateway port requires remembering to update this hardcoded `15432` too

.github/workflows/docker-compose-ci.yml
```diff
@@ -0,0 +1,124 @@
...
+          done
+          echo "Timed out waiting for the cluster to become healthy." >&2
+          exit 1
+
+      - name: Run PostgreSQL compatibility smoke test
+        run: |
+          set -euo pipefail
+          run_sql() {
+            docker compose exec -T -e PGPASSWORD=postgres multigres \
+              psql -h 127.0.0.1 -p 15432 -U postgres -d postgres -v ON_ERROR_STOP=1 "$@"
```
Verolop: please use the configured gateway port here as well, so the CI smoke test stays in sync if `MULTIGRES_GATEWAY_PG_PORT` changes

Verolop (commented):
this is super useful, thanks! Some of the configs are prone to conflict with the release.yaml settings, so I left some comments to tighten a few release/port-configuration details before merging

.github/workflows/docker-build.yml
GuptaManan100: Removed the `tags: ["v*"]` trigger from `.github/workflows/docker-build.yml`. `release.yml` is now the sole publisher of `multigres` on release tags (with SBOM, provenance, and cosign signing), and this workflow goes back to only firing on `main` pushes and `workflow_dispatch`.

Dockerfile.cluster
GuptaManan100: Fixed.

docker-compose.yml
GuptaManan100: Done

.github/workflows/docker-compose-ci.yml
GuptaManan100: Fixed

GuptaManan100:
Thanks for catching the overlap with `release.yml`. Addressed all four inline comments. Folded `multigres-cluster` into `release.yml`'s build matrix too.

Verolop (approved):
lgtm

PR #1053, opened 2026-05-28
refactor(multipooler): graceful shutdown advertises cohort ineligibility

dweitzman:
Replace the LEADERSHIP_SIGNAL_REQUESTING_DEMOTION announcement with COHORT_ELIGIBILITY_SIGNAL_INELIGIBLE; multiorch's LeaderNeedsReplacement now treats an INELIGIBLE leader as needing replacement. REQUESTING_DEMOTION stays in place for Recruit's post-crash demote, where term correlation matters.

### DIFF
--- a/go/services/multiorch/recovery/types/pooler_signals.go
+++ b/go/services/multiorch/recovery/types/pooler_signals.go
@@ -29,9 +29,20 @@ import (
 // signalled that it should be replaced via a new election. It reads the
 // self-reported AvailabilityStatus from the consensus status RPC.
 //
-// The staleness check (leadership_status.primary_term == consensus primary_term)
-// ensures we don't act on a REQUESTING_DEMOTION signal left over from a previous
-// election cycle that was never cleared (e.g. after a process restart).
+// Two independent signals trigger replacement:
+//
+//   - CohortEligibilityStatus.Signal == INELIGIBLE: the pooler is unwilling to
+//     remain in the cohort (e.g. graceful shutdown advertises this before
+//     stopping postgres). Not term-gated — eligibility is a current preference
+//     and staleness comes from the freshness of the surrounding health
+//     snapshot; a node that doesn't want to be in the cohort certainly should
+//     not continue as leader.
+//
+//   - LeadershipStatus.Signal == REQUESTING_DEMOTION with leader_term ==
+//     current primary term: emitted by Recruit's primary-demote path after a
+//     postgres crash so the coordinator can correlate the request with the
+//     specific term and avoid acting on a leftover signal from a previous
+//     election cycle that was never cleared (e.g. after a process restart).
 //
 // TODO: once the coordinator synthesizes AvailabilityStatus into
 // PoolerHealthState directly (see clustermetadata.proto TODO), read from
@@ -39,7 +50,11 @@ import (
 // coordinator-synthesized signals (e.g. TEMPORARILY_UNAVAILABLE for unreachable
 // nodes) are handled here too.
 func LeaderNeedsReplacement(p *multiorchdatapb.PoolerHealthState) bool {
-	leadershipStatus := p.GetAvailabilityStatus().GetLeadershipStatus()
+	av := p.GetAvailabilityStatus()
+	if PoolerIsCohortIneligible(av) {
+		return true
+	}
+	leadershipStatus := av.GetLeadershipStatus()
 	if leadershipStatus == nil {
 		return false
 	}
--- a/go/services/multipooler/manager/graceful_shutdown.go
+++ b/go/services/multipooler/manager/graceful_shutdown.go
@@ -44,25 +44,16 @@ var pgctldStopModes = []struct {
 	{"immediate", 5 * time.Second},
 }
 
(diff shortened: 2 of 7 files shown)


go/services/multipooler/manager/graceful_shutdown.go
```diff
@@ -84,26 +79,9 @@ func (pm *MultiPoolerManager) GracefulShutdown(ctx context.Context) {
...
-		case primaryTerm != 0:
-			if err := pm.setResignedLeaderAtTerm(lockCtx, primaryTerm); err != nil {
-				pm.logger.WarnContext(lockCtx, "failed to record resigned primary term",
-					"error", err)
-			}
-		}
-	}
+	// Advertise cohort ineligibility before stopping postgres just in case
+	// stopping is slow. We're favoring speed of failover rather than grace.
+	pm.setCohortEligibility(clustermetadatapb.CohortEligibilitySignal_COHORT_ELIGIBILITY_SIGNAL_INELIGIBLE)
```
dweitzman, line 111: This is the main change

go/services/multiorch/recovery/types/pooler_signals.go
```diff
@@ -29,17 +29,32 @@ import (
...
 //
 // TODO: once the coordinator synthesizes AvailabilityStatus into
 // PoolerHealthState directly (see clustermetadata.proto TODO), read from
 // PoolerHealthState.AvailabilityStatus instead of ConsensusStatus so that
 // coordinator-synthesized signals (e.g. TEMPORARILY_UNAVAILABLE for unreachable
 // nodes) are handled here too.
 func LeaderNeedsReplacement(p *multiorchdatapb.PoolerHealthState) bool {
-	leadershipStatus := p.GetAvailabilityStatus().GetLeadershipStatus()
+	av := p.GetAvailabilityStatus()
+	if PoolerIsCohortIneligible(av) {
```
dweitzman, line 54: And this is related

go/test/endtoend/multiorch/cohort_rotation_test.go
```diff
@@ -0,0 +1,199 @@
...
+// ADD), both of which have unit coverage but were not previously exercised
+// end-to-end. Write-buffering during failover is covered by
+// TestDeadPrimaryRecovery and intentionally out of scope here.
+//
+// Why primary last: each follower drain doesn't trigger leader failover, so
+// chaining them is cheap. Draining the primary is the single failover event
+// in the test — running it after the cohort has already grown to 5 means
+// LeaderResignedAnalyzer can pick from the two new poolers as recruitment
+// candidates and the replacement is fast.
+func TestCohortRotation_FullReplacement(t *testing.T) {
```
dweitzman, line 96: This e2e test expands and shrinks the cohort to prove the removing cohort members works

go/services/multipooler/manager/graceful_shutdown.go
```diff
@@ -44,28 +44,23 @@ var pgctldStopModes = []struct {
...
-// same lock — holding it across both serialises against any concurrent
-// consensus operation (Recruit, Propose, etc.).
+// The action lock is held for the whole sequence because pgctld.Stop is
+// gated behind the protectedPgctldClient action-lock check.
 func (pm *MultiPoolerManager) GracefulShutdown(ctx context.Context) {
 	pm.logger.InfoContext(ctx, "graceful shutdown starting")
 
+	// TODO: issue a bounded CHECKPOINT before resigning so there's a recent
+	// checkpoint if the failover process doesn't go smoothly and rewinds are
+	// needed.
```
mkindahl, line 62: This would optimize recovery time at the expense of shutdown time, which seems like the wrong trade-off.

mkindahl (approved):
Change looks good. But it raises the question on what we will use the `REQUESTING_DEMOTION` for? On shutdown, all poolers will make themselves ineligible, including the primary, and this will be treated as a leader resignation by the orchestrator. Right now the only usage for `REQUESTING_DEMOTION` is to announce an emergency shutdown because the Postgres process crashed, which is basically the same as the pooler being ineligible to be leader.

dweitzman:
> Change looks good. But it raises the question on what we will use the `REQUESTING_DEMOTION` for? On shutdown, all poolers will make themselves ineligible, including the primary, and this will be treated as a leader resignation by the orchestrator. Right now the only usage for `REQUESTING_DEMOTION` is to announce an emergency shutdown because the Postgres process crashed, which is basically the same as the pooler being ineligible to be leader.

Today we allow re-promotion after a crash or something else that caused becoming a replica. In the sense we're now informing of a demotion rather than requesting one. An interesting question is whether to just self re-promote instead of signaling demotion in a way that allows re promotion. I don't have any strong feelings on that. We probably want to notice if a leader is crashing repeatedly and replace it, but there could be multiple approaches

go/services/multipooler/manager/graceful_shutdown.go
dweitzman: If a checkpoint happens in parallel with draining it might not be that disruptive, and having a checkpoint seems like it provides both more safety if things go wrong during promotion and better performance with the promotion itself if we make that force a checkpoint, which we may want to do to lower the risk of rewind failing

PR #1056, opened 2026-05-29
refactor(test): extract shared query-serving harness code into suiteutil

GuptaManan100:
## Summary

- Hoists the machinery duplicated across the `pgregresstest`, `pgproto`, and `sqllogictest` end-to-end suites into `go/test/endtoend/suiteutil`, so the patch verify/generate flow, corpus discovery, subprocess helpers, per-file run scaffolding, and the differential pass/divergence report all have a single implementation.
- Pure refactor: no change to test behavior, gating, or the on-disk `results.json` shape that `detect-regressions.sh` consumes. Each suite keeps the parts that genuinely differ (normalization, runner invocation, report wording).
- Adds unit tests for the extracted logic (previously untested because it only ran inside cluster-backed suites).

## Description

The three suites had grown three independent copies of the same scaffolding. This PR consolidates them in two steps (one commit each):

**1. Patch verify/generate (`suiteutil/patch.go`).** `pgregresstest` and `pgproto` each implemented the same "PostgreSQL is the oracle, known divergences are recorded as per-case unified-diff patches" flow. The byte-oriented core — `PatchMode` + `PatchModeFromEnv`, `VerifyPatch`, `ApplyPatch`, `GenerateDiff`, `SplitPatchComment`, `FileExists` — now lives in `suiteutil`. Callers normalize their bytes first (whitespace collapse for `.out` files; SQLSTATE trace reduction for wire traces — these stay per-suite) and compute their own patch paths. Comment-preamble preservation is now always-on, which is a no-op for `pgregresstest` (its patches have no preamble) and keeps `pgproto`'s explanatory patch comments across regeneration.

**2. Corpus / runner / report scaffolding.** `pgproto` and `sqllogictest` were near-twins. Extracted:
- `suiteutil/corpus.go` — `GlobToRegexp` (was character-for-character identical) and `ListCorpusFiles(dir, globEnvVar, defaultGlob)` (identical except the env-var name, now a parameter).
- `suiteutil/output.go` — `IsExitError`, `TruncateOutput(s, limit)`, `FirstLine` (the per-suite `maxOutputBytes` stays local and is passed in).
- `suiteutil/run.go` — `RunWithTimeout` (generic per-file bounded-context wrapper), `NewSchemaResetterWithCleanup` (the timeout + connect + fatal + cleanup boilerplate each suite ran twice), and `RunCorpusLoop` (deadline-aware iteration with a progress callback).
- `suiteutil/diffreport.go` — the shared serialized shapes (`PerRun` / `FileReport` / `SuiteReport`) and `NewSuiteReport`, which owns the pass/divergence counting and status assignment that was identical across suites. Each suite still computes its own per-file verdict (intrinsic exit-code for `sqllogictest`, differential trace match + patch dimension for `pgproto`) and renders its own markdown/log summary, since that wording legitimately differs.

The patch fields on the shared report structs are `omitempty`, so `sqllogictest`'s JSON is byte-identical to before (it never carried them) and `pgproto`'s only drops a zero-valued `passed_via_patch` when no patch applied — and `pgproto` gates on the Go test exit, not on that JSON.

Net ~390 lines of duplication removed across the suites; the added `suiteutil` code is shared rather than triplicated.

## Test plan

- [x] `go vet ./go/test/endtoend/...` clean.
- [x] New `suiteutil` unit tests pass: glob semantics, corpus listing + env override, report counters + JSON-shape guard, patch verify/generate (verify pass, diff-without-patch, absorbed-by-patch, stale patch, generate writes/removes, comment preservation), `SplitPatchComment`, `ApplyPatch`-with-comment.
- [x] `pgproto`'s pure unit tests (`TestReduceErrorLine`, `TestNormalizeTraceErrorParity`) still pass.
- [x] gofumpt + goimports clean on all changed files.
- [ ] Full extended-query-serving suites (cluster-backed, gated by `RUN_EXTENDED_QUERY_SERVING_TESTS`) — not run locally; behavior preserved by construction and covered by the new unit tests.


haritabh17 (approved):
🚀

PR #1057, opened 2026-05-29
fix(multiorch): exclude unseen poolers from replica-connected check

mkindahl:
A pooler added by PoolerWatcher has StreamSnapshotsReceived==0 until its health stream delivers the first snapshot. If the recovery tick fires in that window, allReplicasConnectedToLeader would count the new pooler as a replica (via MultiPooler.Type fallback) but mark it as not connected (IsLastCheckValid==false), making the check return false even when all known-healthy replicas are still streaming WAL from the leader's postgres.

LeaderIsDeadAnalyzer uses ReplicasConnectedToLeader as the gate on its suppression window: when false, the "wait until replicas disconnect before failing over" logic is skipped entirely, risking a premature AppointLeaderAction during a transient leader blip.

Fix: skip poolers with StreamSnapshotsReceived==0 in allReplicasConnectedToLeader before incrementing replicaCount. This correctly distinguishes "never observed" (skip both counts) from "was healthy, now unreachable" (count but not connected).

Update test fixtures to set StreamSnapshotsReceived=1 on replicas that model previously-healthy nodes, and add a regression test that directly encodes the startup-race scenario.

Fixes MUL-552

### DIFF
--- a/go/services/multiorch/recovery/analysis/generator.go
+++ b/go/services/multiorch/recovery/analysis/generator.go
@@ -333,6 +333,13 @@ func (g *AnalysisGenerator) allReplicasConnectedToLeader(
 			continue
 		}
 
+		// Skip poolers that have never reported health — we have no information
+		// about their replication state, so they must not influence the
+		// connected-count check used to suppress failover.
+		if pooler.StreamSnapshotsReceived == 0 {
+			continue
+		}
+
 		replicaCount++
 
 		// Check if replica is connected to the leader's postgres
--- a/go/services/multiorch/recovery/analysis/generator_test.go
+++ b/go/services/multiorch/recovery/analysis/generator_test.go
@@ -894,7 +894,8 @@ func TestAllReplicasConnectedToLeader(t *testing.T) {
 					Shard:      "shard1",
 				},
 			},
-			IsLastCheckValid: true,
+			IsLastCheckValid:        true,
+			StreamSnapshotsReceived: 1,
 			Status: &multipoolermanagerdatapb.Status{
 				PoolerType: clustermetadatapb.PoolerType_REPLICA,
 				ReplicationStatus: &multipoolermanagerdatapb.StandbyReplicationStatus{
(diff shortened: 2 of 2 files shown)


go/services/multiorch/recovery/analysis/generator.go
```diff
@@ -333,6 +333,13 @@ func (g *AnalysisGenerator) allReplicasConnectedToLeader(
 			continue
 		}
 
+		// Skip poolers that have never reported health — we have no information
```
dweitzman, line 336: I'm wondering if the `allReplicasConnectedToLeader()` gate is maybe too strong to begin with?

What do you think of this definition of unreachable for a leader:
- Orch can't reach the leader
- Orch can reach at least a quorum's worth of followers
- Orch can't find at least a quorum's worth of followers that are successfully connected after they've been trying for some minimum amount of time since learning who the leader was

If orch can't reach either the leader _or_ a quorum's worth of followers then the shard's problem is probably more like "consensus is stuck" than a dead leader

mkindahl: > I'm wondering if the `allReplicasConnectedToLeader()` gate is maybe too strong to begin with?

I agree that we do not need all replicas connected to start a consensus round. My reasoning is that we can have a temporary blip in the connectivity for one of the standbys without it affecting the situation. The standby will eventually reconnect and catch up with the leader.

I think we mean the same thing here, but I elaborate below.

> What do you think of this definition of unreachable for a leader:
> 
> * Orch can't reach the leader

Then orch should fall back on the standbys and ask them. This is an indirect "ping".

> * Orch can reach at least a quorum's worth of followers

I don't think this is sufficient since we need to consider what the orchestrator will do if there is a problem.

If there is a problem with the leader, the orchestrator will want to revoke the term to elect a new leader. So to have a fast failover, it needs to be have enough followers to be able to revoke the term. We should probably have some margin as well so that we know we can revoke the term if we have to.

> * Orch can't find at least a quorum's worth of followers that are successfully connected after they've been trying for some minimum amount of time since learning who the leader was

This is a slightly different definition, but I think I agree. This is more about how we decide if the standby is connected:

Socket connected and data streaming -> definitely connected

Socket connected but no data streaming -> suspicious, have a grace period 

Socket connected but no data streaming and grace period is out -> declare dead and make sure to exclude the leader from any possible interaction with followers (that is, revoke the term).

Socket denied -> dead for sure (pooler responded and said no).

> If orch can't reach either the leader _or_ a quorum's worth of followers then the shard's problem is probably more like "consensus is stuck" than a dead leader

If we do not have enough followers to revoke the term, yes, then consensus is stuck, but if we can revoke the term, then this is "leader is dead".

mkindahl (commented):
Even though you have a good point and we should revisit the definition, this PR just avoids counting standbys for which we have not received a status report as "alive".

dweitzman (approved):

PR #1058, opened 2026-05-29
test(pgproto): add adversarial corpus sequences and fix error SQLSTATEs

GuptaManan100:
## Summary

- Expands the pgproto wire-protocol conformance corpus (added in #1050) with the six adversarial sequences from that PR's follow-up checklist: Bind before Parse, Execute against a closed portal, a pipelined batch that errors mid-batch, multi-statement simple Query, Describe of unknown statements, and Flush behaviour.
- Fixes two real multigateway bugs the new files surfaced: Bind/Describe of a nonexistent prepared statement or portal returned the internal `MTD05`/`MTD06` codes instead of PostgreSQL's `26000`/`34000`. Drivers branch on these SQLSTATEs.
- Adds coverage for two features the gateway does not implement yet — fast-path FunctionCall and `COPY ... TO STDOUT` — recording each divergence as a patch that carries the gateway's current failure, so the tests ship now and flip to a fix-signal when support lands.
- Net result: the suite is 11 corpus files, 3 absorbed by patches (`error_recovery`, `copy`, `function_call`), zero unpatched divergences. Verified locally with `make pgproto`.

## Problem

The extended-query handlers blanket-wrapped every handler error as an internal `MTDxx` code, clobbering the structured SQLSTATE the handler had already produced — the same class of bug #1050 fixed for the Parse path (`MTD04` → `42601`). A Bind referencing a never-Parsed statement returned `MTD05` instead of `26000` (invalid_sql_statement_name); a Describe of an unknown statement or an unknown portal both returned `MTD06`, not even distinguishing `26000` (statement) from `34000` (invalid_cursor_name, portal). Clients read `MTDxx` as a proxy bug and branch on the standard SQLSTATEs, so this was a genuine protocol divergence from PostgreSQL.

## Solution

Introduce `preserveExtendedQueryError(err, fallback)` in `conn.go`: forward a structured `*PgDiagnostic` unchanged, and wrap only opaque errors with the fallback `MTDxx`. Applied to all four extended-query error sites (Parse, Bind, Describe, deferred Describe), unifying the pattern #1050 added inline for Parse. The inner handlers already produced correctly-coded diagnostics — which is why Execute against a closed portal already matched at `34000` — so the fix is purely to stop discarding them. New unit tests in `extended_query_sqlstate_test.go` pin the `26000`/`34000` preservation and the `MTD05` opaque-error fallback.

## Unsupported-feature tests (carried as patches)

`function_call.pgproto` (fast-path FunctionCall) and a new `COPY ... TO STDOUT` block in `copy.pgproto` exercise paths the gateway rejects today — FunctionCall has no `'F'` handler (`ErrorResponse(MTD03)` + connection teardown), and COPY TO STDOUT is refused at plan time (`0A000` feature_not_supported). Each divergence is recorded in a patch (`function_call.patch`, `copy.patch`) with a comment preamble explaining the current behaviour and noting the patch should be deleted once the feature is implemented — at which point the gateway will match PostgreSQL, the patch will stop applying, and the suite will flag it. The corpus files were each validated against a real PostgreSQL 17 oracle.

## Test plan

- [x] Each new `.pgproto` file validated against a standalone PostgreSQL 17.
- [x] `make pgproto` (verify mode) → 11/11 files match, 3 via patch, 0 unpatched divergences.
- [x] New `extended_query_sqlstate_test.go` unit tests pass; existing `go/common/pgprotocol/server` unit tests pass.
- [x] Confirmed the SQLSTATE fix is load-bearing: without it, `bind_before_parse` and `describe_unknown` turn the gate red.

### DIFF
--- a/go/common/pgprotocol/server/conn.go
+++ b/go/common/pgprotocol/server/conn.go
@@ -928,6 +928,25 @@ func (c *Conn) handleQuery() error {
 	return c.writeReadyForQuery()
 }
 
+// preserveExtendedQueryError returns err unchanged when it already carries a
+// structured PostgreSQL diagnostic, so the client sees the real SQLSTATE the
+// handler produced (e.g. 42601 syntax_error from the parser on Parse, 26000
+// invalid_sql_statement_name for a missing prepared statement on Bind/Describe,
+// 34000 invalid_cursor_name for a missing portal on Describe). Only an opaque
+// error with no SQLSTATE is wrapped with fallback, which marks a genuinely
+// internal failure of the extended-query step. Drivers branch on these
+// SQLSTATEs, so blanket-wrapping a structured error as an internal MTDxx code
+// (which clients read as a proxy bug) would be a real protocol divergence from
+// PostgreSQL — see the pgproto conformance suite's bind_before_parse and
+// describe_unknown corpus files.
+func preserveExtendedQueryError(err error, fallback *mterrors.MTError) error {
+	var diag *mterrors.PgDiagnostic
+	if errors.As(err, &diag) {
+		return err
+	}
+	return fallback.NewWithDetail(err.Error())
+}
+
 // handleParse handles a 'P' (Parse) message - extended query protocol.
 // Parse message format:
 // - Statement name (string, null-terminated)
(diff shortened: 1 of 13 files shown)


haritabh17 (approved):
🚀

PR #1067, opened 2026-06-01
fix(multipooler): respect context deadline when a backend cancel doesn't take effect

GuptaManan100:
## Summary

- A hung statement (e.g. an `UPDATE` / `SELECT ... FOR UPDATE` that blocks on a row lock once `NOWAIT` is dropped) could ignore its context deadline, hang forever, and hold the connection — blocking subsequent attempts that need it.
- The fix bounds the post-cancel wait: if `pg_cancel_backend` doesn't actually stop the statement within a short grace period, the connection is force-closed so the call returns near the deadline.
- Adds a regression test that reproduces the hang and verifies the deadline is now honored.

## Problem

In the internal query path (`InternalQueryService.Query` / `InternalTx.Query` → `reserved.Conn.Query` → `regular.Conn.Query`), query execution goes through `execWithContextCancel` / `execOnce`. When the context deadline fires, these call `handleContextCancellation`, which runs `pg_cancel_backend(pid)` via the admin pool.

`pg_cancel_backend` is best-effort: it returns `true` when the cancel signal is *delivered*, not when the query actually stops. PostgreSQL only acts on a cancel at `CHECK_FOR_INTERRUPTS` points, and cancel requests can be lost or race. On a successful cancel the original code deliberately left the connection open and then blocked on `<-ch` waiting for the op goroutine to return, with no upper bound. If the backend never aborted the statement, the call hung forever despite the expired context — and kept the connection checked out, so later attempts that needed it blocked too. This is the "drop NOWAIT and you can no longer cancel a hung query" symptom.

## Solution

Bound the wait after a cancel is requested. A new `drainOpAfterCancel` helper waits for the op goroutine to drain, but if it doesn't within `postCancelDrainGrace` (2s) it force-closes the connection so the read unwinds and the call returns `context.DeadlineExceeded`. It's used at both `execOnce` and `execWithContextCancel`. The common case — a cancel that takes effect promptly — is unchanged: the connection still drains and is recycled rather than force-closed.

The fix sits at the `regular.Conn` layer, so it covers both one-shot internal queries and the `InternalTx` / reserved-connection transaction path.

## Test plan

- `TestHungQuery_ContextTimeoutNotRespected_AfterSuccessfulCancel` fails (watchdog fires) without the fix and passes with it.
- `regular`, `reserved`, and `executor` package tests pass; `regular` is also clean under `-race`.

### DIFF
--- a/go/services/multipooler/pools/regular/regular_conn.go
+++ b/go/services/multipooler/pools/regular/regular_conn.go
@@ -559,6 +559,34 @@ func (c *Conn) handleContextCancellation() {
 	}
 }
 
+// postCancelDrainGrace bounds how long we wait for an in-flight operation to
+// unwind after handleContextCancellation has requested a backend cancel.
+// pg_cancel_backend only reports that the cancel signal was *delivered*, not
+// that the statement actually stopped (PostgreSQL acts on it at
+// CHECK_FOR_INTERRUPTS points, and cancel requests can be lost or race). Without
+// an upper bound, a hung statement that never honors the cancel would block the
+// caller forever, ignoring the already-expired context deadline.
+const postCancelDrainGrace = 2 * time.Second
+
+// drainOpAfterCancel waits for the op goroutine (publishing on ch) to return
+// after a backend cancel has been requested. Because pg_cancel_backend is
+// best-effort, the statement may never abort on its own; if it does not drain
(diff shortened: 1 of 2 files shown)


haritabh17 (approved):
🚀

PR #1068, opened 2026-06-01
fix(multipooler,multiorch): prevent cascade re-elections after failover

mkindahl:
## Problem

After a primary failure the coordinator promotes a survivor, and that promotion is a long-running operation that must finish before the new leader is usable. It can stall for two very different reasons — replaying accumulated WAL, or committing the new leader's consensus rule, which under `AT_LEAST_2` blocks on a synchronous-standby ACK (see the two scenarios below). Two bugs let a fast-RPC-style timeout abandon the promotion partway through that window. Abandoning it is not free: by then `pg_promote()` has **irreversibly forked the timeline**, so clearing `promotionInProgress` and firing a fresh election leaves a timeline-forked node the cluster can no longer cleanly reconcile — cascading re-elections in one scenario, an unbreakable rewind/ACK deadlock in the other.

**Bug 1 — `propose()` applied `RemoteOperationTimeout` (15 s) to the leader's Propose RPC.**

The leader's Propose RPC is intentionally a long-running blocking call: it drives `pg_promote()`, waits for the node to finish coming up (`pg_isready`), and commits the new leader's rule before returning. Applying a 15-second deadline designed for fast RPCs meant the gRPC deadline was propagated to the multipooler's server context. When the promotion window exceeded 15 s the server context was cancelled, `promotionInProgress` was cleared prematurely, and `LeaderIsDeadAnalyzer` immediately fired a new election.

**Bug 2 — `waitForPromotionComplete` had a hardcoded 30 s inner timeout.**

Even without Bug 1, a 30-second inner `context.WithTimeout` inside `waitForPromotionComplete` would fire before WAL replay finished on large databases, triggering the same premature clear of `promotionInProgress`.

### Scenario 1 — cascade re-election during failover (MUL-557)

**Observed — 2026-05-29 AZ-outage test (mg-scale12, zone-a pod deletion):**

The cascade is visible in the zone-d multiorch log. After earlier elections at terms 7 and 8, zone-d elected p-28c00303 as leader at term 9. The Propose RPC carried a 15 s `RemoteOperationTimeout` deadline. Exactly 15 seconds later:

```
13:58:13  primary.promotion started  leader=p-28c00303  term=9
           ...WAL replay in progress...
13:58:28  primary.promotion failed   leader=p-28c00303  term=9
          rpc error: code = DeadlineExceeded
                     desc = stream terminated by RST_STREAM with error code: CANCEL
13:58:28  recovery action failed  problem=LeaderIsDead
          failed to appoint leader: leader p-28c00303 failed to accept proposal
13:58:29  executing appoint leader action  ← term 12 starts immediately
```

`RST_STREAM with error code: CANCEL` is the gRPC wire signal for a client-side deadline cancellation. The multiorch's 15 s deadline fired and cancelled the stream; this propagated as context cancellation into the multipooler's Propose handler. With `promotionInProgress` cleared by the cancellation, `LeaderIsDeadAnalyzer` immediately re-fired, and zone-d started term 12 one second later.

The full election sequence over the 90 seconds following the zone-a pod deletion:

```
13:56:41  term  7  zone-b  leader=p-704bce7b  → FAILED  (postgres socket not yet open)
13:56:57  term  8  zone-b  leader=p-fe445f40  → SUCCESS (WAL replay completed within 15 s window)
13:58:12  term  9  zone-d  leader=p-28c00303  → FAILED  (WAL replay exceeded 15 s; RST_STREAM CANCEL)
13:58:17  term 10  zone-b  leader=p-1b3f9f4d  → SUCCESS (concurrent with term 9; different leader)
13:58:29  term 12  zone-d  leader=p-28c00303  → FAILED  (cohort already accepted term 10)
```

**Cascade mechanism**

```mermaid
sequenceDiagram
    participant ORCH as multiorch<br/>(zone-d)
    participant ANA as LeaderIsDeadAnalyzer
    participant MP as p-28c00303<br/>(multipooler)
    participant PG as postgres<br/>(p-28c00303)

    Note over ORCH,PG: t=0 (13:58:12) — LeaderIsDead fires, term 9 elected<br/>leader=p-28c00303, Propose RPC sent with 15 s deadline

    ORCH->>MP: Propose(term=9) [deadline: +15 s]
    MP->>MP: promotionInProgress = true
    MP-->>ORCH: health: PoolerStatus=PROMOTING, PostgresReady=false
    Note over ANA: PromotingPrimaryID=p-28c00303 → suppress ✓
    MP->>PG: SELECT pg_promote()
    PG-->>MP: ok (t ≈ 3 s)
    Note over MP,PG: WAL replay in progress (t = 3–15 s)...

    loop every ~1 s (analysis cycle)
        ORCH->>ANA: Analyze(shard)
        ANA-->>ORCH: no problem (PROMOTING suppresses)
    end

    Note over ORCH,MP: t = 15 s (13:58:28) — RemoteOperationTimeout fires
    ORCH--xMP: gRPC RST_STREAM CANCEL
    MP->>MP: ctx cancelled → waitForPromotionComplete returns error
    MP->>MP: defer: promotionInProgress = false
    MP-->>ORCH: health: PoolerStatus=RUNNING, PostgresReady=false
    Note over ANA: PromotingPrimaryID = nil
    MP-->>ORCH: Propose → DeadlineExceeded ✗
    Note over ORCH: primary.promotion failed

    ORCH->>ANA: Analyze(shard)
    ANA-->>ORCH: LeaderIsDead 🔴<br/>(RUNNING, not ready, PromotingPrimaryID=nil)
    Note over ORCH,PG: t = 16 s (13:58:29) — term 12 starts immediately<br/>Cascade locked in. PG still replaying WAL.
```

### Scenario 2 — AZ loss + restore leaves two write-capable primaries (MUL-1004)

The same failover machinery has a second, higher-impact manifestation, and here the stall is **not** WAL replay. In an EKS 3-AZ / `AT_LEAST_2` cluster, losing the primary AZ (cordon + delete pod) promotes a survivor. `Recruit` has already stopped replication and drained WAL replay, so `pg_promote()` completes in well under a second — but it **advances the timeline (TL4 → TL5)**. The new leader then tries to commit the consensus rule that finalizes proposal 4.0; under `AT_LEAST_2` that durable commit blocks in PostgreSQL's `SyncRepWaitForLSN`, waiting for a sync-standby ACK. The only surviving follower is stranded on TL4 and cannot stream from a TL5 leader without a `pg_rewind`, so no ACK ever arrives: the leader's LSN stays **frozen at the fork point (`0/80000A0`)**, proposal 4.0 stays **UNDECIDED**, and it never advertises `rewind_ready`. When the AZ is **restored** (uncordon), the original primary's pod returns and comes back up as a writable primary on its old timeline (TL4). Because the leader never becomes `rewind_ready`, the stale-primary demote **and** the replica `pg_rewind` are deferred for the entire ~900 s restore window, leaving **two locally write-capable primaries** — verified directly: the leader's LSN never moves off `0/80000A0`.

```mermaid
sequenceDiagram
    autonumber
    participant K8s as AZ and Kubernetes
    participant Orch as multiorch
    participant Old as old primary p-22f023e7 TL4
    participant Cand as follower p-f16b048d TL4
    participant Ldr as new leader p-c3032d9b
    Note over Old,Ldr: Steady state. Old is primary on TL4, the others stream from it
    Note over K8s,Ldr: PHASE 1, AZ loss
    K8s->>Old: cordon AZ nodes then delete primary pod
    Old--xOrch: heartbeats stop
    Orch-->>Orch: LeaderIsDead, then appoint leader
    Note over Orch,Ldr: PHASE 2, promotion succeeds but the rule-commit cannot finalize
    Orch->>Cand: Recruit, fence, stop replication, drain WAL replay
    Orch->>Ldr: Recruit, fence, stop replication, drain WAL replay
    Note over Cand,Ldr: WAL is fully replayed during Recruit, so Promote has nothing to replay
    Orch->>Ldr: Promote
    Ldr->>Ldr: pg_promote bumps timeline TL4 to TL5, completes in under 1s
    Note over Ldr: now primary on TL5, LSN at fork point 0/80000A0
    Ldr->>Ldr: async CHECKPOINT on TL5 then MarkSelfRewindReady
    Note over Ldr: rewind_ready is set HERE. It needs a completed TL5 checkpoint while the promote term is still current. The checkpoint never completes, LSN stays frozen at 0/80000A0, so rewind_ready is never set
    Ldr->>Ldr: commit rule for proposal 4.0
    Note over Ldr: AT_LEAST_2 commit blocks in SyncRepWaitForLSN, it needs a sync-standby ACK
    Orch->>Cand: FixReplication SetPrimary, point follower at TL5 then start pg_rewind
    Cand--xOrch: WAL receiver cannot attach to TL5, pg_rewind deferred until leader is rewind_ready
    Note over K8s,Ldr: PHASE 3, AZ restored, split-brain
    K8s->>Old: uncordon, AZ restored
    Note over Old: pod returns, postgres starts as PRIMARY on TL4, writable
    loop every recovery cycle, about 900s
        Orch->>Old: SetPrimary, demote stale leader
        Old--xOrch: leader not yet rewind-ready, deferring demote
        Orch->>Cand: FixReplication, SetPrimary then pg_rewind
        Cand--xOrch: leader not yet rewind-ready, deferring pg_rewind
    end
    Note over Old,Ldr: DUAL PRIMARY. p-22f023e7 on TL4 and p-c3032d9b on TL5 both write-capable for 900s
```

Why the restore window never converges — each recovery step is blocked on the next, so none can go first:

```mermaid
flowchart TD
    A["1. proposal 4.0 durable commit"] -->|needs| B["2. a sync-standby ACK"]
    B -->|needs| C["3. follower streams from leader on TL5"]
    C -->|needs| D["4. pg_rewind onto the leader"]
    D -->|needs| E["5. leader advertises rewind_ready"]
    E -->|needs a settled term| A
```

## Fix

> **`pg_promote()` is a point of no return.** The instant the leader's `Promote` begins, PostgreSQL ends recovery and **irreversibly forks the timeline**: it selects the next timeline id (`findNewestTimeLine() + 1`), writes a timeline history file, and deletes `standby.signal` so the node can never re-enter standby on a restart (`xlog.c` / `timeline.c`). There is no "un-promote." A timeout that abandons a `Promote` *after* this point therefore cannot roll it back — it leaves a committed primary on a fresh timeline. Worse, if the coordinator then appoints a *different* leader, that node forks its **own** timeline and the two diverge from the same parent: neither can stream from the other without a `pg_rewind`, and reconciling them can deadlock (Scenario 2). An abandoned promotion can only rejoin via `pg_rewind`, which **discards its post-fork writes** (risking loss of acknowledged writes) — and if that rewind is gated on the new leader's `rewind_ready`, it never rejoins at all, leaving a standing split-brain. Leader selection is only reversible up to `pg_promote()` — `Recruit` merely fences and stops replication, forking nothing. So the fix's core job is to make a promotion atomic with respect to elections: once `Promote` starts it must never be interrupted, which is why `POSTGRES_STATUS_PROMOTING` is held until the node is ready and re-elections are suppressed for that whole window.

The root fix is in the multipooler: `waitForPromotionComplete` no longer imposes its own deadline. It polls `pg_is_in_recovery()` and `pg_isready` using the caller's context, so `promotionInProgress` stays `true` — and the health stream keeps broadcasting `POSTGRES_STATUS_PROMOTING` — for the entire window from `pg_promote()` through WAL replay until postgres accepts connections. `LeaderIsDeadAnalyzer` suppresses re-elections whenever `PromotingPrimaryID` is set, so the cascade cannot start while WAL replay is in progress.

The second fix is in multiorch: `propose()` no longer wraps the leader's Propose RPC with `RemoteOperationTimeout`. Because gRPC propagates the client deadline to the server context, the old 15 s cap was the actual binding constraint — shorter than even the 30 s inner timeout. Non-leader `SetTermPrimary` calls keep the short deadline since they are fast.

Together, `POSTGRES_STATUS_PROMOTING` is now held until `PostgresReady=true`, and both the multiorch and multipooler contexts allow as much time as the outer `AppointLeaderAction` timeout permits.

```mermaid
sequenceDiagram
    participant ORCH as multiorch
    participant ANA as LeaderIsDeadAnalyzer
    participant MP as leader multipooler
    participant PG as postgres

    Note over ORCH,PG: LeaderIsDead fires — Propose RPC sent with outer ctx (no 15 s cap)

    ORCH->>MP: Propose(term=N) [outer ctx]
    MP->>MP: promotionInProgress = true
    MP-->>ORCH: health: PoolerStatus=PROMOTING, PostgresReady=false
    Note over ANA: PromotingPrimaryID set → suppress ✓
    MP->>PG: SELECT pg_promote()
    PG-->>MP: ok (~3 s)
    Note over MP,PG: WAL replay in progress (seconds to minutes)...

    loop every 100 ms
        ORCH->>ANA: Analyze(shard)
        ANA-->>ORCH: no problem (PROMOTING suppresses) ✓
        MP->>PG: pg_is_in_recovery() / pg_isready
        PG-->>MP: not ready yet
    end

    PG-->>MP: pg_isready = true ✓
    MP->>MP: defer: promotionInProgress = false
    MP-->>ORCH: health: PoolerStatus=RUNNING, PostgresReady=true
    MP-->>ORCH: Propose → success ✅

    ORCH->>ANA: Analyze(shard)
    ANA-->>ORCH: no problem (leader reachable and ready) ✅
```

### Changes

**`go/services/multipooler/internal/manager/manager.go`**

- Removed the hardcoded 30 s `context.WithTimeout` from `waitForPromotionComplete`. The function now uses the caller's context directly so `POSTGRES_STATUS_PROMOTING` is held for the full window — `pg_promote()` call, WAL replay, and `pg_isready`.
- Replaced three OTel spans with `promotion.wal_replay` and `promotion.postgres_ready` eventlog events (Started/Success/Failed) so each phase is observable through the structured event log.

**`go/services/multiorch/consensus/rule_change.go`**

- `propose()` no longer applies `RemoteOperationTimeout` to the leader's Propose RPC. The leader Propose intentionally blocks until postgres is ready; non-leader `SetTermPrimary` calls keep the short deadline since they are fast.

**`go/common/eventlog/events.go`**

- Added `PromotionWalReplay` (`promotion.wal_replay`) and `PromotionPostgresReady` (`promotion.postgres_ready`) event types. Started fires at the beginning of each phase; Success fires on completion; Failed fires on context cancellation or query error.

**`go/cmd/pgctld/testutil/grpc_helpers.go`**

- Added `StatusFunc func(*pb.StatusRequest) (*pb.StatusResponse, error)` to `MockPgCtldService` to allow sequential response control in tests (e.g. returning `Ready=false` N times then `Ready=true`).

## Observability

The two new eventlog event types let operators reconstruct the WAL replay window from structured logs:

| Event type | Outcome | When |
|---|---|---|
| `promotion.wal_replay` | `started` | `waitForPromotionComplete` begins polling `pg_is_in_recovery()` |
| `promotion.wal_replay` | `success` | `pg_is_in_recovery()` first returns `false` |
| `promotion.wal_replay` | `failed` | Context cancelled or query error |
| `promotion.postgres_ready` | `started` | WAL replay done; now polling `pg_isready` |
| `promotion.postgres_ready` | `success` | `pg_isready` returns true |
| `promotion.postgres_ready` | `failed` | Context cancelled |

These pair with the existing `consensus.promote` Started/Success/Failed events (emitted in `rpc_consensus.go`) to give end-to-end visibility across the Propose RPC boundary.

## Tests

| Test | Type | What it verifies |
|------|------|-----------------|
| `TestWaitForPromotionComplete_KeepsPollingUntilPostgresReady` | Unit | Regression: `waitForPromotionComplete` keeps polling when `pg_isready` returns false, catching a reintroduced inner timeout |
| `TestWaitForPromotionComplete_ContextCancellationReturnsError` | Unit | Context cancellation propagates cleanly so `promotionInProgress` is cleared |
| `LeaderIsDeadAnalyzer/suppresses_LeaderIsDead_during_WAL_replay` | Unit | `PromotingPrimaryID` set with `PostgresReady=false` suppresses re-election |
| `TestPromotionElectsExactlyOneLeader` | E2E | After a primary kill, `rule_history` shows exactly one promotion (cascade would show two or more) |
| `TestPromotingStatusClearedOnlyWhenReady` | E2E | `POSTGRES_STATUS_PROMOTING` never clears before `PostgresReady=true` |

Fixes MUL-557
Fixes MUL-1004

### DIFF
--- a/go/services/multipooler/internal/manager/manager.go
+++ b/go/services/multipooler/internal/manager/manager.go
@@ -28,6 +28,7 @@ import (
 	"github.com/multigres/multigres/go/common/backup"
 	commonconsensus "github.com/multigres/multigres/go/common/consensus"
 	"github.com/multigres/multigres/go/common/constants"
+	"github.com/multigres/multigres/go/common/eventlog"
 	"github.com/multigres/multigres/go/common/mterrors"
 	"github.com/multigres/multigres/go/common/pgprotocol/client"
 	"github.com/multigres/multigres/go/common/servenv"
@@ -1343,9 +1344,10 @@ func (pm *MultipoolerManager) promoteStandbyToPrimary(ctx context.Context, state
 		pm.broadcastHealth()
 	}()
 
-	if err := pm.exec(ctx, "SELECT pg_promote()"); err != nil {
-		pm.logger.ErrorContext(ctx, "Failed to call pg_promote()", "error", err)
-		return mterrors.Wrap(err, "failed to promote standby")
+	pgPromoteErr := pm.exec(ctx, "SELECT pg_promote()")
+	if pgPromoteErr != nil {
+		pm.logger.ErrorContext(ctx, "Failed to call pg_promote()", "error", pgPromoteErr)
+		return mterrors.Wrap(pgPromoteErr, "failed to promote standby")
 	}
 
 	// Wait for promotion to complete: pg_is_in_recovery()=false AND postgres_ready=true.
@@ -1470,38 +1472,54 @@ func (pm *MultipoolerManager) dropUnloggedTablesAfterPromotion(ctx context.Conte
 // is accepting connections. Both conditions are required: pg_is_in_recovery()=false
 // confirms the WAL-level promotion, and postgres_ready=true confirms clients can
 // connect. Clearing promotionInProgress only when both are true ensures multiorch's
-// PrimaryIsDeadAnalyzer suppression window matches the full visibility gap.
+// LeaderNeedsReplacementAnalyzer suppression window covers the full gap — including
+// WAL replay which can take several minutes on large databases.
+//
+// The caller's context controls the timeout; no inner deadline is imposed here.
 func (pm *MultipoolerManager) waitForPromotionComplete(ctx context.Context) error {
 	ticker := time.NewTicker(100 * time.Millisecond)
 	defer ticker.Stop()
 
-	promotionTimeout := 30 * time.Second
-	promotionCtx, cancel := context.WithTimeout(ctx, promotionTimeout)
-	defer cancel()
-
-	promotedFromRecovery := false
+	// Phase A: wait for postgres to leave recovery mode (WAL replay).
+	eventlog.Emit(ctx, pm.logger, eventlog.Started, eventlog.PromotionWalReplay{})
 	for {
 		select {
-		case <-promotionCtx.Done():
-			pm.logger.ErrorContext(ctx, "Timeout waiting for promotion to complete")
+		case <-ctx.Done():
+			eventlog.Emit(ctx, pm.logger, eventlog.Failed, eventlog.PromotionWalReplay{}, "error", ctx.Err())
+			pm.logger.ErrorContext(ctx, "Context cancelled waiting for promotion to complete", "error", ctx.Err())
 			return mterrors.New(mtrpcpb.Code_DEADLINE_EXCEEDED,
-				fmt.Sprintf("timeout waiting for promotion to complete after %v", promotionTimeout))
+				fmt.Sprintf("promotion wait cancelled: %v", ctx.Err()))
 
 		case <-ticker.C:
-			if !promotedFromRecovery {
-				pgMode, err := pm.postgresMode(promotionCtx)
-				if err != nil {
-					pm.logger.ErrorContext(ctx, "Failed to check recovery status during promotion", "error", err)
-					return mterrors.Wrap(err, "failed to check recovery status")
-				}
-				if pgMode.OutOfRecovery() {
-					pm.logger.InfoContext(ctx, "Postgres left recovery mode, waiting for connections to be accepted")
-					promotedFromRecovery = true
-				}
+			pgMode, err := pm.postgresMode(ctx)
+			if err != nil {
+				pm.logger.ErrorContext(ctx, "Failed to check recovery status during promotion", "error", err)
+				eventlog.Emit(ctx, pm.logger, eventlog.Failed, eventlog.PromotionWalReplay{}, "error", err)
+				return mterrors.Wrap(err, "failed to check recovery status")
 			}
+			if pgMode.OutOfRecovery() {
+				pm.logger.InfoContext(ctx, "Postgres left recovery mode, waiting for connections to be accepted")
+				eventlog.Emit(ctx, pm.logger, eventlog.Success, eventlog.PromotionWalReplay{})
+				goto waitReady
+			}
+		}
+	}
 
-			if promotedFromRecovery && pm.isPostgresReady(promotionCtx) {
+waitReady:
+	// Phase B: wait for postgres to accept connections.
+	eventlog.Emit(ctx, pm.logger, eventlog.Started, eventlog.PromotionPostgresReady{})
+	for {
+		select {
+		case <-ctx.Done():
+			eventlog.Emit(ctx, pm.logger, eventlog.Failed, eventlog.PromotionPostgresReady{}, "error", ctx.Err())
+			pm.logger.ErrorContext(ctx, "Context cancelled waiting for promotion to complete", "error", ctx.Err())
+			return mterrors.New(mtrpcpb.Code_DEADLINE_EXCEEDED,
+				fmt.Sprintf("promotion wait cancelled: %v", ctx.Err()))
+
+		case <-ticker.C:
+			if pm.isPostgresReady(ctx) {
 				pm.logger.InfoContext(ctx, "Promotion completed successfully - node is now primary and accepting connections")
+				eventlog.Emit(ctx, pm.logger, eventlog.Success, eventlog.PromotionPostgresReady{})
 				return nil
 			}
 		}
--- a/go/cmd/pgctld/testutil/grpc_helpers.go
+++ b/go/cmd/pgctld/testutil/grpc_helpers.go
@@ -64,6 +64,12 @@ type MockPgCtldService struct {
 	InitDirError            error
 	PgRewindError           error
 	StopRestoreCommandError error
+
+	// StatusFunc, if non-nil, is called for each Status request instead of the
+	// default behaviour. It lets tests return different values on successive
+	// calls, e.g. to simulate postgres transitioning from not-ready to ready
+	// during WAL replay. Called while the mutex is held.
+	StatusFunc func(*pb.StatusRequest) (*pb.StatusResponse, error)
 }
 
 func (m *MockPgCtldService) Start(ctx context.Context, req *pb.StartRequest) (*pb.StartResponse, error) {
@@ -137,6 +143,9 @@ func (m *MockPgCtldService) Status(ctx context.Context, req *pb.StatusRequest) (
 	m.StatusCalls = append(m.StatusCalls, req)
 	if m.StatusError != nil {
 		return nil, m.StatusError
+	}
+	if m.StatusFunc != nil {
+		return m.StatusFunc(req)
 	}
 	if m.StatusResponse != nil {
 		return m.StatusResponse, nil
--- a/go/common/eventlog/events.go
+++ b/go/common/eventlog/events.go
@@ -162,3 +162,22 @@ func (ConsensusSetPrimary) EventType() string { return "consensus.set_primary" }
 func (e ConsensusSetPrimary) LogAttrs() []slog.Attr {
 	return []slog.Attr{slog.String("rule", e.Rule)}
 }
+
+// PromotionWalReplay is emitted by the multipooler leader node while it waits
+// for postgres to leave recovery mode (complete WAL replay) after pg_promote().
+// Started fires when the polling loop begins; Success fires when
+// pg_is_in_recovery() first returns false; Failed fires on context cancellation
+// or a query error.
+type PromotionWalReplay struct{}
+
+func (PromotionWalReplay) EventType() string     { return "promotion.wal_replay" }
+func (PromotionWalReplay) LogAttrs() []slog.Attr { return nil }
+
+// PromotionPostgresReady is emitted by the multipooler leader node while it
+// waits for postgres to start accepting connections after WAL replay completes.
+// Started fires when pg_is_in_recovery() first returns false; Success fires
+// when pg_isready succeeds; Failed fires on context cancellation.
+type PromotionPostgresReady struct{}
+
+func (PromotionPostgresReady) EventType() string     { return "promotion.postgres_ready" }
+func (PromotionPostgresReady) LogAttrs() []slog.Attr { return nil }
--- a/go/pb/clustermetadata/clustermetadata.pb.go
+++ b/go/pb/clustermetadata/clustermetadata.pb.go
@@ -21,12 +21,13 @@
 package clustermetadata
 
 import (
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
-	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
 	reflect "reflect"
 	sync "sync"
 	unsafe "unsafe"
+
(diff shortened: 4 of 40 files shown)


dweitzman:
> After a primary failure, the newly promoted postgres can spend minutes replaying WAL before it starts accepting connections

If that's true, there's a bug at a lower level?... `Recruit()` waits for all received WAL to replay (or "stabilize" as the code says) so there shouldn't be any replaying necessary while promoting a primary

mkindahl:
> > After a primary failure, the newly promoted postgres can spend minutes replaying WAL before it starts accepting connections
> 
> If that's true, there's a bug at a lower level?... `Recruit()` waits for all received WAL to replay (or "stabilize" as the code says) so there shouldn't be any replaying necessary while promoting a primary

There is a partial problem in the `Recruit()` phase as well, but I would like to investigate this separately. The problem here is that the `Recruit()` waits for the LSN to stop incrementing: it does not wait for the WAL to be completely replayed. If there is a pause in I/O read from disk, LSN will stop incrementing and `Recruit()` will assume that WAL is fully replayed. The I/O gap can be as a result of I/O throttling at disk level, a network problem for network mounted disks, or something else. We really need to find a better method to decide if WAL is fully replayed (see MUL-645).

Since we can have bugs, however, we need to deal with the case that `Recruit()` fails to wait until the WAL is replayed, which makes this PR relevant as a defensive measure.

The particular scenario we have here is that the cluster is scaled from 3 to 12 nodes and the backup is old. That forces the new nodes to replay WAL, and while they are replaying, the primary crashes and there are no other candidate nodes than the ones that are replaying WAL.

As a result, the best candidate is picked based on the `receive_lsn`, but that also means that it still has work to replay before being ready to work. Timing out the promotion at this point to pick another primary will either pick the same candidate again, or pick another one that also has WAL to replay, so it is just a waste of time.

The best we can do in this situation is to wait for the promotion to finish. Note that this is not the common case, it is a rare case and should not normally trigger.

An argument is that we should make a selection based on which node has the smallest amount to replay, with the reasoning that it will be ready faster, but this do not make this PR unnecessary since the situation can still occur and we still need to wait for the promotion to finish.

go/services/multipooler/internal/manager/manager.go
```diff
@@ -1470,38 +1472,54 @@ func (pm *MultipoolerManager) dropUnloggedTablesAfterPromotion(ctx context.Conte
...
+			pgMode, err := pm.postgresMode(ctx)
+			if err != nil {
+				pm.logger.ErrorContext(ctx, "Failed to check recovery status during promotion", "error", err)
+				eventlog.Emit(ctx, pm.logger, eventlog.Failed, eventlog.PromotionWalReplay{}, "error", err)
+				return mterrors.Wrap(err, "failed to check recovery status")
+			}
+			if pgMode.OutOfRecovery() {
+				pm.logger.InfoContext(ctx, "Postgres left recovery mode, waiting for connections to be accepted")
+				eventlog.Emit(ctx, pm.logger, eventlog.Success, eventlog.PromotionWalReplay{})
+				goto waitReady
```
dweitzman, line 1503: Maybe break these into subroutines instead of using a `goto`?

`.waitUntilOutOutRecovery()` and `.waitUntilAcceptingConnections()` or something like that?

dweitzman:
A few disconnected thoughts:
* Somewhat related to all this, I think rule_store should probably actually report the replay LSN, not the receive LSN, since the replay LSN is what corresponds to the rule it read
* I haven't been able to reproduce a scenario locally where waiting for WAL replay to stabilize didn't work. Do you have a repro or a failure on a recent version of the code, by any chance? I was trying to write an e2e test that would trigger logic to detect problems (draft at https://github.com/multigres/multigres/pull/1268) but it didn't seem to be reproducing for me in my first few attempts to write a test.
* it's still not obvious to me why pg_promote() would take more than 30 seconds. Even if waiting for wal replay to stabilize is imperfect, it seems like it shouldn't be normal for there to be so much unreplayed WAL that it takes 30+ seconds to replay it. If that's the case, it doesn't feel productive to patch around it because at that point the consensus algorithm is broken

mkindahl:
Closing in favor of #1287, #1303, #1307, and #1266

PR #1069, opened 2026-06-01
fix(multipooler): restore primary_conninfo after pg_rewind in RewindToSource

mkindahl:
pg_rewind syncs postgresql.auto.conf from the source server, which does not have primary_conninfo set (it is the primary). After the actual rewind runs, the replica restarts as standby but the WAL receiver has no primary to connect to, leaving it permanently idle.

Fix: call setPrimaryConnInfoLocked after the restart in RewindToSource, matching what demoteStalePrimaryLocked already does. This re-establishes primary_conninfo and triggers pg_reload_conf so the WAL receiver connects immediately.

Observed in mg-scale12 AZ-outage test (2026-05-29): newly added replicas that entered the pg_rewind path via fixNotReplicating ended up with no streaming WAL receiver after restart.

Also add two regression tests:
- TestRewindToSource_RestoresPrimaryConnInfo: verifies setPrimaryConnInfoLocked is called with the source host/port after a successful rewind.
- TestPgCtldService_StopRewindStart: verifies the pgctld stop→pg_rewind→start sequence completes without error across all three RPC handlers.

Fixes MUL-559

### DIFF
--- a/go/cmd/pgctld/command/server_test.go
+++ b/go/cmd/pgctld/command/server_test.go
@@ -642,6 +642,72 @@ func TestPgCtldService_Status_IncludesPgBackRest(t *testing.T) {
 	assert.Equal(t, int32(5), resp.PgbackrestStatus.RestartCount)
 }
 
+// TestPgCtldService_StopRewindStart verifies the stop→pg_rewind→start lifecycle.
+// This is a regression test for a bug where the pgctld service would fail to start
+// postgres after a stop+pg_rewind sequence. The fix ensures that the pgctld RPC
+// handlers have no internal "closed" state that prevents re-starting postgres after
+// a stop-for-rewind flow.
+func TestPgCtldService_StopRewindStart(t *testing.T) {
+	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelWarn}))
+
+	baseDir := t.TempDir()
+	poolerDir := baseDir
+	dataDir := filepath.Join(baseDir, "pg_data")
+
+	// Set env vars expected by pgctld
+	t.Setenv(constants.PgDataDirEnvVar, dataDir)
+
+	// Create mock PostgreSQL binaries (including pg_rewind)
+	binDir := filepath.Join(baseDir, "bin")
+	require.NoError(t, os.MkdirAll(binDir, 0o755))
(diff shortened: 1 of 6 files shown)


go/services/multipooler/manager/rpc_manager.go
```diff
@@ -970,6 +970,17 @@ func (pm *MultiPoolerManager) RewindToSource(ctx context.Context, source *cluste
 		return nil, mterrors.Wrap(err, "failed to reconnect to database after pg_rewind")
 	}
 
+	// Re-establish primary_conninfo. pg_rewind syncs files from the source including
+	// postgresql.auto.conf, which does not have primary_conninfo set on the primary.
+	// Without this, the WAL receiver has no primary to connect to after restart.
+	pm.logger.InfoContext(ctx, "Re-establishing primary_conninfo after pg_rewind",
+		"source_host", source.Hostname,
+		"source_port", port)
+	if err := pm.setPrimaryConnInfoLocked(ctx, source.Hostname, port, false, false); err != nil {
```
mkindahl, line 979: This is the key change. The rest are just tests to make sure that it works.

go/cmd/pgctld/command/server_test.go
```diff
@@ -642,6 +642,72 @@ func TestPgCtldService_Status_IncludesPgBackRest(t *testing.T) {
...
+		SourceHost: "127.0.0.1",
+		SourcePort: 5433,
+		DryRun:     true,
+	})
+	require.NoError(t, err, "PgRewind must succeed after Stop")
+	require.NotNil(t, rewindResp)
+
+	// Step 4: Start postgres again as standby. This is the critical step: the service
+	// must be able to restart after stop+rewind without requiring a pgctld process restart.
+	restartResp, err := service.Restart(ctx, &pb.RestartRequest{
```
cuongdo, line 698: After this Restart, is there value in verifying that streaming replication is working?

cuongdo (approved):
lgtm, with one comment

go/cmd/pgctld/command/server_test.go
mkindahl: For this particular test, we just verify that the server can be restarted after running PgRewind. To test streaming replication, we would have to configure and start it, which is out of scope for this test.

This is, however, something we should test so I added a test for the full sequence from stopping, through rewind, and back up to replication to make sure that we have that path covered.

PR #1081, opened 2026-06-02
feat(ci): Integrate Trunk Flaky Tests

mmatheson:
Adds a shared composite action (.github/actions/upload-test-results) that uploads CI test results to Trunk Flaky Tests, authenticating with the non-secret public repo identifier so it works without repository secrets to support PRs from forks.


rafael:
I haven't gotten a chance to review. Small detail - could you fix the DCO by signing the commit?

.github/actions/upload-test-results/action.yaml
```diff
@@ -0,0 +1,83 @@
...
+      Whether to allow Trunk to quarantine failing tests. When 'true', the
+      uploader step exits non-zero only if there are non-quarantined failures,
+      so it can gate the job in place of the (continue-on-error) test step. The
+      quarantine decision itself is Trunk's; this only permits it.
+    required: false
+    default: "false"
+  org-slug:
+    description: Trunk organization slug. Non-secret; safe to commit.
+    required: false
+    default: supabase
```
rafael, line 56: Where do we get this and the repo-id identifier? I don't have this set yet the multigres trunk account and I don't seem to have access to the `supabase` org.

.github/workflows/test-short.yml
```diff
@@ -47,6 +47,10 @@ jobs:
         run: go install gotest.tools/gotestsum@v1.13.0
 
       - name: Run short tests
+        id: short-tests
+        # continue-on-error lets the Trunk upload step below gate the job
+        # instead: it passes when every failure is quarantined, fails otherwise.
+        continue-on-error: true
         run: |
           gotestsum \
```
rafael, line 55: One flag here: to stop the bleeding with flakes, we introduced an initial flake-detection mechanism in: https://github.com/multigres/multigres/pull/937.

The way we run `gotestsum` here dedups the failures, so the output the uploader gets won't include them, it'll be the last clean successful run after the retries.

We could iterate on this in a separate PR. One quick idea is a new action specific to Trunk.

.github/actions/upload-test-results/action.yaml
mmatheson: I sent you the invite, you should have access at https://app.trunk.io/supabase/flaky-tests/repo/46ad0a36-669e-4212-ac23-3c321807a74e

.github/workflows/test-short.yml
mmatheson: Since gotestsum includes all runs in the test report, using --rerun fails will help us detect flaky tests - both the fail and the pass will be reported and analyzed.

So, as-is this should still work for flake detection on our end. Once we have validated that and set up quarantining, you should be able to deprecate the rerun and analysis work from #937

rafael: ah perfect. I thought that was not in the report.

rafael (approved):
LGTM

PR #1082, opened 2026-06-02
fix(multipooler): increase promotion timeout and add OTel spans to election flow

mkindahl:
## Summary

- Introduces `RuleWriteTimeout = 30 s` and applies it to the three election-flow RPCs — Recruit, Propose, and SetTermPrimary — and to the `rule_history` postgres write inside `updateRule` for all callers. For promotions the ack arrives after the full SetTermPrimary round-trip; for primary-side cohort changes standbys are already streaming and the ack is nearly immediate either way.
- Action timeouts for `AppointLeader` and `ShardInit` are computed from the constant (`2×RuleWriteTimeout + 5 s = 65 s`) so they automatically track any future adjustment.
- Adds `DoUpdateRule` on `MultiPoolerManager`, which wraps `updateRule` with a `consensus/rule-write` OTel span. All callers (`Propose`, `UpdateConsensusRule`) go through `DoUpdateRule` so every rule write is traced. Two further spans cover the promotion path: `consensus/pg-promote` and `consensus/standby-ack`.
- Adds `consensus/recruit` on the pooler side with three sub-spans: `consensus/stop-replication` (pauseReplication / emergencyDemoteLocked), `consensus/stabilize` (waitForReplayStabilize), and `consensus/accept-revocation` (AcceptRevocation). Together with the coordinator-side `consensus/recruit`, this makes the full pooler execution time visible in addition to the round-trip time.
- All pooler spans correlate with coordinator spans via gRPC trace propagation already wired in `grpccommon.NewClient`.

## Background

The `rule_history` write in `updateRule` uses `synchronous_commit = on`, which blocks until at least one streaming standby sends a WAL acknowledgment. The write is the durability gate: it proves the new consensus rule has reached a quorum of servers before the election is declared complete.

The problem is timing. The Recruit round runs before Propose, and it intentionally clears replication on every cohort member — standbys call `pauseReplication()` to stop streaming, the old primary is demoted to a standalone standby. By the time `Propose` runs and `pg_promote()` completes, **zero standbys are streaming**. `SetTermPrimary` — which tells standbys to reconnect — is dispatched concurrently with `Propose`, so the ack can only arrive after the full `SetTermPrimary` round-trip, which includes optional `pg_rewind` if a standby has diverged.

In the 9-pooler AZ-freeze test this took **8–14 seconds** against a 15 s `RemoteOperationTimeout`. One promotion timed out. Any scenario where standby reconnection is slower — `pg_rewind`, pod restart, cross-AZ latency under load — exceeds the old budget and fails the election.

Because the timeout is a deadline, not a sleep, increasing it does not slow promotions: postgres returns immediately when an ack arrives.

## Sequence of events

```mermaid
sequenceDiagram
    participant C as Coordinator
    participant L as Leader Candidate
    participant Lpg as Leader postgres
    participant S as Standbys

    Note over C,S: Recruit phase
    par
        C->>L: Recruit(term=N)
        L->>L: pauseReplication()<br/>clear primary_conninfo, stop streaming
        L-->>C: RecruitResponse(WAL pos)
    and
        C->>S: Recruit(term=N)
        S->>S: pauseReplication()<br/>clear primary_conninfo, stop streaming
        S-->>C: RecruitResponse(WAL pos)
    end

    Note over C: quorum assembled — build proposal

    Note over C,S: Propose + SetTermPrimary dispatched concurrently
    par
        C->>L: Propose(term=N, leader=L)
        L->>Lpg: pg_promote()
        Lpg-->>L: postgres is primary ✓
        Note over L,Lpg: INSERT rule_history<br/>synchronous_commit = on<br/>BLOCKS — zero standbys streaming<br/>waiting for WAL ack
        Lpg-->>S: WAL record (streaming replication)
        S-->>Lpg: WAL ack ✓
        Note over L,Lpg: write unblocks (8–14 s in logs)
        L-->>C: ProposeResponse ✓
    and
        C->>S: SetTermPrimary(leader=L)
        S->>S: set primary_conninfo → L
        S->>Lpg: connect replication stream
        S-->>C: SetTermPrimaryResponse ✓
    end
```

## Changes

| File | Change |
|---|---|
| `go/common/timeouts/rpc.go` | Add `RuleWriteTimeout = 30 s` |
| `go/services/multiorch/consensus/rule_change.go` | Use `RuleWriteTimeout` for Recruit, Propose, SetTermPrimary |
| `go/services/multiorch/recovery/actions/appoint_leader.go` | Action timeout = `2×RuleWriteTimeout + 5 s` |
| `go/services/multiorch/recovery/actions/shard_init_action.go` | Action timeout = `2×RuleWriteTimeout + 5 s` |
| `go/services/multipooler/manager/manager.go` | Add `DoUpdateRule` with `consensus/rule-write` span; add `consensus/pg-promote` span |
| `go/services/multipooler/manager/rpc_consensus.go` | Use `DoUpdateRule` in `Propose`; add `consensus/recruit` span with `consensus/stop-replication`, `consensus/stabilize`, `consensus/accept-revocation` sub-spans |
| `go/services/multipooler/manager/rpc_manager.go` | Use `DoUpdateRule` in `UpdateConsensusRule` handler |
| `go/services/multipooler/manager/rule_store.go` | Use `RuleWriteTimeout` for all rule writes; add `consensus/standby-ack` span |

## Test plan

- [x] Unit tests: `go test -short ./go/common/timeouts/... ./go/services/multiorch/... ./go/services/multipooler/...`
- [ ] Integration: `multiorch` bootstrap and failover tests pass with the new timeouts
- [ ] Verify in Tempo that `consensus/pg-promote`, `consensus/rule-write`, and `consensus/standby-ack` appear as children of the pooler's `consensus/propose` span during a promotion
- [ ] Verify that `consensus/stop-replication`, `consensus/stabilize`, and `consensus/accept-revocation` appear as children of the pooler's `consensus/recruit` span
- [ ] Confirm `consensus/standby-ack` span duration matches the 8–14 s observed in the AZ-freeze log analysis

Fixes MUL-554

### DIFF
--- a/go/common/timeouts/rpc.go
+++ b/go/common/timeouts/rpc.go
@@ -21,6 +21,13 @@ import "time"
 // RPC calls, etcd data fetches, and synchronous replication health checks.
 const RemoteOperationTimeout = 15 * time.Second
 
+// RuleWriteTimeout is the timeout for rule writes and the election-flow RPCs
+// (Recruit, Propose, SetTermPrimary). For promotions the rule_history write
+// blocks until a sync-standby WAL ack arrives after the full SetTermPrimary
+// round-trip, which includes optional pg_rewind. 60 s gives ~45 s of headroom
+// beyond the typical 8–14 s observed in the 9-pooler AZ-freeze test.
+const RuleWriteTimeout = 30 * time.Second
+
 // DefaultHealthStreamStalenessTimeout is the default staleness watchdog timeout for
 // ManagerHealthStream connections. If no message is received within this window
 // the client reconnects and the server advertises this value in the start response.
--- a/go/services/multiorch/consensus/rule_change.go
+++ b/go/services/multiorch/consensus/rule_change.go
@@ -251,7 +251,7 @@ func (r *coordinatorLedRuleChange) recruit(
 	p *multiorchdatapb.PoolerHealthState,
 	revocation *clustermetadatapb.TermRevocation,
 ) *clustermetadatapb.ConsensusStatus {
-	rpcCtx, cancel := context.WithTimeout(ctx, timeouts.RemoteOperationTimeout)
+	rpcCtx, cancel := context.WithTimeout(ctx, timeouts.RuleWriteTimeout)
 	defer cancel()
 	resp, err := r.coordinator.rpcClient.Recruit(rpcCtx, p.MultiPooler, &consensusdatapb.RecruitRequest{
 		TermRevocation: revocation,
@@ -286,7 +286,7 @@ func (r *coordinatorLedRuleChange) propose(
 	req *consensusdatapb.ProposeRequest,
 	isLeader bool,
 ) error {
-	rpcCtx, cancel := context.WithTimeout(ctx, timeouts.RemoteOperationTimeout)
+	rpcCtx, cancel := context.WithTimeout(ctx, timeouts.RuleWriteTimeout)
 	defer cancel()
 	if isLeader {
 		_, err := r.coordinator.rpcClient.Propose(rpcCtx, p.MultiPooler, req)
--- a/go/services/multiorch/recovery/actions/appoint_leader.go
+++ b/go/services/multiorch/recovery/actions/appoint_leader.go
@@ -23,6 +23,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"github.com/multigres/multigres/go/common/mterrors"
+	"github.com/multigres/multigres/go/common/timeouts"
 	"github.com/multigres/multigres/go/common/topoclient"
 	commontypes "github.com/multigres/multigres/go/common/types"
 	"github.com/multigres/multigres/go/services/multiorch/config"
@@ -154,7 +155,10 @@ func (a *AppointLeaderAction) Metadata() types.RecoveryMetadata {
 	return types.RecoveryMetadata{
 		Name:        "AppointLeader",
 		Description: "Elect a new primary for the shard using consensus",
-		Timeout:     60 * time.Second,
+		// Two sequential phases each bounded by RuleWriteTimeout
+		// (Recruit, then concurrent Propose/SetTermPrimary), plus margin so the
+		// action context does not race its own phases to the deadline.
+		Timeout:     2*timeouts.RuleWriteTimeout + 5*time.Second,
 		LockTimeout: 15 * time.Second,
 		Retryable:   true, // can retry if it fails
 	}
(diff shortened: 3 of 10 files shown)


go/common/timeouts/rpc.go
```diff
@@ -21,6 +21,13 @@ import "time"
 // RPC calls, etcd data fetches, and synchronous replication health checks.
 const RemoteOperationTimeout = 15 * time.Second
 
+// RuleWriteTimeout is the timeout for rule writes and the election-flow RPCs
+// (Recruit, Propose, SetTermPrimary). For promotions the rule_history write
+// blocks until a sync-standby WAL ack arrives after the full SetTermPrimary
+// round-trip, which includes optional pg_rewind. 60 s gives ~45 s of headroom
+// beyond the typical 8–14 s observed in the 9-pooler AZ-freeze test.
+const RuleWriteTimeout = 60 * time.Second
+
```
GuptaManan100: Raising this from 15s → 60s means that when sync replication is genuinely broken (no standby will ever ack), each attempt now blocks a full 60s before failing instead of 15s - ~4× longer worst-case stall on a dead cohort.

Just pointing it out.

GuptaManan100 (approved):
The test plan says the new spans appear "as children of the pooler's `consensus/propose` span", but there is no `consensus/propose` span in the `Propose` handler. The actual parent is the auto-instrumented gRPC server span — `consensus/rule-write` is the child, and `consensus/pg-promote` / `consensus/standby-ack` are its grandchildren. Worth fixing the Tempo verification steps so they don't send the reader looking for a span that doesn't exist.

go/common/timeouts/rpc.go
mkindahl: Good point.

Doing some back-of-the-envelop calculations, the logs show 8-14s with occasional overruns, so if we assume that the mean is 11s and the standard deviation is 3 seconds (which would cover ~67% of the cases), we should be able to cover 99.9% of the cases with a little more than three standard deviations, meaning that 20-30s should be sufficient.

Updating it to 30s and keeping fingers crossed. We have OTel after this, so we should be able to start measuring.

mkindahl: Also updating the AppointLeader timeout margin to 5 seconds.

dweitzman:
In my local testing, a complete coordinator-led rule change takes under 1 second (with a few minor fixes). Seems like something else is wrong if it's taking >10 seconds in the absence of severe network lag?...

mkindahl:
> In my local testing, a complete coordinator-led rule change takes under 1 second (with a few minor fixes). Seems like something else is wrong if it's taking >10 seconds in the absence of severe network lag?...

In the logs they all show taking 8-14 seconds. Not surprised that you can get it down locally, but we don't have good figures yet, so we need to get metrics to understand what the actual time is and adapt accordingly.

PR #1086, opened 2026-06-03
feat(pgregresstest): run pgvector's regression suite through multigateway

GuptaManan100:
## Summary

Adds an external-extension suite to `pgregresstest`: clones an out-of-tree PGXS extension, builds it against the from-source PostgreSQL, and runs its pg_regress suite through multigateway. pgvector v0.8.1 is the first; enrolling another is a 2-line catalog edit (`externalSpecs` + flip the row to covered). Runs under `RUN_EXTENDED_QUERY_SERVING_TESTS`, so CI picks it up with no workflow changes.

## Result: 11/14 pgvector tests pass

The 3 failures share one root cause this surfaced — a multigateway bug (#1090): **`SET` is applied to local session state and not validated against Postgres until the next query** (`apply_session_state.go`). So `SET hnsw.ef_search = 1001` doesn't raise its range error at SET time; it surfaces later on an unrelated statement. That fails `hnsw_vector` and `ivfflat_vector` directly, and `ivfflat_bit` downstream — the deferred error is flushed onto a later `DROP TABLE`, aborting it and leaking a table into the next test.

Fix tracked in #1090; the 3 tests stay red until then (no divergence patches committed to mask them).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ export PGPROTO_VER
 CMDS = multigateway multipooler pgctld multiorch multigres multiadmin portpoolserver
 BIN_DIR = bin
 
-.PHONY: all build build-all clean images install test test-coverage pgregress pgregress-update-patches pgproto pgproto-update-patches proto tools parser help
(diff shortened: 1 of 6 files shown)


haritabh17 (approved):
🚀

PR #1087, opened 2026-06-03
fix(multigateway): handle DISCARD ALL at the gateway instead of forwarding it

GuptaManan100:
## Summary

- Handle `DISCARD ALL` entirely at the gateway and stop forwarding it to a shared pooled backend, which fixes prepared-statement tracking going stale after `DISCARD ALL`.
- Import a small suite of pooling/resilience tests ported from PgBouncer's `test_prepared.py`, run them in normal integration CI (no longer behind the extended-tests gate).

## Problem

`DISCARD ALL` was planned as `CloseCursorRoute(CloseAll)`, which forwarded the literal statement to one pooled backend. Since PG's `DISCARD ALL` runs `DEALLOCATE ALL`, that backend silently dropped its prepared statements while neither the gateway's consolidator nor the multipooler's per-connection tracking were told. Result: re-`PREPARE` of the same name failed with "already exists", and the desynced backend tracking poisoned the pooled connection for later sessions (`ensurePrepared` would skip the `Parse`, so `Bind` failed). A reserved backend was also left with stale temp-table / HOLD-cursor reasons.

## Solution

A new gateway-local `DiscardAllPrimitive` reproduces PG's `DISCARD ALL` against the virtual session state instead of a backend: it rejects the statement inside a transaction block (SQLSTATE 25001), clears the prepared-statement consolidator (`DEALLOCATE ALL`), resets session GUCs (`RESET ALL`), drops LISTEN subscriptions (`UNLISTEN *`), and releases any reserved connection back to the pool via the existing `ReleaseReservedConnection` RPC (which rolls back, discards temp tables, and unpins cursors). Prepared statements already parsed on pooled backends are intentionally left in place — they are a pool-wide dedup cache, and not running `DEALLOCATE ALL` on a shared backend is exactly what keeps the multipooler's tracking accurate.

Verified against PostgreSQL 17 as the oracle: `TestPreparedStatementResetSemantics` now passes through the gateway for both `DEALLOCATE ALL` and `DISCARD ALL` (EXECUTE fails + re-PREPARE succeeds), alongside the prepared-statement swap and large-message relay tests.

### DIFF
--- a/go/services/multigateway/engine/copy_statement_test.go
+++ b/go/services/multigateway/engine/copy_statement_test.go
@@ -55,6 +55,10 @@ type mockIExecute struct {
 	copyAbortCalled atomic.Int32
 	copyAbortErr    error
 
+	// ReleaseAllReservedConnections behavior
+	releaseAllCalled atomic.Int32
+	releaseAllErr    error
+
 	// CopyOutInitiate behavior (TO STDOUT direction)
 	copyOutInitiateErr     error
 	copyOutInitiateFormat  int16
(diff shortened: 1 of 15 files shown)


haritabh17 (approved):
🚀

PR #1088, opened 2026-06-03
feat(multigateway): block user changes to synchronous_commit

GuptaManan100:
## Summary

- Block users from assigning a value to cluster-managed GUCs through the gateway. The set is a small `restrictedGUCs` map (currently just `synchronous_commit`); the guard covers `SET` / `SET LOCAL` / `SET … FROM CURRENT`, `ALTER DATABASE … SET`, `ALTER ROLE … SET`, and `set_config(...)` (both `is_local` variants). `ALTER SYSTEM` was already blocked as a Tier 2 statement.
- Reverts stay allowed (`RESET`, `SET … TO DEFAULT`, `RESET ALL`) since they can only restore the managed value.
- `synchronous_commit` is the first entry: durability is owned by the multipooler rule store (the sole writer, pinned to `on`), so a session lowering it silently weakens the durability guarantee — a footgun we take away rather than a load-bearing API break. Rejected at plan time with `feature_not_supported` (`0A000`), pointing users at `RESET`.

## Description

The guard lives in the multigateway planner (`checkRestrictedGUCChange` for the SET/ALTER paths, plus a `restrictedGUCError` check in the existing `set_config` expression walker), wired into `planUnsupportedConstructs` so both the simple and extended query protocols are covered. The normalizer keeps the `set_config` name literal (parameterizing only the value) so the guard still applies on the cached/normalized path; this costs at most one cache entry per distinct GUC name since only values churn. To restrict another GUC, add a line to the `restrictedGUCs` map. Includes docs and a decision log; may later be reopened behind an explicit opt-in (e.g. batching transactions async, then `on` for the final commit).


haritabh17 (approved):
🚀

PR #1089, opened 2026-06-03
fix(consensus): prefer postgres-ready poolers during leader election

mkindahl:
## Problem

During a failover, candidate selection picks the node with the highest WAL position. A freshly recreated pod has its PVC data intact (valid LSN) but postgres has not yet finished initialising. When that pod wins the election, the `Propose` fails immediately because the promotion hook cannot connect to the postgres Unix socket:

```
propose failed: could not write rule: promotion hook: reconnect dial failed:
failed to connect to Unix socket .s.PGSQL.5432: no such file or directory
```

Without this fix, the election fails and stalls until the next analysis cycle. With this fix, a ready pod is preferred when available, avoiding the failure. If no pod is ready, the election still fails and retries in the next cycle.

### Without this fix

```mermaid
sequenceDiagram
    participant orch as multiorch
    box old primary pod
    participant old_pool as multipooler
    participant old_pg as postgres
    end
    box new pod
    participant new_pool as multipooler
    participant new_pg as postgres
    end
    box standby pod
    participant stby_pool as multipooler
    participant stby_pg as postgres
    end

    activate old_pg
    activate stby_pg

    Note over old_pg,stby_pg: Normal operation — all postgres processes running

    old_pool-->>orch: health (READY)
    stby_pool-->>orch: health (READY)

    Note over old_pool,old_pg: pod deleted / AZ outage
    deactivate old_pg

    activate new_pool
    Note over new_pg: pod recreated, PVC intact<br/>crash recovery running
    activate new_pg

    new_pool-->>orch: health (STARTING — socket not open)
    stby_pool-->>orch: health (READY)

    Note over orch: LeaderIsDeadAnalyzer fires<br/>→ runFailover()

    orch->>new_pool: Recruit
    new_pool-->>orch: ConsensusStatus {lsn: A/BC}

    orch->>stby_pool: Recruit
    stby_pool-->>orch: ConsensusStatus {lsn: A/BC}

    Note over orch: LSNs tied (A/BC = A/BC)<br/>no readiness tiebreaker<br/>→ new pod elected

    orch->>new_pool: Promote
    new_pool-xnew_pg: pg_promote() — socket not open
    new_pool-->>orch: error: dial .s.PGSQL.5432: no such file or directory

    Note over orch: Propose fails — stalls ~30 s<br/>until next analysis cycle
```

### With this fix — STARTING, not INELIGIBLE

A natural first instinct is to mark a STARTING pod as `INELIGIBLE` and filter it out of the candidate set entirely. That would be unsafe. Because all pods use PVCs, the STARTING pod has flushed every synchronous write up to its crash — under `AT_LEAST_2` the STARTING pod may be the only node that acknowledged the most recent transaction, with the READY standby still catching up. Filtering it out elects the READY standby at a lower LSN and loses that acknowledged write.

The correct fix keeps the STARTING node in the candidate set and uses its LSN normally, only deprioritising it as a tiebreaker when another node is tied on LSN and READY.

```mermaid
sequenceDiagram
    participant client
    participant primary as postgres (primary)
    participant stby_a as postgres A (STARTING)
    participant stby_b as postgres B (READY)
    participant orch as multiorch

    Note over primary,stby_b: T1 — both standbys acknowledge (AT_LEAST_2 satisfied)
    client->>primary: INSERT T1
    primary-->>stby_a: WAL → LSN B
    primary-->>stby_b: WAL → LSN B
    stby_a-->>primary: flush ACK
    stby_b-->>primary: flush ACK
    primary-->>client: OK — T1 durable at LSN B

    Note over primary,stby_b: T2 — standby B lags, only standby A flushes
    client->>primary: INSERT T2
    primary-->>stby_a: WAL → LSN C
    primary-->>stby_b: WAL → LSN C (in-flight — not yet flushed)
    stby_a-->>primary: flush ACK — AT_LEAST_2 satisfied
    primary-->>client: OK — T2 durable at LSN C

    Note over primary: primary crashes

    Note over stby_a: STARTING — crash recovery<br/>LSN C on PVC
    Note over stby_b: READY — LSN B only<br/>(never received T2)

    alt STARTING treated as INELIGIBLE — broken
        orch->>stby_b: Recruit (only B eligible — A filtered out)
        stby_b-->>orch: ConsensusStatus lsn=B
        orch->>stby_b: Promote (LSN B elected)
        stby_b-->>orch: OK — new leader at LSN B
        rect rgb(255, 150, 150)
            Note over stby_a,stby_b: T2 lost — client was told OK<br/>but data is gone (durability violated)
        end
    else STARTING kept as deprioritised candidate — correct
        orch->>stby_a: Recruit
        stby_a-->>orch: ConsensusStatus lsn=C
        orch->>stby_b: Recruit
        stby_b-->>orch: ConsensusStatus lsn=B
        Note over orch: LSN C > B — standby A wins<br/>despite STARTING signal
        orch->>stby_a: Promote (LSN C)
        stby_a-->>orch: OK — T2 preserved
        rect rgb(150, 220, 150)
            Note over stby_a,stby_b: T2 preserved — durability maintained
        end
    end
```

### Pooler: publish `LeadershipAvailability` signal

A new `LeadershipAvailability` field (field 3 on `AvailabilityStatus`) carries a `READY` or `STARTING` signal. The postgres monitor already polls pgctld on every tick; a new `postgresReadyForCohort atomic.Bool` stores whether postgres is running and `pg_isready` passes. `buildLeadershipAvailability()` maps this to:

- `READY` — postgres is accepting connections, safe to promote
- `STARTING` — postgres is still initialising (crash recovery, socket not yet open)

A human-readable `reason` string is included for debugging. The signal flows via the health stream to `PoolerHealthState.AvailabilityStatus` in the orchestrator.

`buildCohortEligibilityStatus` is unaffected — postgres readiness does not change cohort membership preference. The two signals are orthogonal: `CohortEligibilityStatus` covers administrative state (graceful shutdown, `StopReplication`); `LeadershipAvailability` covers transient startup state.

### Orchestrator: sort tied candidates by readiness

WAL position is always the primary criterion. Under synchronous replication, all nodes in the cohort typically share the same LSN, so the tie-breaker matters in practice.

`leadershipAwareDiscoverer(healthByID)` replaces direct use of `DiscoverMostAdvancedTimeline`:

```go
candidates := commonconsensus.DiscoverMostAdvancedTimeline(recruited, outgoingRule)
sort.SliceStable(candidates, func(i, j int) bool {
    return isReady(candidates[i].GetId()) && !isReady(candidates[j].GetId())
})
return candidates
```

`DiscoverMostAdvancedTimeline` returns only the nodes tied at the highest LSN. The sort puts `READY` nodes first within that tied set. `buildFailoverProposal` picks index 0, so the first `READY` node wins. If all tied nodes are `STARTING` the sort is a no-op and the election proceeds normally — WAL position is never sacrificed for readiness.

The same `Discoverer` is passed to both `BuildSafeProposal` (actual election) and `CheckProposalPossible` (pre-vote feasibility check), keeping both gates consistent.

## Test

`TestFailoverSkipsCohortIneligibleNode` (`go/test/endtoend/multiorch/`):

1. 3-node cluster bootstraps (primary + 2 standbys).
2. `StopReplication(RECEIVER_ONLY)` on `standbyIneligible` — sets `walReceiverManuallyStopped`, causing it to publish `STARTING` via `buildLeadershipAvailability`. Postgres stays running so Recruit RPCs succeed; the node still contributes to quorum.
3. Wait until both the `STARTING` signal and `primary_conninfo == ""` are confirmed in a single Status RPC call.
4. Write one row via the primary and wait until `standbyEligible`'s WAL position is strictly ahead of `standbyIneligible`'s frozen LSN. This closes the LSN race: `standbyEligible` is now both more advanced and `READY`, so the sort places it first on both criteria.
5. Kill primary postgres → failover.
6. Assert `standbyEligible` is elected, `standbyIneligible` is not.
7. Cleanup: re-enable postgres restarts, `StartReplication` on `standbyIneligible`, `RequireRecovery`.

Closes MUL-558

### DIFF
--- a/go/services/multiorch/consensus/coordinator.go
+++ b/go/services/multiorch/consensus/coordinator.go
@@ -17,6 +17,7 @@ package consensus
 import (
 	"context"
 	"log/slog"
+	"sort"
 	"sync"
 
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -120,8 +121,30 @@ func (c *Coordinator) runFailover(ctx context.Context, cohort []*multiorchdatapb
 		return mterrors.Errorf(mtrpcpb.Code_FAILED_PRECONDITION, "%v", err)
 	}
 
-	poolerByID, _ := buildCohortMaps(cohort)
+	poolerByID, healthByID := buildCohortMaps(cohort)
+	less := leadershipLess(healthByID)
 	buildProposal := func(r commonconsensus.RecruitmentResult) (*consensusdatapb.CoordinatorProposal, error) {
+		// Sort the eligible leaders so that READY nodes are preferred over
+		// STARTING nodes when multiple candidates share the highest LSN. A
+		// STARTING candidate's postgres is still initialising (starting, crash
+		// recovery, or socket not open), so electing it could cause a promote
+		// failure and force a new round of elections. Preferring an
+		// already-READY candidate avoids that delay.
+		//
+		// WAL position is always the primary criterion: a STARTING node still
+		// wins if it holds a strictly higher LSN than every READY node.
+		//
+		// We put this logic in the proposal builder rather than the proposal
+		// core because it only affects the ordering of tied leaders when
+		// actually sending promote to it.
+		//
+		// This works because buildFailoverProposal only looks at the first
+		// eligible leader, so as long as all READY nodes are sorted before
+		// STARTING nodes, we will prefer READY leaders without affecting the
+		// quorum logic.
+		sort.SliceStable(r.EligibleLeaders, func(i, j int) bool {
+			return less(r.EligibleLeaders[i], r.EligibleLeaders[j])
+		})
 		return buildFailoverProposal(r, poolerByID)
 	}
 	tryBuildProposal := func(rev *clustermetadatapb.TermRevocation, statuses []*clustermetadatapb.ConsensusStatus) (*consensusdatapb.CoordinatorProposal, error) {
@@ -278,6 +301,30 @@ func (c *Coordinator) GetBootstrapPolicy(ctx context.Context, database string) (
 
 	c.policyCache.Store(database, db.BootstrapDurabilityPolicy)
 	return db.BootstrapDurabilityPolicy, nil
+}
+
+// leadershipLess returns a less function for sort.SliceStable that orders
+// ConsensusStatus entries so READY nodes appear before STARTING nodes. It is
+// used as the tiebreaker when calling BuildSafeProposal when multiple nodes
+// share the highest LSN.
+//
+// UNKNOWN (the proto zero value, emitted by older poolers that do not publish
+// the field) is treated as READY for backward compatibility. (We might want to
+// change this in the future since we should prefer READY poolers before poolers
+// with UNKNOWN state, which could be ready or not.)
+//
+// Recruited nodes participate in the outgoing-cohort quorum check regardless of
+// their LeadershipAvailability — this tiebreaker only affects which tied node
+// is proposed as leader, not the quorum denominator.
+func leadershipLess(healthByID map[string]*multiorchdatapb.PoolerHealthState) func(a, b *clustermetadatapb.ConsensusStatus) bool {
+	isReady := func(id *clustermetadatapb.ID) bool {
+		h := healthByID[topoclient.ClusterIDString(id)]
+		sig := h.GetAvailabilityStatus().GetLeadershipAvailability().GetSignal()
+		return sig != clustermetadatapb.LeadershipAvailabilitySignal_LEADERSHIP_AVAILABILITY_SIGNAL_STARTING
+	}
+	return func(a, b *clustermetadatapb.ConsensusStatus) bool {
+		return isReady(a.GetId()) && !isReady(b.GetId())
+	}
 }
 
 // poolerIDs extracts the clustermetadata IDs from a slice of PoolerHealthState.
--- a/go/services/multiorch/consensus/coordinator.go
+++ b/go/services/multiorch/consensus/coordinator.go
@@ -122,18 +122,25 @@ func (c *Coordinator) runFailover(ctx context.Context, cohort []*multiorchdatapb
 	}
 
 	poolerByID, healthByID := buildCohortMaps(cohort)
-	less := poolerHealthStateLess(healthByID)
+	availLess := leadershipLess(healthByID)
+	signalLess := poolerHealthStateLess(healthByID)
 	buildProposal := func(r commonconsensus.RecruitmentResult) (*consensusdatapb.CoordinatorProposal, error) {
-		// Prefer non-resigning nodes when multiple candidates share the highest
-		// LSN. A node signalling REQUESTING_DEMOTION has explicitly asked to be
-		// replaced. Electing it again immediately defeats the purpose of the
-		// switchover.
-		//
-		// A resigning node still wins if it holds a strictly higher LSN than
-		// every other node, but consensus status serves as a tiebreaker when
-		// multiple nodes are at the same WAL position.
+		// Sort tied candidates with two orthogonal tiebreakers applied in order:
+		//   1. LeadershipAvailability: READY before STARTING (postgres not yet ready
+		//      for promotion — crash recovery still running, socket not open).
+		//   2. LeadershipSignal: non-resigning before REQUESTING_DEMOTION (node has
+		//      explicitly asked to be replaced via SwitchPrimary).
+		// WAL position is always the primary criterion; both tiebreakers only
+		// affect the ordering within the tied-LSN set.
 		sort.SliceStable(r.EligibleLeaders, func(i, j int) bool {
-			return less(r.EligibleLeaders[i], r.EligibleLeaders[j])
+			a, b := r.EligibleLeaders[i], r.EligibleLeaders[j]
+			if availLess(a, b) {
+				return true
+			}
+			if availLess(b, a) {
+				return false
+			}
+			return signalLess(a, b)
 		})
 		return buildFailoverProposal(r, poolerByID)
 	}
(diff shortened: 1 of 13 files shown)


go/common/consensus/proposals.go
```diff
@@ -65,14 +65,42 @@ type RecruitmentResult struct {
...
+// which isEligible returns true are passed to the base Discoverer. Because
+// the filter runs before WAL-position comparison, the base Discoverer always
+// selects the most advanced node among those that are actually promotable —
+// rather than selecting the globally most-advanced node and then failing when
+// it turns out to be unpromotable.
+//
+// The quorum checks in buildProposalCore operate on the full recruited set
+// and are unaffected by this filter; pre-filtering here affects only leader
+// candidate selection.
+func WithCandidateFilter(base Discoverer, isEligible func(*clustermetadatapb.ID) bool) Discoverer {
```
dweitzman: This doesn't seem safe from a consensus invariants perspective. For safe proposals, using the most advanced recruited pooler ensures it'll be a timeline that contains all durable transactions. For externally-certified proposals, the filter is intended to allow the external certification to be _more_ strict than what's discovered (if desired), in case the poolers are running and out-of-date backup or something like that.

mkindahl: This was leftovers from the previous solution that we discussed. Removed.

go/services/multiorch/consensus/coordinator.go
```diff
@@ -268,6 +270,42 @@ func (c *Coordinator) GetBootstrapPolicy(ctx context.Context, database string) (
...
+// so the first READY node in the tied set wins. If all tied nodes are STARTING
+// the sort is a no-op and the election proceeds normally.
+//
+// UNKNOWN (the proto zero value, emitted by older poolers that do not publish
+// the field) is treated as READY for backward compatibility.
+//
+// Recruited nodes participate in the outgoing-cohort quorum check regardless
+// of their LeadershipAvailability — this discoverer only affects which tied
+// node is proposed as leader, not the quorum denominator (cf. MUL-505).
+func leadershipAwareDiscoverer(healthByID map[string]*multiorchdatapb.PoolerHealthState) commonconsensus.Discoverer {
```
dweitzman: I think the most appropriate place for this sort of logic would be in `buildFailoverProposal` rather than changing proposals.go

proto/clustermetadata.proto
```diff
@@ -660,3 +666,31 @@ enum CohortEligibilitySignal {
 message CohortEligibilityStatus {
   CohortEligibilitySignal signal = 1;
 }
+
+// LeadershipAvailabilitySignal describes whether a pooler is ready to be
+// promoted to consensus leader.
+enum LeadershipAvailabilitySignal {
```
dweitzman, line 915: This signal doesn't seem like it's really leadership-specific? I think it might be useful to make this more of a signal of recruitability, like a `recruit_ineligible` boolean or similar

Then we could also reuse it as a signal when pg_rewind is catching up and between the last checkpoint and the fork point

mkindahl: > This signal doesn't seem like it's really leadership-specific? I think it might be useful to make this more of a signal of recruitability, like a recruit_ineligible boolean or similar

It was intended to pick the node most suitable as the leader when there are multiple alternatives, so it is not for the Recruit phase.

It will only affect if the node is picked as leader and not affect if the node is included in the Recruit.

> Then we could also reuse it as a signal when pg_rewind is catching up and between the last checkpoint and the fork point

This is a signal used to pick the leader, not about whether the node can serve queries, and recovery is about ability to handle queries.

If the node is in recovery its pooler status should be set to `NOT_SERVING` so that we do not send queries to it since it is not ready to serve traffic. Once it is properly replicating, we can start to send queries to it by setting it to `SERVING`.

As previously discussed, I don't think we should filter out servers from the Recruit phase since that might lead to throwing away transactions that were durable.

mkindahl: Hmm... this should probably be in a different proto file. It is a health snapshot message.

mkindahl: It was in the right proto file.

go/services/multiorch/consensus/coordinator.go
mkindahl: Moved it into buildSafeProposalCore, which does a stable sort on the leader availability so that known available leaders are preferred.

go/services/multipooler/internal/manager/manager.go
```diff
@@ -444,9 +452,16 @@ func (pm *MultiPoolerManager) openLocked(ctx context.Context, targetServingStatu
...
+					"postgres_running", newState.postgresRunning,
+					"postgres_ready", newState.postgresReady)
 				pm.broadcastHealth()
 			}
+			// Track postgres readiness for cohort eligibility. When postgres is
+			// not ready (starting, in crash recovery, socket not open), replication
+			// is not running and this node cannot participate in write quorum.
+			// buildCohortEligibilityStatus reads this atomically so the health
+			// stream reflects the current state without needing a lock.
+			pm.postgresReadyForCohort.Store(newState.postgresReady)
```
dweitzman: Something feels off here. `postgresReady` is already computed in for every status stream message, so this is saving a cached version of information that's currently recomputed with every status update anyway

What's the intended final state for when/how we compute and share this health information?

mkindahl: Yeah, it wasn't needed. Removed.

proto/clustermetadata.proto
dweitzman: Seems like these proto3 changes aren't needed, since the health already reports `postgres_ready`?

go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -927,6 +952,15 @@ func (pm *MultipoolerManager) setPrimaryLocked(ctx context.Context, req *consens
 		pm.logger.InfoContext(ctx, "SetPrimary: updating standby primary_conninfo", //nolint:sloglint // message intentionally starts with an operation name or proper noun
 			"new_leader", leader.GetId().GetName(),
 			"incoming_position", incomingPosition)
+		// A coordinator-initiated SetPrimary overrides a prior manual
+		// StopReplication. walReceiverManuallyStopped prevents this node from
+		// being elected leader (INELIGIBLE), but once the coordinator has elected
+		// a different leader, this node must follow it. Clear the flag so that
+		// setPrimaryConnInfoLocked can configure replication toward the new leader.
+		if pm.walReceiverManuallyStopped.CompareAndSwap(true, false) {
```
dweitzman, line 960: Clearing `walReceiverManuallyStopped` feels like it breaks my intuition for the contract of `StopReplication`, if replication can resume without anyone calling `StartReplication`.

Can you say more about this change?

PR #1091, opened 2026-06-03
fix(multipooler): wait for postgres socket before calling pg_promote()

mkindahl:
When a pod is recreated with PVC data intact, the multipooler can successfully respond to a Recruit RPC (postgres is running and answers SQL) but by the time the Propose RPC arrives the socket may no longer exist. The connection pool tries to reconnect for pg_promote() and fails:

    propose failed: could not write rule: promotion hook: reconnect dial
    failed: failed to connect to Unix socket .s.PGSQL.5432: no such file
    or directory

Add waitForPostgresReady, which polls isPostgresReady at 200ms intervals until postgres is accepting connections or the context deadline (the Propose RPC timeout) is reached. Call it at the start of promoteStandbyToPrimary, after the already-promoted early-exit, so the wait only runs when pg_promote() is actually about to be called.

If postgres becomes ready within the Propose timeout the promotion proceeds without waiting for the next analysis cycle. If postgres is already ready the fast path returns immediately with no polling overhead.

Change is confined to go/services/multipooler/manager/manager.go.

Tests (rpc_consensus_test.go):

  - already ready returns immediately: fast path, no polling loop entered
  - polls until postgres becomes ready: queue of 2 not-ready + 1 ready pgctld responses; StatusCallCount confirms exactly 3 calls were made
  - context deadline exceeded: permanently not-ready; error wraps context.DeadlineExceeded at 350ms

Test infrastructure (go/cmd/pgctld/testutil/grpc_helpers.go):

  - MockPgCtldService.StatusResponseQueue: dequeues one response per Status call to enable sequenced mock responses
  - MockPgCtldService.StatusCallCount(): call count accessor under the internal mutex
  - setupManagerWithPgctld helper: variant of setupManagerWithMockDB that accepts a custom pgctld mock

Part of MUL-558

### DIFF
--- a/go/services/multipooler/manager/manager.go
+++ b/go/services/multipooler/manager/manager.go
@@ -1335,6 +1335,17 @@ func (pm *MultiPoolerManager) promoteStandbyToPrimary(ctx context.Context, state
 		return nil
 	}
 
+	// Wait for postgres to be ready before attempting promotion. A freshly
+	// recreated pod may have valid WAL data on its PVC (giving it a valid
+	// LSN in the Recruit response) but postgres may still be initialising —
+	// the socket may not yet exist or pg_isready may not yet pass. Without
+	// this wait the connection attempt inside pg_promote() fails immediately
+	// with "no such file or directory". We block here rather than failing and
+	// waiting for the next analysis cycle.
+	if err := pm.waitForPostgresReady(ctx); err != nil {
+		return mterrors.Wrap(err, "postgres not ready for promotion")
+	}
+
 	// Call pg_promote() to promote standby to primary
 	pm.logger.InfoContext(ctx, "PostgreSQL promotion needed")
 	pm.logger.InfoContext(ctx, "Calling pg_promote() to promote standby to primary")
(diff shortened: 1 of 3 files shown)


go/services/multipooler/manager/manager.go
```diff
@@ -1335,6 +1335,17 @@ func (pm *MultiPoolerManager) promoteStandbyToPrimary(ctx context.Context, state
...
 	}
 
+	// Wait for postgres to be ready before attempting promotion. A freshly
+	// recreated pod may have valid WAL data on its PVC (giving it a valid
+	// LSN in the Recruit response) but postgres may still be initialising —
+	// the socket may not yet exist or pg_isready may not yet pass. Without
+	// this wait the connection attempt inside pg_promote() fails immediately
+	// with "no such file or directory". We block here rather than failing and
+	// waiting for the next analysis cycle.
+	if err := pm.waitForPostgresReady(ctx); err != nil {
```
dweitzman, line 1345: How was `Recruit()` able to read from postgres prior to the promotion if there was no socket file?

mkindahl: Logs don't show that, but it is clear that the Recruit successfully executed and retrieved the LSN and Propose failed with an error.

mkindahl: Regardless of reason for the failures, I see no harm in waiting for the socket to be available again. We already have timeouts on the operation, so in practically all circumstances it should be available shortly.

mkindahl: > How was `Recruit()` able to read from postgres prior to the promotion if there was no socket file?

The most likely reason is that the `Recruit()` stops the servers after it has answered and then it takes some time for the servers to start again when the `Propose()` and `SetTermPrimary()`, but this is not verified by the logs. I am a little surprised that I don't see `SetTermPrimary` failures as well, so might be an idea to look into.

dweitzman: `Recruit()` does the read from postgres after restarting it: https://github.com/multigres/multigres/blob/43db3a4873922eeebc463f8e249d0b5f214e93f1/go/services/multipooler/manager/rpc_consensus.go#L364

That implies to me that postgres should be accepting connections at the time `Recruit()` returns

dweitzman: Assuming there was a need to do waiting (which still seems odd to me given that Recruit() already needed to read from postgres after restarting), I don't think this would be the right place to do it. It should ideally be in Recruit(), or after that maybe in Propose() before the action lock is taken

The premise of accepting Recruit()ment is that a pooler is agreeing to participate in establishing a new rule, so I would much prefer if the pooler was in a functional state at the end of Recruit(). Accepting recruitment while postgres isn't running and ready to promote feels like it goes against the spirit of recruitment.

mkindahl: > Assuming there was a need to do waiting (which still seems odd to me given that Recruit() already needed to read from postgres after restarting), I don't think this would be the right place to do it. It should ideally be in Recruit(), or after that maybe in Propose() before the action lock is taken
> 
> The premise of accepting Recruit()ment is that a pooler is agreeing to participate in establishing a new rule, so I would much prefer if the pooler was in a functional state at the end of Recruit(). Accepting recruitment while postgres isn't running and ready to promote feels like it goes against the spirit of recruitment.

From the logs, the `Recruit()` is already successful and is exiting the recruit stage, so it is in a functional state at the end of `Recruit()` already.

Something happens between `Recruit()` and `Propose()` that causes the socket to be unavailable. From the logs you can see that the server is up approximately 740ms after the failed `Propose()` call, but it is not clear what caused it to go down.

We need to decide how to handle these transient states. A failure between `Recruit()` and `Propose()` can happen for a few different reasons:
- It is another stop or crash. It might quickly recover, in which case it should not be a problem to use it as primary.
- If it does not recover, however, we need to pick another pooler.
- It is a transient "blip", then we should just wait it out.

We discussed this and decided to leave this for now and re-visit it if it turns out to be a common occurrence.

mkindahl:
Closing with the decision to check logs and see if this is a common occurrence. We might re-open it later in that case.

PR #1092, opened 2026-06-03
feat(multiorch): record and assert leader appointment latency

dweitzman:
Two related changes:

1. **Instrument leader appointment latency.** The `primary.promotion` event now records `proposed_term` and the `recruit_ms` / `propose_ms` split (recruit = start → leader selected, propose = leader selected → commit). `Started` now fires before recruitment so it marks the true start, and recruitment failures emit a terminal `Failed` instead of returning silently. The dead-primary e2e asserts each phase stays fast, isolating appointment latency from the much larger detection/grace window.

2. **Poll for replication state with backoff.** `waitForReceiverDisconnect` and `waitForReplicationPause` used fixed-interval tickers whose first tick fires before the first check, adding that interval to every recruit even though the state usually settles in milliseconds. Both now share a `pollForReplicationStatus` helper that checks immediately, then backs off exponentially via the `retry` helper. This change makes recruit a few hundred milliseconds faster.

## Validation

Dead-primary e2e now asserts recruit and propose each stay within per-phase bounds

### DIFF
--- a/go/services/multipooler/manager/pg_replication.go
+++ b/go/services/multipooler/manager/pg_replication.go
@@ -353,45 +353,93 @@ func (pm *MultiPoolerManager) queryReplicationStatus(ctx context.Context) (*mult
 	return status, nil
 }
 
-// waitForReplicationPause polls until WAL replay is paused and returns the status at that moment.
-// This ensures the LSN returned represents the exact point at which replication stopped.
-func (pm *MultiPoolerManager) waitForReplicationPause(ctx context.Context) (*multipoolermanagerdatapb.StandbyReplicationStatus, error) {
-	// Create a context with timeout for the polling loop
+// pollForReplicationStatus polls poll on a 10s budget until it reports done,
+// returning the accompanying status. The first attempt runs immediately and
+// subsequent attempts back off exponentially from a small base: a state that
+// settles within milliseconds (the common case for these replication waits)
+// returns promptly instead of paying a fixed poll interval on the failover
+// critical path, while the backoff avoids hammering postgres when it takes
+// longer.
+//
+// poll returns (status, done, err): done==true stops the loop and returns
+// status; a non-nil err aborts immediately, except that an err observed after
+// the context expired mid-poll is reported as the canonical timeout. On
+// budget/context exhaustion, onTimeout (if non-nil) is invoked for diagnostics
+// and a canonical DEADLINE_EXCEEDED (timeoutMsg) or wrapped cancellation
+// (cancelMsg) is returned.
+func (pm *MultiPoolerManager) pollForReplicationStatus(
+	ctx context.Context,
+	timeoutMsg, cancelMsg string,
+	onTimeout func(cause error),
+	poll func(ctx context.Context) (status *multipoolermanagerdatapb.StandbyReplicationStatus, done bool, err error),
+) (*multipoolermanagerdatapb.StandbyReplicationStatus, error) {
 	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
 
-	ticker := time.NewTicker(100 * time.Millisecond)
-	defer ticker.Stop()
+	fail := func(cause error) (*multipoolermanagerdatapb.StandbyReplicationStatus, error) {
+		if onTimeout != nil {
+			onTimeout(cause)
+		}
+		if errors.Is(cause, context.DeadlineExceeded) {
+			return nil, mterrors.New(mtrpcpb.Code_DEADLINE_EXCEEDED, timeoutMsg)
+		}
+		return nil, mterrors.Wrap(cause, cancelMsg)
+	}
 
-	for {
-		select {
-		case <-waitCtx.Done():
-			if waitCtx.Err() == context.DeadlineExceeded {
-				pm.logger.ErrorContext(ctx, "Timeout waiting for WAL replay to pause")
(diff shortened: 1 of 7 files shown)


go/services/multipooler/manager/pg_replication.go
```diff
@@ -353,45 +353,93 @@ func (pm *MultiPoolerManager) queryReplicationStatus(ctx context.Context) (*mult
...
+		return nil, mterrors.Wrap(cause, cancelMsg)
+	}
 
-	for {
-		select {
-		case <-waitCtx.Done():
-			if waitCtx.Err() == context.DeadlineExceeded {
-				pm.logger.ErrorContext(ctx, "Timeout waiting for WAL replay to pause")
-				return nil, mterrors.New(mtrpcpb.Code_DEADLINE_EXCEEDED, "timeout waiting for WAL replay to pause")
+	r := retry.New(5*time.Millisecond, 500*time.Millisecond)
```
dweitzman, line 389: The switch from 100ms to exponential backoff makes this faster

go/services/multipooler/manager/pg_replication.go
```diff
@@ -542,51 +590,30 @@ func (pm *MultiPoolerManager) waitForReceiverDisconnect(ctx context.Context) (*m
...
-		case <-ticker.C:
-			// Re-check waitCtx before issuing a query: when waitCtx expires at the
-			// same tick boundary, select may pick this branch and the subsequent
-			// pm.query would surface an opaque "pool ctx expired" instead of the
-			// real timeout cause.
-			if err := waitCtx.Err(); err != nil {
-				return timedOut(err)
-			}
-
+	return pm.pollForReplicationStatus(ctx,
```
dweitzman, line 593: The switch from 500ms polling to exponential backoff makes this significantly faster

go/test/endtoend/multiorch/dead_primary_recovery_test.go
```diff
@@ -423,6 +425,54 @@ func TestDeadPrimaryRecovery(t *testing.T) {
...
+		// Per-phase regression guards, separate because the phases have very
+		// different baselines and failure modes. Both exclude the failure-
+		// detection grace period that dominates the test's overall 30s budget.
+		//
+		// Recruit needs a few round trips + WAL replay stabilizing.
+		// Propose needs pg_promote() to succeed and the rule entry to replicate.
+		// The test bounds are deliberately tight to catch regressions that would
+		// make a simpler failover slow.
+		const (
+			maxRecruit = 500 * time.Millisecond
```
dweitzman, line 457: This is one of the motivations for these changes: having the e2e tests assert some reasonable bound on how long recruit & propose should take under controlled conditions

In practice, on my laptop recruit takes maybe a few 10s of milliseconds and propose takes maybe a few hundred. Here's what the test run at https://github.com/multigres/multigres/actions/runs/26900485132/job/79351193362 printed (about twice as long as my laptop, give or take):

```
      dead_primary_recovery_test.go:467: Leader appointment (new_primary=pooler-1, proposed_term=3): recruit=60ms, propose=685ms
      dead_primary_recovery_test.go:467: Leader appointment (new_primary=pooler-2, proposed_term=2): recruit=58ms, propose=705ms
      dead_primary_recovery_test.go:467: Leader appointment (new_primary=pooler-2, proposed_term=4): recruit=56ms, propose=611ms
      dead_primary_recovery_test.go:467: Leader appointment (new_primary=pooler-1, proposed_term=6): recruit=59ms, propose=1.492s
```

Slower than my laptop but feels comfortably within the bounds of the test assertion values

cuongdo: This tight bound makes me nervous about potential future CI flakes. We should keep an eye out for those.

cuongdo (approved):
lgtm, though I'm less familiar with the consensus code

go/test/endtoend/multiorch/dead_primary_recovery_test.go
dweitzman: I'm fine with raising it if there are flakes. The goal is to assert that failovers are relatively fast, so I feel like it's better to start too low and raise the value instead of starting too high and not noticing if we make a change that leads to slower failovers

mkindahl (approved):
LGTM

go/test/endtoend/multiorch/dead_primary_recovery_test.go
mkindahl: I think that it is important to track and follow up on the failover times. Failover times on local machines (quite powerful ones that we have compared to what are used in the fleet) or test environment (prone to sudden stops) are not good indicators. We need to track this in the fleet and in tests, monitor it as necessary, and investigate the cases that are too long.

I don't know what timeouts we need, so I am focusing on collecting actual information. IMHO, having short timeouts in tests just leaves us with flaky tests that usually will be ignored over time since they are considered "always flaky" and increases the risk of missing an important failure.

I added a [summary table startup times](https://github.com/multigres/multigres/actions/runs/26900485132?pr=1092) precisely for this reason. It takes the measurements that we _actually_ have, presents the max and min and compute a confidence interval for the parameter so that we can tailor it and not get more flakes than we can handle.

Here is an [updated version](https://github.com/multigres/multigres/actions/runs/27012878715?pr=1104) that includes failover and recovery, so that you can see what figures we currently have.

PR #1093, opened 2026-06-03
feat(multipooler): raise and surface max_wal_senders for multi-cell clusters

dweitzman:
1. **fix(pgctld):** raise default `max_wal_senders`/`max_replication_slots` 5 → 25.
   With 5, a primary can't support more than 5 standbys — extras are refused
   (`number of requested standby connections exceeds "max_wal_senders"`) and
   never stream. Easy to hit in multi-cell (3 cells × 3 poolers → 8 standbys).

2. **feat(multipooler):** surface `max_wal_senders` in `PrimaryStatus`. The
   replication page now shows **"N connected / M max"** on the WAL senders card,
   so capacity exhaustion is obvious ("5 connected / 5 max" explains itself).

### DIFF
--- a/go/services/multipooler/manager/rpc_manager.go
+++ b/go/services/multipooler/manager/rpc_manager.go
@@ -31,6 +31,7 @@ import (
 	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
 	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
+	"github.com/multigres/multigres/go/services/multipooler/executor"
 )
 
 // broadcastHealth broadcasts the current health state to all subscribers.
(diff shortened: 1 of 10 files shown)


go/services/multipooler/manager/rpc_manager.go
```diff
@@ -546,12 +547,32 @@ func (pm *MultiPoolerManager) getPrimaryStatusInternal(ctx context.Context) (*mu
...
+	var (
+		syncStandbyNames string
+		syncCommit       string
+		maxWalSenders    int32
+	)
+	if err := executor.ScanSingleRow(result, &syncStandbyNames, &syncCommit, &maxWalSenders); err != nil {
+		return nil, mterrors.Wrap(err, "failed to scan replication settings")
+	}
+
+	syncConfig, err := parseSyncReplicationConfig(syncStandbyNames, syncCommit)
```
mkindahl, line 570: Wouldn't it be easier to just update `pm.getSynchronousReplicationConfig` to return an extra field and add the retrieval of `max_wal_senders` to that function? Then we have a single function we do the retrieval of the sync replication configuration. With this we would have two different methods that are likely to go out of sync as the code evolves.

go/services/pgctld/postgresconfig_gen.go
```diff
@@ -91,8 +91,8 @@ func GeneratePostgresServerConfig(poolerDir string, pgUser string, extraConfFile
 	cnf.MinWalSize = "1GB"
 	cnf.MaxWalSize = "4GB"
 	cnf.CheckpointCompletionTarget = 0.9
-	cnf.MaxWalSenders = 5
-	cnf.MaxReplicationSlots = 5
+	cnf.MaxWalSenders = 25
+	cnf.MaxReplicationSlots = 25
```
mkindahl, line 95: This is sufficient for now, but as the system grows we might to have to dynamically adapt this to changing circumstances. Otherwise we would have to keep updating the parameters to match the "biggest" user.

We probably need a mechanism to adapt the parameter as the size of the deployment changes.

mkindahl (approved):
It seems like a roundabout way to have two different code paths for retrieving the replication parameters: one that fetches a fixed list of parameters in one location, and a "generic" function that retrieves a different set and is used everywhere else.

Approving since I don't think this is a blocker.

go/services/multipooler/manager/rpc_manager.go
dweitzman: I considered that, but I think there's an argument to be made that `max_wal_senders` is not exactly part of the synchronous replication config. It applies even for non-synchronous replication, for example.

I'm not opposed to giving the message a more generic name and putting more into it, though

dweitzman:
> It seems like a roundabout way to have two different code paths for retrieving the replication parameters: one that fetches a fixed list of parameters in one location, and a "generic" function that retrieves a different set and is used everywhere else.
> 
> Approving since I don't think this is a blocker.

The `status` includes things that aren't config like connected followers count, so I see the reason to have different methods for status vs config. The config method is actually only used by test code at the moment: I'm open to deleting it entirely but that would be a bigger PR and I didn't want to grow the scope

go/services/pgctld/postgresconfig_gen.go
dweitzman: Agreed that this is just buying runway without solving the problem. 25 was arbitrary, but it feels like a number that buys a decent amount of runway without being extreme

go/services/multipooler/manager/rpc_manager.go
mkindahl: My main thinking is that it is strange to have two different functions with similar purpose: returning replication configuration information. They return slightly different information, but it might be easier to maintain by keeping a single function that provides information that can be used by both. It might return more information than necessary for both usages, but at the bottom this is just a single select, so it should not make a difference performance-wise, but is easier to maintain.

PR #1094, opened 2026-06-04
test(pgbouncertests): pooling & resilience tests ported from PgBouncer

GuptaManan100:
## Summary

- Adds the `pgbouncertests` end-to-end suite — pooling, session-state, and fault-recovery behavior on the `multigateway → multipooler → postgres` path, with scenarios ported from [PgBouncer's test suite](https://github.com/pgbouncer/pgbouncer/tree/master/test) (ISC-licensed).
- Covers backend crash recovery under load + transparent reconnect, pool exhaustion (multigres queues rather than rejecting), reserved-connection pin/release + idle-in-transaction timeout, cancel-race, `COPY` interrupted by a backend crash, and connecting to a nonexistent database (differential vs. PostgreSQL).
- Test-only: no product or proto changes. Builds on #1087, which landed the `DISCARD ALL` gateway fix, the Group 1 prepared-statement tests, and the `shardsetup.WithMultipoolerExtraArgs` harness option.

## Description

The tests are **black-box**: they speak only the PostgreSQL wire protocol to the multigateway and observe only what a client can — query results, errors, and `pg_backend_pid()` read from the client's own rows. Pooled-backend rotation is induced with concurrent churn (`startPoolChurn`) and asserted via distinct `pg_backend_pid()` values rather than forcing a specific internal swap, so a test is never green-by-luck. Fault tests inject crashes with `KillPostgres`, and each runs in its own `shardsetup.NewIsolated` cluster so the damage can't leak into the shared-cluster tests.

The suite runs in the standard e2e integration job (no opt-in label); tests self-skip in `-short` or when the PostgreSQL binaries are absent. See the package `README.md` for the file→scenario map, the approach, and known divergences from PostgreSQL — notably that the gateway serves its backend database for an unknown requested database instead of rejecting with `3D000` (guarded by `nonexistent_db_test.go`).

### DIFF
--- a/go/test/endtoend/queryserving/pgbouncertests/README.md
+++ b/go/test/endtoend/queryserving/pgbouncertests/README.md
@@ -1,551 +1,76 @@
-# pgbouncertests — pooling & resilience suite (triage)
-
-> **Status: PORTING COMPLETE — all (A) groups landed.** This document started as
-> Step 1 of the PgBouncer-port effort. It categorizes every PgBouncer test file
-> as **(A) port**, **(B) out of scope**, or **(C) already covered**, and tracks
-> the net-new Go tests to land. The triage has been reviewed; the resolved
-> decisions are recorded in [Review decisions](#review-decisions). The point of
-> the triage is to land only the net-new coverage and avoid re-porting what we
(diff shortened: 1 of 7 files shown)


haritabh17 (approved):
🚀

PR #1096, opened 2026-06-04
test(pgregresstest): run pg_cron's regression suite through multigateway

GuptaManan100:
## Summary

- Enrolls **pg_cron** (Citus, pinned `v1.6.4`) as a covered external extension in the pgregress suite, running its shipped `pg_cron-test` through multigateway — mirroring how pgvector is run.
- Generalizes `ExternalExtension` with four per-extension knobs, since pg_cron diverges from the pgvector baseline: `TestSubdir` (pg_cron ships fixtures at the repo root, not under `test/`), `CreateExtension` (pg_cron's test manages the extension itself, so the harness must not preload it), `ServerConfigFile` (pg_cron's background worker needs `shared_preload_libraries = 'pg_cron'`, applied to the cluster at initdb time only when the external suite runs), and `ScratchDatabases` (see below).
- Adds `testdata/pg17/external/pg_cron.conf` and a captured `pg_cron-test.patch` (just the 4 `CREATE`/`DROP DATABASE` lines).

Verified green in both patch-generate and verify (CI) modes: `pg_cron/pg_cron-test ✅`, 1/1.

## On `CREATE DATABASE` / `DROP DATABASE`

pg_cron's `pg_cron-test.sql` uses `CREATE`/`DROP DATABASE` to set up two scratch databases (`pgcron_dbno`, `pgcron_dbyes`) for its cross-database scheduling tests. multigateway blocks both as Tier 2 statements **by design** — multigres is one-database-per-postgres-instance, and adding a database is a provisioning operation (a new cluster), not a query-path operation (see `docs/query_serving/unsafe_statement_rejection.md`).

Key observation: the test never actually *connects to* those databases. It references them purely as **metadata** — it passes the names to `cron.schedule_in_database` / `cron.alter_job(database := ...)` and reads their ACLs from the shared `pg_database` catalog, all over the same `postgres` connection. The only thing that would open a session against them is pg_cron's in-process launcher (local to postgres, not via the gateway), and the test doesn't wait for it.

So rather than edit upstream's test (which would weaken its assertions), the harness front-loads the two databases **directly on the primary** (bypassing the gateway, via the new `ScratchDatabases` knob) before the suite and drops them after. The catalog/CONNECT-privilege/`alter_job` checks then run for real and match upstream; only the test's own `CREATE`/`DROP DATABASE` statements remain in the patch as the documented gateway block. This is a **test-only** accommodation and is commented as such — it does not represent multi-database support; when `CREATE DATABASE` becomes a real provisioning feature, this hook is revisited.

### DIFF
--- a/go/test/endtoend/pgregresstest/README.md
+++ b/go/test/endtoend/pgregresstest/README.md
@@ -143,8 +143,8 @@ as catalog entries move to `covered` and their suites run.
 
 The external suite runs the pg_regress suites of extensions that live **outside**
 the PostgreSQL source tree (separate repositories), executed through
-multigateway. pgvector is the first: catalog name `vector`, pinned to a tag in
-`externalSpecs` (`extensions.go`).
+multigateway. The covered set lives in `externalSpecs` (`extensions.go`), each
+pinned to a tag: `vector` (pgvector) and `pg_cron` (Citus pg_cron).
 
 How it works, and how it differs from the contrib suite:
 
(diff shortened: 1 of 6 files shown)


haritabh17 (approved):
🚀

PR #1098, opened 2026-06-04
fix(multipooler): wait out reopen window instead of surfacing transient pool-close

GuptaManan100:
## Summary

- Stabilizes `TestWrappedPreparedStatementExecution/multigateway` by closing the race in `withReopenRetry` that surfaced a transient closed-pool error to clients during a pool reopen.
- Alternative to #1031 — same bug, simpler model. 
## Problem

`reopenConnections()` runs `Close -> Open` on the conn pool manager to refresh file descriptors after a PostgreSQL restart. A query whose acquisition lands in the close-but-not-yet-open window sees `ErrPoolClosed` (or "manager is closed") and surfaces it to the client: `pq: ... connection pool is closed`. The old generation-counter scheme didn't actually fix this — on the reopen path it made the same retry decision as the plain `closed` flag, so the window stayed open.

## Solution

Distinguish a transient reopen from a terminal shutdown explicitly. `CloseForReopen()` sets `closed=true` — so `getOrCreateUserPool` still refuses to build a pool against a half-open manager, which is what prevents an orphaned pool wired to a nil admin pool — and additionally marks a reopen window with a completion channel that the paired `Open()` closes.

`withReopenRetry` then waits out the window and retries against the fresh pools, honoring the request context. A closed manager with no window in progress stays terminal and surfaces unchanged. Retries are bounded by `maxPoolClosedRetries`. The reopen path in `MultiPoolerManager` calls `CloseForReopen` (vs `Close` for genuine shutdown) via a `forReopen` flag on `closeConnectionsLocked`.

Tests: `TestWithReopenRetry` suite rewritten around `CloseForReopen`/`Open` (op-path retry, lookup-path wait, terminal close, pre-closed fail-fast, retry cap, ctx cancellation); passes `-race -count=5`.

### DIFF
--- a/go/services/multipooler/connpoolmanager/config.go
+++ b/go/services/multipooler/connpoolmanager/config.go
@@ -606,6 +606,6 @@ func (c *Config) NewManager(logger *slog.Logger) *Manager {
 		logger:  logger,
 		metrics: metrics,
 	}
-	mgr.closed.Store(true) // Manager is closed until Open() is called
+	mgr.setLifecycle(lifecycleClosed) // Manager is closed until Open() is called
 	return mgr
 }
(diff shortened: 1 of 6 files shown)


mkindahl (approved):
LGTM

haritabh17 (approved):

PR #1100, opened 2026-06-04
refactor: delete redundant MultiPoolerConsensus.Status RPC

dweitzman:
ConsensusStatus is already embedded in MultiPoolerManager.Status responses (and in ManagerHealthStream snapshots), so the dedicated unary RPC on the consensus service was a duplicate surface. Removing it consolidates orch's status reads onto a single proto envelope and shrinks the consensus service to mutating operations (Recruit, Propose, SetTermPrimary, UpdateConsensusRule).

Callers migrated to MultiPoolerManager.Status, extracting .GetConsensusStatus() from the richer response.


go/services/multiorch/consensus/rule_change_certified.go
```diff
@@ -167,9 +168,9 @@ func (c *Coordinator) refreshShardConsensusStatuses(
 		wg.Go(func() {
 			rpcCtx, cancel := context.WithTimeout(ctx, timeouts.RemoteOperationTimeout)
 			defer cancel()
-			resp, err := c.rpcClient.ConsensusStatus(rpcCtx, p, &consensusdatapb.StatusRequest{})
+			resp, err := c.rpcClient.Status(rpcCtx, p, &multipoolermanagerdatapb.StatusRequest{})
```
dweitzman, line 171: This is the main impacted thing. It's not impossible that `Status()` could be slower than `ConsensusStatus()` because it returns more information, but if that's the case then maybe in the short term it'd make more sense to have `Status()` take parameters indicating what information is needed instead of having an entire additional RPC

In the long term, getting consensus status on-demand seems like it should use the same RPC as streaming status rather than having multiple separate endpoitns.

go/test/endtoend/multipooler/rpc_manager_api_test.go
```diff
@@ -21,14 +21,60 @@ import (
...
 
+	"github.com/multigres/multigres/go/common/consensus"
 	"github.com/multigres/multigres/go/test/endtoend/shardsetup"
 	"github.com/multigres/multigres/go/test/utils"
 
-	consensusdatapb "github.com/multigres/multigres/go/pb/consensusdata"
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 )
 
+// TestManagerStatus_NodeIdentityAndConsensus verifies that MultiPoolerManager.Status
```
dweitzman, line 31: These tests are moved from rpc_consensus_test.go since they're testing an RPC endpoint that's not part of the consensus gRPC service

rafael (approved):

PR #1101, opened 2026-06-05
feat(multipooler): publish CurrentLeadership observation to etcd

dweitzman:
Adds MultiPooler.current_leadership so consumers can identify the
leader from a single MultiPooler record instead of inferring from
PoolerType. Replicas leave it nil; only the leader sets it.

A future follow-up PR will change multigateway to read the current leadership instead of the pooler type.

This PR sets up some invariants / assertions to make sure that pooler type and the new current_leadership field are consistent: a primary must have a current_leadership, any other pooler type must not.

- New clustermetadata.LeaderObservation (leader_id + leader_rule_number).
  Coexists with multipoolerservice.LeaderObservation on the health
  stream; a follow-up will migrate the gateway and dedupe.
- MutablePoolerRecordState.CurrentLeadership; Mutate enforces the
  Type ↔ CurrentLeadership invariant (PRIMARY iff CurrentLeadership
  is set and names this pooler).
- StateManager.SetState now takes *ShardRule. Type and
  CurrentLeadership are derived together in deriveTypeAndObs.
- latestRule() merges consensusState's view with the rule_store
  cache; SetState callers pick it explicitly.
- StopTopoRegistration clears CurrentLeadership alongside DRAINED.
- TestPoolerRecord_Mutate_LeadershipInvariant covers each rejection
  branch.

changeTypeLocked's poolerType argument becomes vestigial under the
new SetState semantics. A follow-up PR will rewrite it and its callers.


go/services/multipooler/internal/manager/pooler_record.go
```diff
@@ -172,23 +193,49 @@ func (r *poolerRecord) Mutate(ctx context.Context, fn func(*MutablePoolerRecordS
...
 	next.LifecycleStatus = state.LifecycleStatus
 	r.desired.Store(next)
+	return nil
+}
+
+// validateState enforces the Type ↔ CurrentLeadership consistency
+// invariant: a pooler is the leader iff Type == PRIMARY iff
+// CurrentLeadership is set and names this pooler. Any deviation is a
+// caller bug.
+func (r *poolerRecord) validateState(state *MutablePoolerRecordState) error {
```
dweitzman, line 238: As an intermediate state, PoolerType and LeadershipObservation need to be consistent. If/when we remove references to PoolerType we can drop it from the record and remove this validation code

go/services/multipooler/internal/manager/manager.go
```diff
@@ -814,6 +814,35 @@ func (pm *MultiPoolerManager) getCurrentTermNumber(ctx context.Context) (int64,
...
 }
 
+// latestRule returns the higher-numbered rule between consensusState's
+// recorded replicationPrimary and rule_store's cached position, either
+// of which may be nil. Used by callers that need "the freshest rule
+// this pooler knows about" — RPCs update consensusState first; postgres
+// WAL replay populates rule_store first; the two can briefly disagree.
+// Returns nil if both sources are empty (cold boot before any
+// observation or RPC).
+func (pm *MultiPoolerManager) latestRule() *clustermetadatapb.ShardRule {
```
dweitzman: We can derive the pooler type from the highest rule know to this pooler. For a stale primary that's still trying to catch up on replication, the highest known rule might have come from SetTermPrimary rather than from the WAL

go/services/multipooler/internal/manager/state_manager.go
```diff
@@ -120,16 +170,34 @@ func (ssm *StateManager) SetState(ctx context.Context, poolerType clustermetadat
...
-		"type", poolerType, "status", servingStatus)
+		"type", newType, "status", servingStatus)
 
 	return nil
 }
+
+// deriveTypeAndObs returns the PoolerType and LeaderObservation implied
+// by rule for the pooler identified by selfID. See StateManager docs for
+// the derivation rules.
+func deriveTypeAndObs(rule *clustermetadatapb.ShardRule, selfID *clustermetadatapb.ID) (clustermetadatapb.PoolerType, *clustermetadatapb.LeaderObservation) {
```
dweitzman: This is what converts the highest known rule into the current pooler's PoolerType + leader observation

go/services/multipooler/internal/manager/manager.go
```diff
@@ -825,6 +825,35 @@ func (pm *MultiPoolerManager) getCurrentTermNumber(ctx context.Context) (int64,
 	return pm.consensusState.GetCurrentTermNumber(ctx)
 }
 
+// latestRule returns the higher-numbered rule between consensusState's
+// recorded replicationPrimary and rule_store's cached position, either
+// of which may be nil. Used by callers that need "the freshest rule
+// this pooler knows about" — RPCs update consensusState first; postgres
+// WAL replay populates rule_store first; the two can briefly disagree.
```
mkindahl: Shouldn't we ensure that these two fields are updated atomically using, e.g., a lock? If they disagree on who is the leader, there is a risk that two poolers use the wrong leader because of the order of updates, for example.

go/services/multipooler/internal/manager/pooler_record.go
```diff
@@ -46,6 +49,11 @@ type MutablePoolerRecordState struct {
 	Type            clustermetadatapb.PoolerType
 	ServingStatus   clustermetadatapb.PoolerServingStatus
 	LifecycleStatus *clustermetadatapb.PoolerLifecycle
+	// CurrentLeadership is the pooler's most recent recorded consensus
+	// observation. Set ONLY when this pooler currently considers itself the
+	// leader of its shard; replicas leave it nil. Published into etcd so
+	// multigateway can bootstrap leader routing on discovery.
+	CurrentLeadership *clustermetadatapb.LeaderObservation
```
mkindahl: Why do you not want to set this for non-leaders?

It can be useful for debugging and to understand if two poolers have different ideas of who the leader is. I can, for example, imagine that we end up in a broken state and want to reconstruct the leader for the records even if, for example, the record for the leader is missing. Having the standbys keeping information about their view of the leader can help debug the system and figure out what went wrong.

The LeaderObservation contains the ID, so the check for isLeader is simply to check if the ID matches the ID of the pooler. Since the rule number is there as well, there is no risk of confusion about who is the leader and in what term/rule.

rafael (approved):

rafael (approved):

go/services/multipooler/internal/manager/state_manager.go
```diff
@@ -36,13 +38,32 @@ type StateAware interface {
...
+// waits for completion, then updates the record via Mutate. The Mutate
+// also schedules an asynchronous publish so callers cannot forget to
 // reflect the new state to etcd.
+//
+// Callers do NOT pass a rule to SetState. The contract is "consensusState
+// and rule_store are the source of truth; SetState reflects what they
+// currently say." Code paths that learn of a new rule (e.g. SetTermPrimary
+// receiving an RPC) must record it via the appropriate path
+// (consensusState.RecordTermPrimary, rule_store updates) before calling
+// SetState. Closing this loop in the StateManager prevents accidents
```
mkindahl: "Code paths that learn of a new rule (e.g. SetTermPrimary receiving an RPC) must record it via the appropriate path (consensusState.RecordTermPrimary, rule_store updates) before calling SetState" sounds like something that is easy to make mistakes around. Is there no way to, for example, let SetTermPrimary update the consensusState when it learns about a new rule? That would at least avoid these kinds of mistakes.

mkindahl (approved):
I think the patch is OK, but I have two questions inline that might be good to consider.

proto/clustermetadata.proto
```diff
@@ -185,6 +185,16 @@ message MultiPooler {
...
   PoolerLifecycle lifecycle_status = 12;
+
+  // current_leadership is set ONLY when this pooler currently considers
+  // itself the leader of its shard. Replicas leave this empty so etcd
+  // doesn't carry the leader's view of itself in every pooler's record.
+  //
+  // Consumers (multigateway) will read this on discovery to bootstrap
+  // leader routing without relying on `type` as a hint. Best-effort:
+  // empty until the pooler has been told (via SetTermPrimary / Propose
+  // / Recruit) that it is the leader; cleared again on demotion.
```
dweitzman: TODO: add a comment that the reason this is only on leaders is to avoid a high volume of etcd writes

dweitzman: Done

go/services/multipooler/internal/manager/manager.go
dweitzman: Rule number ordering resolves conflicts between rule observations from different places, like WAL (rule store) vs SetPrimary RPCs (replicationPrimary)

There's an unavoidable delay between SetPrimary and receiving new WAL entries from that primary, so inconsistency between the rule store vs replicationPrimary is unavoidable

go/services/multipooler/internal/manager/pooler_record.go
dweitzman: Clarified the comment in the proto3 definition: the reason is to avoid unnecessary etcd churn which can be expensive if every replica is updating its etcd record on every failover

go/services/multipooler/internal/manager/state_manager.go
dweitzman: What happens in practice is that SetTermPrimary() calls `RecordTermPrimary()` before `SetState` and promotion does `UpdateRule()` before `SetState`, and on success both `RecordTermPrimary` and `UpdateRule` would have caused the value returned by `latestRule` to bump by updated the replicationPrimary and cached position observation respectively

This is a little subtle and I don't love it, but it also seems a little weird to me to take the consensus role as an input here when we should already know this pooler's consensus role based on previous rule changes and/or SetTermPrimary calls

I'm hoping this can be simplified in the future by separating consensus role from serving status. We could say consensus role depends only on the replication primary + WAL / cached WAL and is owned by the consensus engine, serving status is owned by query serving, and either one changing is independent of the other. I haven't tried to prototype that, though, it's just a hand-wavy hope as this point

dweitzman: I decided to revert this aspect of the PR for now and keep having the callers explicitly pass in the pooler type

This PR is focused on adding LeadershipObservation to etcd, and a follow-up PR will deal with what it means to make the consensus information authoritative, in terms of calling conventions and the impact on the postgres monitor

PR #1102, opened 2026-06-05
feat(multipooler): database-aware connection pooling + CREATE DATABASE (prototype)

GuptaManan100:
> **Prototype — not for merge.** This is an exploratory prototype to see what it takes to support multiple logical databases and `CREATE DATABASE` through the pooler. It builds, vets, and passes unit tests for the touched packages, but has not had live end-to-end validation.

## Summary

- A single multipooler can now serve **many logical databases** on the same PostgreSQL instance: connection pools are keyed by `(database, user)` instead of just `user`.
- The client's startup `database` is plumbed end-to-end (gateway → pooler) via a new `ExecuteOptions.database` field, instead of being ignored in favor of the pooler's one configured DB.
- `CREATE DATABASE` / `DROP DATABASE` are no longer rejected — they run as ordinary statements, so a freshly created database is immediately reachable by reconnecting with that dbname.

## Description

Today the gateway reads the client's requested database but only uses it for plan-cache keys, metrics, and tracing — routing always lands on the one database each multipooler is configured for. And the connection pool manager dials a single `connConfig.Database` for every pool. So regardless of the `dbname` a client sends, it effectively talks to `postgres`.

This prototype takes the **multipooler-serves-many-databases** approach (one Postgres instance, many logical DBs):

- **`proto/query.proto`** — adds `string database = 9` to `ExecuteOptions`.
- **Gateway (`scatterconn`)** — a `newExecuteOptions(conn, state)` helper stamps `Database: conn.Database()` on every outbound query RPC. All 13 inline construction sites now go through it so no path can silently omit the database (which would route to the wrong pool).
- **`connpoolmanager`** — pools are keyed by a new `poolKey{database, user}`. Public methods take a `database` arg, per-pool dials override the dbname, and the rebalancer, GC, metrics, and stats all key by `(database, user)`. An empty database resolves to the pooler's configured default, so internal machinery (heartbeat, logical-replication listener) and the admin/maintenance pool stay on the default DB and the multigres internal schema is untouched.
- **`executor`** — extracts the database from options and threads it into every pool acquisition.
- **`planner/unsafe_stmt.go`** — drops the `CREATE DATABASE` / `DROP DATABASE` rejections (other Tier-2 guards remain). PostgreSQL still enforces its own constraints (can't run inside a transaction block, can't drop the DB you're connected to).

## Behavior change worth flagging

Previously the gateway ignored the client's `dbname`, so every connection landed on the configured database. Now `dbname` is honored: connecting to a database that doesn't exist in PostgreSQL surfaces a real "database does not exist" error instead of silently using `postgres`. That's the intended multi-DB semantics, but it's a visible change for clients that relied on the old pass-through.

## Not done yet

- Live end-to-end validation (spin a cluster, `CREATE DATABASE foo`, reconnect with `dbname=foo`).
- The `TestRuleStorePG_*` failures in `multipooler/manager` are pre-existing (they fail on `main` too), unrelated to this change.

### DIFF
--- a/go/pb/query/query.pb.go
+++ b/go/pb/query/query.pb.go
@@ -1111,8 +1111,14 @@ type ExecuteOptions struct {
 	// functions can map virtual PIDs (visible to clients) back to real
 	// backend PIDs via pg_stat_activity.
 	ClientConnectionId uint32 `protobuf:"varint,8,opt,name=client_connection_id,json=clientConnectionId,proto3" json:"client_connection_id,omitempty"`
-	unknownFields      protoimpl.UnknownFields
-	sizeCache          protoimpl.SizeCache
+	// database is the PostgreSQL database the client connected to (from the
+	// startup packet). A single multipooler serves many logical databases on
+	// the same PostgreSQL instance; this selects which (database, user)
+	// connection pool the query runs on. When empty, the multipooler falls
+	// back to its configured default database.
+	Database      string `protobuf:"bytes,9,opt,name=database,proto3" json:"database,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
 }
 
 func (x *ExecuteOptions) Reset() {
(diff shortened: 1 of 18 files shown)


GuptaManan100:
Just a prototype for showcasing work involved.

PR #1103, opened 2026-06-05
test(pgregresstest): run pgmq's regression suite through multigateway

GuptaManan100:
## Summary

- Enroll [tembo-io/pgmq](https://github.com/tembo-io/pgmq) (v1.11.1) as a covered external extension, running all five of its shipped pg_regress files (`base`, `fifo_tests`, `fifo_rr_tests`, `fifo_per_group_tests`, `topic_tests`) through multigateway. All pass with **no patches** — pgmq's output matches upstream's expected output exactly.
- Add two reusable knobs to `ExternalExtension`, both first needed by pgmq: `BuildSubdir` (the PGXS Makefile lives under `pgmq-extension/`, not the repo root) and `DependsOn` (build-only prerequisite extensions).
- Build `pg_partman` (v5.4.3) as a build-only dependency of pgmq via `DependsOn`, so `base.sql`'s partitioned-queue tests (`create_partitioned` → partman's `create_parent`) run for real. It is not tested on its own (it ships a pgTAP suite, not pg_regress) and is recorded as `StatusUnsupported` in the catalog.

## Description

`ExternalBuildList` resolves the selected extensions plus their dependencies (dependencies first, deduped) and honors `PGEXTERNAL_TESTS`, so a narrowed run (e.g. `PGEXTERNAL_TESTS="pgmq"`) builds only what it needs. The build phase iterates this list; the test phase still iterates only the covered set.

One behavior worth flagging: `base.sql` calls `pgmq.create_unlogged()`, and it succeeds even though multigres rejects unlogged tables. That rejection is top-level-statement-only, and pgmq issues the `CREATE UNLOGGED TABLE` as dynamic SQL inside a plpgsql function — the gateway only sees `SELECT pgmq.create_unlogged(...)`, so the guard never fires.

Verified locally with `RUN_PGEXTERNAL=1 PGEXTERNAL_TESTS="pgmq"`: pg_partman and pgmq clone/build/install cleanly against the from-source PostgreSQL, and pg_regress reports all 5 tests pass with an empty `regression.diffs`.

### DIFF
--- a/go/test/endtoend/pgbuilder/builder.go
+++ b/go/test/endtoend/pgbuilder/builder.go
@@ -249,7 +249,13 @@ func (b *Builder) InstallContrib(t *testing.T, ctx context.Context) error {
 // ABI-consistency guarantee Build provides for regress.so). The checkout is
 // per-run, so a shallow clone happens once per invocation; pinning the tag keeps
 // the suite reproducible.
-func (b *Builder) InstallExternalExtension(t *testing.T, ctx context.Context, name, repo, tag string) (string, error) {
+//
+// buildSubdir is the directory within the checkout that holds the PGXS Makefile,
+// relative to the clone root. It is empty for extensions whose Makefile sits at
+// the repo root (pgvector, pg_cron) and set when it lives in a subdirectory
+// (pgmq: "pgmq-extension"). The returned path is always the clone root, so the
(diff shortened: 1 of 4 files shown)


haritabh17 (approved):
🚀

PR #1104, opened 2026-06-05
test(shardsetup): extend timing table with failover and recovery events

mkindahl:
Record elapsed time for WaitForNewPrimary (failover election visible to multipooler) and RequireRecovery (replication fix / rewind) in the TimingCollector summary table. Both functions already hold a *ShardSetup so they can write to setup.Timings / s.Timings directly — no call-site changes required.

Also rename .github/scripts/startup-timing-summary.py to timing-summary.py (the table now covers more than startup) and switch the confidence interval from 95% to 99.9% (t-crit values updated to the 99.95th percentile).

### DIFF
--- a/.github/scripts/timing-summary.py
+++ b/.github/scripts/timing-summary.py
@@ -14,11 +14,11 @@
 # limitations under the License.
 #
 # Reads integration-test-results.jsonl (go test -json output), extracts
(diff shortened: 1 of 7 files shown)


GuptaManan100 (approved):
LGTM!

PR #1106, opened 2026-06-05
feat(cluster): pre-install pgvector in the cluster image

GuptaManan100:
## Summary

- Pre-install pgvector in the all-in-one `multigres-cluster` image so compatibility suites can `CREATE EXTENSION vector` against the bundled PostgreSQL without building it at runtime.
- Install it from the PostgreSQL APT repo (`postgresql-17-pgvector`) that the Dockerfile already wires up for pgBackRest — the PGDG package is built for these exact pg17 binaries, so no build toolchain is added to the image.
- Extend the existing docker-compose smoke test to verify pgvector is actually usable through multigateway, failing loudly if it's missing.

## Description

The cluster image is Debian-based (`postgres:17.7`), so unlike the Alpine source-build approach used elsewhere, pgvector is available as a prebuilt PGDG package. It's folded into the same apt install that already pulls in pgBackRest and procps, keeping it to a single layer with no extra build dependencies.

The smoke-test step in `docker-compose-ci.yml` now runs a small pgvector round-trip through the gateway. The `vector(3)` column type and the `<->` distance operator only resolve when the extension is installed, so with `ON_ERROR_STOP=1` already set, a missing pgvector turns into a hard failure of the step. Verified locally end-to-end: built the image, brought up the compose cluster, and ran the SQL through multigateway (`CREATE EXTENSION` / insert / nearest-neighbor / drop) — all green.

### DIFF
--- a/.github/workflows/docker-compose-ci.yml
+++ b/.github/workflows/docker-compose-ci.yml
@@ -93,6 +93,18 @@ jobs:
           SELECT label FROM compat_check WHERE id = 1;
           DROP TABLE compat_check;
           SQL
+          echo "== pgvector extension =="
+          # The vector(3) column type and the <-> distance operator only resolve
+          # if pgvector is installed in the image, so ON_ERROR_STOP=1 turns a
+          # missing extension into a hard failure of this step.
+          run_sql <<'SQL'
+          CREATE EXTENSION vector;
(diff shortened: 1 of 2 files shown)


rafael (approved):
LGTM

ferhatelmas:
Amazing, thanks ❤️

GuptaManan100:
Our pleasure! ❤️

PR #1107, opened 2026-06-05
refactor(testtiming): record test timings as testing.T attributes

dweitzman:
Replace the t.Log timing table + Python summary script with Go 1.25 testing.T attributes. A shared go/tools/testtiming package owns the wire contract (Record to emit, Parse to consume); go/tools/testtiming/summary renders the percentile table for CI using gonum for the stats.


go/tools/testtiming/summary/main.go
```diff
@@ -0,0 +1,205 @@
...
+	for _, label := range labels {
+		vals := data[label]
+		limit := limits[label]
+		mean, margin := meanCI95(vals)
+
+		// stat.Quantile requires sorted input; sort once and reuse for min/max.
+		sorted := append([]float64(nil), vals...)
+		sort.Float64s(sorted)
+		minV, maxV := sorted[0], sorted[len(sorted)-1]
+		p50 := stat.Quantile(0.50, stat.LinInterp, sorted, nil)
```
dweitzman, line 192: `gonum/stat` is doing the heavy lifting here on the percentile math

go/tools/testtiming/summary/main_test.go
dweitzman, line 1: This PR is a net reduction is production code size, but an increase in overall code because of the two added test files

go/tools/testtiming/summary/main.go
```diff
@@ -0,0 +1,205 @@
...
+	warnPct = 70.0
+	critPct = 90.0
+)
+
+// poolerSuffix matches the per-instance suffix that varies between otherwise
+// identical operations (e.g. "manager ready: pooler-3"), so they aggregate into
+// a single row.
+var poolerSuffix = regexp.MustCompile(`: pooler-\d+$`)
+
+// meanCI95 returns the mean and the half-width of a 95% CI on the mean, assuming
```
mkindahl: I think it is better to use 99% rather than 95%. With 95% we will get one out of 20 "flakes" if we use that upper bound, which is a little too often. With 99% we would have one out of 100 flakes, which is manageable.

go/tools/testtiming/summary/main.go
```diff
@@ -0,0 +1,205 @@
...
+// normality. For n<2 the half-width is 0 (a CI cannot be computed from a single
+// sample). The t critical value comes from the exact Student's t quantile.
+func meanCI95(values []float64) (mean, halfWidth float64) {
+	n := len(values)
+	mean = stat.Mean(values, nil)
+	if n < 2 {
+		return mean, 0.0
+	}
+	tCrit := distuv.StudentsT{Mu: 0, Sigma: 1, Nu: float64(n - 1)}.Quantile(0.975)
```
mkindahl: Thank you for using Student's rather than Gaussian. Since we have few samples, this will show the situation better.  Here is the fix to use 99% instead of 95%.
```suggestion
	tCrit := distuv.StudentsT{Mu: 0, Sigma: 1, Nu: float64(n - 1)}.Quantile(0.995)
```

go/tools/testtiming/summary/main.go
```diff
@@ -0,0 +1,205 @@
...
+		data[label] = append(data[label], m.Elapsed.Seconds())
+		limits[label] = m.Limit.Seconds()
+	}
+	return data, limits
+}
+
+func writeTable(out io.Writer, data map[string][]float64, limits map[string]float64) error {
+	w := bufio.NewWriter(out)
+	fmt.Fprint(w, "## Test Timing\n\n")
+	fmt.Fprint(w, "| Operation | Timeout | mean ±95%CI | min | p50 | p95 | p99 | max |\n")
```
mkindahl:
```suggestion
	fmt.Fprint(w, "| Operation | Timeout | mean ±99%CI | min | p50 | p95 | p99 | max |\n")
```

mkindahl (approved):
Suggest to use 99% CI instead of 95%. If we start to follow that advice, we will get 1 out of 100 flakes rather than 1 out of 20 flakes.

go/tools/testtiming/summary/main.go
dweitzman: Done

PR #1108, opened 2026-06-06
fix(pgctld): treat EPERM as alive and probe connectivity for status

### LINKED ISSUE #628: Connection pool closes after ~5-15min idle when pgctld runs as non-postgres OS user (MonitorPostgres EPERM on kill)
## Summary

When pgctld runs as a different OS user than the `postgres` process (e.g., a dedicated `multigres` user), `MonitorPostgres` closes the connection pool after ~5-15 minutes of idle time — even though PostgreSQL is running and healthy.

**Reproduction: 100% reliable under this condition.**

If pgctld runs as the `postgres` OS user (same UID as the PG process), the pool stays alive and this does not trigger.

## When Does This Happen

This affects the common evaluation/benchmarking setup where:
- PostgreSQL is installed via system package (runs as `postgres` OS user, data in `/var/lib/postgresql/...`)
- Multigres runs as a separate OS user (e.g., `multigres`)
- pgctld is started as the `multigres` user

In a setup where pgctld manages PG itself (starts and owns the PG process), this likely doesn't trigger since pgctld and PG share the same UID.

## Steps to Reproduce

```bash
# 1. PostgreSQL running as 'postgres' OS user (standard apt install)
sudo -u postgres pg_lsclusters  # PG up, port 5432

# 2. Start pgctld as 'multigres' user (NOT as postgres)
su - multigres -c '/path/to/pgctld server \
  --pooler-dir /home/multigres/data/pooler_test \
  --pg-port 5432 --grpc-port 15470 --http-port 15400 &'

# 3. Start multipooler + multigateway normally

# 4. Verify it works
PGPASSWORD=x psql -h 127.0.0.1 -p 15432 -U postgres -c 'SELECT 1'
# → returns 1 ✓

# 5. Wait 5-15 minutes with no traffic

# 6. Try again
PGPASSWORD=x psql -h 127.0.0.1 -p 15432 -U postgres -c 'SELECT 1'
# → ERROR: connection pool is closed

# Meanwhile, direct PG still works fine:
psql -h 127.0.0.1 -p 5432 -U postgres -c 'SELECT 1'
# → returns 1 ✓
```

## Root Cause

`MonitorPostgres` calls `pg_ctl status` every 5 seconds to verify PG is alive. Internally, `pg_ctl status` does `kill(pid, 0)` to check if the process exists. When pgctld runs as a different user than the PG process, this syscall returns `EPERM` — the OS rejects the signal check.

`MonitorPostgres` interprets this as "PostgreSQL is not running" and tries to restart it. When `pg_ctl start` also fails (PG is already running, started by a different user), it logs the error and loops every 5 seconds. Eventually the pool is marked closed:

```json
{"level":"INFO","msg":"MonitorPostgres: Attempting to restart PostgresSQL"}
{"level":"ERROR","msg":"MonitorPostgres: failed to start PostgreSQL, will retry",
 "error":"failed to start PostgreSQL with pg_ctl: exit status 1"}
... (every 5s) ...
{"level":"INFO","msg":"reserved pool closed","user":"postgres"}
```

The design issue: `MonitorPostgres` conflates "I can't manage PG via pg_ctl" with "PG is actually down." A TCP health check (`SELECT 1` on port 5432) would correctly determine PG is healthy regardless of OS user ownership.

## Workaround

Run pgctld as the `postgres` OS user:

```bash
# Allow postgres user to access multigres binaries
chmod 755 /home/multigres /home/multigres/multigres/bin
chmod 755 /home/multigres/multigres/bin/pgctld

# Run pgctld as postgres
su - postgres -c '/home/multigres/multigres/bin/pgctld server \
  --pooler-dir /home/multigres/multigres_local/data/pooler_test \
  --pg-port 5432 --grpc-port 15470 --http-port 15400 &'
```

## Expected Behavior

`MonitorPostgres` should use a TCP health check (connect to PG on the configured port and run `SELECT 1`) as the primary liveness check, with `pg_ctl status` as an optional management-plane check. Pool closure should not occur when the database is actively accepting connections.

## Environment

- **Multigres:** `main` branch, commit `66af929`
- **PostgreSQL:** 17.8 (Ubuntu apt.postgresql.org, running as `postgres` OS user)
- **OS:** Ubuntu 24.04, kernel 6.8.0-90-generic
- **Hardware:** Hetzner CCX43 (16 dedicated AMD EPYC vCPU, 64GB RAM)
- **Setup:** Single-node, pgctld + multipooler + multigateway + etcd

vatsalpatel:
## Summary

Fixes #628. When `pgctld` runs as a **different OS user** than the `postgres` process (e.g. a dedicated `multigres` user managing a system-package PostgreSQL), the connection pool was torn down after ~5-15 min of idle time even though PostgreSQL was healthy and accepting connections - surfacing as `ERROR: connection pool is closed`.

## Root cause

multipooler's `MonitorPostgres` loop derives PG liveness from pgctld's `Status()` RPC (`GetStatusWithResult`). That function short-circuited to `STOPPED` whenever a local process-existence check failed - **before** ever running the `pg_isready` connectivity check that would have seen PG is up.

The process check (`isProcessRunning`) uses `kill(pid, 0)` (`process.Signal(syscall.Signal(0))`), which returns `EPERM` when the postmaster is owned by another OS user. The code treated `EPERM` (process exists, not ours) identically to `ESRCH` (process gone) - both -> `false`.

So pgctld reported `STOPPED` -> multipooler ran `remedialActionStartPostgres` -> `pg_ctl start` failed (`exit status 1`) -> it flapped every 5s and the pool was eventually closed.

## Changes (all in pgctld)

- **`isProcessRunning` is EPERM-aware** - `EPERM` now means "alive" (the kernel did a permission check against a real process). Only `ESRCH`/`ErrProcessDone` means gone. Also corrects the `isPostgreSQLRunning` guards used by start/stop/restart/reload.
- **`GetStatusWithResult` makes connectivity authoritative** - the cheap PID check stays as the fast path, but when it can't confirm a postmaster (cross-user signal failure *or* unreadable PGDATA), it falls back to a `pg_isready` probe before declaring `STOPPED`. A reachable PostgreSQL is reported `RUNNING` regardless of process ownership (and logs a warning explaining the likely cause). Frozen-process detection (`exists but not accepting connections -> STOPPED`) is preserved.

| process check | pg_isready | result | vs. before |
| --- | --- | --- | --- |
| true | true | RUNNING | same (fast path) |
| true | false | STOPPED (frozen) | same |
| false | true | **RUNNING** | **fixed** (was STOPPED) |
| false | false | STOPPED | same |

## Testing

- New `TestIsProcessRunning`: self-pid -> alive; killed+reaped pid -> dead; `kill(1, 0)` EPERM -> alive (skips when run as root). Fails before the fix, passes after.
- `status_test.go`: updated the `stopped` case to model a genuinely stopped server (no process **and** `pg_isready` fails), and added `reachable_without_pidfile` - the direct regression test (no PID file -> reachable -> `Running`).
- `go/cmd/pgctld/command`: 221 passing; `gofumpt`/`go vet` clean.

## Known limitation

The fallback probes `pg_isready` over the pooler's unix socket (`pgctld.PostgresSocketDir`), which is where a pgctld-managed PG places its socket. A fully external PostgreSQL whose socket lives elsewhere (e.g. `/var/run/postgresql`) won't be probed by this path - but pgctld can't manage (start/stop) such an instance anyway, and `ensurePGDATAPermissions` already assumes pgctld owns PGDATA. Out of scope here.


vatsalpatel:
@mkindahl Is there anything I can do to make the review easier? For example, splitting the EPERM liveness from the status connectivity probe?

cuongdo:
Hi @vatsalpatel! 

Thanks for the PR. Could you address the following the integration test failures in `go/test/endtoend/pgctl`, which look related to the changes.

vatsalpatel:
Hi @cuongdo 
I have fixed the tests. Please review the changes

go/cmd/pgctld/command/start.go
```diff
@@ -241,20 +241,60 @@ func ensurePGDATAPermissions(logger *slog.Logger, dataDir string) error {
...
+	if isProcessRunning(pid) {
+		return pgAlive, pid
 	}
+	return pgDown, 0
+}
 
-	return isProcessRunning(pid)
+func isPostgreSQLRunning(dataDir string) bool {
+	liveness, _ := postgresLiveness(dataDir)
+	return liveness == pgAlive
```
cuongdo: This function discards the `pgLiveness` tri-state value for true/false, losing the new information `postgresLiveness` provides. This means that the bug you're fixing for `status` remains for the other `pgctld` commands such as `stop` and `restart`. I suggest replacing calls to `isPostgreSQLRunning` with `postgresLiveness`, just like you did in `status.go`.

vatsalpatel: Resolved @cuongdo

vatsalpatel:
Addressed the changes @cuongdo 
Please review again
Also I had to force push since committing reformatted imports across go/pb/, the amended commit drops the auto formatted files and brings the diff back down to the 5 files with the tri-state change

vatsalpatel:
The tests were failing after merging main, pushed a commit fixing them

PR #1109, opened 2026-06-08
fix(multipooler): return honest 40001 instead of MTF01 for transactions aborted by failover

GuptaManan100:
## Summary

- A client holding a transaction open across a planned failover for longer than the connection-drain grace period (`--connpool-drain-grace-period`, default 3s) would get `MTF01` ("planned failover in progress, will be retried automatically") leaked to the client. The error is misleading: the transaction was already rolled back, nothing buffers or retries it, and to the app the query had "seemed to succeed."
- The fix makes the pooler return an honest, retryable `40001` (`serialization_failure`) instead, so clients/ORMs retry the whole transaction.

## Problem

`MTF01` is a pre-execution admission gate (`StartRequest`). For an in-flight op on an existing reserved connection (a COMMIT/ROLLBACK or a mid-transaction statement), the gateway routes it back to the specific pooler that holds the connection. If a planned failover demoted that pooler and the drain grace period expired while the client sat idle-in-transaction, the reserved connection was force-closed (transaction rolled back) and the pooler demoted to REPLICA. The late op then hit `StartRequest`'s serving-status / demotion checks and returned `MTF01` — un-buffered, straight to the client. `MTF01`'s contract ("will be retried automatically") is wrong here: the transaction no longer exists, so only the client can replay it.

This only affects transactions / multi-statement batches. Plain autocommit statements are buffered and unaffected.

## Solution

`StartRequest`'s `allowOnShutdown` flag is already set exactly for in-flight ops on an existing reserved connection (`ConcludeTransaction`, `DiscardTempTables`, `ReleaseReservedConnection`, and mid-transaction statements). It now also bypasses the serving-status and demotion checks — not just the `shuttingDown` check — so the reserved connection's existence becomes the real gate:

- Connection survived the drain → the op completes on the live backend (postgres is demoted only after the drain finishes).
- Connection was force-closed → the executor returns the new `40001` (`serialization_failure`) so the client retries.

Adds `mterrors.NewTransactionAborted` and updates the graceful-drain design doc + `StartRequest` comments.

Tests: unit coverage for the gate (`pooler_test.go`) and the executor's connection-gone path (`executor_test.go`), plus an e2e regression test (`transaction_failover_test.go`) that holds a transaction across a real failover. Verified it fails before this change (leaks `MTF01`) and passes after (returns `40001`).

### DIFF
--- a/go/common/mterrors/code.go
+++ b/go/common/mterrors/code.go
@@ -41,6 +41,7 @@ const (
 	PgSSQueryCanceled           = "57014" // query_canceled
 	PgSSInternalError           = "XX000" // internal_error
 	PgSSReadOnlyTransaction     = "25006" // read_only_sql_transaction
+	PgSSSerializationFailure    = "40001" // serialization_failure
 )
 
 // NewQueryCanceled creates a PgDiagnostic for an explicit cancel request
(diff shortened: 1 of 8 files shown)


go/common/mterrors/code.go
```diff
@@ -57,6 +58,21 @@ func NewStatementTimeout() *PgDiagnostic {
...
+// force-closed when a planned failover drain exceeded its grace period while the
+// client sat idle-in-transaction. SQLSTATE 40001 (serialization_failure) so
+// clients and ORMs retry the whole transaction; the data was rolled back, so a
+// retry is the correct recovery.
+//
+// This is deliberately NOT MTF01: the transaction is gone, so there is nothing
+// to buffer or auto-retry inside the cluster — only the client can replay it.
+func NewTransactionAborted(reservedConnID uint64) *PgDiagnostic {
+	return NewPgError("ERROR", PgSSSerializationFailure,
+		"transaction aborted: connection terminated during a planned failover; please retry the transaction",
```
haritabh17: nit: A reserved connection held solely for say`temp tables` is not in a `transaction`, so the error message and SQLSTATE are semantically inaccurate.

haritabh17 (approved):
Rest LGTM 🚀

go/common/mterrors/code.go
GuptaManan100: Correct, I should use the reservation reasons to fix the error message! Good point, it didn't cross my mind!

PR #1110, opened 2026-06-08
test(pgregresstest): support pgrx extensions and enroll pg_graphql

GuptaManan100:
## Summary

- Add a pgrx (Rust/cargo-pgrx) build path to the external extension harness and enroll Supabase [pg_graphql](https://github.com/supabase/pg_graphql) (v1.6.1) as the first covered pgrx extension. Its full pg_regress suite runs through multigateway: **121/121** (120 clean + 1 documented patch).
- New reusable `ExternalExtension` knobs: `BuildSystem` (`pgxs`/`pgrx`) + `PgrxVersion` (pins cargo-pgrx to the crate's pgrx dep), `FixturesFile` (psql-loads a setup file through the gateway before the suite), and `ContribDeps` (installs contrib modules a suite needs, e.g. citext).
- CI provisions the Rust toolchain (`dtolnay/rust-toolchain`, pinned by SHA); the harness installs the matching cargo-pgrx itself.

## Description

For pgrx extensions, `InstallExternalExtension` installs the pinned cargo-pgrx, runs `cargo pgrx init --pgNN <pg_config>` to adopt the from-source PostgreSQL, then `cargo pgrx install --pg-config <pg_config>` so the `.so` links against the exact server ABI (the same guarantee the PGXS path gives). pg_graphql's `bin/installcheck` loads `test/fixtures.sql` first (it creates the extension and sets the graphql schema comment) and its tests `create extension citext`, hence `FixturesFile` and `ContribDeps`.

The one patch (`resolve_error_mutation_no_field`) is **not** a multigres bug — multigres rolls the transaction back correctly and reflects committed DDL correctly. pg_graphql caches the reflected schema in backend-local memory and only invalidates on committed DDL; a `begin; create table; resolve; rollback;` test poisons that backend's cache, and because multigres pools and reuses backends (unlike stock pg_regress, which gets a fresh backend per test), a later test sees the stale schema. It's the same backend-local-state class as hypopg and narrow (rolled-back in-txn DDL only). The full rationale lives in the patch file's header comment.

### DIFF
--- a/go/test/endtoend/pgregresstest/testdata/pg17/patches/external/pg_graphql/resolve_error_mutation_no_field.patch
+++ b/go/test/endtoend/pgregresstest/testdata/pg17/patches/external/pg_graphql/resolve_error_mutation_no_field.patch
@@ -1,0 +1,32 @@
+# Known divergence: pg_graphql's backend-local schema cache goes stale on a
+# reused pooled backend. This is NOT a multigres bug.
+#
+# pg_graphql caches the reflected GraphQL schema in backend-process memory and
+# only invalidates that cache on COMMITTED DDL. An earlier test shaped like
+# `begin; create table account(...); graphql.resolve(...); rollback;` builds the
+# cache from the uncommitted table but stamps it with the unchanged committed
+# schema version, leaving the backend holding a schema for a table that no longer
+# exists (so a Mutation type is present).
+#
+# On stock PostgreSQL each pg_regress test runs on a fresh, short-lived backend,
+# so that poisoned cache dies with the test and this test sees an empty schema
+# ("Unknown type Mutation"). multigres pools and REUSES backends, so this test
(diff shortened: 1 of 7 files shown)


go/test/endtoend/pgregresstest/testdata/pg17/patches/external/pg_graphql/resolve_error_mutation_no_field.patch
```diff
@@ -0,0 +1,32 @@
+# Known divergence: pg_graphql's backend-local schema cache goes stale on a
+# reused pooled backend. This is NOT a multigres bug.
```
GuptaManan100, line 2: There is one patch here, worth looking at why. I've done my best to explain it.

haritabh17 (approved):

PR #1111, opened 2026-06-08
ci: benchmark postgres target in weekly pgbench so overhead summary is non-empty

GuptaManan100:
## Summary

- The weekly pgbench Slack summary builds a postgres-vs-multigateway overhead comparison, but the workflow never set `PGBENCH_TARGETS`, so `parseTargets()` defaulted to multigateway-only. With no `postgres` rows the comparison was always empty — every run posted "multigateway overhead 0% avg" with a header-only table.
- Set `PGBENCH_TARGETS: "postgres,multigateway"` so both targets are benchmarked and the overhead table has data.
- Harden the summary `jq`: when no comparison can be built it now posts a `:warning:` message naming the targets actually present, instead of silently posting 0%.
- Stop PR-labeled runs from overwriting the baseline, so one-off checks don't pollute the week-over-week regression comparison.

## Problem

Confirmed against the latest scheduled run's artifact: `results.json` contained 10 rows, all `target: "multigateway"`, zero `postgres`. The summary `jq` builds `$pg_tps` from `select(.target == "postgres")` → empty, so `$comparisons` is empty and it falls into the `else 0` branch. The job still succeeds, which is why the breakage went unnoticed — the data was simply missing a baseline.

## Solution

Provide the postgres target explicitly via env, and make the missing-baseline case loud rather than silent. Both `jq` branches were verified locally against the real artifact (multigateway-only → warning; synthetic two-target → proper table).

Baseline-save is now gated on `github.event_name != 'pull_request'` (matching the Slack summary step), so `Run Benchmarks` PR runs are safe throwaway checks while scheduled and `workflow_dispatch` runs still update the baseline.

Note: the regression-detection script only ever compares `multigateway` rows, so it keeps working across the one-run baseline transition from single- to two-target results.

### DIFF
--- a/.github/workflows/test-pgbench.yml
+++ b/.github/workflows/test-pgbench.yml
@@ -266,7 +266,10 @@ jobs:
             "$SLACK_WEBHOOK_URL"
 
       - name: Prepare baseline for cache save
-        if: ${{ !cancelled() && steps.results.outputs.path != '' }}
+        if: >-
+          !cancelled() &&
+          github.event_name != 'pull_request' &&
+          steps.results.outputs.path != ''
         env:
           RESULTS_PATH: ${{ steps.results.outputs.path }}
         run: |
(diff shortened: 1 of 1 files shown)


niconosenzo (approved):
lgtm

haritabh17 (approved):
🚀

PR #1112, opened 2026-06-08
fix(gateway): forward upstream PgDiagnostic on credential-lookup failure

mkindahl:
When the credential provider returns a *mterrors.PgDiagnostic (e.g. "ERROR: planned failover in progress" from an upstream pooler that is mid-election), authenticateSCRAM was discarding it and sending the client a generic "FATAL: password authentication failed" (28P01). This masked a transient cluster condition as a permanent auth failure.

In the mg-scale12 AZ-outage test (2026-05-29), MUL-557 left the cluster stuck in a cascade re-election loop; every credential-lookup hit a pooler in "planned failover in progress" state, and clients saw 28P01 with no indication that retrying would succeed.

Forward the PgDiagnostic verbatim, promoting its severity to FATAL (required for connection teardown during the auth phase). Opaque transport/parse errors that do not carry a PgDiagnostic continue to fail closed with the opaque password-auth message to avoid leaking whether the user exists.

Fixes MUL-585


haritabh17 (approved):
🚀

PR #1113, opened 2026-06-08
fix(multiorch): eliminate blocking recruit drain; propose to full cohort after quorum

mkindahl:
## Problem

`coordinatorLedRuleChange.Run()` launches one Recruit goroutine per cohort member. Phase 2 collected Recruit responses with a **blocking drain**: it called `<-recruited` for every remaining goroutine and waited for each to complete before dispatching Promote results. Any node whose Recruit RPC was slow or timed out would stall `Run()` for the full RPC duration, even though quorum was already satisfied in Phase 1 and the leader's Promote was ready to go.

## Fix

Remove the blocking drain entirely. As soon as Phase 1 elects a leader, dispatch Promote/SetPrimary to every cohort member unconditionally — including nodes whose Recruit goroutines are still in flight or have failed. Recruit goroutines still in flight run to completion on their own without blocking the proposal phase. Node health can change during the recruit phase, so Recruit is sent to every cohort member unconditionally and the response race determines leader selection.

Also adds latency instrumentation to the eventlog calls: `recruit_ms` (time to quorum) and `promote_ms` (proposal phase duration) are now emitted on the terminal Success/Failed events, and a Started event is emitted before any RPCs go out so the full appointment can be traced end-to-end.

## Tests

- **`TestRun_SlowRecruitDoesNotBlockAfterQuorum`** — mp2's Recruit is configured with a 20 s blocking delay; mp1 alone satisfies quorum. Asserts that `Run()` completes in under 2 s and that mp2's Recruit RPC was still issued (no pre-filtering on health status).
- **`TestRun_StragglersGetSetTermPrimary`** — mp2's Recruit is gated and released from inside `tryBuildProposal` after Phase 1 commits. Asserts that mp1 receives `Promote` (as leader) and mp2 receives `SetPrimary`.
- **`RecruitDelays` and `RecruitGates`** added to `FakeClient` to support delay injection and precise goroutine sequencing in tests.

Fixes MUL-555

### DIFF
--- a/go/services/multiorch/consensus/rule_change.go
+++ b/go/services/multiorch/consensus/rule_change.go
@@ -116,16 +116,26 @@ func (r *coordinatorLedRuleChange) Run(
 	// selected → commit). Monotonic, so unaffected by wall-clock adjustments.
 	start := time.Now()
 	var selectedAt time.Time
-	eventlog.Emit(ctx, r.coordinator.logger, eventlog.Started, eventlog.PrimaryPromotion{
+	primaryPromotion := eventlog.PrimaryPromotion{
 		ProposedTerm: proposedTerm,
 		Reason:       r.reason,
-	})
+	}
+	eventlog.Emit(ctx, r.coordinator.logger, eventlog.Started, primaryPromotion)
 
-	// Recruit all nodes concurrently.
+	// Phase 1: Recruit all nodes concurrently.
+	//
+	// Each recruit returns a ConsensusStatus that reflects the node's current WAL
+	// position and durability commitments. The coordinator accumulates these and
+	// tries to build a proposal after each arrival, until one succeeds. This
+	// lets the coordinator select the most advanced node that can still satisfy
+	// the durability policy, without waiting for the slowest nodes in the
+	// cohort to respond.
+
 	type recruitResult struct {
 		pooler *multiorchdatapb.PoolerHealthState
 		cs     *clustermetadatapb.ConsensusStatus // nil if recruit failed
 	}
+
 	recruited := make(chan recruitResult, len(cohort))
 	for _, p := range cohort {
 		go func() {
@@ -135,29 +145,12 @@ func (r *coordinatorLedRuleChange) Run(
 
 	var (
 		statuses  []*clustermetadatapb.ConsensusStatus
-		recruits  []*multiorchdatapb.PoolerHealthState
 		propReq   *consensusdatapb.PromoteRequest
 		leaderKey string
 	)
 
-	type promoteResult struct {
-		poolerName string
-		isLeader   bool
-		err        error
-	}
-	promoteResults := make(chan promoteResult, len(cohort))
-
-	// dispatchProposal sends Promote to the designated leader and SetPrimary
-	// to every other cohort member. r.promote handles the dispatch internally.
-	dispatchProposal := func(p *multiorchdatapb.PoolerHealthState) {
-		isLeader := topoclient.ClusterIDString(p.MultiPooler.Id) == leaderKey
-		go func() {
-			err := r.promote(ctx, p, propReq, isLeader)
-			promoteResults <- promoteResult{poolerName: p.MultiPooler.Id.Name, isLeader: isLeader, err: err}
-		}()
-	}
-
-	// Phase 1: collect recruits until tryBuildProposal succeeds (or we run out).
+	// Phase 2: collect recruits until tryBuildProposal succeeds, or all nodes
+	// have responded without reaching quorum.
 	//
 	// We commit to the leader at the FIRST successful tryBuildProposal and stream
 	// Proposes from there — recruits arriving after that point don't change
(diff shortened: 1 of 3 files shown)


go/services/multiorch/consensus/rule_change.go
```diff
@@ -284,6 +284,14 @@ func (r *coordinatorLedRuleChange) recruit(
 	p *multiorchdatapb.PoolerHealthState,
 	revocation *clustermetadatapb.TermRevocation,
 ) *clustermetadatapb.ConsensusStatus {
+	// The health stream already knows this node is unreachable — skip the RPC
+	// entirely. Attempting the call would block for RemoteOperationTimeout (15s)
+	// before returning nil, which would stall Phase 2 for every failed node.
+	if !p.IsLastCheckValid {
```
dweitzman: There's existing logic for pruning the `liveCohort` at https://github.com/multigres/multigres/blob/17a037e4056c6c0a62a422981e3a84425e86bd41/go/services/multiorch/consensus/coordinator.go#L102

Some thoughts:
- there's probably a need to be more crisp about what "live" means, but the "seen in the last 10s" threshold already in the code doesn't feel totally unreasonable. It's not being used for bootstrap / `certified` rule changes where an external agent is supplying the list of cohort members
- `IsLastCheckValid` doesn't feel like an appropriate measure of reachability since it [can be false](https://github.com/multigres/multigres/blob/17a037e4056c6c0a62a422981e3a84425e86bd41/go/services/multiorch/recovery/health_stream.go#L471) during momentary connectivity blip

dweitzman: In terms of a more crisp definition of `live`, maybe the thing to do is have `ConsensusStatus` also include some kind of `RecruitIneligible` boolean or something in that spirit?

That would allow reachable poolers to advertise that they can't be recruited (maybe they're still in the middle of postgres setup) and unreachable poolers to be synthetically marked as non-recruitable

mkindahl: As we discussed, filtering on health status (alive / dead) to decide what poolers to try to recruit is probably not what we want for the following reasons:

1. We can have too few pooler available to propose safely, meaning that if we revoke the current rule we can get stuck. If we do not revoke the current rule, the shard might recover.
2. Some poolers that are not healthy when we start the Recruit can become healthy while we recruit poolers, and in that case it makes sense to include them.
3. Some poolers that are healthy when we start the Recruit might become unhealthy and not respond in the Recruit phase. In this case, we might end up with fewer poolers than necessary to finish the Recruit.

As mentioned in the discussion, I think that we need to:
1. Gate the start of the recruit round on that we have enough healthy poolers to actually be able to move forward once we have recruited them. This means that if the leader is suspected dead, we should not try to revoke the current rule until we know that we can promote a new rule. I think this is out of scope for this PR.
2. We should include all poolers in the cohort in the recruit phase. This is related to what you mention here, but I think we can submit a separate PR for this once we have sorted it out.
3. We should move over from the recruit phase to the propose phase once we have enough poolers responding to avoid stragglers blocking the leader promotion. Stragglers will be configured as standbys once they respond. This solves the original problem this PR was going to fix, so I have updated the PR with this approach instead.

mkindahl: Issue 2 is now available as PR #1131

go/services/multiorch/consensus/rule_change.go
dweitzman: I think actually this entire logic could actually be simplified -- only Propose() needs to wait for recruitment and that should have happened anyway for the chosen leader to be eligible

I think once there's a proposal available, it might fine to:
- unconditionally dispatch all the SetTermPrimary() in background goroutines -- no waiting for recruitment needed
- unconditionally dispatch the Propose(), since the eligible leader should have been recruited already with no additional waiting needed

Then the overall rule change method can return immediately when any of:
  - Propose() succeeds, overall success
  - Propose() fails, overall failure
  - optional TODO: overall failure can happen early if so many SetTermPrimary() calls failed that it looks like Propose() couldn't achieve quorum. This doesn't seem essential right now

dweitzman (approved):

go/services/multiorch/consensus/rule_change.go
mkindahl: Yes, this is what is discussed [above](https://github.com/multigres/multigres/pull/1113#discussion_r3379952215).

mkindahl: Now implemented in the last change. The rule change goes through these steps:

1. **Recruit (fan-out):** launch one Recruit goroutine per cohort member. No filtering on health status.
2. **Collect:** receive Recruit responses until tryBuildProposal succeeds and quorum is reach (or all nodes have responded without reaching quorum).
3. **Promote:** dispatch Promote/SetPrimary to every cohort member concurrently.
4. **Wait:** collect results from all cohort members until the leader's result arrives.

go/services/multiorch/consensus/rule_change.go
mkindahl: We are now mostly covering issue 2 in this PR as well. I'll rebase #1131 on this PR once merged, but don't expect any changes.

PR #1114, opened 2026-06-08
refactor(multipooler): reserve-and-run portals via reservation_options

GuptaManan100:
## Summary

- Adds `reservation_options` to `PortalStreamExecuteRequest` so the multipooler can reserve a backend **and** run the portal in one RPC, mirroring the simple-protocol `StreamExecute` path.
- Replaces the gateway's no-op `SELECT 1` "reserve trick" on the extended-protocol path (a fake query + extra round trip) with a real reserve-and-run.
- Closes the simple-vs-extended asymmetry: per-statement reservation features can now attach their reason to the real portal query instead of special-casing the portal path.

## Description

On `main`, `PortalStreamExecute` could only run a portal on a connection that was *already* reserved. When a portal needed a fresh reservation (open transaction or pending temp table), `ScatterConn` faked it by first running a throwaway `StreamExecute(ctx, target, "SELECT 1", ..., reservationOpts, ...)` to create the reserved connection, then ran the real portal on it by `reserved_connection_id`. That's a wasted round trip and splits "establish the reservation" from "run the statement."

This threads `ReservationOptions` through the portal path end to end:

- **proto / queryservice / poolergateway**: `PortalStreamExecute` carries `reservation_options`.
- **multipooler executor**: `portalExecuteWithReserved` reserves-and-runs atomically — runs `BEGIN`/`BeginWithQuery` first when `ReasonTransaction` is set (after the `ensurePrepared` Parse), ORs the reasons onto the connection, then Bind/Executes the real portal. Existing reservations get the new reasons OR'd in. Suspended-portal (`maxRows > 0`) handling is unchanged.
- **grpc handler**: validates reasons at the trust boundary and forwards `reservation_options`.
- **gateway `ScatterConn.PortalStreamExecute`**: builds reservation options from transaction / temp-table state the same way the simple path does and passes them on the portal RPC; deletes the `SELECT 1` / `noopEo` / `noopCallback` scaffolding.

Behavior-preserving for transactions and temp tables. Verified against the existing extended-protocol suites (cursors, `WITH HOLD`, `CREATE TEMP TABLE`, suspended portals) plus new focused `scatterconn` unit tests asserting reservation happens via the portal RPC with no `SELECT 1` round trip.


haritabh17 (approved):
Looks good

PR #1119, opened 2026-06-09
feat(consensus): structured observability events for recruit and propose rounds

mkindahl:
## Problem

When a Propose round fails — because postgres became unavailable between Recruit and Propose, or because a standby's SetTermPrimary failed — there was no structured signal in the event log. The only evidence was unstructured log lines scattered across multipooler and multiorch. Correlating a failing Propose round with postgres state changes (socket gone, postgres restarting) required manual grep across multiple services.

### Three new consensus events (multipooler)

| Event                   | Handler          | Emitted when                                           |
| ----------------------- | ---------------- | ------------------------------------------------------ |
| `consensus.recruit`     | `Recruit`        | A Recruit attempt starts / succeeds / fails            |
| `consensus.promote`     | `Propose`        | The designated leader attempts pg_promote + rule write |
| `consensus.set_primary` | `SetTermPrimary` | A standby reconfigures primary_conninfo                |

All three carry a `rule_number` field formatted as `"coordinator_term.leader_subterm"` (e.g. `"6.0"`), which is the correlation key tying the three events together across nodes in the same election round. Each event emits Started before the work begins, then Failed (with `error`) or Success on completion.

`consensus.recruit` is a rename of the pre-existing `term.begin` event; the field name changes from `new_term` (int) to `rule_number` (string).

For `consensus.promote`, input validation and the non-leader identity check run before the Started event so routing mistakes do not generate spurious started+failed pairs.

For `consensus.set_primary`, the event is only emitted after both no-op gates pass (rule not revoked, incoming rule strictly higher than local position), so idempotent retries do not generate events.

### FormatRuleNumber helper

`commonconsensus.FormatRuleNumber(rn)` renders a `*RuleNumber` as `"coordinator_term.leader_subterm"` for logs. Nil renders as `"none"`. Used by all three new event types.

## Files changed

| File                                                        | Change                                                                                                             |
| ----------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------ |
| `go/common/consensus/compare.go`                            | Add `FormatRuleNumber`                                                                                             |
| `go/common/eventlog/events.go`                              | Rename `TermBegin` → `ConsensusRecruit`; replace `TermPromotion` with `ConsensusPromote` and `ConsensusSetPrimary` |
| `go/services/multipooler/internal/manager/rpc_consensus.go` | Emit events in `Recruit`, `Propose`, `SetTermPrimary`; move validation before event setup in `Propose`             |
| `go/test/endtoend/multiorch/demote_stale_primary_test.go`   | Update `WaitForEvent` assertion from `term.begin` to `consensus.recruit`                                           |

### DIFF
--- a/go/common/consensus/compare.go
+++ b/go/common/consensus/compare.go
@@ -16,6 +16,8 @@
 package consensus
 
 import (
+	"fmt"
+
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	"github.com/multigres/multigres/go/tools/pgutil"
 )
@@ -51,6 +53,17 @@ func CompareRuleNumbers(a, b *clustermetadatapb.RuleNumber) int {
 		return 1
 	}
 	return 0
+}
+
+// FormatRuleNumber renders a RuleNumber for human-readable logs as
+// "coordinator_term.leader_subterm" (e.g. "7.2"). The proto's default string
+// form is verbose and not log-friendly, so prefer this helper wherever a rule
+// number is logged. A nil RuleNumber renders as "none".
+func FormatRuleNumber(rn *clustermetadatapb.RuleNumber) string {
+	if rn == nil {
+		return "none"
+	}
+	return fmt.Sprintf("%d.%d", rn.GetCoordinatorTerm(), rn.GetLeaderSubterm())
 }
 
 // MostAdvancedPosition returns the highest-ranked PoolerPosition among the
--- a/go/common/eventlog/events.go
+++ b/go/common/eventlog/events.go
@@ -121,17 +121,44 @@ func (e BackupLeaseLost) LogAttrs() []slog.Attr {
 	return []slog.Attr{slog.String("holder", e.Holder)}
 }
 
-type TermBegin struct {
-	NewTerm      int64
+type ConsensusRecruit struct {
+	Rule         string
 	PreviousTerm int64
 	RevokedRole  string // "primary" | "standby" | "" (empty = no revoke)
 }
 
-func (TermBegin) EventType() string { return "term.begin" }
-func (e TermBegin) LogAttrs() []slog.Attr {
+func (ConsensusRecruit) EventType() string { return "consensus.recruit" }
+func (e ConsensusRecruit) LogAttrs() []slog.Attr {
 	return []slog.Attr{
-		slog.Int64("new_term", e.NewTerm),
+		slog.String("rule", e.Rule),
 		slog.Int64("previous_term", e.PreviousTerm),
 		slog.String("revoked_role", e.RevokedRole),
 	}
+}
+
+// ConsensusPromote is emitted by a multipooler node when it executes the leader
+// path of the Propose phase: it calls pg_promote and writes the new rule entry.
+// Together with consensus.recruit events from Recruit, these events let operators
+// reconstruct the full lifecycle of a term election and correlate Propose-phase
+// failures with postgres state changes.
+type ConsensusPromote struct {
+	Rule string
+}
+
+func (ConsensusPromote) EventType() string { return "consensus.promote" }
+func (e ConsensusPromote) LogAttrs() []slog.Attr {
+	return []slog.Attr{slog.String("rule", e.Rule)}
+}
+
+// ConsensusSetPrimary is emitted by a multipooler node when it executes the
+// standby path of the Propose phase: it configures primary_conninfo to replicate
+// from the new leader. Together with consensus.recruit and consensus.promote
+// events, these events cover the full Propose round across all participating nodes.
+type ConsensusSetPrimary struct {
+	Rule string
+}
+
+func (ConsensusSetPrimary) EventType() string { return "consensus.set_primary" }
+func (e ConsensusSetPrimary) LogAttrs() []slog.Attr {
+	return []slog.Attr{slog.String("rule", e.Rule)}
 }
--- a/go/services/multipooler/internal/manager/rpc_consensus.go
+++ b/go/services/multipooler/internal/manager/rpc_consensus.go
@@ -333,7 +333,9 @@ func (pm *MultiPoolerManager) Recruit(ctx context.Context, req *consensusdatapb.
 		return nil, mterrors.Wrap(err, "failed to determine role for recruit")
 	}
 
-	termEvent := eventlog.TermBegin{NewTerm: revokedBelowTerm}
+	termEvent := eventlog.ConsensusRecruit{
+		Rule: commonconsensus.FormatRuleNumber(revocation.GetOutgoingRule()),
+	}
 	eventlog.Emit(ctx, pm.logger, eventlog.Started, termEvent)
 
 	// Stop replication participation.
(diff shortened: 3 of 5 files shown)


mkindahl:
This was added to be able to track and debug errors like what appeared in #1091.

go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -682,6 +698,26 @@ func (pm *MultiPoolerManager) SetTermPrimary(ctx context.Context, req *consensus
...
+	setTermEvent := eventlog.TermPromotion{
+		Term: rule.GetRuleNumber().GetCoordinatorTerm(),
+		Role: "standby",
+	}
+	eventlog.Emit(ctx, pm.logger, eventlog.Started, setTermEvent)
+
+	resp, err := pm.setTermPrimaryLocked(ctx, req)
+	if err != nil {
+		eventlog.Emit(ctx, pm.logger, eventlog.Failed, setTermEvent, "error", err)
+		return nil, err
```
haritabh17: Can this emit misleading `term.promotion` Success event for no-op early returns??

go/services/multiorch/consensus/rule_change.go
```diff
@@ -248,13 +249,17 @@ func (r *coordinatorLedRuleChange) Run(
 			}
 			r.coordinator.logger.WarnContext(ctx, "Propose failed for non-leader",
 				"pooler", pr.poolerName, "error", pr.err)
+			eventlog.Emit(ctx, r.coordinator.logger, eventlog.Failed, eventlog.PrimaryPromotion{
+				NewPrimary:   newPrimary,
```
haritabh17: For each non-leader Propose failure, it'll emit primary.promotion Failed. Since the overall promotion may still succeed (leader succeeds), a single promotion round can produce multiple primary.promotion Failed events followed by a primary.promotion Success inflate failure counts in monitoring.

Although the events do include the "standby" key-value pair to distinguish them from the terminal event. 

This is intentional?

haritabh17 (approved):
🚀

go/services/multipooler/internal/manager/rpc_consensus.go
dweitzman: This placement would also emit events that don't reflect an ongoing promotion, like new pods being started up, informing replicas of decisions that were already made while they were unreachable, etc..

Emitting here and calling this a promotion event feels a little misleading. If you wanted to distinguish an ongoing promotion from something like a fix replication action, the `SetTermLeader` RPC would need more context

go/common/eventlog/events.go
```diff
@@ -135,3 +135,23 @@ func (e TermBegin) LogAttrs() []slog.Attr {
...
+
+// TermPromotion is emitted by a multipooler node when it executes its part of
+// the Propose phase of a term election. The leader emits this for its Propose
+// RPC (which calls pg_promote and writes the rule); each standby emits this
+// for its SetTermPrimary RPC (which configures primary_conninfo). Together with
+// term.begin events from Recruit, these events let operators reconstruct the
+// full lifecycle of a term election on each participating node and correlate
+// Propose-phase failures with postgres state changes (see postgres.state).
+type TermPromotion struct {
+	Term int64
```
dweitzman: This should probably be a full RuleNumber rather than just a coordinator term.

I've already written code to finish consensus and do leader-subterm leader changes (that's a tongue twister!) for stuck rule propagation, just haven't sent it out because I've been trying to trickle out mailing PRs at a rate that's manageable for code reviewers.

go/services/multipooler/internal/manager/manager.go
```diff
@@ -448,8 +448,17 @@ func (pm *MultiPoolerManager) openLocked(ctx context.Context, targetServingStatu
 			//   - postgres going down: allows PrimaryIsDeadAnalyzer to detect failure promptly
 			//   - postgres coming back up: allows FixReplication to see IsInitialized=true quickly
 			if !postgresStateEqual(newState, prevState) {
+				// Include the current term so that postgres stop/start events can
```
dweitzman: As a general rule I'd like to keep consensus internal knowledge to the consensus/ directory and avoid leaking it into general multipooler code

The use of `term` and what's being printed here also feels a little off / misleading. The two things that exist are the rule number in WAL and the revocation, and the fact that they include something called a coordinator term or leader subterm feels like an internal implementation detail that doesn't need to leak out here.

In this case, a revocation doesn't represent the term of a pooler from a leadership perspective. I think maybe what you want to print here is closer to `ruleStore.CachedPosition().RuleNumber()`?

go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -446,6 +446,22 @@ func (pm *MultiPoolerManager) Propose(ctx context.Context, req *consensusdatapb.
 	}
 	defer pm.actionLock.Release(ctx)
 
+	proposeEvent := eventlog.TermPromotion{
+		Term: req.GetProposal().GetTermRevocation().GetRevokedBelowTerm(),
```
dweitzman: Some questions:
- Do you want this to log even if the request fails validation checks?
- Do you want this to log the rule number being proposed? If so, that's in `req.GetProposal().GetProposedRule().GetRuleNumber()`

go/services/multipooler/internal/manager/rpc_consensus.go
mkindahl: I modified this to not emit success events in cases that are no-ops.

go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -733,6 +749,27 @@ func (pm *MultiPoolerManager) SetTermPrimary(ctx context.Context, req *consensus
 		return &consensusdatapb.SetTermPrimaryResponse{ConsensusStatus: cs}, nil
 	}
 
+	// Both no-op gates passed — this call will actually reconfigure replication.
+	setTermEvent := eventlog.TermPromotion{
+		Term: revocation.GetRevokedBelowTerm(),
```
dweitzman: For here, the rule number would likely want to come from the `req`

go/common/eventlog/events.go
mkindahl: I have no opinion on what to log as long as we can correlate the events to each others, but the existing events `term.begin` uses the field `NewTerm`, so are you suggesting that I should change that too?

go/services/multipooler/internal/manager/manager.go
mkindahl: I used "term" since this is what #1092 used but have no opinion on what to use as long as we consistently use this for all the events so that we can correlate what is happening.

As we discussed separately, we'll use "rule_number" instead of "term" everywhere.

go/services/multipooler/internal/manager/rpc_consensus.go
mkindahl: Moving the validation checks to the `Propose` function instead. Now `proposeLocked` only contain error checks that involve a valid rule.

go/services/multipooler/internal/manager/rpc_consensus.go
mkindahl: Fixed.

go/common/eventlog/events.go
mkindahl: As discussed, we now use `consensus.set_term`, `consensus.promote` and `consensus.recruit` and use rule-number instead of term.

go/services/multiorch/consensus/rule_change.go
mkindahl: No, it does not seem reasonable to emit "fake" failure events. It will just make it harder to debug. Removed.

go/common/consensus/compare.go
```diff
@@ -53,6 +55,17 @@ func CompareRuleNumbers(a, b *clustermetadatapb.RuleNumber) int {
...
+// "coordinator_term.leader_subterm" (e.g. "7.2"). The proto's default string
+// form is verbose and not log-friendly, so prefer this helper wherever a rule
+// number is logged. A nil RuleNumber renders as "none".
+func FormatRuleNumber(rn *clustermetadatapb.RuleNumber) string {
+	if rn == nil {
+		return "none"
+	}
+	return fmt.Sprintf("%d.%d", rn.GetCoordinatorTerm(), rn.GetLeaderSubterm())
+}
+
```
mkindahl, line 68: This was copied from #1124.

go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -483,6 +485,34 @@ func (pm *MultiPoolerManager) Propose(ctx context.Context, req *consensusdatapb.
...
 			"proposal.proposed_rule.creation_time is required")
 	}
+	// Propose is only valid for the designated leader. Non-leaders should
+	// receive the leader's identity via SetTermPrimary, which handles
+	// replication setup without requiring a prior Recruit.
+	if !proto.Equal(pm.serviceID, proposalLeader.GetId()) {
+		return nil, mterrors.Errorf(mtrpcpb.Code_INVALID_ARGUMENT,
+			"Propose received on %s but proposal_leader is %s; non-leaders should be told via SetTermPrimary",
+			pm.serviceID.GetName(), proposalLeader.GetId().GetName())
+	}
```
mkindahl, line 495: This was just moved.

PR #1120, opened 2026-06-09
refactor(multigateway): registry-driven lifecycle for gateway-managed variables

haritabh17:
## Summary

This is a refactoring-only PR for gateway-managed variable (GMV) handling. It centralizes the lifecycle, planner, and engine plumbing around the existing `statement_timeout` GMV implementation.

The refactor makes GMV support modular:

- **`gmvLifecycle` interface + `gatewayManagedVariablesLocked()` registry** — transaction/savepoint lifecycle methods iterate the registry instead of naming variables in each method. The fixed-size array makes a forgotten registry entry a compile-time error.
- **`handler.IsGatewayManagedVariable` / `gatewayManagedVariableNames`** — single source of truth for the managed-variable set, shared by the planner and engine.
- **`ApplyGatewayManagedVariable`** — routes SET / `set_config()` of a GMV to gateway-local state with set-time validation, so the engine no longer special-cases variable names.
- **`set_config('<gmv>', v, false)`** updates gateway state instead of leaking into `SessionSettings`; **`set_config('<gmv>', v, true)`** is tracked as a transaction-local override, including bound parameters resolved at execute time.
- **RESET ALL** resets every registered GMV via the registry instead of naming each variable.
- **SHOW** canonicalizes quoted GMV names at plan time so the executor's case-sensitive switch cannot be missed.
- **GMV documentation** — adds `docs/query_serving/gateway_managed_variables.md` and trims duplicated generic GMV content from the session-setting and statement-timeout docs.

## Test report

- Multigateway unit tests pass (`go/services/multigateway/...`).
- Parser AST unit tests pass (`go/common/parser/ast/...`).
- Race-enabled tests pass for the changed engine/executor/planner packages.

### DIFF
--- a/go/services/multigateway/engine/apply_session_state.go
+++ b/go/services/multigateway/engine/apply_session_state.go
@@ -62,15 +62,20 @@ type ApplySessionState struct {
 	SilentTracking bool
 
 	// BindRefs, when non-nil, marks this primitive as a deferred-resolution
-	// set_config(): name, value, and/or is_local are decoded from the
-	// portal's wire-protocol Bind values at execute time before
-	// SessionSettings is updated. Nil-valued fields use the literal already
-	// in VariableStmt.Name / Args[0] / LiteralIsLocal.
+	// set_config(): name, value, and/or is_local are resolved at execute
+	// time before SessionSettings is updated. Nil-valued fields use the
+	// literal already in VariableStmt.Name / Args[0] / VariableStmt.IsLocal.
 	//
-	// Reachable only via PortalStreamExecute — StreamExecute is the simple
-	// protocol path, which has no binds and therefore never sets BindRefs.
-	// Set automatically by NewApplySessionStateFromBind for the
-	// extended-protocol `SELECT set_config($1, ...)` shape.
+	// Two resolution sources, one per protocol path:
+	//   - PortalStreamExecute decodes the portal's wire-protocol Bind values
+	//     (extended-protocol `SELECT set_config($1, ...)`).
+	//   - StreamExecute resolves against the normalizer-extracted bindVars
+	//     (simple protocol: ast.Normalize parameterizes the value of a
+	//     set_config(..., true) call, so the cached plan for a gateway-managed
+	//     variable carries a ValueParam that must be re-resolved from each
+	//     execution's literals — never baked into the plan).
+	//
+	// Set automatically by NewApplySessionStateFromBind.
 	BindRefs *BoundSetConfigRefs
 }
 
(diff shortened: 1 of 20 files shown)


go/services/multigateway/handler/connection_state.go
```diff
@@ -181,6 +189,17 @@ type savepointFrame struct {
...
+type gmvLifecycle interface {
+	Snapshot()
+	RestoreFromDepth(int)
+	PopFrom(int)
+	ClearSnapshots()
+	// ResetLocal clears the transaction-local override at COMMIT/ROLLBACK.
+	// Part of the interface so the lifecycle methods can iterate generically
+	// rather than naming each variable (which is easy to forget for a new GMV).
+	ResetLocal()
+}
```
GuptaManan100: This interface makes sense to live in the `go/services/multigateway/handler/gateway_managed_variable.go` file along with the implementation

GuptaManan100 (commented):
Just a couple of things - 
1. We should start a doc for gateway managed variables so that they are documented and the reasoning behind it. Also `SHOW` vs `current_setting()` divergence should be documented there.
2. No `NAMEDATALEN`/ASCII normalization - PG declares `application_name` with `GUC_IS_NAME`, so it truncates to 63 bytes and replaces non-printable ASCII with `?`. Should we do something similar? or just document the difference

Otherwise it seems solid!

GuptaManan100:
Also this seems like a rework of https://github.com/multigres/multigres/pull/1003, should we close that one?

go/services/multigateway/handler/connection_state.go
haritabh17: Agree, updating

GuptaManan100 (commented):
`DISCARD ALL` still resets GMVs via the variable-specific `ResetStatementTimeout()`, while `RESET ALL` was migrated to the new registry method `ResetGatewayManagedVariables()`.

go/services/multigateway/engine/apply_session_state.go
```diff
@@ -205,25 +220,119 @@ func (s *ApplySessionState) executeSetWithBinds(
...
+//
+// Mirrors executeSetWithBinds' resolution order: is_local first (it decides
+// whether tracking happens at all), then name, then value only if the
+// tracker write will fire. Errors propagate up through the Sequence, which
+// aborts before the trailing Route fires.
+func (s *ApplySessionState) executeSetWithNormalizedBinds(
+	ctx context.Context,
+	conn *server.Conn,
+	state *handler.MultiGatewayConnectionState,
+	bindVars []*ast.A_Const,
```
GuptaManan100, line 207: `executeSetWithBinds` and `executeSetWithNormalizedBinds` are near-identical - same is_local → name → leaky-local-guard → value → `applyTracked` shape, differing only in the resolution primitive (`DecodeBindAsBool/Text` vs `normalizedBindConst` + type asserts). Not blocking, but a small resolver abstraction (or a `resolve(ref, what)` closure passed in) would stop the two paths from drifting as GMV logic evolves.

GuptaManan100 (approved):
Rest LGTM!

go/services/multigateway/engine/apply_session_state.go
haritabh17: Ack, updating

PR #1121, opened 2026-06-09
feat(multigateway): support session-level advisory locks under pooling

### LINKED ISSUE #1126: multigateway: pg_advisory_unlock can return false after pg_advisory_lock in the same client session
### Summary

`pg_advisory_unlock(...)` can return `false` after `pg_advisory_lock(...)` succeeds in the same client session when connecting through `multigateway`.

This differs from PostgreSQL behaviour. Session-level advisory locks should remain owned by the same client session until they are explicitly released or the session ends.

### Reproduction

Start a local Multigres cluster and run the check through the local `multigateway` PostgreSQL endpoint:

```bash
export PGURI='postgresql://postgres:postgres@localhost:15432/postgres?sslmode=disable'

psql "$PGURI" -X -v ON_ERROR_STOP=1 -At <<'SQL'
SELECT pg_advisory_lock(424242);
SELECT pg_advisory_unlock(424242);
SQL
```

### Expected behaviour

The second statement should return:

```text
t
```

### Actual behavior

The second statement can return:

```text
f
```

In interactive use, PostgreSQL may also emit:

```text
WARNING: you don't own a lock of type ExclusiveLock
```

### Why this matters

PostgreSQL applications commonly use session-level advisory locks for migrations, job coordination, leader election, and mutual exclusion.

If the lock and unlock are not executed with the same backend-session affinity, application code that works on PostgreSQL can leave locks behind or fail to release locks it just acquired.

### Additional check

After a failed unlock, leaked advisory locks may be visible with:

```sql
SELECT locktype, mode, granted
FROM pg_locks
WHERE locktype = 'advisory';
```

### Notes

This issue concerns session-level advisory lock functions, including:

```sql
pg_advisory_lock(...)
pg_advisory_lock_shared(...)
pg_try_advisory_lock(...)
pg_try_advisory_lock_shared(...)
pg_advisory_unlock(...)
pg_advisory_unlock_shared(...)
pg_advisory_unlock_all()
```

Transaction-scoped advisory locks such as `pg_advisory_xact_lock(...)` may not be affected because their lifetime is tied to a transaction.

GuptaManan100:
## Summary

- Session-level advisory locks (`pg_advisory_lock` / `_shared` and the `try` variants) now work through the gateway: the acquiring statement pins the client session to its backend via a new `ReasonSessionAdvisoryLock` reserved-connection reason, so the lock stays visible in `pg_locks` and usable across subsequent queries.
- The multipooler decides when to unpin **authoritatively**, by probing `pg_locks` rather than counting at the gateway — correct for reference-counted stacking, a failed `pg_try_advisory_lock`, and `pg_advisory_unlock_all()`. The probe is gated on statements that actually touch advisory locks, so it stays off the per-query hot path.
- A backend never returns to the pool still holding a lock: `DISCARD ALL` and client disconnect scrub via `pg_advisory_unlock_all()` on the reserved-connection release path (the narrow fix — prepared-statement tracking stays in sync).
- Works on both the simple and extended (portal) protocols. Transaction-level `pg_advisory_xact_lock*` is intentionally out of scope (released at transaction end).

## Description

Detection is best-effort: the planner sees advisory calls in the statement text (the common case) but not locks taken/released inside a PL/pgSQL function, trigger, or dynamic SQL — the same limitation as temp tables created via dynamic SQL. A missed acquire just isn't pinned; a missed release keeps the session pinned conservatively until the next observed advisory statement, `DISCARD ALL`, or disconnect — never a leak.

Design notes: `docs/query_serving/advisory_locks_design.md`.

Covered by unit tests (planner detection/routing, executor probe gating) and end-to-end tests in `queryserving` across both protocols, including pinning, cross-query visibility, reference-counted unlock, `pg_advisory_unlock_all`, `DISCARD ALL`, and no-leak-on-disconnect.

Closes https://github.com/multigres/multigres/issues/1126

### DIFF
--- a/go/services/multipooler/internal/executor/executor.go
+++ b/go/services/multipooler/internal/executor/executor.go
@@ -610,13 +610,21 @@ func (e *Executor) maybeUnpinSessionAdvisoryLock(ctx context.Context, rc reserve
 		return false
 	}
 
+	// SELECT EXISTS always returns exactly one row, so an empty result is
+	// unexpected — treat it like a probe failure and stay pinned rather than
+	// fall through with held=false and risk releasing a backend that may still
+	// hold the client's locks.
+	if len(results) == 0 {
+		e.logger.WarnContext(ctx, "advisory-lock probe returned no rows; keeping connection pinned",
+			"reserved_conn_id", rc.ConnID())
+		return false
+	}
+
 	var held bool
-	if len(results) > 0 {
-		if scanErr := ScanSingleRow(results[0], &held); scanErr != nil {
-			e.logger.WarnContext(ctx, "advisory-lock probe returned unexpected result; keeping connection pinned",
-				"reserved_conn_id", rc.ConnID(), "error", scanErr)
-			return false
-		}
+	if scanErr := ScanSingleRow(results[0], &held); scanErr != nil {
+		e.logger.WarnContext(ctx, "advisory-lock probe returned unexpected result; keeping connection pinned",
+			"reserved_conn_id", rc.ConnID(), "error", scanErr)
+		return false
 	}
 	if held {
 		return false
--- a/go/services/multipooler/internal/executor/executor.go
+++ b/go/services/multipooler/internal/executor/executor.go
@@ -24,6 +24,7 @@ import (
 	"log/slog"
 	"strings"
 
+	"github.com/multigres/multigres/go/common/constants"
 	"github.com/multigres/multigres/go/common/mterrors"
 	"github.com/multigres/multigres/go/common/pgprotocol/client"
 	"github.com/multigres/multigres/go/common/preparedstatement"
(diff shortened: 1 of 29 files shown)


go/services/multipooler/internal/executor/executor.go
```diff
@@ -559,9 +569,70 @@ func (e *Executor) streamExecuteOnReservedConn(
...
+	}
+
+	var held bool
+	if len(results) > 0 {
+		if scanErr := ScanSingleRow(results[0], &held); scanErr != nil {
+			e.logger.WarnContext(ctx, "advisory-lock probe returned unexpected result; keeping connection pinned",
+				"reserved_conn_id", rc.ConnID(), "error", scanErr)
+			return false
+		}
+	}
```
haritabh17, line 628: In maybeUnpinSessionAdvisoryLock, when rc.Query() succeeds but returns an empty results slice (len(results) == 0), the held variable retains its zero value (false), causing the code to proceed to unpin the connection and release it back to the pool — potentially leaking the client's advisory locks to the next session that reuses that backend ??

haritabh17 (approved):
Rest LGTM 🚀

go/services/multipooler/internal/executor/executor.go
GuptaManan100: Good catch! Fixed.

PR #1122, opened 2026-06-09
feat(multipooler): two-stage graceful drain that serves single queries

GuptaManan100:
## Summary

- Two-stage graceful drain in the multipooler: when a primary transitions to `NOT_SERVING`, stage 1 drains in-flight transactions while still serving single autocommit queries on the still-primary; stage 2 then drains the remaining in-flight single queries, so the pooler reports `NOT_SERVING` with zero in-flight work and the subsequent postgres demotion kills nothing.
- The connection pool manager gains a reserved-only drain (`WaitForReservedDrain`, backed by a separate `reservedCount`) alongside the combined one, so stage 1 waits for transactions without being held up by transient single-query borrows. Each borrow/reserve event now touches exactly one counter under a single lock.
- `StartRequest`'s admission bool becomes a `RequestKind` (`ExistingReserved` / `SingleQuery` / `NewReservation`), gated per drain stage. The gRPC handlers classify each request from the signal it actually carries — reservation reasons (`StreamExecute`), `MaxRows` (portals), always-reserves (COPY).
- The multigateway's failover buffer no longer proactively blocks single queries during a failover: they are sent straight to a draining pooler and buffered only reactively (on `MTF01`), so they keep flowing instead of stalling. New transactions stay proactively buffered.

## Testing

Unit tests cover the reserved-only drain (including that a regular borrow does not hold up the reserved wait, and channel re-arm), the full `RequestKind` × drain-phase admission matrix, the two-stage `OnStateChange` lifecycle with grace-period force-close in both stages, and the `admissionKind` / `isSingleQuery` classifiers. `-race` clean on the concurrency-heavy packages. End-to-end failover survival (continuous traffic across a planned failover with zero client-visible errors) is covered by the existing `queryserving` buffer tests.

### DIFF
--- a/docs/query_serving/graceful_drain_design.md
+++ b/docs/query_serving/graceful_drain_design.md
@@ -3,23 +3,37 @@
 ## Overview
 
 When multipooler transitions to `NOT_SERVING` (e.g., during demotion
-from primary to replica), it must reject new queries immediately while
-allowing existing reserved connections — transactions, portals, COPY
-operations — to finish cleanly. Without graceful drain, a sudden
-cutover would abort in-flight transactions, causing client errors and
-client-visible errors.
+from primary to replica), it must let in-flight work finish cleanly
+before the cutover, while steering new work to the new primary. A naive
+"reject everything immediately" cutover would abort in-flight
+transactions and needlessly bounce simple queries that the still-primary
+could have served.
+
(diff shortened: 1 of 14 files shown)


mkindahl (approved):
LGTM

PR #1124, opened 2026-06-09
refactor(consensus): rename Propose→Promote and SetTermPrimary→SetPrimary RPCs

dweitzman:
Rename the two MultiPoolerConsensus RPCs:

- Propose → Promote: the RPC tells the designated leader to promote its postgres, so Promote describes the effect at the service boundary.
- SetTermPrimary → SetPrimary: shorter and clearer.

The abstract consensus terminology in go/common/consensus stays "propose" — consensus proposes a rule change to the cohort and is agnostic to a "primary"; promotion is the postgres-writability concern the RPC effects. The multiorch coordinator's dispatch layer (which drives the Promote RPC) and the appointment event-latency phase are renamed to promote/promote_ms accordingly.

Also log the full rule number where one is in scope rather than just the coordinator term: add consensus.FormatRuleNumber (renders "term.subterm", e.g. "7.2") and use it in the Promote, demote-stale-leader, and rule-change logs. Recruit keeps revoked_below_term — it acts on a TermRevocation, not a rule.


mkindahl (approved):
LGTM

PR #1125, opened 2026-06-09
fix(multipooler): don't strand WAL replay paused after a failed SetTermPrimary

dweitzman:
setPrimaryConnInfoLocked paused WAL replay before writing primary_conninfo and only resumed on success, so a failed write/reload left replay paused. The pause was unnecessary and the stop/start bool params were left over from RPCs deleted in #999.

Reduce it to: set conninfo, reload, then wait until the WAL receiver targets the new primary (matched on the receiver's conninfo to ignore a stale one).

Replay pause/resume now belongs to StopReplication/StartReplication. The postgres monitor resumes a standby left with replay paused unless it was paused explicitly, tracked by a new replayManuallyPaused flag.

### DIFF
--- a/go/services/multipooler/internal/manager/manager.go
+++ b/go/services/multipooler/internal/manager/manager.go
@@ -190,6 +190,13 @@ type MultiPoolerManager struct {
 	// process restart implicitly clears it.
 	walReceiverManuallyStopped atomic.Bool
 
+	// replayManuallyPaused is set when StopReplication paused WAL replay
+	// (REPLAY_ONLY / REPLAY_AND_RECEIVER modes) — an explicit signal that apply
+	// should stay paused. It tells the postgres monitor not to self-heal the
+	// pause by resuming replay. Cleared by StartReplication. Not persisted; a
+	// process restart implicitly clears it.
+	replayManuallyPaused atomic.Bool
+
 	// pgMonitorLastLoggedReason tracks the last logged reason in the monitor to avoid duplicate logs.
 	pgMonitorLastLoggedReason string
 
@@ -1613,6 +1620,9 @@ type postgresState struct {
 	backupsAvailable         bool
 	isPrimary                bool
 	bootstrapSentinelPresent bool
+	// replayPaused reports whether WAL replay (apply) is paused. Only meaningful
+	// for a standby (postgresRunning && !isPrimary); false otherwise.
+	replayPaused bool
 	// primaryTerm is the coordinator term at which this pooler is the primary
 	// per the highest known rule. 0 if we are not the primary for that rule.
 	primaryTerm int64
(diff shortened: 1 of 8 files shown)


go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -769,10 +769,9 @@ func (pm *MultiPoolerManager) SetTermPrimary(ctx context.Context, req *consensus
 		pm.logger.InfoContext(ctx, "SetTermPrimary: updating standby primary_conninfo",
 			"new_leader", leader.GetId().GetName(),
 			"incoming_rule", rule.GetRuleNumber())
-		// Already a standby with an active stream; pause replay, swap
-		// conninfo, resume on the new primary.
-		if err := pm.setPrimaryConnInfoLocked(ctx, leader.GetHost(), port,
```
dweitzman, line 774: SetTermPrimary() was pausing replication, changing primaryconninfo, then unpausing

If the change of primary conninfo failed, replication could get stuck paused

go/services/multipooler/internal/manager/manager.go
```diff
@@ -1934,6 +1954,17 @@ func (pm *MultiPoolerManager) determineRemedialAction(ctx context.Context, curre
...
 			// when no divergence, conclusive when there is.
+			//
+			// Self-heal paused WAL replay. setPrimaryConnInfoLocked no longer
+			// resumes replay as a side effect of writing conninfo (replay
+			// pause/resume is owned by StopReplication/StartReplication), so a
+			// standby left paused by an interrupted op would otherwise keep
+			// receiving WAL without ever applying it. Resume it unless apply was
+			// explicitly paused (replayManuallyPaused); a RECEIVER_ONLY stop
+			// doesn't pause replay, and REPLAY_AND_RECEIVER sets this flag too.
+			if currentState.replayPaused && !pm.replayManuallyPaused.Load() {
```
dweitzman, line 1965: Here's a new remedial action to recover from anything that somehow leads replay paused

If replay is paused and we haven't noted in the pooler manager's state that it's _supposed_ to be paused, automatically resume

go/services/multipooler/internal/manager/manager.go
```diff
@@ -2028,11 +2059,16 @@ func (pm *MultiPoolerManager) takeRemedialAction(ctx context.Context, action rem
...
+		if err := pm.setPrimaryConnInfoLocked(ctx, targetHost, targetPort); err != nil {
 			pm.logger.ErrorContext(ctx, "MonitorPostgres: failed to reconcile primary_conninfo", "error", err)
 		}
 
+	case remedialActionResumeReplay:
+		pm.logger.InfoContext(ctx, "MonitorPostgres: WAL replay paused on standby without an explicit stop; resuming")
+		if err := pm.resumeWALReplay(ctx); err != nil {
+			pm.logger.ErrorContext(ctx, "MonitorPostgres: failed to resume WAL replay", "error", err)
+		}
+
```
mkindahl, line 2071: If would be great if you just had two functions that could be called so that you can do something like

```go
pm.pauseWALReplay()
defer pm.resumeWALReplay()
```

if you want to temporarily pause WAL replay, and

```go
pm.pauseWALReplay()
```

If you want to permanently pause WAL replay and the functions would internally deal with all the necessary state transitions and error message.

mkindahl (approved):
LGTM. One suggestion for improvement to ease maintenance.

PR #1129, opened 2026-06-10
test(sqllogictest): run only the postgres-passing subset against multigateway in CI

GuptaManan100:
## Summary

- The weekly SQLLogicTest differential suite ran all 622 corpus files × 2 protocols × 2 targets and could not finish within the 170m test timeout (it panicked at ~150/622 files), so it never wrote `results.json` — which surfaced as the recurring "suite failed to produce results" Slack alert.
- Restrict the CI run to the 157 corpus files that pass against the standalone PostgreSQL baseline, run them against multigateway only, and skip the postgres baseline entirely. This cuts the run from a multi-hour job that times out to ~117m that completes.
- Commit the postgres-passing allowlist so the subset is reproducible and pinned to the corpus commit; the full 622-file differential run remains available for local/manual use.

## Problem

`TestPostgreSQLSqlLogicTest` runs every corpus file against both a standalone PostgreSQL baseline and multigateway, under both wire protocols — four runner invocations per file. At 622 files that does not fit in the 170m `go test -timeout`, so the run hit Go's hard timeout panic before the loop's graceful "write partial report" path could run. No `results.json` was produced, the artifact upload found nothing, and the weekly Slack job reported an infrastructure failure.

## Solution

For any corpus file that passes against the standalone PostgreSQL baseline, the file's embedded expected output *is* the postgres-correct output. So running that file against multigateway and checking it passes is equivalent to a live diff against postgres — no live baseline needed.

This adds `SLT_REGRESSION_MODE=1` (set in the workflow) which:

- filters the corpus to the committed allowlist in `testdata/postgres_passing.txt` (157 files, identical across both protocols),
- skips Phase 4 (standalone postgres) and its schema resetter, and
- synthesizes each file's postgres verdict as passing — true by definition of the allowlist.

The report shape is unchanged, so the existing divergence detection (`postgres.passed == true && multigateway.passed == false`) now means exactly "an allowlisted file regressed on multigateway," and the existing Slack divergence/weekly-summary steps keep working. Because the run now completes, `results.json` is produced and the false "infra failure" alert goes away.

Default behavior (the full differential run, used locally) is unchanged. `filterToPostgresPassing` reports via `t.Errorf` if any allowlisted path is missing from the pinned corpus, catching allowlist drift. The allowlist was generated from the last successful full run (#26817719179); the header documents how to regenerate it.

### DIFF
--- a/.github/workflows/test-sqllogictest.yml
+++ b/.github/workflows/test-sqllogictest.yml
@@ -69,6 +69,11 @@ jobs:
         env:
           RUN_EXTENDED_QUERY_SERVING_TESTS: "1"
           TEST_PRINT_LOGS: "1"
+          # Run only the committed postgres-passing allowlist, against
+          # multigateway only (the postgres baseline is assumed passing by
+          # definition of the allowlist). This keeps the run well within the
+          # timeout; the full 622-file differential run is for local/manual use.
+          SLT_REGRESSION_MODE: "1"
         run: |
           go test -json -v -run TestPostgreSQLSqlLogicTest \
             ./go/test/endtoend/queryserving/sqllogictest/... \
--- a/go/test/endtoend/queryserving/sqllogictest/corpus.go
+++ b/go/test/endtoend/queryserving/sqllogictest/corpus.go
@@ -15,11 +15,14 @@
 package sqllogictest
 
 import (
+	"bufio"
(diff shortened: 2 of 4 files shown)


mkindahl (approved):
LGTM

PR #1131, opened 2026-06-10
fix(multiorch): include all poolers in recruit cohort regardless of health state

mkindahl:
Pre-filtering poolers by LastSeen staleness before sending recruits is incorrect: an unhealthy pooler may recover during the fan-out, the most-advanced pooler may have a stale heartbeat, and a healthy pooler may become unreachable anyway. Recruit RPCs to unresponsive poolers already return nil status and are skipped by the proposal-building loop without blocking progress (MUL-624); the non-blocking Phase 2 drain in rule_change.go (PR #1113) ensures slow Recruit RPCs do not stall Run() after quorum is reached.

Fixes MUL-624

### DIFF
--- a/go/services/multiorch/consensus/coordinator.go
+++ b/go/services/multiorch/consensus/coordinator.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"log/slog"
 	"sync"
-	"time"
 
 	"google.golang.org/protobuf/types/known/timestamppb"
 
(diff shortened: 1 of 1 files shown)


go/services/multiorch/consensus/coordinator.go
```diff
@@ -75,59 +75,49 @@ func (c *Coordinator) AppointLeader(ctx context.Context, shardKey *clustermetada
...
 // that needs rewinding is not fatal — the SetTermPrimary path runs pg_rewind
 // before the node serves writes or replicates.
 func (c *Coordinator) runFailover(ctx context.Context, cohort []*multiorchdatapb.PoolerHealthState, reason string) error {
-	liveCohort := make([]*multiorchdatapb.PoolerHealthState, 0, len(cohort))
 	for _, p := range cohort {
 		last := p.GetLastSeen()
-		if last == nil || time.Since(last.AsTime()) > failoverCohortReachableThreshold {
-			c.logger.InfoContext(ctx, "Excluding unreachable pooler from failover cohort",
+		if last == nil || time.Since(last.AsTime()) > 10*time.Second {
+			c.logger.DebugContext(ctx, "Pooler has stale heartbeat but is included in recruit cohort",
```
dweitzman: I'm not sure this log line is worth the added code, but I'll defer to your judgement

dweitzman (approved):

GuptaManan100 (approved):

go/services/multiorch/consensus/coordinator.go
mkindahl: I'll merge #1113 first and this as a follow. I don't think it makes sense to collect health status only to print a debug message like this, so will only do it if health information is already available in an easy to consume form.

mkindahl: It is a pure cleanup PR now.

PR #1134, opened 2026-06-10
fix(multipooler): keep pooled session state consistent across transaction rollback

haritabh17:
## Problem

PostgreSQL reverts a session-level `SET` / `set_config(…, is_local=false)` when the transaction it ran in is **rolled back** (and on `ROLLBACK TO SAVEPOINT`). The multipooler's cached connection state (`connstate`) did **not** revert — it kept the value the client set inside the transaction.

Because the pool decides whether to re-apply settings to a reused backend by comparing interned setting-pointers for equality, a backend whose cached state still "matched" was handed to the next session **without re-applying** — so a GUC the client set and then rolled back silently leaked into unrelated later sessions.

Surfaced by the PostGIS topology suite, where a `client_min_messages` set-then-rolled-back leaked **902** `NOTICE: SRID value -1 converted…` messages through the gateway that never appear on a direct connection. The regression test reproduces it with no PostGIS dependency.

## Fix

Make the pooler's cached `connstate` track PostgreSQL's transactional GUC semantics:

- `connstate`: add a transaction-scoped snapshot (`TxnSnapshot` + `SnapshotForTxn()` / `RestoreFromTxn()`) capturing the committed baseline (settings by interned pointer, role by value).
- reserved connection: snapshot at `BEGIN`, restore on `ROLLBACK`, drop on `COMMIT` — zero extra round trips, covering full rollback including `SET ROLE`.
- executor `ConcludeTransaction` (COMMIT): reconcile `connstate` against the gateway's authoritative session settings before COMMIT, covering the `ROLLBACK TO SAVEPOINT` → `COMMIT` hole (idempotent via the interned-pointer short-circuit).

## Tests

`TestGUCDivergenceRepro` and `TestGUCSavepointDivergenceRepro` (non-PostGIS): a poisoning session sets a GUC inside a transaction and rolls it back; a victim session then observes leaked NOTICEs via the gateway (0 direct). Both fail with the fix reverted.

### DIFF
--- a/go/services/multipooler/internal/connstate/connection_state.go
+++ b/go/services/multipooler/internal/connstate/connection_state.go
@@ -107,6 +107,49 @@ func (s *ConnectionState) Clone() *ConnectionState {
 	return clone
 }
 
+// TxnSnapshot captures the parts of ConnectionState that PostgreSQL rolls back
+// together with the surrounding transaction: session GUCs set via SET, and the
+// current role set via SET ROLE / SET SESSION AUTHORIZATION. A session SET (or
+// SET ROLE) issued inside a transaction is reverted on ROLLBACK and kept on
+// COMMIT; the pool caches that state in ConnectionState and relies on it (via
+// interned-Settings pointer identity) to decide when to re-apply settings to a
+// backend. A reserved connection snapshots its state when it opens a
+// transaction and restores the snapshot if the transaction rolls back, so the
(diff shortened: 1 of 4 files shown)


haritabh17:
Superseded by #1138 — branch renamed to `haritabhgupta/mul-627-fix-leaked-session-gucs-after-rollback-in-pooling` (renaming a fork branch closes the cross-fork PR).

PR #1136, opened 2026-06-10
feat(multipooler): refresh pooled connections on ALTER DATABASE/ROLE SET

haritabh17:
## Problem

PostgreSQL reads per-database / per-role GUC defaults (`pg_db_role_setting`) **only at session start**. A pooled backend opened before an `ALTER DATABASE/ROLE … SET` keeps the old defaults forever — `RESET` / `RESET ALL` / `DISCARD ALL` revert to the session-start value, never re-reading the catalog (and `pg_reload_conf()` only re-reads `postgresql.conf`, not the catalog). So a pooled session diverged from a fresh connection.

Surfaced by the PostGIS `totopogeom` test: `CREATE EXTENSION postgis_topology` runs `ALTER DATABASE … SET search_path = …, topology` internally (`topology.AddToSearchPath`), but stale pooled backends then resolved unqualified `topology.*` functions against the pre-extension `search_path` and failed.

## Fix — lazy generation-invalidation, planner-triggered, transaction-correct

- **Planner** flags statements that change per-db/role defaults: `ALTER DATABASE/ROLE … SET/RESET`, and `CREATE EXTENSION` for a small hardcoded allowlist of extensions whose install script runs such an `ALTER` (today: `postgis_topology` — the `ALTER` runs server-side, invisible to the gateway, so the extension name is the only signal). The flag rides the execute request (`ExecuteOptions.invalidates_connection_defaults`). `PlanPortal` routes these through `Plan` so the extended protocol carries it too.
- **Executor** bumps the pool's defaults generation on success, honoring transactional-DDL semantics: immediately when the connection is idle (autocommit / non-transactional reserved), or deferred to a successful `COMMIT` when it ran inside a transaction. `ROLLBACK` discards the pending bump; `ROLLBACK TO SAVEPOINT` conservatively keeps it (an extra refresh is harmless; a missed one leaks stale defaults).
- **connpool** refresh is lazy and non-disruptive: `InvalidateConnectionDefaults` bumps a generation counter without closing any connection; each pooled connection records the generation it was established under and is transparently reconnected on its **next borrow** if stale, so the fresh backend re-reads `pg_db_role_setting`. Hot-path cost is one atomic load plus a compare.

Documented deviations (`docs/query_serving/session_settings.md`): more eager than stock pooling (which never refreshes); replica poolers refresh lazily because the `ALTER` executes on the primary; the bump over-approximates across a pooler's user/admin pools.

## Tests

connpool lazy-reconnect (incl. settings re-application, `-race`); planner detection + allowlist; executor durability-aware decision (autocommit vs defer-to-COMMIT, failed block); e2e regression for autocommit and in-transaction `ALTER DATABASE … SET` — verified non-vacuous (fails with the refresh disabled, returning the stale `"$user", public`).

### DIFF
--- a/docs/query_serving/session_settings.md
+++ b/docs/query_serving/session_settings.md
@@ -34,6 +34,72 @@ Behaviour:
 - `SHOW var` reports the current gateway-managed value in PostgreSQL-compatible form.
 - `RESET ALL` clears gateway-managed variables alongside `SessionSettings`.
 
+## Per-Database / Per-Role Default Changes (ALTER DATABASE/ROLE … SET)
+
+`ALTER DATABASE … SET` and `ALTER ROLE … SET` change *per-database* / *per-role*
+session GUC defaults, stored in the `pg_db_role_setting` catalog. PostgreSQL
+applies these defaults to a session **only at session start** (they become the
+session's GUC *reset values*). A connection pooler therefore has a problem:
+backends opened **before** such an `ALTER` keep the old defaults until they
+reconnect — `RESET`, `RESET ALL`, and `DISCARD ALL` all revert to the
+session-start reset value, never re-reading `pg_db_role_setting`.
+
+To keep pooled sessions consistent with a fresh connection, the multigateway
+flags statements that change these defaults and the multipooler refreshes its
+pooled connections:
+
+1. **Planner detection** (`planner/connection_defaults.go`,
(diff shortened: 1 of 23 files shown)


haritabh17:
Superseded by #1140 — branch renamed to `haritabhgupta/mul-629-refresh-pooled-backends-after-alter-database-role-set`.

PR #1138, opened 2026-06-10
fix(multipooler): reconcile reserved session state at the release boundary

haritabh17:
## Summary

- Fixes a session-GUC leak: a `SET` / `SET ROLE` applied inside a transaction on a reserved (pinned) connection is reverted by PostgreSQL on `ROLLBACK` / `ROLLBACK TO SAVEPOINT`, but the pooler's `connstate` cache kept the stale value — so a recycled connection ran later sessions with the wrong GUCs (e.g. leaked NOTICEs that should have been suppressed).
- Restores the pool's core invariant — **connstate must truthfully describe the backend's real session state** — at the two boundaries where it can break (the mid-transaction statement path and reservation release), adding no backend round trips on the release path.
- Adds `ReservationOptions.mark_session_state_untrusted` so the gateway can tell the multipooler when `ROLLBACK TO SAVEPOINT` reverted backend GUCs invisibly; the reserved connection is then reconciled from the gateway's authoritative `SessionSettings`.
- Removes vestigial `connstate.User` tracking (the current role lives in `Settings` as the `"role"` GUC).

## Problem

The regular pool buckets connections by their interned `connstate.Settings` pointer and, on checkout, diffs the desired settings against that cached value — both are correct only if connstate truthfully describes the backend. A session GUC set while a reserved connection is inside a transaction is reverted by PostgreSQL on `ROLLBACK` (and on `ROLLBACK TO SAVEPOINT`), but connstate retained the applied value. Because the pool reuses a connection with matching settings **without re-applying**, the recycled connection could run a later session with stale GUCs. Concretely, `SET client_min_messages TO error` inside a transaction that rolled back left the backend at the default while the pool believed it was `error` — leaking NOTICEs (902 in the PostGIS topology suite through multigateway vs 0 against PostgreSQL directly).

## Solution

**Statement boundary (mid-transaction):** `ROLLBACK TO SAVEPOINT` reverts backend GUCs without the pooler observing the new values. The gateway signals this via the new `ReservationOptions.mark_session_state_untrusted`; the multipooler marks the reserved connection untrusted. The next statement reconciles the backend against the gateway's authoritative `ExecuteOptions.SessionSettings` through the normal diff-apply — which self-heals, because a stale connstate no longer pointer-equals the gateway's desired settings, so the fast-path no-op cannot fire and the full apply (RESET removed vars + SET desired) brings the backend back into agreement.

**Release boundary (reservation ends):** once the reservation ends the gateway and backend are already aligned — PostgreSQL reverted the backend, and the gateway reverted its own tracked `SessionSettings` via its savepoint stack — so only the pooler's connstate cache can still lie. The release finalizer therefore syncs connstate to the gateway's authoritative settings **in memory, with no backend SQL**, and clears the untrusted flag. Full `ROLLBACK` instead restores the in-memory transaction snapshot (zero round trips) and clears the flag in lock-step. The gateway's settings are threaded to every clean-release site — transaction conclude, portal completion, advisory-lock release, COPY, temp-table discard, and session teardown (`ReleaseReservedConnection`) — so an untrusted cache is synced to the truth rather than wrongly cleared to clean (which would leak the backend's real GUCs to the next client).

**Release reasons:** whether a reason permits reuse is classified once next to the enum (`preventsReuse()`): timeout / kill / error — and any unknown future reason, failing safe — taint and close; commit / rollback / portal-complete / advisory-unlock finalize and recycle. The lent-accounting callback fires only after finalization, so a finalizing backend stays counted as lent.

## Tests

- Deterministic, PostGIS-free regression repros for plain rollback and savepoint-rollback-then-commit (`guc_divergence_repro_test.go`, `set_config_connstate_repro_test.go`): zero leaks and zero cache↔backend mismatches on both the direct-primary and multigateway paths.
- Release-finalizer unit tests (`release_finalizer_test.go`): trusted no-SQL, untrusted sync-from-gateway, untrusted-without-gateway-settings clears to clean, missing-cache taints, dirty-reason taints, and lent-accounting ordering.
- Executor unit tests for the mark-untrusted plumbing and for `ReleaseReservedConnection` syncing connstate from the gateway at teardown.
- `-short` multipooler suite passes; gofumpt + `go vet` clean.

### DIFF
--- a/go/services/multipooler/internal/connstate/connection_state.go
+++ b/go/services/multipooler/internal/connstate/connection_state.go
@@ -113,9 +113,13 @@ func (s *ConnectionState) Clone() *ConnectionState {
 // session state.
 //
 // Settings is interned and immutable (it is replaced wholesale via SetSettings,
-// never mutated in place), so it is captured by reference. Prepared statements
-// are intentionally NOT captured here — their lifecycle is owned by the
-// prepared-statement consolidator, not this cache.
+// never mutated in place), so it is captured by reference. PreparedStatements
+// are intentionally NOT captured here. Although ConnectionState tracks which
+// prepared statements have been parsed on this backend connection, PostgreSQL
+// treats prepared statements as session-level objects: PREPARE/Parse survives
+// ROLLBACK and ROLLBACK TO SAVEPOINT, and DEALLOCATE/Close is not undone by
+// rollback. Restoring them from a transaction snapshot would make the pool's
+// cache diverge from the backend.
 type TxnSnapshot struct {
 	settings *Settings
 }
--- a/go/services/multipooler/internal/connstate/connection_state.go
+++ b/go/services/multipooler/internal/connstate/connection_state.go
@@ -31,15 +31,10 @@ type ConnectionState struct {
 	// mu protects all mutable fields in this struct.
 	mu sync.Mutex
 
-	// User is the current role set via SET ROLE.
-	// This is tracked separately from Settings and is NOT used for pool bucket routing.
-	// The role affects PostgreSQL's role-based access control (RLS policies, object
-	// permissions, stored procedure access, etc.).
-	// Empty string means no role has been set (using connection's default role).
-	User string
-
 	// Settings contains session variables (SET commands).
 	// This is the key for connection pool bucket assignment.
+	// The current role is tracked as the "role" variable inside Settings (it is
+	// a session GUC), not as a separate field.
 	Settings *Settings
 
 	// PreparedStatements stores prepared statements by name.
@@ -94,7 +89,6 @@ func (s *ConnectionState) Clone() *ConnectionState {
 	defer s.mu.Unlock()
 
 	clone := &ConnectionState{
-		User:               s.User,
 		PreparedStatements: make(map[string]*query.PreparedStatement, len(s.PreparedStatements)),
 	}
 
@@ -108,22 +102,22 @@ func (s *ConnectionState) Clone() *ConnectionState {
 }
 
 // TxnSnapshot captures the parts of ConnectionState that PostgreSQL rolls back
-// together with the surrounding transaction: session GUCs set via SET, and the
-// current role set via SET ROLE / SET SESSION AUTHORIZATION. A session SET (or
-// SET ROLE) issued inside a transaction is reverted on ROLLBACK and kept on
-// COMMIT; the pool caches that state in ConnectionState and relies on it (via
-// interned-Settings pointer identity) to decide when to re-apply settings to a
-// backend. A reserved connection snapshots its state when it opens a
-// transaction and restores the snapshot if the transaction rolls back, so the
-// pool's cached view never diverges from the backend's real session state.
+// together with the surrounding transaction: session GUCs set via SET,
+// including the current role ("role" is tracked as a Settings variable). A
+// session SET (or SET ROLE) issued inside a transaction is reverted on
+// ROLLBACK and kept on COMMIT; the pool caches that state in ConnectionState
+// and relies on it (via interned-Settings pointer identity) to decide when to
+// re-apply settings to a backend. A reserved connection snapshots its state
+// when it opens a transaction and restores the snapshot if the transaction
+// rolls back, so the pool's cached view never diverges from the backend's real
+// session state.
 //
 // Settings is interned and immutable (it is replaced wholesale via SetSettings,
-// never mutated in place), so it is captured by reference; User is copied by
-// value. Prepared statements are intentionally NOT captured here — their
-// lifecycle is owned by the prepared-statement consolidator, not this cache.
+// never mutated in place), so it is captured by reference. Prepared statements
+// are intentionally NOT captured here — their lifecycle is owned by the
+// prepared-statement consolidator, not this cache.
 type TxnSnapshot struct {
 	settings *Settings
-	user     string
 }
 
 // SnapshotForTxn captures the transaction-revertible parts of the state.
@@ -134,7 +128,7 @@ func (s *ConnectionState) SnapshotForTxn() *TxnSnapshot {
 	}
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	return &TxnSnapshot{settings: s.Settings, user: s.User}
+	return &TxnSnapshot{settings: s.Settings}
 }
 
 // RestoreFromTxn restores state captured by SnapshotForTxn. It is called after a
@@ -147,7 +141,6 @@ func (s *ConnectionState) RestoreFromTxn(snap *TxnSnapshot) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.Settings = snap.settings
-	s.User = snap.user
 }
 
 // Close cleans up the connection state.
@@ -159,7 +152,6 @@ func (s *ConnectionState) Close() {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	s.User = ""
 	s.Settings = nil
 	s.PreparedStatements = nil
 }
@@ -182,46 +174,6 @@ func (s *ConnectionState) SetSettings(settings *Settings) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.Settings = settings
-}
-
-// --- User/Role Methods ---
-
-// GetUser returns the current user role set via SET ROLE.
-// Returns empty string if no role has been set.
-func (s *ConnectionState) GetUser() string {
-	if s == nil {
-		return ""
-	}
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	return s.User
-}
-
-// SetUser sets the current user role.
-// This should be called after executing SET ROLE on the connection.
-func (s *ConnectionState) SetUser(user string) {
-	if s == nil {
-		return
-	}
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	s.User = user
-}
-
-// ClearUser clears the current user role.
-// This should be called after executing RESET ROLE on the connection.
-func (s *ConnectionState) ClearUser() {
-	s.SetUser("")
-}
-
-// HasUser returns true if a user role has been set.
-func (s *ConnectionState) HasUser() bool {
-	if s == nil {
-		return false
-	}
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	return s.User != ""
 }
 
 // --- Prepared Statement Methods ---
--- a/go/services/multipooler/internal/connstate/connection_state.go
+++ b/go/services/multipooler/internal/connstate/connection_state.go
@@ -107,6 +107,49 @@ func (s *ConnectionState) Clone() *ConnectionState {
 	return clone
 }
 
+// TxnSnapshot captures the parts of ConnectionState that PostgreSQL rolls back
+// together with the surrounding transaction: session GUCs set via SET, and the
+// current role set via SET ROLE / SET SESSION AUTHORIZATION. A session SET (or
+// SET ROLE) issued inside a transaction is reverted on ROLLBACK and kept on
+// COMMIT; the pool caches that state in ConnectionState and relies on it (via
+// interned-Settings pointer identity) to decide when to re-apply settings to a
+// backend. A reserved connection snapshots its state when it opens a
+// transaction and restores the snapshot if the transaction rolls back, so the
+// pool's cached view never diverges from the backend's real session state.
+//
+// Settings is interned and immutable (it is replaced wholesale via SetSettings,
+// never mutated in place), so it is captured by reference; User is copied by
+// value. Prepared statements are intentionally NOT captured here — their
+// lifecycle is owned by the prepared-statement consolidator, not this cache.
+type TxnSnapshot struct {
+	settings *Settings
+	user     string
+}
+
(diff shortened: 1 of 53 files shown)


go/services/multipooler/internal/pools/reserved/reserved_conn.go
```diff
@@ -53,6 +53,15 @@ type Conn struct {
...
 	reservedProps *ReservationProperties
 
+	// txnSnapshot captures the session state (GUCs, role) at the moment this
+	// connection opened its current transaction. PostgreSQL reverts session
+	// SET / SET ROLE issued inside a transaction on ROLLBACK; restoring this
+	// snapshot on rollback keeps the pool's cached connstate in lock-step with
+	// the backend, so a recycled connection is never reused with stale settings.
+	// nil when not in a transaction. Accessed only from the transaction-control
+	// methods, which the gateway serializes per reserved connection.
+	txnSnapshot *connstate.TxnSnapshot
```
GuptaManan100, line 63: ~~While your fix is the right step, it's not complete, this has to be a list, with the name of the save point too. And then we need to track the SAVEPOINT queries as well, so that the connection pooler can add a snapshot whenever a new SAVEPOINT is added so that we can get ROLLBACK TO SAVEPOINT to work correctly~~

EDIT: 👇 I got it

GuptaManan100 (commented):
Nope scratch that, I got it, ROLLBACK TO SAVEPOINT still keeps the transaction alive, you only need to restore settings before the Transaction started, otherwise the current settings are applied again so they are correct anyways, so looks good to me!

go/services/multipooler/internal/executor/executor.go
```diff
@@ -1716,6 +1722,28 @@ func (e *Executor) ConcludeTransaction(
...
+		// connstate already matches, so this costs nothing unless a savepoint
+		// rollback actually left them diverged.
+		//
+		// Guarded like the other reserved-conn ApplySettingsToConn sites: when the
+		// gateway sends no session settings, leave the backend untouched. Applying
+		// nil settings would RESET ALL inside the still-open transaction and the
+		// COMMIT would persist that reset.
+		if options.SessionSettings != nil {
+			if err := e.poolManager.ApplySettingsToConn(ctx, reservedConn.Conn(), e.sessionSettingsForPool(options.SessionSettings)); err != nil {
+				reservedConn.Release(reserved.ReleaseError)
```
GuptaManan100: One more thing occurred to me, if we are doing this, why don't we just do this, on both commit and rollback 😆? That way whenever the connection is returned it will be consistent 😆

haritabh17: Yupp right 😄 - updating

go/services/multipooler/internal/pools/reserved/reserved_conn.go
```diff
@@ -53,6 +54,30 @@ type Conn struct {
...
```
GuptaManan100, line 63: `txnSnapshot` is the one new field with no synchronization - the comment says it's safe because the gateway serializes txn-control per reserved conn. Can you confirm nothing off that path reaches it? Specifically `KillConnection` / idle-killer / `Close` running concurrently with an in-flight `Rollback`. The other two new fields took a mutex / atomic; calling out explicitly why this one doesn't would help.

go/services/multipooler/internal/pools/reserved/reserved_pool.go
```diff
@@ -358,25 +369,58 @@ func (p *Pool) release(rc *Conn, reason ReleaseReason) {
...
 
-	// Replication conns cannot be returned to the pool: their socket has
-	// replication=database in its startup packet and is bound to a walsender
-	// backend that may own a slot. Taint regardless of release reason so the
-	// upcoming Recycle frees the cap slot AND closes the socket.
-	if rc.reservedProps != nil && protoutil.HasLogicalReplicationReason(rc.reservedProps.Reasons) {
+	// Dirty releases and replication sockets must never be reused.
+	if dirty || (rc.reservedProps != nil && protoutil.HasLogicalReplicationReason(rc.reservedProps.Reasons)) {
+		rc.pooled.Taint()
+	} else if err := p.finalizeCleanRelease(rc); err != nil {
```
GuptaManan100: Worth a one-line comment on the synchronous cost here: for an untrusted (savepoint-rollback) connection, `finalizeCleanRelease` issues a `RESET ALL` + re-apply round-trip **before this release returns**, and since `ConcludeTransaction` calls `ReleaseClean` synchronously, that round-trip is on the COMMIT/ROLLBACK RPC's critical path.

go/services/multipooler/internal/connstate/connection_state.go
```diff
@@ -107,78 +101,79 @@ func (s *ConnectionState) Clone() *ConnectionState {
...
+// and relies on it (via interned-Settings pointer identity) to decide when to
+// re-apply settings to a backend. A reserved connection snapshots its state
+// when it opens a transaction and restores the snapshot if the transaction
+// rolls back, so the pool's cached view never diverges from the backend's real
+// session state.
+//
+// Settings is interned and immutable (it is replaced wholesale via SetSettings,
+// never mutated in place), so it is captured by reference. Prepared statements
+// are intentionally NOT captured here — their lifecycle is owned by the
+// prepared-statement consolidator, not this cache.
```
GuptaManan100: Not sure about that. If we create a prepared statement as part of a transaction, is the creation of the prepared statement rolled back on rollback? If it is then we need that too.

PreparedStatementConsolidation is only keeping track of the prepared statements seen so far so that they get the same name. But each individual connection is tracking the prepared statements created on that connection.

go/services/multipooler/internal/pools/reserved/properties.go
```diff
@@ -147,6 +147,19 @@ const (
...
+// of finalized and recycled into the regular pool. Unknown reasons fail safe
+// to taint.
+func (r ReleaseReason) preventsReuse() bool {
+	switch r {
+	case ReleaseCommit, ReleaseRollback, ReleasePortalComplete, ReleaseAdvisoryUnlock:
+		return false
+	default:
+		return true
+	}
+}
```
GuptaManan100, line 161: Not really sure what this is for

go/services/multipooler/internal/pools/reserved/reserved_conn.go
```diff
@@ -53,6 +53,22 @@ type Conn struct {
...
+
+	// sessionStateUntrusted is set when PostgreSQL may have reverted backend
+	// session state without the pooler's connstate cache observing the exact new
+	// value (e.g. successful ROLLBACK TO SAVEPOINT). While set, release
+	// finalization syncs connstate to the gateway's authoritative session
+	// settings instead of trusting the stale cache.
+	sessionStateUntrusted atomic.Bool
```
GuptaManan100: This can be a bool, very much like the reservation reasons and others.

go/services/multipooler/internal/pools/reserved/reserved_pool.go
```diff
@@ -31,13 +31,20 @@ import (
...
 	// InactivityTimeout is the maximum duration a reserved connection can be inactive
 	// (no client activity) before being killed. A value of 0 means no timeout.
 	// Default: 30s
 	InactivityTimeout time.Duration
 
+	// ReleaseFinalizationTimeout bounds clean-release reconciliation before the
+	// backend is recycled into the regular pool. On timeout/failure the backend is
+	// tainted/closed instead of reused. Default: 5s.
+	ReleaseFinalizationTimeout time.Duration
+
```
GuptaManan100: This is unused 😆, it is only set, never read.

go/services/multipooler/internal/pools/reserved/reserved_pool.go
```diff
@@ -358,27 +373,51 @@ func (p *Pool) release(rc *Conn, reason ReleaseReason) {
...
-		rc.pooled.Taint()
 	}
 
-	// Replication conns cannot be returned to the pool: their socket has
-	// replication=database in its startup packet and is bound to a walsender
-	// backend that may own a slot. Taint regardless of release reason so the
-	// upcoming Recycle frees the cap slot AND closes the socket.
-	if rc.reservedProps != nil && protoutil.HasLogicalReplicationReason(rc.reservedProps.Reasons) {
+	// Uncertain-state releases and replication sockets must never be reused.
+	if reason.preventsReuse() || (rc.reservedProps != nil && protoutil.HasLogicalReplicationReason(rc.reservedProps.Reasons)) {
```
GuptaManan100, line 368: I know we didn't add it in this PR, but its a better design to mark the connection tainted when it is reserved for LogicalReplicationReason instead of doing it at the release boundary.

go/services/multipooler/internal/pools/reserved/reserved_pool.go
```diff
@@ -358,27 +373,51 @@ func (p *Pool) release(rc *Conn, reason ReleaseReason) {
...
+		rc.pooled.Taint()
+	} else if err := p.finalizeCleanRelease(rc, gatewaySessionSettings); err != nil {
+		p.logger.Warn("reserved clean-release finalization failed; tainting backend",
```
GuptaManan100, line 371: This works too, I was originally thinking the exact opposite, that we apply the settings we have in our state to the connection, but this is good too 👍

GuptaManan100 (approved):
Rest LGTM!

go/services/multipooler/internal/pools/reserved/reserved_pool.go
haritabh17: `finalizeCleanRelease` no longer does backend I/O, i.e. no `RESET ALL` then re-apply; it only interns gateway session settings and updates the in-memory `connstate`, so there is no extra backend round trip on the `COMMIT`/`ROLLBACK` critical path.

go/services/multipooler/internal/pools/reserved/reserved_conn.go
haritabh17: Traced the access paths. `txnSnapshot` is only touched by transaction lifecycle methods (`BeginWithQuery`, `SnapshotTxnState`, `Commit`, `Rollback`). 

KillConnection, idle-killer, and Close do not read/write it; they only kill/taint/close/release. If they race with rollback, they may cause the query to fail or make the release CAS no-op, but they don’t race on `txnSnapshot`. The underlying `ConnectionState` restore uses its own mutex.

go/services/multipooler/internal/connstate/connection_state.go
haritabh17: Verified this against PG that prepared statements are session-scoped, not transaction-scoped: PREPARE survives both full ROLLBACK and ROLLBACK TO SAVEPOINT, and DEALLOCATE is also not undone by rollback/savepoint rollback. The comment wording is not clear though, fixed that and also added following testcases:
 - PREPARE inside transaction survives full ROLLBACK
 - PREPARE after SAVEPOINT survives ROLLBACK TO SAVEPOINT                                                              
 - DEALLOCATE inside transaction is not undone by full ROLLBACK                                                        
 - DEALLOCATE inside savepoint is not undone by ROLLBACK TO SAVEPOINT                                                  
 - Multigres-specific backend connstate guard: prepared statement parsed on a reserved backend inside a txn remains usable after rollback, proving prepared statements should not be restored from TxnSnapshot

go/services/multipooler/internal/pools/reserved/properties.go
haritabh17: This helper is the release-boundary safety classification. Clean releases run `finalizeCleanRelease`, while timeout/kill/error/unknown reasons taint as protocol/backend state may be uncertain.

Used in `release` :
```
   if reason.preventsReuse() || ... {                                                    
       rc.pooled.Taint()                                                                                               
   } else if err := p.finalizeCleanRelease(rc, gatewaySessionSettings); err != nil {
       ...                                                                                                             
   }                                                                                                                   
```

go/services/multipooler/internal/pools/reserved/reserved_conn.go
haritabh17: Agree. updating

go/services/multipooler/internal/pools/reserved/reserved_pool.go
haritabh17: This is leftover from the earlier release-finalization design when finalization could do backend I/O and needed a timeout. The current finalizer is in-memory only, so there’s nothing to bound. Removing entirely.

go/services/multipooler/internal/pools/reserved/reserved_pool.go
haritabh17: Thats a better idea, let me do that in a follow up PR as that would need adding a new primitive e.g. `pooled.TaintOnRecycle`.

go/services/multipooler/internal/pools/reserved/reserved_pool.go
haritabh17: Yup, since the gateway already has the authoritative state, this lets us avoid backend I/O during release.

go/services/multipooler/internal/pools/reserved/reserved_pool.go
haritabh17: Refactor here - https://github.com/multigres/multigres/pull/1169

PR #1139, opened 2026-06-10
fix(multigateway): keep reserved connection across statement_timeout cancellation

haritabh17:
## Problem

The multigateway enforces `statement_timeout` with a context deadline. When the deadline fired mid-stream, the gRPC `StreamExecute` returned **before** the trailer carrying the reserved-connection state arrived, so the gateway saw `reservedState == nil`. `ScatterConn` then treated nil as "the reserved connection is gone" and **unconditionally cleared the reservation** — even though the backend (with its session temp tables / open transaction) was still alive. The next statement landed on a different pooled backend, so the session temp state had vanished.

PostgreSQL keeps session temp tables across a statement-level cancellation, so this diverged from PostgreSQL. Surfaced by the PostGIS `interrupt` / `interrupt_relate` / `interrupt_buffer` tests, which do `CREATE TEMP TABLE … AS SELECT` + `SET statement_timeout` + a long query and then read the temp table — failing with `relation "_inputs"/"_time" does not exist`.

## Fix

In `ScatterConn`, only clear the reservation when the stream genuinely lost the backend — i.e. keep it when the error is a cancellation (`context.Canceled` / `DeadlineExceeded`, or `PgDiagnostic.Code == 57014` query-cancelled), the connection is not in a transaction block, and no new reserved id was returned. A cancellation is no longer misread as connection loss. Transaction-safe by construction (gated on not-in-block), so transaction abort/rollback semantics are unchanged.

## Tests

`TestMultiGateway_TempTableSurvivesStatementTimeout` (regression): a CTAS temp table survives a `statement_timeout` cancellation. `TestMultiGateway_StatementTimeoutInTransaction` (guard): a timeout inside an explicit transaction still aborts the transaction and recovers on ROLLBACK. The regression test fails with the fix reverted.

### DIFF
--- a/Dockerfile.cluster
+++ b/Dockerfile.cluster
@@ -22,7 +22,12 @@ FROM golang:1.25-alpine AS builder
 
 WORKDIR /src
 
-RUN apk add --no-cache git=2.52.0-r0 make=4.4.1-r3 bash=5.3.3-r1
+# Intentionally unpinned: the builder base (golang:1.25-alpine) is a rolling
+# tag, so pinning exact apk versions breaks whenever Alpine bumps these
+# packages. These are build-stage tools from the base image's own apk
+# repository, not external supply-chain downloads.
+# hadolint ignore=DL3018
+RUN apk add --no-cache git make bash
 
 COPY go.mod go.sum ./
 RUN go mod download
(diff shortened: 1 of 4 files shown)


GuptaManan100 (approved):
Good catch!

PR #1140, opened 2026-06-10
feat(multipooler): refresh pooled connections on ALTER DATABASE/ROLE SET

haritabh17:
## Problem

PostgreSQL reads per-database / per-role GUC defaults (`pg_db_role_setting`) **only at session start**. A pooled backend opened before an `ALTER DATABASE/ROLE … SET` keeps the old defaults forever — `RESET` / `RESET ALL` / `DISCARD ALL` revert to the session-start value, never re-reading the catalog (and `pg_reload_conf()` only re-reads `postgresql.conf`, not the catalog). So a pooled session diverged from a fresh connection.

Surfaced by the PostGIS `totopogeom` test: `CREATE EXTENSION postgis_topology` runs `ALTER DATABASE … SET search_path = …, topology` internally (`topology.AddToSearchPath`), but stale pooled backends then resolved unqualified `topology.*` functions against the pre-extension `search_path` and failed.

## Fix — lazy generation-invalidation, planner-triggered, transaction-correct

- **Planner** flags statements that change per-db/role defaults: `ALTER DATABASE/ROLE … SET/RESET`, and `CREATE EXTENSION` for a small hardcoded allowlist of extensions whose install script runs such an `ALTER` (today: `postgis_topology` — the `ALTER` runs server-side, invisible to the gateway, so the extension name is the only signal). The flag rides the execute request (`ExecuteOptions.invalidates_connection_defaults`). `PlanPortal` routes these through `Plan` so the extended protocol carries it too.
- **Executor** bumps the pool's defaults generation on success, honoring transactional-DDL semantics: immediately when the connection is idle (autocommit / non-transactional reserved), or deferred to a successful `COMMIT` when it ran inside a transaction. `ROLLBACK` discards the pending bump; `ROLLBACK TO SAVEPOINT` conservatively keeps it (an extra refresh is harmless; a missed one leaks stale defaults).
- **connpool** refresh is lazy and non-disruptive: `InvalidateConnectionDefaults` bumps a generation counter without closing any connection; each pooled connection records the generation it was established under and is transparently reconnected on its **next borrow** if stale, so the fresh backend re-reads `pg_db_role_setting`. Hot-path cost is one atomic load plus a compare.

Documented deviations (`docs/query_serving/session_settings.md`): more eager than stock pooling (which never refreshes); replica poolers refresh lazily because the `ALTER` executes on the primary; the bump over-approximates across a pooler's user/admin pools.

## Tests

connpool lazy-reconnect (incl. settings re-application, `-race`); planner detection + allowlist; executor durability-aware decision (autocommit vs defer-to-COMMIT, failed block); e2e regression for autocommit and in-transaction `ALTER DATABASE … SET` — verified non-vacuous (fails with the refresh disabled, returning the stale `"$user", public`).

### DIFF
--- a/docs/query_serving/session_settings.md
+++ b/docs/query_serving/session_settings.md
@@ -27,6 +27,72 @@ or forwarded to PostgreSQL. See
 contract and [`statement_timeout_design.md`](./statement_timeout_design.md) for
 the concrete `statement_timeout` behavior.
 
+## Per-Database / Per-Role Default Changes (ALTER DATABASE/ROLE … SET)
+
+`ALTER DATABASE … SET` and `ALTER ROLE … SET` change _per-database_ / _per-role_
+session GUC defaults, stored in the `pg_db_role_setting` catalog. PostgreSQL
+applies these defaults to a session **only at session start** (they become the
+session's GUC _reset values_). A connection pooler therefore has a problem:
+backends opened **before** such an `ALTER` keep the old defaults until they
+reconnect — `RESET`, `RESET ALL`, and `DISCARD ALL` all revert to the
+session-start reset value, never re-reading `pg_db_role_setting`.
+
+To keep pooled sessions consistent with a fresh connection, the multigateway
+flags statements that change these defaults and the multipooler refreshes its
+pooled connections:
+
+1. **Planner detection** (`planner/connection_defaults.go`,
+   `statementChangesConnectionDefaults`): a statement is flagged when it is
+   - `ALTER DATABASE … SET`/`RESET` or `ALTER ROLE … SET`/`RESET` (any form —
+     every one mutates `pg_db_role_setting`), or
+   - `CREATE EXTENSION <name>` for a `name` on a small hardcoded allowlist of
+     extensions whose install script runs such an `ALTER` internally
+     (`extensionsAlteringConnectionDefaults`; today: `postgis_topology`, whose
+     `topology.AddToSearchPath` runs `ALTER DATABASE … SET search_path`). The
+     `ALTER` runs server-side inside the extension, invisible to the gateway,
+     so the extension name is the only signal — hence an explicit allowlist
+     rather than general detection.
+
+   The flag rides the execute request as
+   `ExecuteOptions.InvalidatesConnectionDefaults`.
+
+2. **Durability-aware bump** (multipooler executor): on **successful**
+   execution, if the connection is now idle (autocommit, or a
+   non-transactional reserved connection) the change is durable, so the
+   executor bumps the pool's _defaults generation_ immediately. If the
+   statement ran inside an explicit transaction the change is durable only at
+   `COMMIT`, so the executor marks the reserved connection pending and bumps on
+   a successful `COMMIT`. `ROLLBACK` discards the mark (no bump). A
+   `ROLLBACK TO SAVEPOINT` keeps the mark — a deliberately conservative choice
+   (an extra refresh is harmless; a missed one would leak stale defaults). This
+   mirrors PostgreSQL's transactional-DDL semantics.
+
+3. **Lazy refresh** (`connpool.Pool`): bumping the generation does not close any
+   connection. Each pooled connection records the generation it was
+   established under; on its **next borrow**, a connection whose generation is
+   stale is transparently reconnected (`refreshIfStale`) so the fresh backend
+   re-reads `pg_db_role_setting`. The hot-path cost is one atomic load plus an
+   int compare; an actual reconnect happens at most once per connection per
+   bump, and only when that connection is next used.
+
+### Deviations and limitations of this behaviour
+
+- **More eager than stock pooling, by design.** A plain connection pooler
+  (pgbouncer/odyssey) never refreshes pooled server connections after an
+  `ALTER DATABASE/ROLE … SET`; the change is observed only as connections
+  naturally cycle. Multigres refreshes them proactively so a pooled session
+  behaves like a fresh PostgreSQL connection.
+- **Standby/replica poolers refresh lazily.** The flagged statement executes
+  on the **primary**, so only the primary's pooler bumps immediately. Replica
+  poolers receive the catalog change via WAL but are not signalled, so their
+  pooled backends pick up the new default on their own lifecycle (idle /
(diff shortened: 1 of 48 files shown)


docs/query_serving/session_settings.md
```diff
@@ -27,6 +27,77 @@ or forwarded to PostgreSQL. See
...
 
+## Per-Database / Per-Role Default Changes (ALTER DATABASE/ROLE … SET)
+
+`ALTER DATABASE … SET` and `ALTER ROLE … SET` change _per-database_ / _per-role_
+session GUC defaults, stored in the `pg_db_role_setting` catalog. PostgreSQL
+applies these defaults to a session **only at session start** (they become the
+session's GUC _reset values_). A connection pooler therefore has a problem:
+backends opened **before** such an `ALTER` keep the old defaults until they
+reconnect — `RESET`, `RESET ALL`, and `DISCARD ALL` all revert to the
+session-start reset value, never re-reading `pg_db_role_setting`.
```
GuptaManan100, line 38: I'm not sure about this reasoning, in the sense that let's say a user runs an `ALTER DATABASE SET` and we reset our connection pooler, we'll still have a an issue where the client that is connected to the multigateway that never disconnected will immediately see new altered values... So it's unpredictable behaviour anyways.

go/services/multigateway/handler/connection_state.go
```diff
@@ -116,6 +116,14 @@ type MultiGatewayConnectionState struct {
 	// user SQL or at release. One-shot: cleared after it is consumed.
 	PendingMarkSessionStateUntrusted bool
 
+	// PendingInvalidateConnectionDefaults is set by the planner (via Route) when
+	// the current statement changes per-database/role session GUC defaults
+	// (ALTER DATABASE/ROLE ... SET, or an allowlisted CREATE EXTENSION).
+	// ScatterConn consumes it to set ExecuteOptions.InvalidatesConnectionDefaults
+	// so the multipooler refreshes its pooled connections once the change is
+	// durable. One-shot: cleared after the request is sent.
+	PendingInvalidateConnectionDefaults bool
```
GuptaManan100, line 125: Looking at the connection state struct, this is getting awfully big with info related to planning. We shouldn't be using conn state to port info from the planner to the execution options. I am fixing this as part of my refactor, but just flagging in case that lands first.

go/services/multipooler/internal/executor/executor.go
```diff
@@ -166,6 +175,60 @@ func (e *Executor) buildReservedStateFromAPI(rc reservedConnAPI) *query.Reserved
...
+	if marker != nil && txnStatus != protocol.TxnStatusIdle {
+		// In a transaction block (or a failed block whose COMMIT will roll back):
+		// defer to COMMIT.
+		marker.MarkPendingDefaultsInvalidation()
+		return
+	}
+	e.poolManager.InvalidateConnectionDefaults()
+	if refreshConn != nil && !refreshConn.IsReleased() {
+		_ = refreshConn.RefreshDefaultsIfStale(ctx)
+	}
```
GuptaManan100, line 222: This seems wrong. If the connection is not released, you should not be refreshing its defaults, that will lose all state, like temptable, open cursor etc.

GuptaManan100:
```sql
CREATE TEMP TABLE t (x int);          -- reserves backend (ReasonTempTable), Idle
ALTER DATABASE db SET search_path=…;  -- flagged → inline reconnect → temp table gone
SELECT * FROM t;                      -- ERROR: relation "t" does not exist
```

go/services/multipooler/internal/executor/executor.go
```diff
@@ -321,7 +386,8 @@ func (e *Executor) StreamExecute(
 			}
 		}
 
-		return e.streamExecuteOnReservedConn(ctx, reservedConn, sql, reservationOptions, e.sessionSettingsFromOptions(options), callback)
+		rs, err := e.streamExecuteOnReservedConn(ctx, reservedConn, sql, options, reservationOptions, e.sessionSettingsFromOptions(options), callback)
+		return rs, err
```
GuptaManan100, line 390: Needless unpack to rs, err, we can return directly

go/services/multipooler/internal/executor/executor.go
```diff
@@ -1758,10 +1840,22 @@ func (e *Executor) ConcludeTransaction(
...
+		preCommitTxnStatus := reservedConn.TxnStatus()
 		if err := reservedConn.Commit(ctx); err != nil {
 			reservedConn.Release(reserved.ReleaseError, nil)
 			return nil, nil, err
 		}
+		if shouldBumpDefaultsAfterCommit(pendingDefaultsInvalidation, preCommitTxnStatus) {
+			e.poolManager.InvalidateConnectionDefaults()
+			if !reservedConn.IsReleased() {
+				_ = reservedConn.RefreshDefaultsIfStale(ctx)
+			}
```
GuptaManan100, line 1857: This seems wrong too...

go/test/endtoend/queryserving/connection_defaults_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	var before string
+	require.NoError(t, c.QueryRowContext(ctx, "SHOW search_path").Scan(&before))
+	require.NotEqual(t, newSearchPath, before)
+
+	_, err = c.ExecContext(ctx, fmt.Sprintf("ALTER DATABASE %s SET search_path TO %s", dbName, newSearchPath))
+	require.NoError(t, err)
+
+	var after string
+	require.NoError(t, c.QueryRowContext(ctx, "SHOW search_path").Scan(&after))
```
GuptaManan100, line 205: We should access the temporary table afterwards, I believe that would surface the bug I'm talking about 👆

haritabh17:
We discussed and decided to not fix this issue related to `ALTER DATABASE ... SET` ... Because :

1. We found a cheaper and much simpler workaround to support postgis test suite.
2. Did not find any other usecase where the deviated behaviour from PG following `ALTER DB / ROLE ... SET` was causing any issues.

For now this is documented as a known deviation from PG - which is that after an `ALTER DB / ROLE ... SET` command the pooled connections will continue to have the older defaults and hence even new connections from client to gateway will not see the new defaults. 

Mitigation -- they can always do a SET on the new conns.

PR #1141, opened 2026-06-10
test(pgregresstest): run index_advisor, plpgsql_check, pgjwt, pgsodium, and pgtap suites through multigateway

haritabh17:
## Summary

Enrolls five more external extensions in the `pgregresstest` compatibility harness, running their shipped test suites through multigateway — **all of their tests pass**, with zero regressions in the existing suites:

| Extension | Harness | Tests | Pin |
|---|---|---|---|
| `index_advisor` | pg_regress | 7/7 ✅ | v0.2.0 (+ `hypopg` 1.4.2 built as its dependency) |
| `plpgsql_check` | pg_regress | 5/5 ✅ | v2.9.1 |
| `pgjwt` | pgTAP | 23/23 assertions ✅ | commit `f3d82fd` (upstream never tags) |
| `pgsodium` | pgTAP | ~480 assertions ✅ | commit `38d2289` (v3.1.9 fixtures predate PG17 array types) |
| `pgtap` | pg_regress | 33/33 ✅ | v1.3.4 (promoted from build-only dependency) |

## Harness changes

- **pgbuilder**: commit pinning (init + fetch-by-SHA) for untagged upstreams; `PkgConfigDeps` resolves system libs (libsodium) via pkg-config into `COPT` for PGXS builds.
- **New spec knobs**: `Commit`, `RegressTests` (explicit REGRESS list — plpgsql_check ships per-major `-14`…`-19` test files), `ExpectedSubdir`, `PreloadLibraries`, `PkgConfigDeps`; `ExcludeGlobs` now also filters the pg_regress wildcard.
- **Merged preload snippet**: `shared_preload_libraries` is composed from the union of selected extensions' `PreloadLibraries` into one generated conf snippet — per-extension conf files are last-write-wins per GUC and would clobber each other (pg_cron's preload line moved out of `pg_cron.conf`).
- **Phase-scoped server config**: external conf snippets now apply only to the external phase's cluster (new `ShardSetup.AddPgInitdbExtraConfFiles` + the existing pre-external reinit), so regression/isolation/contrib run on a **stock** cluster. This matters: plpgsql_check's cursor-leak detection is not inert (emits WARNINGs the core `plpgsql` test doesn't expect) and **cannot be disabled — turning `cursors_leaks` off hangs any exception-trapping plpgsql function, an upstream plpgsql_check bug reproduced on stock PostgreSQL 17.6 (no multigres) with both v2.9.1 and master**.
- **Generalized between-extension reset**: drops all extensions (except plpgsql) and all user schemas (except `public` and multipooler's `multigres`), not just `public` — extensions whose control file pins a non-public schema (pgmq, pg_graphql) leaked into pgtap's catalog-introspection assertions.
- pg_regress invoked with explicit `--expecteddir` (its default is the CWD, not `--inputdir`); `PG_CONFIGURE_EXTRA_ARGS` env for host-specific `./configure` flags; CI installs `libsodium-dev pkg-config`.

## Known divergences (documented in patch headers / spec comments)

- `plpgsql_check`: gateway rejects `LOAD` by design (the preload makes it a no-op) + one parser error-wording hunk → 3 patches.
- `pgtap/throwtap`: 4 of 97 assertions pass SQL-level prepared-statement names, which pgTAP runs as `EXECUTE` inside plpgsql — invisible to the gateway-owned `PREPARE` → narrow patch (the file completes; the other 93 assertions stay verified).
- `pgtap`: `performs_ok`/`performs_within`/`resultset`/`valueset` excluded — same prepared-statement pattern, but the error aborts each file's wrapping transaction, so a patch would have to absorb the whole file.

Catalog notes document why investigated extensions stay uncovered: `pgaudit` (\connect as password-auth users), `supabase_vault` (autocommit pgTAP — pooler limit), `http` (live HTTP server), `pg_jsonschema` (no SQL suite), `pg_net` (pytest + network), `hypopg`'s own suite (backend-local autocommit). PostGIS is out of scope (separate workstream).

## Validation (full local runs)

- **Regression 161/222** and **isolation 91/117** — failure lists byte-identical to the pre-change baseline (the core `plpgsql` test is unaffected by the preloads thanks to phase scoping)
- **Contrib 99/99**, **External 222/222** (175 existing + 47 new tests)
- External-only (`RUN_PGEXTERNAL=1`) and combined (`RUN_EXTENDED_QUERY_SERVING_TESTS=1`) paths both validated end-to-end
- New unit tests guard catalog/spec coherence (covered ⇒ spec exists, exactly one of Tag/Commit, DependsOn resolves, exclude/explicit-list selection modes)
- `golangci-lint`, `go vet`, `gofumpt` clean


GuptaManan100 (approved):
Love it! reviewed the patches specifically. All good!

PR #1149, opened 2026-06-11
refactor(multiorch): extract Run phases into methods on coordinatorLedRuleChange

mkindahl:
Break the four phases of Run() into separate methods — dispatchRecruit, collectRecruitsAndBuildProposal, dispatchPromote, and waitForPromotes — and move the recruitResult/promoteResult type definitions to package scope. Also hoists recruitMs computation so it is declared once and shared between the error and success event paths.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,6 +14,7 @@
 
 version: "2"
 run:
+  timeout: 10m
   build-tags:
     - ruleguard
 linters:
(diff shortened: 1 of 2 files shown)


rafael (approved):
Linter is not happy. Change makes sense to me.

mkindahl:
> Linter is not happy. Change makes sense to me.

Yeah, but it is happy locally. Investigating.

mkindahl:
> > Linter is not happy. Change makes sense to me.
> 
> Yeah, but it is happy locally. Investigating.

Seems we are exceeding the 5 minute timeout for a run, so updating it to 10 minutes.

PR #1150, opened 2026-06-11
feat(multipooler): map gateway vpids to backend pids via multigres.backend_vpid

haritabh17:
## Summary

Adds `multigres.backend_vpid`, a Multigres sidecar metadata table that maps active PostgreSQL backend PIDs to the multigateway virtual PIDs (`ClientConnectionId`) they are currently serving when backend VPID tracking is enabled.

## Current behavior

- The multipooler creates `multigres.backend_vpid` during sidecar schema bootstrap, alongside the other `multigres` schema tables.
- `multigres.backend_vpid` is unlogged because its rows are ephemeral active-session metadata. After a PostgreSQL promotion, the unlogged-table sweep may drop it, so the multipooler recreates the empty table before marking the promoted primary serving.
- Runtime writes to `multigres.backend_vpid` are gated by `--backend-vpid-tracking-enabled` and default to disabled for production.
- The pgregress isolation phase starts multipooler with backend VPID tracking enabled; other pgregress phases use the default disabled setting.
- When tracking is enabled, the multipooler upserts `(backend_pid, vpid)` when it hands a backend to a gateway session:
  - regular pooled connection checkout
  - newly reserved backend before any `BEGIN`
- Mapping writes happen only at autocommit handoff points, so rows are visible to other sessions for lock-wait probes.
- `ConnectionState` caches the active tracked vpid for each backend connection and skips duplicate upserts for the same active association.
- Rows are deleted at release/recycle boundaries so the table represents active gateway-to-backend associations.
- Clean reserved-connection release runs VPID cleanup through the existing reserved `Release(...)` flow.
- The pgregress isolation shim reads `multigres.backend_vpid` joined with `pg_stat_activity` to translate multigateway virtual PIDs into real PostgreSQL backend PIDs for `pg_blocking_pids` / `pg_safe_snapshot_blocking_pids` checks.

## Included fix

Settings containing `role` or `session_authorization` are re-applied when a pooled connection is reused, even when the interned settings pointer matches. PostgreSQL resolves these names to role OIDs at `SET` time, so reapplying keeps the backend aligned with the current catalog if a role name is dropped and recreated.

## Validation

- Multipooler package tests
- Promotion-path manager tests
- PostgreSQL compatibility test package compile/skip validation


go/services/multipooler/internal/executor/vpid_tracking.go
```diff
@@ -0,0 +1,139 @@
...
+// current role (e.g. a client SET ROLE to an unprivileged role).
+//
+// If the upsert fails, the query still runs; the affected behavior is lock-wait
+// translation for this gateway/backend association. No-op when the caller has
+// no client connection id, or this connection already recorded the same active
+// association.
+func (e *Executor) trackVpidOnRegular(ctx context.Context, conn *regular.Conn, options *query.ExecuteOptions) {
+	if options == nil || options.ClientConnectionId == 0 {
+		return
+	}
```
GuptaManan100, line 80: We should still guard this against a flag so that we can control to only do this work for the isolation tests, and not for production. Cause in production we don't read these vpids.

GuptaManan100 (commented):
The vpid tracking writes are now unconditional on the latency-sensitive query path: an upsert on every regular-conn checkout serving a new vpid, plus a DELETE on every recycle of a tracked regular conn, up to 2 extra round trips per query. The per-conn `trackedVpid` cache only elides this when the same session reuses the same backend.

The old design gated all of this behind `--vpid-stamp-enabled` (off in prod, on only for the pgregress suite); this PR removes that gate. We should keep it in my opinion.

And a small nit: PR description doesn't match the diff. The body claims this introduces a new `go/common/multigresschema` package and that the isolation table uses "shared DDL in `multigresschema`". Neither is in the PR: there is no `multigresschema` package, and the DDL is an inline string literal in `createBackendVpidTable` (`pg_multischema.go`);

go/services/multipooler/internal/executor/vpid_tracking.go
haritabh17: Ack. I reintroduced a runtime gate `--backend-vpid-tracking-enabled` now defaults to false. The table is still always created during sidecar schema bootstrap, but writes/cleanup only run when the flag is enabled. The
 pgregress isolation phase enables the flag when starting/restarting the cluster.

Also, updated the PR body to match the current design: the table is created in sidecar schema bootstrap and there is no `multigresschema` package.

shedges:
Overall idea looks interesting. Had couple of questions
1/ Will this work with replica's which are in standby mode and we are trying to insert row into Unlogged tables
2/ we are granting Insert/Delete/Update to Public on mapping table. Any client can corrupt the table
3/ dead tuple bloat can be issue over the time as frequent release (delete) backend  will leave dead tuples.

shedges (approved):
LGTM 🚀

PR #1152, opened 2026-06-11
test(multipooler): attempt to fix ADD_Success flakiness via post-test rule cleanup

dweitzman:
The manager's reconcile loop re-applies synchronous_standby_names from the committed rule after RestoreGUCs resets it. This could cause ValidateCleanState to fail when the rule is left modified after the test completes.

Fix: introduce setupUpdateConsensusRuleTest which registers a t.Cleanup to reset the rule instead of just resetting the GUC.


rafael (approved):
LGTM

PR #1154, opened 2026-06-12
fix(multipooler): check reservation reasons in the portal stage-1 admission gate

GuptaManan100:
## Problem

The two-stage graceful drain serves single autocommit queries during stage 1, gating each request on whether it will create a reserved connection. For `PortalStreamExecute` the gate only checked `MaxRows > 0`, but the executor reserves a connection on `MaxRows > 0` **or** non-zero reservation `reasons`.

This is reachable on the extended-query path: after a deferred `BEGIN`, the first portal folds the `BEGIN` in with `Reasons = ReasonTransaction` while `MaxRows == 0`. If a drain starts and that portal arrives during stage 1, the gate admits it as a single query, but the executor then opens a reserved connection — so the drain degrades to a force-close at `gracePeriod` instead of only serving autocommits.

(Found in review on #1122 by @haritabh17. Thank you!)

## Solution

Mirror the executor's reserve decision (`MaxRows > 0 || reasons != 0`) in both gates:

- Pooler admission gate (`grpcpoolerservice`): extract a `portalReserves(options, reservationOptions)` helper and use it to compute the `RequestKind`, so a reasons-carrying portal is classified as a new reservation and rejected/buffered during the drain.
- Gateway (`poolergateway`): apply the same check in `PortalStreamExecute` so a reserving portal is proactively buffered rather than sent through.

Adds `TestPortalReserves` covering the regression scenario (fetch-all portal with a folded-in `BEGIN`).

### DIFF
--- a/go/services/multigateway/poolergateway/pooler_gateway.go
+++ b/go/services/multigateway/poolergateway/pooler_gateway.go
@@ -267,10 +267,12 @@ func (pg *PoolerGateway) PortalStreamExecute(
 	callback func(context.Context, *sqltypes.Result) error,
 ) (*query.ReservedState, error) {
 	var state *query.ReservedState
-	// A portal reserves a connection only as a suspendable cursor (MaxRows > 0);
-	// a fetch-all portal (MaxRows == 0) runs on a pooled connection and is a
-	// single query — mirrors the pooler's classification.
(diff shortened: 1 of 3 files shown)


haritabh17 (approved):
Awesome!! 🚀

PR #1155, opened 2026-06-12
feat(observability): generate Prometheus metric catalog and keep-list

GuptaManan100:
## Summary

- Add a generator (`go/tools/metricsgen`) that statically scans the codebase for OpenTelemetry instrument definitions and emits an authoritative catalog of every metric Multigres exports (`go/observability/metriccatalog`).
- Prometheus names are computed with `github.com/prometheus/otlptranslator` — the same library the OTel Prometheus exporter uses — so the catalog stays in lock-step with what is actually scraped (unit/`_total` suffixes, histogram `_bucket`/`_count`/`_sum` expansion) instead of a hand-maintained transform that can drift.
- Exposes a global `PrometheusKeepListRegex` plus a per-binary `PrometheusKeepListByBinary` (derived from each `cmd/*` import graph) for Prometheus relabel `keep` rules.
- Wired into `make metrics` and `build-all`, so the existing `validate-generated-files` CI job fails if a metric is added without regenerating.

## Description

Each catalog entry records the OTel name, instrument kind, unit, defining package, source location, exported Prometheus base name, and the concrete series. Binaries are attributed by walking each `cmd/*` main's import graph, so shared `go/common` metrics (`rpcclient`, `topoclient`) list every binary that links them. This is static import reachability ("may expose"), the safe over-approximating direction for a keep list. Regenerate with `make metrics`.


haritabh17 (approved):
LGTM 🚀

PR #1156, opened 2026-06-12
feat(multigateway): resolve dynamic-name set_config for pg_dump

GuptaManan100:
## Summary

- pg_dump on PG17+ fails through the gateway: its startup probe `SELECT set_config(name, 'view, foreign-table', false) FROM pg_settings WHERE name = 'restrict_nonsystem_relation_kind'` is rejected because the `name` argument is a column reference, not a literal or bound parameter, so the session-scoped setting can't be tracked for pool-rotation replay.
- Adds a resolve-and-apply path for a `SELECT` whose target list is entirely `set_config(...)` with a dynamic argument: run the argument projection once to learn the concrete `(name, value, is_local)` tuples, track the session-scoped ones, and apply them with literals.
- pg_dump — and any version-guarded `set_config` over `pg_settings` — now works through the gateway.

## Problem

The gateway must know a session-scoped `set_config`'s GUC name up front to track it in `SessionSettings` and replay it after pool rotation. pg_dump's probe resolves the name per row from a `pg_settings` column, so the planner couldn't extract it and rejected the statement — blocking pg_dump before it emits any schema.

## Solution

When the analyzer sees a target list made entirely of `set_config(...)` with an argument that isn't a literal or bound parameter, the planner emits a `ResolveTrackSetConfig` primitive:

1. **Resolve** — run the projection of the call arguments (`set_config(a,b,c)` → `a, b, c`) once. The first resolution is the source of truth; re-running the original dynamic query could resolve to different rows/values.
2. **Track** — record the session-scoped (`is_local=false`) tuples in `SessionSettings` for replay.
3. **Apply** — run `set_config(<captured literals>, true)` and forward PostgreSQL's authoritative result. `is_local := true` returns the new value without leaving a session GUC on the pooled backend (which would leak to the next client); persistence comes from tracking + replay.

The literal/bound fast path and the rejection of dynamic `set_config` mixed with other target-list entries are unchanged.

## Testing

e2e `TestPgDump` (real `pg_dump` binary) and `TestDynamicSetConfig` (pool-rotation, zero-row no-op, multi-column, `is_local=true` transaction-scoped), plus analyzer/plan-shape/`sqltypes` unit tests; existing session-settings and extended-protocol suites still pass.

### DIFF
--- a/go/services/multigateway/planner/select_stmt.go
+++ b/go/services/multigateway/planner/select_stmt.go
@@ -17,6 +17,8 @@ package planner
 import (
 	"fmt"
 
+	"github.com/multigres/multigres/go/common/constants"
+	"github.com/multigres/multigres/go/common/mterrors"
 	"github.com/multigres/multigres/go/common/parser/ast"
 	"github.com/multigres/multigres/go/common/pgprotocol/server"
 	"github.com/multigres/multigres/go/services/multigateway/engine"
@@ -49,8 +51,12 @@ func (p *Planner) planSelectStmt(
 	stmt *ast.SelectStmt,
 	conn *server.Conn,
 	setConfigs []setConfigCall,
+	dynamicSetConfig bool,
 	opts PlannerOptions,
 ) (*engine.Plan, error) {
+	if dynamicSetConfig {
+		return p.planResolveSetConfig(sql, stmt)
+	}
 	if len(setConfigs) == 0 {
 		return p.planDefault(sql, stmt, conn, opts)
 	}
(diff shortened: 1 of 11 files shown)


go/services/multigateway/planner/select_stmt.go
```diff
@@ -49,8 +51,12 @@ func (p *Planner) planSelectStmt(
 	stmt *ast.SelectStmt,
 	conn *server.Conn,
 	setConfigs []setConfigCall,
+	dynamicSetConfig bool,
 	opts PlannerOptions,
 ) (*engine.Plan, error) {
+	if dynamicSetConfig {
+		return p.planResolveSetConfig(sql, stmt)
+	}
```
haritabh17, line 59: Not forwarding the `opts` parameter can result in session not pinned for advisory locks right?

haritabh17 (approved):
Rest LGTM 🚀

go/services/multigateway/planner/select_stmt.go
GuptaManan100: Yes, you're 100% right!

GuptaManan100: It would have to be something like `set_config('x', pg_try_advisory_lock(1)::text, false))`, cause this path only runs when set_config is the only function call 🤔

shedges (approved):

PR #1157, opened 2026-06-12
feat(multigres): support unlogged tables for short-lived workloads

GuptaManan100:
Unlogged tables were disabled because failover can leave their state inconsistent across hosts. This re-enables them for short-lived workloads with clear, documented failover semantics.

- **On promotion**, the new primary best-effort drops every unlogged table (plain `DROP TABLE`, never `CASCADE`). A table with no dependents disappears — clients get `42P01` and rebuild from scratch; one a view/function depends on is left empty. This replaces PostgreSQL's silent "table is now empty" with an explicit signal.
- **On create**, the gateway emits a `WARNING` (SQLSTATE `01000`) linking the docs, on both the simple (`Plan`) and extended (`PlanPortal`) query protocols. The statement still succeeds.
- Behavior documented in `docs/query_serving/unlogged_tables.md`, with unit + end-to-end coverage.

Not covered: `ALTER TABLE … SET UNLOGGED`. Extension-owned unlogged tables are swept like user tables (usually left empty due to dependencies).


go/services/multigateway/planner/planner.go
```diff
@@ -221,12 +221,36 @@ func (p *Planner) Plan(
 		return nil, err
 	}
 
+	// CREATE UNLOGGED TABLE (and its CREATE TABLE AS / SELECT INTO forms) routes
+	// normally, but unlogged data is never replicated and is lost on failover, so
+	// prepend a WARNING notice that points the user at the failover behaviour doc.
+	if isUnloggedCreate(stmt) {
```
haritabh17: The `CREATE UNLOGGED TABLE ... AS EXECUTE` (wrapped execute path) might get skipped here due to the The `tryUnwrapWrappedExecute` early-return before the `isUnloggedCreate` check.

go/services/multigateway/planner/planner.go
```diff
@@ -221,12 +221,36 @@ func (p *Planner) Plan(
...
+// isUnloggedCreate reports whether stmt creates an UNLOGGED relation, across the
+// plain CREATE TABLE, CREATE TABLE AS, and SELECT INTO forms.
+func isUnloggedCreate(stmt ast.Stmt) bool {
+	switch s := stmt.(type) {
+	case *ast.CreateStmt:
+		return s.Relation != nil && s.Relation.RelPersistence == ast.RELPERSISTENCE_UNLOGGED
+	case *ast.CreateTableAsStmt:
+		return s.Into != nil && s.Into.Rel != nil && s.Into.Rel.RelPersistence == ast.RELPERSISTENCE_UNLOGGED
+	case *ast.SelectStmt:
+		return s.IntoClause != nil && s.IntoClause.Rel != nil && s.IntoClause.Rel.RelPersistence == ast.RELPERSISTENCE_UNLOGGED
```
haritabh17, line 262: While at it lets also cover `UNLOGGED SEQUENCES`? Or fine to cover in another PR as well.

haritabh17 (approved):
Rest LGTM 🚀

docs/query_serving/unlogged_tables.md
```diff
@@ -0,0 +1,107 @@
...
+the WAL, so it is never streamed to standbys. After a failover the new primary
+has the table's catalog entry but no data — PostgreSQL resets every unlogged
+table to empty as part of promotion. To avoid silently presenting an empty table
+as if nothing happened, the freshly promoted primary **best-effort drops** every
+unlogged table so clients get an explicit error and can rebuild from scratch.
+
+## Background
+
+An `UNLOGGED` table skips WAL for its contents. This makes writes faster but has
+two consequences:
```
mkindahl, line 41: Might be useful to elaborate on why you want to use an unlogged table rather than a temporary table. This would give good context for the user, and also point to temporary tables as the solution when they can use it.

go/services/multigateway/planner/planner.go
mkindahl: Materialized views can be unlogged as well.

mkindahl (approved):
This covers unlogged tables, but also sequences and materialized views can be unlogged. Also suggest to add to the documentation why you might want to use an unlogged table, to steer users away from it and give alternatives when it is not necessary to use an unlogged table.

go/services/multigateway/planner/planner.go
GuptaManan100: Apparently its not allowed - 
```
=# CREATE UNLOGGED MATERIALIZED VIEW foo AS select 1;
ERROR:  materialized views cannot be UNLOGGED
```

docs/query_serving/unlogged_tables.md
GuptaManan100: Done!

PR #1162, opened 2026-06-12
feat(multigateway): route from pooler self_leadership in topology

dweitzman:
With this change, pooler.Type is no longer consulted for leader vs follower identity anywhere in multigateway. It's still used in targeting for now, though

LeadershipObservation is used instead

### DIFF
--- a/go/services/multigateway/poolergateway/load_balancer.go
+++ b/go/services/multigateway/poolergateway/load_balancer.go
@@ -43,11 +43,14 @@ type shardKey struct {
 // It creates connections based on discovery events and destroys them when poolers
 // are removed from discovery.
 //
-// Leader identity is tracked per-shard via the leaders map, populated only from
-// LeaderObservation messages on health streams. The pooler.Type field from etcd
-// topology is a hint about intended role — never the source of truth for which
-// pooler leads a shard. A leaders entry survives connection lifecycle events:
-// removing a pooler does not erase consensus's choice of leader.
+// Leader identity is tracked per-shard via the leaders map, populated from two
+// sources that both carry a rule number: a pooler's self_leadership in its
+// topology record (read on discovery) and LeaderObservation messages on health
+// streams. The most-authoritative observation (highest rule number) wins,
+// regardless of source. The pooler.Type field from etcd topology is never
+// consulted for leader identity — only self_leadership is. A leaders entry
+// survives connection lifecycle events: removing a pooler does not erase
+// consensus's choice of leader.
 type LoadBalancer struct {
 	// localCell is the cell where this gateway is running
 	localCell string
@@ -64,10 +67,12 @@ type LoadBalancer struct {
 	// connections maps pooler ID to PoolerConnection
 	connections map[string]*PoolerConnection
 
-	// leaders maps shard key to the highest-term LeaderObservation seen for that
-	// shard. Populated only from health-stream observations; never from topology.
-	// The leader_id field of the observation may name a pooler we are not
-	// currently connected to — GetConnection handles that case at read time.
+	// leaders maps shard key to the most-authoritative LeaderObservation seen
+	// for that shard (highest rule number wins). Populated from a pooler's
+	// self_leadership topology record and from LeaderObservation messages on
+	// health streams. The leader_id field of the observation may name a pooler
+	// we are not currently connected to — GetConnection handles that case at
+	// read time.
 	leaders map[shardKey]*clustermetadatapb.LeaderObservation
 
 	// onPrimaryServing is called when a new primary is detected via health stream.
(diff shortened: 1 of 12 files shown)


go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -539,18 +528,26 @@ func matchesShardTarget(conn *PoolerConnection, target *query.Target) bool {
...
+// signal carried on the health observation.
+func matchesReplicaTarget(conn *PoolerConnection, target *query.Target) bool {
 	if !matchesShardTarget(conn, target) {
 		return false
 	}
-	if confirmedLeaderID != "" {
-		return id != confirmedLeaderID
-	}
-	return conn.Type() == clustermetadatapb.PoolerType_REPLICA
+	return !conn.believesSelfLeader()
```
dweitzman, line 550: This shouldn't really change behavior, but now the way of thinking about whether a pooler is eligible for targeting as a REPLICA is:
- not the consensus leader
- not a stale leader

(In some future PR I'd like to change Target's enum so it's not PoolerType either)

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -173,37 +173,33 @@ func (lb *LoadBalancer) AddPooler(pooler *clustermetadatapb.MultiPooler) error {
...
+	obs := pooler.GetSelfLeadership()
+	if obs == nil {
 		return
 	}
-	if lb.leaders[key] != nil {
+	merged := commonconsensus.MostAuthoritativeObservation(lb.leaders[key], obs)
+	if merged == lb.leaders[key] {
 		return
 	}
-	lb.leaders[key] = &clustermetadatapb.LeaderObservation{
```
dweitzman, line 200: Previously this could be seeded with a synthetic LeaderObservation at coordinator term 0, leader subterm 0

Now that etcd has a leader observation (`pooler.GetSelfLeadership()` above) that's used directly

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -267,18 +263,11 @@ func (lb *LoadBalancer) GetConnection(target *query.Target) (*PoolerConnection,
...
-	if commonconsensus.CompareRuleNumbers(leaderObs.GetLeaderRuleNumber(), nil) > 0 {
-		confirmedLeaderID = poolerIDString(leaderObs.LeaderId)
-	}
+	// A pooler that believes itself the leader — the current leader or a stale
+	// leader — is excluded; see matchesReplicaTarget.
 	var candidates []*PoolerConnection
-	for id, conn := range lb.connections {
-		if matchesReplicaTarget(id, conn, target, confirmedLeaderID) {
+	for _, conn := range lb.connections {
+		if matchesReplicaTarget(conn, target) {
```
dweitzman, line 270: matchesReplicaTarget() is excluding any pooler that considers itself a leader (either the real leader or a stale leader) so it doesn't need a confirmedLeaderID as input anymore

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -539,18 +528,26 @@ func matchesShardTarget(conn *PoolerConnection, target *query.Target) bool {
...
+// leader. Leadership is judged per-connection from the better of the pooler's
+// etcd self_leadership and its health-stream observation (see
+// believesSelfLeader), never from the topology Type label.
+//
+// Excluding self-believed leaders covers both the current leader and a stale
+// leader — a pooler whose own view still names it the leader even though a
+// newer leader has been observed. Either would serve reads as a primary or from
+// a diverged timeline, so neither is a replica candidate.
+//
+// TODO(replica-quality): eligibility is otherwise binary. Longer term, treat a
```
dweitzman, line 541: Added a TODO here that the choice of replica in load balancing will probably eventually want to factor in things like replication lag, etc.

go/tools/ruleguard/rules.go
```diff
@@ -204,3 +204,24 @@ func disallowWallClockInConsensus(m dsl.Matcher) {
...
+
+// disallowMultiPoolerTypeForRouting flags reads of a MultiPooler record's Type
+// field in the multigateway, which must derive leader identity from consensus
+// state (self_leadership), never from the topology role label. The PoolerType
+// routing label on a query.Target is a different field and is unaffected;
+// constructing a record with a Type (struct literal) is also unaffected — only
+// reading .Type off a discovered MultiPooler / MultiPoolerInfo is disallowed.
+//
+// Use GetSelfLeadership() != nil instead.
+func disallowMultiPoolerTypeForRouting(m dsl.Matcher) {
```
dweitzman, line 217: Defense against a regression

rafael (approved):

PR #1163, opened 2026-06-12
refactor(topoclient): introduce typed ComponentID for serialized component IDs

dweitzman:
To reduce the chances of mistakes. Raw strings for structured data makes me a little nervous, but sometimes we need to do that because we the proto3 message doesn't make a good map key and can't serialize itself to a string.


go/services/multiadmin/backup.go
```diff
@@ -47,8 +47,8 @@ func (s *MultiAdminServer) Backup(ctx context.Context, req *multiadminpb.BackupR
 	}
 
 	// Generate a job ID that will be stored in pgbackrest and can be queried after restart.
-	// Format: YYYYMMDD-HHMMSS.microseconds_<pooler_name>
-	jobID := backup.GenerateJobID(pooler.Id.Name)
+	// Format: YYYYMMDD-HHMMSS.microseconds_<multipooler-cell-name>
+	jobID := backup.GenerateJobID(pooler.Id)
```
dweitzman, line 51: This is actually a slight behavior change, but it makes job names more consistent with what the pooler would have done

rafael (approved):
I did not look in depth, the overall refactor makes sense to me

PR #1165, opened 2026-06-13
refactor(multiorch): derive leader identity from consensus, not PoolerType

dweitzman:
## Summary

Leader/primary identity across multiorch is now derived from **consensus state**
(the highest known shard rule) rather than the `PoolerType` topology/health label.
`commonconsensus.HighestKnownRule(statuses)` is the single method that names the
leader.

## Bugs fixed

All variants of "judge leadership by PoolerType / self-claim / reachable-only"
instead of consensus identity, and all bite during the post-failover window
before the new leader's health has propagated:

1. **AppointLeader suppressed needed failovers.** The "primary already exists,
   skip appointment" guard keyed off `Status.PoolerType == PRIMARY`. A stale
   leader still in postgres-primary mode (reachable + ready, but not the consensus
   leader) was treated as the healthy primary, so `Execute` returned early and the
   failover never happened.

2. **Stale leader interpreted as primary in analysis.** `findHighestTermRawLeader`
   only considered poolers that *self-claimed* leadership, so a newly elected
   leader that hadn't reported health yet was invisible while a stale self-claimer
   was picked. `HighestKnownRule` reads each pooler's rule (incl. replication
   primary), so a follower replicating from the new leader reveals it.

3. **Stale-leader analyzer could demote the wrong node.** It used
   `HighestTermReachableLeader` — the highest-term *reachable* self-claimer. If the
   real new leader wasn't reachable yet, a stale-but-reachable leader became the
   "most advanced," and the analyzer flagged the real leader as stale. Now keyed
   off the reachability-independent `HighestShardRule`.

4. **FindHealthyPrimary mis-selected / refused during failover.** It accepted any
   postgres-PRIMARY node (could return a stale primary) and errored "multiple
   primaries found" when two nodes transiently reported PRIMARY — failing both
   callers (FixReplication, ReconcileCohort) outright. Now selects the single
   consensus leader.

5. **LeaderIsDead failed over a healthy primary whose pooler was down.** (Latent
   pre-existing bug, surfaced by `#2`.) Failover suppression required a recent
   *direct* leader-postgres observation, which a dead pooler can't provide — so a
   leader whose pooler died while postgres and replicas stayed healthy would fail
   over once the response window lapsed. On `main` this was masked because the old
   self-claim-only discovery (bug `#2`) lost track of such a leader entirely and
   `LeaderIsDead` bailed; consensus discovery recovers it, exposing the gap. Fix:
   when the leader pooler is unreachable, trust `ReplicasConnectedToLeader` (true
   only when replicas are actively streaming with fresh WAL heartbeats) as proof
   the leader's postgres is alive. The timestamp/threshold gate now applies only
   when the pooler is reachable. (This is what the `TestPoolerDownNoFailover`
   integration failure caught.). Also added some TODOs to LeaderIsDead for future
   changes to handle things like a permanent state where Postgres is up and the
   pooler is crashed, etc.

## Other changes

- **Phantom `0/0` replication primary.** proto3 can't distinguish an unset
  `ReplicationPrimary` from one written with the zero-valued initial rule (`0/0`,
  no leader — the shape of fresh cluster state, e.g. the first backup). Added
  `commonconsensus.ReplicationPrimaryOrNil` (returns nil for `0/0`) + a ruleguard
  banning raw `ConsensusStatus.GetReplicationPrimary()` reads so a phantom entry
  can't masquerade as a real leader.
- **ruleguard** now also forbids reading `MultiPooler.Type` for leader identity in
  multiorch (previously multigateway only).
- **FixReplication** no longer writes a pooler's topology `Type` to mark it broken;
  an infeasible `pg_rewind` returns `FAILED_PRECONDITION` (lifecycle-stage
  signalling to the provisioner is future work).

## Tests

Regression tests added for each bug above (failover-window leader discovery, stale
primary selection, AppointLeader stale-primary guard, LeaderIsDead suppression for
a dead pooler with streaming replicas), plus the phantom-`0/0` case and the
"leader named by a rule but absent from the store" invariant.

## Follow-ups (TODOs in code)

- `HighestKnownRule`: detect equivocation (same rule number naming different
  leaders = split brain) instead of silently keeping the first-seen rule.
- Split `LeaderIsDead` into distinct conditions (leader-unhealthy / unreachable
  with replica quorum / no heartbeats reaching WAL).
- Surface a broken pooler to the provisioner via its lifecycle stage rather than
  orch writing its topology `Type`.
- Re-implementing draining somehow. This PR actually removes the drain logic. If/when we re-add something like that, it seems like the pooler should self-drain since it owns its discovery record and probably it should drain via lifecycle status rather than pooler type


go/services/multiorch/recovery/actions/appoint_leader.go
```diff
@@ -80,19 +81,14 @@ func (a *AppointLeaderAction) Execute(ctx context.Context, problem types.Problem
...
-	// without going through the normal resign → appoint → demote flow), this check
-	// would miss it and proceed with an appointment unnecessarily.
+	// Leadership is judged from consensus (commonconsensus.IsLeader), never the
+	// PoolerType label: a pooler is the existing primary when its own consensus
+	// status names it leader. We also require it to be reachable (IsLastCheckValid)
+	// and PostgreSQL ready; if the pooler is up but Postgres is not ready, or it
+	// has signalled it needs replacement, we still need to trigger failover.
 	for _, pooler := range cohort {
 		if pooler.MultiPooler == nil ||
-			pooler.GetStatus().GetPoolerType() != clustermetadatapb.PoolerType_PRIMARY ||
```
dweitzman, line 95: This seems like a likely bug fix. Seems like previously a stale leader with a valid health check & postgres ready could have be misunderstood as a functioning leader and preventing appoint leader from acting

go/services/multiorch/recovery/actions/fix_replication.go
```diff
@@ -286,13 +271,19 @@ func (a *FixReplicationAction) tryPgRewind(
...
-		a.logger.WarnContext(ctx, "pg_rewind not feasible, marking as DRAINED",
+		// pg_rewind is not feasible (e.g. the required WAL has been recycled): the
+		// replica cannot rejoin and needs replacement.
+		//
+		// TODO: signal this durably via the pooler's lifecycle stage so the
+		// provisioner and orch treat the pooler as broken/needs-replacement. Orch
+		// must not write the pooler's topology Type — the pooler owns its own record
+		// and would clobber an external write. For now we surface an error so the
+		// failure is visible; the next recovery cycle will retry.
+		a.logger.WarnContext(ctx, "pg_rewind not feasible; pooler needs replacement",
```
dweitzman, line 275: For now, the code to drain stuck replicas is removed. It's not clear to me whether it actually worked, but I think the right way to do this would be have poolers self-drain if possible, since they own their registration record and can overwrite it at any time

go/services/multiorch/recovery/analysis/generator.go
```diff
@@ -440,45 +399,39 @@ func (g *AnalysisGenerator) computeShardLevelFields(sa *ShardAnalysis, poolers m
...
 		}
 	}
+	sa.HighestShardRule = commonconsensus.HighestKnownRule(statuses)
 
-	// Determine the highest-term reachable leader (used for stale-leader detection).
-	sa.HighestTermReachableLeader = findHighestTermLeader(sa.Leaders)
-
-	// Compute the highest-term leader in the shard regardless of reachability.
-	// This may differ from HighestTermReachableLeader when the leader pooler is down.
-	topologyPrimary := findHighestTermRawLeader(poolers)
```
dweitzman, line 455: This seems like it fixes a subtle bug as well: previously a stale leader could be interpreted as the topology primary here if, for example, a failover recently happened but we haven't been getting health updates from the real primary

findHighestTermRawLeader() was asking each pooler if it was the leader rather than asking each pooler who it thinks the leader is, which is what HighestShardRule does now

go/services/multiorch/recovery/analysis/stale_leader_analyzer.go
```diff
@@ -58,26 +58,24 @@ func (a *StaleLeaderAnalyzer) Analyze(sa *ShardAnalysis) ([]types.Problem, error
...
+	// Any reachable pooler whose own consensus status says it believes itself the
+	// leader of its term (commonconsensus.IsLeader) but is not that named leader is
+	// a stale leader to be demoted.
+	leaderID := sa.HighestShardRule.GetLeaderId()
+	if leaderID == nil {
 		return nil, nil
 	}
 
-	// A tie in LeaderTerm indicates a consensus bug — skip automatic demotion.
-	if sa.HighestTermReachableLeader == nil {
```
dweitzman, line 67: Again, this is a bug fix: previously if the real consensus leader was unreachable, a stale leader might be misinterpreted as leader so we wouldn't inform it who the real leader is

go/services/multiorch/store/pooler_store.go
```diff
@@ -152,67 +153,60 @@ func (s *PoolerStore) FindPoolerByID(id *clustermetadatapb.ID) (*multiorchdatapb
...
-// is PRIMARY are accepted, so stale topology entries running as standby are skipped.
+// Leadership comes purely from consensus: the highest known rule across the
+// poolers' consensus statuses names the leader (commonconsensus.HighestKnownRule),
+// never the PoolerType topology/health label. A follower can carry that rule via
+// its replication primary, so the leader is found even when its own snapshot is
+// stale. The named leader is then verified via a live Status RPC and must still
+// report itself as the consensus leader — a node that has since resigned or
+// dropped into recovery is rejected. Because there is exactly one highest rule,
+// there is no multiple-primary ambiguity to resolve here.
 func (s *PoolerStore) FindHealthyPrimary(
```
dweitzman: Previously this could have confused a stale primary for a real one, causing issues with fix_replication and reconcile_cohort

go/services/multiorch/recovery/analysis/generator.go
```diff
@@ -253,43 +245,18 @@ func (g *AnalysisGenerator) generateAnalysisForPooler(
...
+	}
 	for _, pooler := range poolers {
-		if pooler == nil || pooler.MultiPooler == nil || pooler.MultiPooler.Id == nil {
-			continue
-		}
-
-		cs := pooler.GetConsensusStatus()
-		isConsensusLeader := commonconsensus.IsLeader(cs)
-		isHealthPrimary := pooler.GetStatus().GetPoolerType() == clustermetadatapb.PoolerType_PRIMARY
-		if !isConsensusLeader && !isHealthPrimary {
```
dweitzman, line 282: The previous logic would treat a stale leader as the leader if either the real leader isn't reachable or is reachable and unhealhty

rafael (approved):
LGTM - 

I re-read the parts we discussed in the walk through. I did not look at the tests very closely

PR #1167, opened 2026-06-15
feat(multipooler): make the consensus rule authoritative for the pooler's role

dweitzman:
## Summary

Postgres Monitor now derives the pooler's role (`PoolerType`) from **consensus state**
— the committed rule plus the recorded replication primary — instead of being inferred
from the observed PostgreSQL recovery mode. The postgres monitor's
`determineRemedialAction` reconciles postgres *to* the rule rather than relabeling
the pooler *from* what postgres happens to be doing.

This is the pooler-side counterpart to the orchestrator-side change (#1165): there
the orchestrator stopped trusting `PoolerType` for leader identity; here the pooler
stops trusting its own observed postgres state for its role. `latestRule` ranks the
pooler's two rule sources via the same `commonconsensus.HighestKnownRule` the
orchestrator uses, so both sides can never disagree about which rule wins.

## What changed

`determineRemedialAction`'s postgres-running path is now a pure function of
*(intended role from the rule, postgres recovery mode, current pooler type)*:

- **rule says REPLICA but postgres is a primary** → demote: restart as a standby of
  the recorded leader (pg_rewind heals timeline divergence); wait if there is no
  recorded primary to rewind against rather than restart blind.
- **rule names us leader but postgres is a standby** → resign so the coordinator
  re-elects; we do not self-promote.
- **role drift / `UNKNOWN` at boot** → apply the rule-derived role via `SetState`
  (transitions the serving components and republishes the label + observation).
- **no rule observed yet** → wait.

The observed-role relabel actions (`AdjustTypeToPrimary` / `AdjustTypeToReplica`)
are removed.

Also drops vestigial error returns from `GetInconsistentRevocation`,
`getCachedConsensusStatus`, and `getInconsistentConsensusStatus`; the latter now
surfaces a real error on `ObservePosition` failure and its callers fall back to the
cached status **explicitly**, rather than the read silently returning stale data.

## Bugs this likely fixes

1. **A rogue primary that never self-demotes.** After a failover where
   the old leader couldn't be reached to demote (or its postgres crashed and came
   back up as a primary after a partial demote), the old monitor took **no action**
   on a postgres-primary whose rule named another leader — it relied entirely on the
   orchestrator's demote RPC landing. If that RPC failed, the node ran indefinitely
   as a second primary on a deposed term. The monitor now self-demotes (restart as
   standby + pg_rewind) when the recorded rule outranks its position.

2. **The real leader flapped to REPLICA on a transient postgres restart.** A brief
   postgres restart on the healthy leader (config reload, quick crash-recovery)
   made the old monitor relabel the consensus leader to REPLICA from observed
   recovery mode — dropping it from the gateway and flapping the label when postgres
   came back. The published role now follows consensus, not transient postgres state.

3. **Premature relabel at boot**, before the pooler had observed its rule (waits for
   `UNKNOWN` now; fuller boot-`UNKNOWN` handling is a follow-up).


go/common/consensus/compare.go
```diff
@@ -107,6 +107,63 @@ func MostAdvancedPosition(statuses []*clustermetadatapb.ConsensusStatus) *cluste
...
+// written with a zero-valued rule, and the initial cluster state is exactly
+// that zero value: rule number 0/0 with no leader (e.g. the first backup is
+// taken at rule 0/0 with an empty cohort and no leader). A 0/0 replication
+// primary therefore means "no leader has been established to replicate from"
+// and is reported as absent. The first appointed leader always advances the
+// rule number past 0/0, so a real replication primary is never dropped.
+//
+// Always use this instead of ConsensusStatus.GetReplicationPrimary() (enforced
+// by ruleguard) so a phantom 0/0 entry never gets mistaken for a real one.
+func ReplicationPrimaryOrNil(cs *clustermetadatapb.ConsensusStatus) *clustermetadatapb.ReplicationPrimary {
```
dweitzman, line 123: This is duplicated from https://github.com/multigres/multigres/pull/1165 and will go away while resolving merge conflicts

go/services/multipooler/internal/manager/manager.go
```diff
@@ -776,6 +776,86 @@ func (pm *MultiPoolerManager) leaderObs(rule *clustermetadatapb.ShardRule) *clus
 	}
 }
 
+// deriveTypeAndObs maps a consensus rule to the pooler role it implies and the
+// self-leadership observation to publish. A nil rule (no rule known yet) yields
+// UNKNOWN; a rule naming this pooler yields PRIMARY plus an observation; any
+// other rule yields REPLICA with no observation. This is the single place the
+// rule is translated into a role.
+func deriveTypeAndObs(rule *clustermetadatapb.ShardRule, selfID *clustermetadatapb.ID) (clustermetadatapb.PoolerType, *clustermetadatapb.LeaderObservation) {
```
dweitzman: The direction of all this is that the highest known shard rule determines the pooler type & leader observation

go/services/multipooler/internal/manager/manager.go
```diff
@@ -776,6 +776,86 @@ func (pm *MultiPoolerManager) leaderObs(rule *clustermetadatapb.ShardRule) *clus
...
+	return t
+}
+
+// staleStandbyDemoteTarget returns the leader this pooler should restart as a
+// standby of, for the case where consensus has moved on but postgres is still
+// running as a primary on a deposed term. Returns nil unless there is a recorded
+// primary with usable contact info, the recorded rule strictly outranks our
+// applied position, the recorded leader is not us, and the rule is not revoked —
+// without a target to rewind against we wait rather than restart blind.
+func (pm *MultiPoolerManager) staleStandbyDemoteTarget() *clustermetadatapb.PoolerAddress {
```
dweitzman: This is a practical thing: replication primary indicates the host/port of the leader, but the rule store entries in WAL do not (at least currently), so we can't rewind a stale primary until after we've learned the host/port of a newer leader.

go/services/multipooler/internal/manager/manager.go
```diff
@@ -1972,24 +2070,53 @@ func (pm *MultiPoolerManager) determineRemedialAction(ctx context.Context, curre
...
+			// observed postgres state.
+			return remedialActionNone
 		}
-		if !currentState.isPrimary && pm.getPoolerType() == clustermetadatapb.PoolerType_PRIMARY {
-			return remedialActionAdjustTypeToReplica
+
+		// Rule says replica but postgres is a primary: restart as a standby of the
+		// recorded leader. Waits if we have no host/port to rewind against rather
+		// than restart blind.
+		if intended == clustermetadatapb.PoolerType_REPLICA && currentState.isPrimary {
```
dweitzman: If postgres is primary but consensus role is follower, try to demote

Previously the remedial action would have been `remedialActionAdjustTypeToPrimary` to reflect postgres state into the PoolerType

go/services/multipooler/internal/manager/manager.go
```diff
@@ -1972,24 +2070,53 @@ func (pm *MultiPoolerManager) determineRemedialAction(ctx context.Context, curre
...
+
+		// Rule names us leader but postgres is a standby: resign so the coordinator
+		// re-elects. We do not self-promote.
+		//
+		// TODO(leader-led-leader-changes): promote here instead of resigning (reuse
+		// Promote's promoteStandbyToPrimary), and disambiguate resign (we lost our
+		// postgres) from promote (newly elected). Embedding the leader host/port in
+		// the WAL rule would also let replicas reconcile without waiting for
+		// SetPrimary.
+		if intended == clustermetadatapb.PoolerType_PRIMARY && !currentState.isPrimary {
```
dweitzman: If postgres is in recovery mode and consensus role is leader, make sure we've resigned the leadership

Previously `remedialActionAdjustTypeToReplica` would have changed the PoolerType to replica instead, treating the postgres state as authoritative

go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -807,8 +782,12 @@ func (pm *MultiPoolerManager) setPrimaryLocked(ctx context.Context, req *consens
 		if _, err := pm.restartAsStandbyLocked(ctx, leader.GetHost(), port); err != nil {
 			return nil, err
 		}
-		if err := pm.resetSynchronousReplication(ctx); err != nil {
-			pm.logger.WarnContext(ctx, "Failed to reset synchronous replication", "error", err)
+		// Clear synchronous_standby_names now that this node is a read-only standby.
+		// Route it through the rule store's SyncStandbyManager (not a direct
+		// ALTER SYSTEM) so the manager stays the sole writer and its cache stays
+		// coherent.
+		if err := pm.rules.ClearSyncStandby(ctx); err != nil {
```
dweitzman, line 789: This is a subtle bug fix that's arguably independent of these changes, but I stuffed it in here. Feel free to ask me to separate it out

go/services/multipooler/internal/manager/manager.go
```diff
@@ -165,9 +163,18 @@ type MultiPoolerManager struct {
...
+	// run pg_rewind to remove potential phantom / non-durable WAL entries.
+	//
+	// We suspect divergence when:
+	//   - a primary learns its term was ended by a failover — it may have local
+	//     WAL the new leader's history never adopted. Set by emergencyDemoteLocked,
+	//     SetPrimary's stale-primary branch, and the monitor's stale-primary demote.
+	//   - a replica cannot replicate despite successfully connecting to the leader,
+	//     implying its timeline diverged. Set via RewindToSource; today orch detects
+	//     this, but in the future a pooler could self-detect it.
+	suspectedDivergence atomic.Bool
```
dweitzman, line 177: I renamed `rewindPending` to `suspectedDivergence` while I was in here, which feels like a more honest name

go/services/multipooler/internal/manager/postgres_monitor.go
dweitzman, line 1: I split out a new file for this, which make make reviewing harder. If you look at the commits within this PR you can see the diff more cleanly

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -0,0 +1,805 @@
...
+	}
+
+	// Acquire action lock before taking remedial action
+	lockCtx, err := pm.actionLock.Acquire(ctx, "MonitorPostgres")
+	if err != nil {
+		pm.logger.InfoContext(ctx, "MonitorPostgres: failed to acquire action lock", "error", err)
+		return postgresState{}, err
+	}
+	defer pm.actionLock.Release(lockCtx)
+
```
dweitzman, line 220: A subtle bug fix: there was leftover code here to block the monitor from acting if a rewind was pending. That was leftover from when `rewindPending` meant postgres was stopped

One reason I renamed it to `divergenceSuspected` is to make the meaning more clear, and we want to allow the monitor to restart postgres even when divergence is suspected, etc.

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -0,0 +1,805 @@
...
+	if state.isPrimary && pm.rules.HasInconsistentGUC(ctx) {
+		return remedialActionReconcileGUC
+	}
+
+	return remedialActionNone
+}
+
+// determineRemedialAction decides what action to take based on discovered state.
+// This is pure decision logic with no side effects.
+func (pm *MultiPoolerManager) determineRemedialAction(ctx context.Context, currentState postgresState) remedialAction {
```
dweitzman, line 484: This method is split up into several smaller methods now

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -0,0 +1,805 @@
...
+		// Unparsable conninfo is itself drift worth fixing.
+		return true
+	}
+	return parsed.GetHost() != targetHost || parsed.GetPort() != targetPort
+}
+
+// determineRoleAction returns the action needed to align this pooler's role with
+// the rule-derived intended role: demote a stale primary, resign when the rule
+// names us leader but postgres is a standby, etc.
+func (pm *MultiPoolerManager) determineRoleAction(intended clustermetadatapb.PoolerType, state postgresState) remedialAction {
```
dweitzman, line 403: A lot of the changes relate to this method. The previous code would try to align PoolerType with the postgres set (if postgres isn't in recovery mode, change the pooler type to PRIMARY, etc.)

Now consensus state is authoritative, and the comments here are trying to be clear about what that means

go/test/endtoend/multiorch/dead_primary_recovery_test.go
```diff
@@ -301,24 +301,30 @@ func TestDeadPrimaryRecovery(t *testing.T) {
 		"all multipoolers should be healthy",
 	)
 
-	// Verify replicas have primary_term = 0 (never been primary)
-	t.Logf("Verifying replicas have primary_term = 0...")
-	for name, inst := range setup.Multipoolers {
-		if name == setup.PrimaryName {
-			continue // Skip primary
+	// After the failovers settle, every pooler — the final primary included —
+	// must agree on who leads the shard.
```
dweitzman, line 305: Snuck in some fixes to dead primary test to try to reduce flakes and better model how consensus works

go/test/endtoend/multiorch/dead_primary_recovery_test.go
```diff
@@ -340,10 +346,15 @@ func TestDeadPrimaryRecovery(t *testing.T) {
...
 
-		var result int
-		err = db.QueryRow("SELECT 1").Scan(&result)
-		require.NoError(t, err, "Should be able to query new primary")
-		assert.Equal(t, 1, result)
+		// Attempt a write to prove the final primary is actually writable. SELECT 1
+		// succeeds on a standby too; a write does not — a node still in recovery
+		// rejects it ("cannot execute ... during recovery"). Roll back so we leave no
+		// state behind and don't block on synchronous-commit acknowledgment.
+		tx, err := db.Begin()
```
dweitzman, line 353: Making sure the primary is actually writable

go/test/endtoend/multiorch/dead_primary_recovery_test.go
```diff
@@ -538,14 +556,23 @@ func TestDeadPrimaryRecovery(t *testing.T) {
 		finalPrimaryInst := setup.GetMultipoolerInstance(finalPrimaryName)
 		require.NotNil(t, finalPrimaryInst)
 
+		// Snapshot the primary's current LSN before reading, so standbys can
```
dweitzman, line 559: This is a flake fix

proto/clustermetadata.proto
```diff
@@ -664,6 +664,14 @@ message AvailabilityStatus {
...
   // non-member signaling INELIGIBLE should be skipped when growing the cohort.
   CohortEligibilityStatus cohort_eligibility_status = 2;
+
+  // suspected_divergence is true when the pooler suspects its local WAL may have
+  // diverged from the cluster's chosen history and a pg_rewind may be needed to
+  // rejoin replication as a standby. This is a best-effort, in-memory signal that's
+  // usually set when a former primary learns an failover happened. A pooler with
+  // suspected divergence may be slower to endorse a new rule proposal because it
+  // needs to stop, run pg_rewind, and restart.
+  bool suspected_divergence = 3;
```
dweitzman, line 674: While I was renaming rewindPending -> suspected_divergence, it seemed like a small and worthwhile change to publish suspected_divergence so orch and consider the relative fitness of different potential participants in endorsing a new rule proposal

go/services/multipooler/internal/manager/consensus/rule_store.go
```diff
@@ -160,6 +166,12 @@ func (rs *ruleStore) ReconcileGUC(ctx context.Context, inRecovery bool) error {
 	})
 }
 
+// ClearSyncStandby clears synchronous_standby_names via SyncStandbyManager so the
+// manager's cache stays coherent. See SyncStandbyManager.Clear.
+func (rs *ruleStore) ClearSyncStandby(ctx context.Context) error {
```
dweitzman, line 171: The rule store has exclusive ownership of the SyncStandbyManager, so this method just helps delegate

rafael (approved):
Overall direction makes sense to me. I did not review in depth

PR #1169, opened 2026-06-16
refactor(connpool): taint logical replication conns on acquire

haritabh17:
## Summary
- add `Pooled.TaintOnRecycle()` for borrowed conns that must stay capacity-accounted but never return to idle pool
- mark logical replication backends tainted immediately after replication-mode acquisition
- remove the logical-replication release-boundary special case

## Tests
- `go test -short ./go/services/multipooler/internal/pools/connpool ./go/services/multipooler/internal/pools/reserved`
- `go test -short ./go/services/multipooler/...`


GuptaManan100 (approved):
Love it!

PR #1170, opened 2026-06-17
refactor(multigateway): unify query planning and thread reservation directives via PlanExecInfo

GuptaManan100:
## Summary

- Collapse the simple and extended (portal) query-planning entry points into a single `Planner.Plan`, so both protocols resolve and execute through one cached `Plan`.
- Replace the dedicated `TempTableRoute` / `AdvisoryLockRoute` primitives and the one-shot `state.Pending*` connection-state fields with a `PlanExecInfo` value threaded through `StreamExecute` / `PortalStreamExecute`.
- Net result: ~300 fewer lines, two routing primitives and several mutex-guarded `Pending*` accessors deleted, and per-query reservation intent no longer leaks onto per-connection state.

## Description

**Unify planning (commit 1).** `PlanPortal` duplicated `Plan`'s statement dispatch — every non-nil branch delegated back to `Plan` and re-derived the same local-vs-forward split the per-statement planners already encode. It's deleted. The executor's `PortalStreamExecute` now resolves a plan (cached iff cacheable) and calls `plan.PortalStreamExecute`, collapsing its three branches into one: a plain Route in portal mode reissues the portal to the multipooler, exactly what the old raw-forward escape hatch did. `Plan` gains `PlanOptions{IsPortal}`, which gates a single plan-time decision — the wrapped-EXECUTE unwrap (EXPLAIN EXECUTE / CREATE TABLE AS EXECUTE) is simple-protocol only. Because the plan cache is shared across protocols, `IsPortal: true` is confined to the non-cacheable path so cached plans stay protocol-agnostic by construction. A side effect of full unification: SQL-level PREPARE/EXECUTE/DEALLOCATE over the extended protocol now get the same gateway-consolidator handling as the simple protocol.

**Thread reservation directives (commit 2).** Temp-table and advisory-lock routing are now a plain `Route` plus a plan-level `PlanExecInfo` carrying the connection-reservation signals (temp-table, advisory-lock pin/recheck, portal pin/release). Those signals are scoped to a single execute call, so they ride on the call rather than on per-connection state — the old `Pending*` fields leaked single-query intent onto the connection. Cursor and rollback primitives augment the threaded `PlanExecInfo` with their runtime-computed portal release set, and `planType` refines the observability label for a `Route` from its `ExecInfo` so spans and query logs keep the previous `TempTableRoute` / `AdvisoryLockRoute` names.


haritabh17 (approved):
🚀 Nice!!!

PR #1171, opened 2026-06-17
test(pgregresstest): Docker patch-regen harness and divergence patches

GuptaManan100:
## Summary

- **Docker patch-regen harness** (`make pgregress-update-patches-docker`): reproduces the CI `ubuntu-24.04` environment (apt deps, Go from `go.mod`, Rust/cargo-pgrx for pg_graphql, clang/libclang for pgrx bindgen, pgbackrest from PGDG for PG17 cluster bootstrap, etcd, and the glibc locales the collate suite needs; runs as a non-root user). This lets the pgregress patch set be regenerated on any host — regenerating directly on macOS produces platform-divergent patches (locale collation, timezone/datestyle, error-cursor positions) that don't match the Linux CI runner.
- **Accepted-divergence patches**: per-test patches recording known, by-design multigres divergences so those regression tests verify cleanly — multigateway parser error wording, error-cursor caret shifts from query rewriting, single-database catalog behavior, by-design pooler rejections (FDW/SERVER/CREATE/DROP DATABASE/LOAD), and the unlogged-sequence warning. Genuine proxy bugs (wrong query plans, internal-error leaks, invalid SQL the gateway wrongly accepts) are deliberately left unpatched so they stay visible as failures to fix.
- **Documentation**: every committed patch now carries a `# Known divergence:` header explaining the why and the what. The patch pipeline strips the preamble before applying and preserves it across regeneration, so the explanations survive future patch updates.

## Notes

The full extended suite was regenerated and run green on Linux through the new harness (regression + isolation + contrib + external). This branch is merged up to date with `main`; the overlap with #1166 (unlogged-table warnings) deduplicated cleanly, leaving the explanatory headers as the net contribution for those files.


docker/Dockerfile.pgregress
```diff
@@ -0,0 +1,128 @@
...
+    && locale-gen
+ENV LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8
+
+# Go toolchain — pinned to go.mod (keep in sync with the `go` directive there).
+# Installed under /usr/local (world-readable), usable by the non-root user below.
+ARG GO_VERSION=1.25.1
+RUN ARCH="$(dpkg --print-architecture)" \
+    && curl -fsSL "https://go.dev/dl/go${GO_VERSION}.linux-${ARCH}.tar.gz" -o /tmp/go.tgz \
+    && tar -C /usr/local -xzf /tmp/go.tgz \
+    && rm /tmp/go.tgz
```
haritabh17, line 98: Lets also have SHA256 checksum verification to be in line with other existing downloads?

docker/Dockerfile.pgregress
```diff
@@ -0,0 +1,128 @@
...
+
+# etcd — topology store the multigres cluster-under-test starts (matches the
+# version setup-test-environment.sh installs in CI).
+ARG ETCD_VERSION=v3.6.4
+RUN ARCH="$(dpkg --print-architecture)" \
+    && curl -fsSL "https://github.com/etcd-io/etcd/releases/download/${ETCD_VERSION}/etcd-${ETCD_VERSION}-linux-${ARCH}.tar.gz" -o /tmp/etcd.tgz \
```
haritabh17, line 112: Same here

docker/Dockerfile.pgregress
```diff
@@ -0,0 +1,128 @@
...
+
+USER ${USER}
+ENV HOME=/home/${USER}
+ENV PATH=/usr/local/go/bin:/home/${USER}/.cargo/bin:$PATH
+ENV GOTOOLCHAIN=local
+
+# Rust toolchain — installed under the non-root user's home so the harness's
+# `cargo install cargo-pgrx` (pinned to the version pg_graphql's crate requires,
+# currently 0.16.1) can write to $CARGO_HOME/bin at runtime.
+RUN curl -fsSL https://sh.rustup.rs | sh -s -- -y --profile minimal --default-toolchain stable
```
haritabh17: And here.

docker/pgregress-generate.sh
```diff
@@ -0,0 +1,121 @@
...
+    git config --global --add safe.directory "*"
+    echo "== make build =="
+    make build
+    echo "== start port pool server =="
+    ./bin/portpoolserver --socket /tmp/multigres-port-pool.sock &
+    PP=$!
+    trap "kill $PP 2>/dev/null || true" EXIT
+    export MULTIGRES_PORT_POOL_ADDR=/tmp/multigres-port-pool.sock
+    echo "== pgregress ('"${PATCH_MODE}"') =="
+    go test -v -timeout '"${TIMEOUT}"' -run TestPostgreSQLRegression ./go/test/endtoend/pgregresstest/...
```
haritabh17: Very minor as its not a risk here - go test -v -timeout '"${TIMEOUT}"' - safer approach would be to pass as env var (-e PATCH_MODE=...) and reference them inside the single-quoted script?

haritabh17 (approved):
Rest LGTM 🚀

docker/Dockerfile.pgregress
GuptaManan100: Fixed

docker/Dockerfile.pgregress
GuptaManan100: Fixed

docker/Dockerfile.pgregress
GuptaManan100: Fixed

docker/pgregress-generate.sh
GuptaManan100: Fixed

PR #1172, opened 2026-06-17
chore(lint): upgrade golangci-lint to v2.12.2, scope gosec to non-test code, clear debt

dweitzman:
**Upgrades the pinned golangci-lint v2.6.1 → v2.12.2** (in
`.github/workflows/golangci-lint.yml`). The newer version pulls in lint debt and
an expanded gosec rule set — including the experimental "via taint analysis"
G7xx rules. This PR cleans up the fallout and sets a deliberate gosec policy so
CI is green on the new version. It started as a pure `--fix` run but now includes
manual changes; the repo is lint-clean (`0 issues`) under v2.12.2.

## CI
- Bump `golangci/golangci-lint-action` pinned `version: v2.6.1` → `v2.12.2` so CI
  actually enforces everything below.

## Automated (`golangci-lint --fix`, no behavior change)
- Modernizations across touched packages: `atomic.AddInt32/LoadInt32(&x)` →
  `atomic.Int32` methods, `WriteString(fmt.Sprintf(...))` → `fmt.Fprintf`,
  manual reverse loops → `slices.Backward`, `slog.NewTextHandler(io.Discard)` →
  `slog.DiscardHandler`.
- sloglint: `Error`/`Info` → `ErrorContext`/`InfoContext` in multiadmin.

## gosec policy
- **Scope G101 (hardcoded credentials) + the G7xx taint-analysis family
  (G703 path traversal, G704 SSRF, G705 XSS, G706 log injection, G710 open
  redirect) to test code only** (`_test.go`, `go/test/`, `testutil/`). On test
  code these are false positives — GUC-name maps, example AWS keys from the
  docs, computed temp paths. **All gosec checks stay active for non-test
  code**, where taint analysis still has value.
- Dropped the now-redundant `//nolint:gosec` directives in test files (kept the
  ones in non-test `root.go`/`postgres.go`, where G101 still applies).
- **Annotated every non-test gosec finding with a justified `#nosec`** after
  review — all were false positives: operator-configured file paths (password
  file, testdata dir, pooler data dir, sampling config, CI
  `GITHUB_STEP_SUMMARY`); embedded-template content / writes after
  `http.NotFound`; a fan-out `cancel` stored and invoked on unsubscribe; dev/CI
  tool URL + log args from the CI runner.

## Test-code findings resolved
The classic gosec checks (G118/G122/G202) stay active in test code. The current
findings were addressed directly:
- **G118 fixes** (real): `writer_validator.Stop()` now calls `cancel()` after
  draining workers; `process.go` derives its stop-timeout context from the
  detached parent instead of `context.Background()`.
- **G122/G202 annotated** as false positives: cleanup/diagnostic `WalkDir` over
  the test's own temp dir; `DROP TABLE` with a constant test table name. (An
  `os.Root` rewrite was considered but rejected — no untrusted symlink TOCTOU on
  a directory the test itself creates.)

## Refactor
- **Moved `s3mock` from `go/tools/` to `go/test/s3mock`.** It's a mock S3 server
  used solely by the multipooler e2e backup tests; it doesn't fit `go/tools/`'s
  charter (generic, non-multigres-specific helpers). The move also brings it
  under the test-code gosec exclusion. (This commit also folds in a `process.go`
  change swapping `context.Background()` for `ctxutil.Detach` — a small behavior
  change, not pure lint.)


.golangci.yml
```diff
@@ -316,6 +316,15 @@ linters:
       - linters:
           - gosec
         text: "G204:"
+      # In test code only, silence G101 (hardcoded credentials) and the gosec
+      # taint-analysis family (G7xx: path traversal, SSRF, XSS, log injection,
+      # open redirect). Tests run against controlled inputs — GUC-name maps,
+      # example AWS keys, computed temp paths — so these fire as false positives.
+      # We keep all of them active for non-test code.
+      - linters:
```
dweitzman, line 324: While I was in here, also excluded tests from some security checks related to tainted user input or hardcoded credentials

rafael (approved):

PR #1173, opened 2026-06-18
test(pgregresstest): run pg_prewarm, pgstattuple, and pg_walinspect suites through multigateway

shedges:
Summary
----

Generalize the contrib test runner so modules whose Makefile is NO_INSTALLCHECK where make installcheck is a no-op and the REGRESS list lives only in the Makefile can run through the harness, by driving pg_regress directly with an explicit REGRESS list instead of only supporting the make installcheck path.

Enroll three more contrib extensions in the pg_regress compatibility harness: `pg_prewarm`, `pgstattuple`, and `pg_walinspect`.

Net result: +113/−3 across catalog, harness, docs, and one patch; coverage grows by three extensions / four tests, added through a single `contribRegressTests` map entry rather than a dedicated code path per extension, and the existing per-module patch pipeline applies to the new modules unchanged.

Description
----
Generalize the contrib runner. `RunContribTests` only knew how to drive make -C contrib/<mod> installcheck. A module whose Makefile sets `NO_INSTALLCHECK` makes that target a no-op and keeps its REGRESS list only in the Makefile, so the harness produced zero tests for it. The runner now consults a `contribRegressTests` map: for a listed module it invokes the built `pg_regress` directly with the explicit REGRESS list (mirroring the Makefile's REGRESS line), reading sql/+expected/ from the source tree and writing results/ into the build tree so the existing `verifyContribModule` patch pipeline applies unchanged. No new test phase, type, or per-extension code path. One map entry enrolls a `NO_INSTALLCHECK` module.

Enroll the extensions. `pg_prewarm` (1 test) is a plain installcheck module. A catalog flip to covered, runs on the stock contrib path, output matches upstream with no patch. `pgstattuple` (1 test) is covered the same way; its only divergence is the test's foreign-data-wrapper section — multigateway blocks `CREATE FOREIGN DATA WRAPPER` / `CREATE SERVER` by design (see `unsafe_stmt.go`), so those statements and the dependent foreign-table checks fail with multigres-specific errors, captured in a per-module patch. While the core tuple/index/page statistics on tables, indexes, sequences, toast tables, and partitions match upstream. `pg_walinspect` (2 tests) is `NO_INSTALLCHECK` because it needs `wal_level=replica`, which the multigres cluster already provides via its standby. So it needs no preload and no dedicated phase, riding the new explicit-REGRESS path on the stock cluster. Its tests assert COUNT(*)>=1 booleans rather than exact LSNs, so they pass unchanged (no patch), including the replication-slot, CHECKPOINT, and role/privilege steps the gateway does not block.

Validated locally (RUN_PGCONTRIB, macOS): pg_prewarm 1/1, pgstattuple 1/1 (foreign-data-wrapper section patched), pg_walinspect 2/2 (no patch) — 4/4 in a single contrib run.

### DIFF
--- a/docs/query_serving/extensions.md
+++ b/docs/query_serving/extensions.md
@@ -22,8 +22,11 @@ complete `pg_available_extensions` inventory.
 | `fuzzystrmatch` | contrib  | -                                                                                              |
 | `hstore`        | contrib  | -                                                                                              |
 | `ltree`         | contrib  | -                                                                                              |
+| `pg_prewarm`    | contrib  | -                                                                                              |
 | `pg_trgm`       | contrib  | -                                                                                              |
+| `pg_walinspect` | contrib  | `NO_INSTALLCHECK`; explicit REGRESS list (cluster provides `wal_level=replica`).               |
 | `pgcrypto`      | contrib  | Requires PostgreSQL to be built with OpenSSL.                                                  |
+| `pgstattuple`   | contrib  | FDW-section patched (gateway blocks `CREATE FOREIGN DATA WRAPPER` / `CREATE SERVER`).          |
 | `unaccent`      | contrib  | -                                                                                              |
 | `uuid-ossp`     | contrib  | Requires PostgreSQL UUID support.                                                              |
 | `http`          | external | pgsql-http; local httpbin; upstream/autocommit suite; timeout-wording patch.                   |
--- a/go/test/endtoend/pgregresstest/testdata/pg17/patches/contrib/pgstattuple/pgstattuple.patch
+++ b/go/test/endtoend/pgregresstest/testdata/pg17/patches/contrib/pgstattuple/pgstattuple.patch
@@ -1,0 +1,61 @@
+# Known divergence: multigateway blocks CREATE FOREIGN DATA WRAPPER and CREATE
+# SERVER by design (foreign-data wrappers open outbound connections, which the
+# connection pooler does not permit; see
+# go/services/multigateway/planner/unsafe_stmt.go). pgstattuple's test creates a
+# dummy FDW, server, and foreign table only to assert its functions reject
+# foreign tables; with the FDW/server creation rejected, those statements and the
+# dependent foreign-table checks instead fail with multigres-specific "does not
+# exist" errors. Not a multigres bug. The rest of the suite (tuple/index/page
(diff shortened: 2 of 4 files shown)


docs/query_serving/extensions.md
```diff
@@ -2,56 +2,35 @@
...
+| `hstore`        | contrib  | -                                                                                   |
+| `ltree`         | contrib  | -                                                                                   |
+| `pg_trgm`       | contrib  | -                                                                                   |
+| `pg_prewarm`    | contrib  | -                                                                                   |
+| `pg_walinspect` | contrib  | `NO_INSTALLCHECK` (needs `wal_level=replica`, already provided by the cluster's standby); run via an explicit REGRESS list. Tests assert `COUNT(*)>=1` rather than exact LSNs, so they pass through the gateway unchanged (no patch). |
+| `pgcrypto`      | contrib  | Requires PostgreSQL to be built with OpenSSL.                                       |
+| `pgstattuple`   | contrib  | Foreign-data-wrapper section of the test is patched: multigateway blocks `CREATE FOREIGN DATA WRAPPER` / `CREATE SERVER` by design, so those checks fail with multigres-specific errors. Core tuple/index/page stats match upstream. |
```
GuptaManan100: This table is alphabetically ordered, we should make sure it stays that way 😄

go/test/endtoend/pgregresstest/testdata/pg17/patches/contrib/pgstattuple/pgstattuple.patch
```diff
@@ -0,0 +1,51 @@
+--- a
```
GuptaManan100, line 11: Let's add a small comment to the top

GuptaManan100 (approved):
Rest LGTM, but we should merge only after we check that the CI is working as expected in the run from `Run Extended Query Serving Tests` after we resolve the merge conflicts

docs/query_serving/extensions.md
shedges: done

go/test/endtoend/pgregresstest/testdata/pg17/patches/contrib/pgstattuple/pgstattuple.patch
shedges: done

PR #1174, opened 2026-06-18
feat: match PostgreSQL statement_timeout error and add max_connections knob

GuptaManan100:
## Summary

- The gateway-managed `statement_timeout` out-of-range error now matches PostgreSQL exactly. It includes the `ms` unit on the value and on both range bounds (e.g. `-1 ms is outside the valid range for parameter "statement_timeout" (0 ms .. 2147483647 ms)`), and it now also rejects values above the upper bound instead of silently overflowing the duration.
- New `MULTIGRES_PG_MAX_CONNECTIONS` env var on the cluster image raises PostgreSQL's `max_connections` and sizes the connection pooler's global capacity to that value minus a 10-connection reserve, so PostgreSQL and the pooler stay consistent. This is backed by a new `CONNPOOL_GLOBAL_CAPACITY` env binding on the multipooler, plus docs and a docker-compose CI check.

### DIFF
--- a/docker/cluster-entrypoint.sh
+++ b/docker/cluster-entrypoint.sh
@@ -37,6 +37,31 @@ NUM_CELLS="${MULTIGRES_NUM_CELLS:-2}"
 # `tenant_db`). Additional cells use consecutive ports (base+1, base+2).
 GATEWAY_PG_PORT="${MULTIGRES_GATEWAY_PG_PORT:-15432}"
 
+# PostgreSQL max_connections. The bundled default is 60 (Supabase Pico preset).
+# Setting this raises PostgreSQL's ceiling AND sizes the connection pooler to
+# match: the pooler's global capacity is set to this value minus POOL_RESERVE,
+# leaving headroom for superuser logins, replication, and the pooler's own admin
+# pool. So the pooler never tries to open more backends than PostgreSQL allows.
+#
+#   MULTIGRES_PG_MAX_CONNECTIONS=100   # postgres=100, pooler global capacity=90
+#
+# Applied only at first init (a fresh data dir).
+PG_MAX_CONNECTIONS="${MULTIGRES_PG_MAX_CONNECTIONS:-}"
+
+# Connections held back from the pooler when MULTIGRES_PG_MAX_CONNECTIONS is set.
+POOL_RESERVE=10
+
+# Extra PostgreSQL configuration, as raw postgresql.conf text (one setting per
+# line) for anything MULTIGRES_PG_MAX_CONNECTIONS doesn't cover. Appended
+# verbatim onto every cell's generated postgresql.conf at data-dir init,
+# last-write-wins, so it overrides the bundled defaults:
+#
+#   MULTIGRES_PG_EXTRA_CONF=$'shared_buffers = 256MB\nwork_mem = 8MB'
+#
+# Applied only at first init (a fresh data dir); changing it after a cluster has
+# already initialized has no effect, matching POSTGRES_INITDB_EXTRA_CONF.
+PG_EXTRA_CONF="${MULTIGRES_PG_EXTRA_CONF:-}"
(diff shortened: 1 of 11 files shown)


docker/cluster-entrypoint.sh
```diff
@@ -108,6 +143,35 @@ if [ ! -f "${CONFIG_PATH}/multigres.yaml" ]; then
...
+# snippet paths pgctld appends at init). The local provisioner spawns pgctld and
+# multipooler with the container's environment, so exporting here reaches every
+# cell. Our snippet is appended last so it wins under postgres' last-write-wins,
+# even if the caller already pointed POSTGRES_INITDB_EXTRA_CONF at their own file.
+if [ -n "${PG_MAX_CONNECTIONS}" ] || [ -n "${PG_EXTRA_CONF}" ]; then
+  extra_conf_file="${CONFIG_PATH}/pg-extra.conf"
+  : >"${extra_conf_file}" # truncate so reruns don't accumulate duplicate lines
+
+  if [ -n "${PG_MAX_CONNECTIONS}" ]; then
+    echo "max_connections = ${PG_MAX_CONNECTIONS}" >>"${extra_conf_file}"
```
shedges, line 160: Is it possible for PG_MAX_CONNECTIONS to be already present in the conf file?

shedges (approved):
Rest LGTM

docker/cluster-entrypoint.sh
GuptaManan100: Yes, then it gets appended, so it overrides it!

go/services/multigateway/handler/statement_timeout.go
```diff
@@ -68,8 +69,11 @@ func ParsePostgresInterval(paramName, value string) (time.Duration, error) {
 		return 0, invalidParamError(paramName, value,
 			`Valid units for this parameter are "us", "ms", "s", "m", "h".`)
 	}
-	if d < 0 {
-		return 0, outOfRangeParamError(paramName, value)
+	// PostgreSQL range-checks in the base unit (milliseconds), and reports the
+	// value converted to it. d < 0 also catches negative sub-millisecond values
+	// that truncate to 0 ms.
+	if ms := d.Milliseconds(); d < 0 || ms > constants.MaxStatementTimeoutMS {
+		return 0, outOfRangeParamError(paramName, ms)
```
haritabh17, line 80: Will `outOfRangeParamError` the message like "0 ms is outside the valid range ...". Because d.Milliseconds() truncates to zero, yielding ms = 0.

docker/cluster-entrypoint.sh
```diff
@@ -108,6 +143,35 @@ if [ ! -f "${CONFIG_PATH}/multigres.yaml" ]; then
...
+    # multipooler reads CONNPOOL_GLOBAL_CAPACITY; keep it below max_connections.
+    export CONNPOOL_GLOBAL_CAPACITY=$((PG_MAX_CONNECTIONS - POOL_RESERVE))
```
haritabh17, line 162: If `MULTIGRES_PG_MAX_CONNECTIONS` gets changed between restarts with a persistent volume, the pooler's `CONNPOOL_GLOBAL_CAPACITY` will reflect the new value while PG's `max_connections` remains at the original value from first init. On every start, also rewrite PG's `max_connections` ??

haritabh17 (approved):
Rest LGTM 🚀

docker/cluster-entrypoint.sh
GuptaManan100: Documented as an issue for now!

go/services/multigateway/handler/statement_timeout.go
GuptaManan100: Fixed! And added e2e tests to verify we match postgres errors exactly in all these cases.

PR #1175, opened 2026-06-18
feat(pgprotocol): support direct SSL negotiation

haritabh17:
## Summary
- Add PostgreSQL 17 `sslnegotiation=direct` support to pgprotocol client and server, including TLS-first startup detection and mandatory `postgresql` ALPN negotiation.
- Wire `--pg-client-sslnegotiation` through multipooler → PostgreSQL connections so poolers can use direct TLS to Postgres.
- Add gateway TLS metrics for negotiated vs direct TLS and direct-TLS rejection reasons, and refresh the generated metric catalog.
- Add unit and end-to-end coverage for direct TLS, ALPN rejection, RequireTLS, SCRAM-PLUS, cancel requests, real `psql`, and multipooler backend TLS.

## Problem
PostgreSQL TLS normally starts after an SSLRequest round trip. For Envoy SSLRequest offload / TLS-first forwarding and PostgreSQL 17 clients using `sslnegotiation=direct`, multigateway must accept a TLS ClientHello as the first byte instead of treating it as an invalid startup packet. Direct TLS also requires ALPN so the server can verify the TLS connection is intended for PostgreSQL and avoid cross-protocol confusion.

## Solution
The server now peeks the first startup byte and, when it sees a TLS handshake record, replays any buffered ClientHello bytes into a server TLS handshake, requires ALPN `postgresql`, and then continues normal startup over the encrypted connection. The client side parses and validates `sslnegotiation`, starts TLS immediately for direct mode, offers and requires ALPN, and multipooler passes the setting into Postgres client configs. This PR also adds direct-TLS metrics and caps golangci-lint concurrency to keep full-repo lint within GitHub runner memory limits.

## Validation
- `make metrics`
- `go run ./go/tools/metricsgen/main -verify`
- `make build`
- `go test -short ./go/common/pgprotocol/client ./go/common/pgprotocol/server -run 'Test(ParseSSLNegotiation|ValidateSSLNegotiation|DirectTLS)' -count=1`
- `MULTIGRES_PORT_POOL_ADDR=/tmp/multigres-port-pool.sock go test ./go/test/endtoend/queryserving -run 'TestMultiGateway_SSL_(DirectNegotiation|NegotiatedStillWorks)' -count=1 -timeout=30m -v`
- `MULTIGRES_PORT_POOL_ADDR=/tmp/multigres-port-pool.sock go test ./go/test/endtoend/multipooler -run 'TestPGClientTLS_DirectNegotiation' -count=1 -timeout=30m -v`
- `golangci-lint run` under a 7 GB Docker memory limit
- commit hook lint passed

### DIFF
--- a/go/common/pgprotocol/server/direct_tls.go
+++ b/go/common/pgprotocol/server/direct_tls.go
@@ -1,0 +1,191 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Direct TLS (PostgreSQL 17 "sslnegotiation=direct") server-side support.
+//
+// PostgreSQL 17 added a TLS-first connection mode: instead of the classic
+// SSLRequest → 'S' → handshake dance, the client opens the TCP connection
+// and immediately sends a TLS ClientHello. The server disambiguates by
+// looking at the first byte — 0x16 is the TLS "handshake" record type and
+// can never start a legitimate startup packet (it would imply a length of
+// hundreds of megabytes, far past MaxStartupPacketLength).
+//
+// This mirrors ProcessSSLStartup in PostgreSQL's
+// src/backend/tcop/backend_startup.c, with the same hard requirements:
+//
+//   - Direct TLS is accepted whenever TLS is configured; there is no
+//     opt-in knob (PG 17 likewise has no GUC for it).
+//   - ALPN ("postgresql", RFC 7301) is MANDATORY for direct connections.
+//     A client that completes the handshake without negotiating it is
+//     rejected as a protocol violation. This is PG's cross-protocol-attack
+//     hardening: without ALPN, a generic TLS client (e.g. an HTTPS client
+//     coaxed into connecting here) could be replayed against the database.
+//     Negotiated (SSLRequest) connections keep ALPN optional, matching PG's
+//     compatibility posture for pre-17 clients.
+//   - When TLS is not configured, the connection is closed without writing
+//     anything: the peer is speaking TLS, so a plaintext ErrorResponse
+//     would be unintelligible garbage. PostgreSQL also rejects silently.
+//
+// This is the server half of the Envoy SSLRequest-offload model (MINT-37):
+// an edge proxy can answer the client's SSLRequest itself and forward the
+// raw TLS byte stream, which the gateway then sees as a direct TLS
+// connection.
+
+package server
+
+import (
+	"bytes"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"slices"
+
+	"github.com/multigres/multigres/go/common/mterrors"
+	"github.com/multigres/multigres/go/common/pgprotocol/protocol"
+)
+
+// directTLSFirstByte is the first byte of every TLS record carrying a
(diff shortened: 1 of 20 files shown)


go/common/pgprotocol/server/startup.go
```diff
@@ -191,31 +212,51 @@ func (c *Conn) handleSSLRequest() error {
...
 			outcome = TLSOutcomeClientAborted
 		}
-		c.metrics().RecordTLSHandshake(c.ctx, outcome, time.Since(handshakeStart))
-		return fmt.Errorf("TLS handshake failed: %w", err)
+		c.metrics().RecordTLSHandshake(c.ctx, negotiation, outcome, time.Since(handshakeStart))
+		return nil, fmt.Errorf("TLS handshake failed: %w", err)
 	}
 	handshakeDuration := time.Since(handshakeStart)
-	c.metrics().RecordTLSHandshake(c.ctx, TLSOutcomeSuccess, handshakeDuration)
+	c.metrics().RecordTLSHandshake(c.ctx, negotiation, TLSOutcomeSuccess, handshakeDuration)
```
shedges, line 260: Not related to this PR: Handshake metrics are useful as we also have TLS connection metrics

go/common/pgprotocol/server/direct_tls.go
```diff
@@ -0,0 +1,191 @@
...
+	//
+	// Divergence from PostgreSQL, deliberately: PG logs a COMMERROR and
+	// closes without telling the client anything. We have a fully
+	// established TLS channel at this point, so we return the FATAL
+	// PgDiagnostic (same wording, SQLSTATE 08P01 protocol_violation) and
+	// let serve()'s startup-error path deliver it through the tunnel —
+	// strictly more debuggable for the operator of a misconfigured client,
+	// and wire-safe because the bytes are encrypted.
+	if tlsConn.ConnectionState().NegotiatedProtocol != protocol.ALPNProtocol {
+		c.metrics().RecordDirectTLSRejected(c.ctx, DirectTLSRejectedReasonNoALPN)
```
shedges, line 141: `tlsConnections` is already incremented inside `completeTLSHandshake`, rejecting the connection here due to a direct TLS violation creates a metrics discrepancy.  Both metrics make sense at an individual  level but they will cause a data mismatch when aggregated at the multigateway level.

go/services/multipooler/internal/connpoolmanager/config.go
```diff
@@ -225,6 +232,10 @@ func NewConfig(reg *viperutil.Registry) *Config {
 			Default:  "",
 			FlagName: "pg-client-sslrootcert",
 		}),
+		pgSSLNegotiation: viperutil.Configure(reg, "connpool.pg.sslnegotiation", viperutil.Options[string]{
+			Default:  string(client.SSLNegotiationPostgres),
```
shedges, line 236: As a optimisation - if Multipooler performs Postgres version check once during startup and if Postgres is PG17+ then set `pg-client-sslnegotiation`to `direct`? It will reduce one round trip between Multipooler and Postgres.

shedges (commented):
Added few minor comments

go/common/pgprotocol/server/direct_tls.go
haritabh17: Ack, updating this so `completeTLSHandshake` only records handshake-attempt metrics now. The admitted TLS connection counter is recorded by the negotiated path after handshake success, and by the direct path only after ALPN validation passes. Missing-ALPN direct TLS still records `tls.handshake.duration{outcome="success"}` and `tls.direct_rejected{reason="alpn_missing"}`, but no longer increments `tls.connections`.

go/services/multipooler/internal/connpoolmanager/config.go
haritabh17: That’s fair, but I guess for now we can keep it aligned with community libpq defaults, i.e. `sslnegotiation=postgres` remains the default, and `direct` is explicit opt-in. 

Auto-switching based on a PG17+ version probe would be a Multigres-specific behavior change and would also need some changes in the default `sslmode` because since `direct` is only valid for TLS-enforcing modes. I am a bit inclined to do it separately if needed.

shedges:
LGTM

Verolop (approved):

PR #1183, opened 2026-06-23
feat(query): replace Target's pooler_type with Mode enum + ShardKey

dweitzman:
## Summary

Reshapes `query.Target` from `{table_group, shard, pooler_type}` to
`{ShardKey, Mode}`, where `ShardKey` is the existing
`clustermetadata.ShardKey` (gaining a `database` field) and `Mode` is a
new enum expressing caller intent (`WRITABLE` / `CONSISTENT` /
`INCONSISTENT`) rather than entity classification (PRIMARY/REPLICA).

The gateway and multiadmin no longer consult `MultiPooler.Type` for
routing — consensus state (`self_leadership` / `LeaderObservation`) is
the only source of truth. A ruleguard rule locks this in.

## What's in here

- **Proto**: new `Mode` enum, `Target` carries `ShardKey + Mode`. Old
  field numbers reserved. `StreamPoolerHealthResponse.target` removed
  (its only consumer was a buffer-drain race check now expressed via
  `serving_status` + `leader_observation`).
- **Database threading**: every gateway `Target` construction site
  carries the connection's database; `matchesShardTarget` filters on
  the full `(database, tableGroup, shard)` tuple. Cross-database
  collisions on `(tableGroup, shard)` names can no longer pick the
  wrong shard.
- **Buffer-drain race fix**: the gateway now requires the pooler's own
  broadcast `leader_observation` to name itself before firing
  `OnLeaderServing` (replacement for the dropped `Target.pooler_type ==
  PRIMARY` check). The multipooler-side promote sequence reordered so
  `leader=self` is broadcast only after `SetState(PRIMARY, SERVING)`
  completes — closes the window structurally.
- **Multiadmin backup**: `findPoolerForBackup` now uses
  `self_leadership` (with highest-rule tiebreak) instead of the
  topology Type label — backups taken from a stale leader on a
  divergent timeline would be unrestorable.
- **Ruleguard**: forbids `clustermetadata.PoolerType` enum mentions in
  gateway production code; extends `disallowMultiPoolerTypeForRouting`
  to cover multiadmin.
- **Renames**: `onPrimaryServing` → `onLeaderServing` (consensus
  vocabulary, not postgres-replication vocabulary).
- **Bug fixed**: four internal `shardKey` constructions dropped the
  `database` field — every lookup missed in production. Caught by
  making test fixtures use `constants.DefaultPostgresDatabase` to
  mirror production shape.

## Follow-ups (logged in memory, not blocking)

- Auth tablegroup discovery (still hardcoded `"default"`)
- Per-database notification fanout
- Manager startup leader-broadcast ordering (analogous race during
  restart recovery, lower practical risk)
- `BackupRequest.force_primary` → `force_leader` proto rename
- `cache.GetByShard` switch once gateway populates shard
- Couple `RecordTermPrimary` + `UpdateLeaderObservation` so the
  broadcast-ordering bug becomes structurally impossible


proto/query.proto
```diff
@@ -187,22 +187,44 @@ message ParameterDescription {
...
+// Why a mode rather than a pooler-type axis: callers care about "can I
+// write?" and "is the read consistent?", not about which fleet (PRIMARY vs
+// REPLICA) the underlying pooler belongs to. Decoupling lets the gateway
+// take advantage of cluster shape — e.g. routing CONSISTENT reads to a
+// caught-up sync standby — without callers having to know.
+enum Mode {
+  MODE_UNSPECIFIED = 0;
+  // WRITABLE: must be the consensus leader. Reads here are also strongly
+  // consistent. This is the only mode that accepts writes.
+  MODE_WRITABLE = 1;
```
dweitzman, line 203: I'm not attached to the values here. I'm also fine with LEADER / FOLLOWER for consensus role, or PRIMARY and STANDBY/REPLICA if folks prefer to think in terms of postgres role

The important thing from my perspective is just making this separate from PoolerType

go/services/multiadmin/backup.go
```diff
@@ -133,20 +140,36 @@ func (s *MultiAdminServer) findPoolerForBackup(ctx context.Context, database, ta
...
-
-		// Find a pooler of the target type
 		for _, info := range poolerInfos {
-			if info.MultiPooler.Type == targetType {
-				return info.MultiPooler, nil
+			poolers = append(poolers, info.MultiPooler)
+		}
+	}
+
+	if forceLeader {
```
dweitzman, line 148: This is a drive-by fix for multiadmin, while I'm in here. Leadership needs to be determined through the most authoritative leadership observation, not just any pooler with a given target type

proto/query.proto
```diff
@@ -187,22 +187,44 @@ message ParameterDescription {
...
+}
 
-  // pooler_type is the type of pooler to route to (PRIMARY, REPLICA)
-  // If not specified (UNKNOWN), defaults to PRIMARY
-  // TODO: Change from PoolerType something else, like a TargetType or ServingType enum.
-  clustermetadata.PoolerType pooler_type = 3;
+// Target identifies the destination for query execution: which shard, and
+// what mode (writability / consistency) the caller needs. The gateway
+// translates the mode into a specific pooler at routing time.
+message Target {
```
dweitzman, line 219: These changes to the Target proto3 message are the essence of the PR, and other changes are mostly the consequences

haritabh17 (approved):
🚀

go/test/endtoend/queryserving/pgbouncertests/nonexistent_db_test.go
```diff
@@ -78,22 +86,17 @@ func TestNonexistentDatabaseConnectBehavior(t *testing.T) {
...
-	require.NoError(t, err,
-		"DIVERGENCE: the multigateway does not validate the connection database, so a nonexistent name is accepted")
-	defer gwConn.Close(ctx)
-
-	var served string
-	require.NoError(t, gwConn.QueryRow(ctx, "SELECT current_database()").Scan(&served))
-	t.Logf("multigateway accepted database %q and served %q instead", nonexistentDB, served)
-	assert.Equal(t, "postgres", served,
-		"the gateway serves its backend database regardless of the requested (nonexistent) name")
+	// Multigateway (DIVERGENCE): the credential lookup is routed to the
```
dweitzman, line 89: @haritabh17 @GuptaManan100 

The changes in this PR impacted this test

PR #1184, opened 2026-06-23
fix(multipooler): gate writable-only work on physical primary, not consensus role

dweitzman:
The heartbeat writer and LISTEN/NOTIFY listener keyed off PoolerType==PRIMARY, which now means "consensus leader" rather than "postgres out of recovery." A leader whose postgres dropped into recovery (the resign window) kept issuing failing heartbeat INSERTs against a read-only standby every interval, spamming the log.

Rework the StateManager around a Mutate(ctx, func(*servingStateMutation)) core that fans the effective (leader, postgresPrimary, servingStatus) tuple out to components and persists topology-backed fields. Components now receive OnStateChange(isConsensusLeader, postgresPrimary, servingStatus); the heartbeat writer and listener require postgresPrimary in addition to leadership. PoolerType is derived from the leadership observation rather than carried explicitly, and the health streamer translates it back to the wire label.

The postgres monitor reconciles physical-primary drift as a remedial action (remedialActionReconcileState, which also subsumes role drift); the resign branch syncs postgresPrimary=false so the writer stops promptly. Mutate dedups on the effective tuple so record-only changes don't re-notify components.

Rename the manager field servingState -> stateManager.

Add a `writable` (`!pg_is_in_recovery`) bit to the health stream so consumers can distinguish "consensus leader" from "accepting writes." The multigateway uses it to hold failover buffer-drain until the new leader's postgres is out of recovery — needed now that leadership is advertised before promotion completes — and the demote paths sync `postgresPrimary=false` so a freshly-demoted standby never advertises `writable=true`.


go/services/multipooler/internal/manager/manager.go
```diff
@@ -1438,6 +1414,15 @@ func (pm *MultiPoolerManager) promoteStandbyToPrimary(ctx context.Context, state
 		return mterrors.Wrap(err, "failed to promote standby")
 	}
 
+	// TODO: Kicking off a checkpoint here is useful, but there's still a race condition where attempting
+	// to pg_rewind against a new primary before the checkpoint completes could crash the replica.
+	// There will be a follow-up PR soon to wait until the new primary is ready to be rewound against.
+	go func() {
+		if err := pm.exec(pm.ctx, "CHECKPOINT"); err != nil {
```
dweitzman: There's a latent rewind bug here that another open PR helps address, and for some reason these changes seem to trigger it more often.

This code in this placement isn't a bad thing to do. It's intentionally placed to cause merge conflict with my other PR so the merge will ultimately do the right thing

go/services/multipooler/internal/manager/state_manager.go
```diff
@@ -86,66 +114,124 @@ func (ssm *StateManager) RegisterAndSync(ctx context.Context, component StateAwa
...
+//
+// The action lock is always required: record.Mutate asserts it for topology
+// writes, and routing physical-state changes through the same gate keeps a
+// recovery flip from racing a role transition. It is asserted up front, before
+// fanning out, so a caller without the lock cannot half-transition the components
+// and only fail later at record.Mutate.
+//
+// PoolerType is derived from the leadership observation, and the poolerRecord
+// enforces the Type ⇔ SelfLeadership invariant.
+func (ssm *StateManager) Mutate(ctx context.Context, fn func(s *servingStateMutation)) error {
```
dweitzman, line 174: A significant change here is replacing SetState on StateManager with Mutate where you can optionally change some or all parts of state

Also, consensus role (leader / follower) and postgres role (primary vs recovery mode) are now explicitly separate

go/services/multipooler/internal/heartbeat/repltracker.go
```diff
@@ -82,9 +82,12 @@ func (rt *ReplTracker) makeNonPrimary() {
...
 // OnStateChange transitions the heartbeat tracker based on the serving state.
-// Starts the heartbeat writer for (PRIMARY, SERVING), stops it otherwise.
-func (rt *ReplTracker) OnStateChange(_ context.Context, poolerType clustermetadatapb.PoolerType, servingStatus clustermetadatapb.PoolerServingStatus) error {
-	if poolerType == clustermetadatapb.PoolerType_PRIMARY && servingStatus == clustermetadatapb.PoolerServingStatus_SERVING {
+// The writer runs only when this pooler is the consensus leader AND postgres is
+// out of recovery AND serving; otherwise the reader runs. Being the leader does
+// not imply a writable postgres — writing heartbeats to a standby fails every
+// interval and spams the log — so postgresPrimary must gate the writer too.
+func (rt *ReplTracker) OnStateChange(_ context.Context, isConsensusLeader, postgresPrimary bool, servingStatus clustermetadatapb.PoolerServingStatus) error {
+	if isConsensusLeader && postgresPrimary && servingStatus == clustermetadatapb.PoolerServingStatus_SERVING {
```
dweitzman, line 90: A bug fix: only try to write heartbeats if postgres in in a writable mode. Simply being consensus leader isn't enough

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -569,27 +583,40 @@ func (pm *MultiPoolerManager) takeRemedialAction(ctx context.Context, action rem
...
-		pm.logger.InfoContext(ctx, "MonitorPostgres: applying consensus role from rule",
-			"intended_role", intended.String())
-		if err := pm.servingState.SetState(ctx, intended, obs, clustermetadatapb.PoolerServingStatus_SERVING); err != nil {
-			pm.logger.ErrorContext(ctx, "MonitorPostgres: failed to apply role from rule", "error", err)
+		pm.logger.InfoContext(ctx, "MonitorPostgres: reconciling state from rule",
+			"intended_role", intended.String(), "postgres_primary", state.isPrimary)
+		if err := pm.stateManager.Mutate(ctx, func(s *servingStateMutation) {
+			s.SelfLeadership = obs
+			s.PostgresPrimary = state.isPrimary
+			if s.ServingStatus == clustermetadatapb.PoolerServingStatus_DRAINING {
```
dweitzman, line 700: A new behavior here: if multipooler is left in DRAINING serving state, it'll automatically transition back to SERVING

The thinking behind that is that DRAINING is a temporary state typically used before restarting postgres for a configuration change. If the action that was trying to restart postgres died, it's better to to resume serving state instead of leaving serving disabled indefinitely

go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -808,7 +808,13 @@ func (pm *MultiPoolerManager) setPrimaryLocked(ctx context.Context, req *consens
 	// stale-leader analyzer to keep firing forever. Promote has the same
 	// step on its replica branch for the same reason.
 	// A REPLICA pooler record carries no self leadership observation.
-	if err := pm.servingState.SetState(ctx, clustermetadatapb.PoolerType_REPLICA, nil, clustermetadatapb.PoolerServingStatus_SERVING); err != nil {
+	// Republish REPLICA (clear any stale PRIMARY self-leadership) so the
+	// stale-leader analyzer stops firing. Only the role changes here; serving
+	// status is owned by the lifecycle and the monitor's reconcileState, not by
+	// "here is your primary" bookkeeping.
+	if err := pm.stateManager.Mutate(ctx, func(s *servingStateMutation) {
```
dweitzman, line 675: Some places in the code like here are now able to do things like transition from leader to follower role without necessary re-enabling SERVING state

So for example, now if you put a pooler into DISABLED serving state and then it got the message that it's need to demote itself, it'll stay disabled

Previously if we tried to implement some kind of drain RPC, receiving a SetPrimary RPC here would have inadvertently flipped serving back on

proto/clustermetadata.proto
```diff
@@ -345,22 +345,27 @@ message PoolerLifecycle {
...
-  NOT_SERVING = 1;
-
-  // BACKUP is the status of a server when it is taking a backup. No queries
-  // can be served in BACKUP mode.
-  BACKUP = 2;
-
-  // RESTORE is the status a server uses when restoring a backup, at
-  // startup time.  No queries can be served in RESTORE mode.
-  RESTORE = 3;
+  // DISABLED means not serving and NOT brought back automatically: the pooler is
```
dweitzman, line 356: These are the modified serving states

go/services/multipooler/internal/manager/manager.go
```diff
@@ -1580,8 +1565,15 @@ func (pm *MultiPoolerManager) updateTopologyAfterPromotion(ctx context.Context,
...
-	// SetState is idempotent — if already at PRIMARY/SERVING it short-circuits.
-	if err := pm.servingState.SetState(ctx, clustermetadatapb.PoolerType_PRIMARY, pm.leaderObs(rule), clustermetadatapb.PoolerServingStatus_SERVING); err != nil {
+	// Promotion has already waited for postgres to leave recovery, so set the
+	// writable state in the same Mutate — this lets the heartbeat writer / LISTEN
+	// start immediately rather than waiting for the next monitor tick to observe
+	// it. Mutate is idempotent — if already at this state it short-circuits.
+	if err := pm.stateManager.Mutate(ctx, func(s *servingStateMutation) {
+		s.SelfLeadership = pm.leaderObs(rule)
+		s.PostgresPrimary = true
+		s.ServingStatus = clustermetadatapb.PoolerServingStatus_SERVING
```
dweitzman, line 1581: Note to self: from the failing coverage tests, I wonder if there's a regression related to this where gate is trying to send traffic to the new primary too early ("cannot execute INSERT in a read-only transaction")

Could be that gate is waiting on the pooler type but not the serving status?...

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -303,6 +303,14 @@ func (lb *loadBalancer) notifyLeaderServingFromSummary(summary *shardSummary, co
 	if health.Target.GetPoolerType() != clustermetadatapb.PoolerType_PRIMARY {
 		return
 	}
+	// Gate write-resume on writability, not just leadership. PoolerType tracks the
```
dweitzman: Slight change here: it's possible to be the consensus leader and not be writable. During `Promote()`, for example, a pooler might be aware it's the leader before it's able to accept writes

Now for draining the failover query buffer we check for writability, not just leadership

go/services/multipooler/internal/manager/state_manager.go
```diff
@@ -27,20 +27,27 @@ import (
...
+	// OnStateChange is called when the effective serving state changes; the
+	// component transitions to match it.
+	//
+	// isConsensusLeader reports whether this pooler is the consensus-elected
+	// leader (the shard's write target). postgresPrimary reports the physical
+	// recovery state (!pg_is_in_recovery): being the leader does NOT imply
+	// postgres is out of recovery, so components that need a writable backend
+	// (heartbeat writer, LISTEN/NOTIFY) must require both. servingStatus is the
+	// serving intent.
+	OnStateChange(ctx context.Context, isConsensusLeader, postgresPrimary bool, servingStatus clustermetadatapb.PoolerServingStatus) error
```
haritabh17, line 42: `postgresPrimary bool` --> `writable bool` to keep consistent with the HB proto??

go/services/multipooler/internal/manager/state_manager.go
```diff
@@ -52,10 +59,31 @@ type StateManager struct {
 	// is the exclusive caller of record.Mutate for Type/ServingStatus.
 	record *poolerRecord
 
+	// postgresPrimary is the physical recovery state (!pg_is_in_recovery), fed by
+	// the postgres monitor via Mutate's PostgresPrimary field. Unlike
+	// Type/ServingStatus it is local-only and never published to topology. Guarded by mu.
+	postgresPrimary bool
```
haritabh17, line 65: `postgresPrimary bool` --> `writable bool` to keep consistent with the HB proto?? Alt, we can even call it exactly what it is from pg-flag, i.e. `pg_is_not_in_recovery `

Basically having uniform in all places, including the proto??

haritabh17 (approved):
🚀

proto/multipoolerservice.proto
```diff
@@ -516,6 +516,12 @@ message StreamPoolerHealthResponse {
   // replication_lag_ns is the current replication lag in nanoseconds,
   // measured via heartbeat timestamps. Zero on a leader or when unknown.
   int64 replication_lag_ns = 6;
+
+  // writable reports whether this pooler can accept writes — i.e. postgres is
+  // out of recovery (!pg_is_in_recovery). A consensus leader mid-promotion is
+  // SERVING (can answer reads) but not yet writable, so clients must gate write
+  // traffic on this rather than on pooler_type/serving_status alone.
+  bool writable = 7;
```
Verolop, line 521: How does this behave during a rolling upgrade when the gateway has this field but the pooler does not? Since this is a proto3 scalar, an omitted value decodes as `false`; does the gateway have another way to tell the difference between old pooler with no writability signal and a pooler reported not writable before using it to hold buffer drain?

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -303,6 +303,14 @@ func (lb *loadBalancer) notifyLeaderServingFromSummary(summary *shardSummary, co
...
+	// consensus term, so a leader can report PoolerType==PRIMARY and SERVING (it can
+	// answer reads) before it has finished promoting, while postgres is still in
+	// recovery. Draining the failover buffer then would route buffered writes to a
+	// read-only node ("cannot ... during recovery" / read-only transaction).
+	if !health.isWritable() {
```
Verolop, line 325: How does this interact with `getConnection(PRIMARY)` below? Since this gates `StopBuffering` on writability, I’m wondering what handles PRIMARY requests selected during the same interval, especially COPY and reserved-connection flows

go/services/multipooler/internal/manager/manager.go
```diff
@@ -516,11 +515,14 @@ func (pm *MultiPoolerManager) openLocked(ctx context.Context, targetServingStatu
...
-	// streamer) and Mutates the record. The publisher (if running, started
-	// by StartTopoRegistration) picks it up and writes to etcd.
+	// Mutate notifies all components (query service, heartbeat, health streamer)
+	// and Mutates the record. The publisher (if running, started by
+	// StartTopoRegistration) picks it up and writes to etcd. Only the serving
+	// status changes here; the role is left as the record already holds it.
 	go pm.runHealthHeartbeat(pm.ctx, timeouts.DefaultHealthHeartbeatInterval)
-	if err := pm.servingState.SetState(ctx, pm.record.Type(), pm.record.SelfLeadership(), targetServingStatus); err != nil {
+	if err := pm.stateManager.Mutate(ctx, func(s *servingStateMutation) {
+		s.ServingStatus = targetServingStatus
```
Verolop, line 531: Is the short-lived `SERVING` + stale/false `writable` state intentional here?

proto/clustermetadata.proto
```diff
@@ -345,22 +345,27 @@ message PoolerLifecycle {
...
+  // stopping/restarting, paused, or has been deliberately drained (operator or
+  // config). The postgres monitor does not re-enable serving from this state; an
+  // explicit action (Open/resume, promotion) does.
+  DISABLED = 1;
```
Verolop, line 360: What is the compatibility contract for these enum names? Numeric `1` remains wire-compatible with the old `NOT_SERVING`, but generated-code callers and JSON/prototext users would observe the rename/removal of `NOT_SERVING`, `BACKUP`, and `RESTORE`, unless those are those considered internal-only at this point (?)

Verolop (commented):
the split between consensus leadership and physical Postgres writability makes sense to me. I left a few questions around the upgrade and routing edges :D

go/services/multigateway/poolergateway/load_balancer.go
dweitzman: `getConnection(PRIMARY)` has no opinion on buffering, so it'll return a connection to the highest known primary regardless of whether the primary is health or writable

For read queries that's helpful since the newly chosen primary is available to serve reads even before it's available to serve writes.

proto/clustermetadata.proto
dweitzman: I've been assuming proto3 compatibility is not a concern yet

proto/multipoolerservice.proto
dweitzman: I'm assuming for now that we don't support rolling upgrades yet and can assume shards are being fully stopped and upgraded atomically

go/services/multipooler/internal/manager/manager.go
dweitzman: Great question

Ideally we'd check the recovery mode status here, but I'm on the fence about whether that's a good idea because we'd probably need to put a wait until the db connections are ready

If this open was from restarting postgres intentionally, whatever caller restarted it should know the intended recovery mode state and set it

If this open was from multipooler crashing and restarting, it might take a run of the postgres monitor for the writable status to get set. I'm not super concern about the crash and restart case, though. It shouldn't happen frequently and it's not that harmful to have a brief blip where `writable` is off

For now I'm inclined to leave this code alone and let it self-heal after a pooler crash, but I've also modified the code above to run the first postgres monitor loop iteration immediately instead of waiting for an initial delay

go/services/multipooler/internal/manager/state_manager.go
dweitzman: Replied to this above

go/services/multipooler/internal/manager/state_manager.go
dweitzman: Re: this and the `pg_is_not_in_recovery` idea below, I'm not feeling decisive about what's the right thing to do

In the health proto, the signal is meant to indicate "are user queries allowed?" which is potentially not the same as not being in recovery mode. During promotion there's a brief period of time where we're out of recovery mode but not yet able to accept user queries, and you could imagine some extension or future upgrade of Postgres finally implementing a "read only" feature that doesn't require being in recovery mode

For now I'm wondering if it makes sense so say that health deals in the idea of being writable and internal state monitoring deals in the idea of being a primary or not?

I like the sentiment behind these suggestions, but not feeling confident about any particular naming approach that feels concise, consistent, and also honest

go/services/multipooler/internal/manager/manager.go
Verolop: got it, that makes sense. I saw that `WithFastStart()` schedules the monitor immediately but asynchronously. Is the goal to remove the full monitor interval delay rather than to make the first reconciliation happen before the `SERVING` publish?

dweitzman: Yes, the `WithFastStart()` is to make up for not doing an immediate check whether postgres is in recovery mode here

Previously after starting the postgres monitor there would be a 5s delay before the first round of self-healing, so I figured doing the first round of self-healing almost immediately could be helpful to react quickly if something changed about postgres state (whether it's in recovery mode, etc.) between closing and re-opening connections to it.

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -626,27 +658,50 @@ func (pm *MultiPoolerManager) takeRemedialAction(ctx context.Context, action rem
...
 		intended, obs := deriveTypeAndObs(pm.latestRule(), pm.serviceID)
-		pm.logger.InfoContext(ctx, "MonitorPostgres: applying consensus role from rule",
-			"intended_role", intended.String())
-		if err := pm.servingState.SetState(ctx, intended, obs, clustermetadatapb.PoolerServingStatus_SERVING); err != nil {
-			pm.logger.ErrorContext(ctx, "MonitorPostgres: failed to apply role from rule", "error", err)
+		pm.logger.InfoContext(ctx, "MonitorPostgres: reconciling state from rule",
+			"intended_role", intended.String(), "postgres_primary", state.isPrimary)
+		if err := pm.stateManager.Mutate(ctx, func(s *servingStateMutation) {
+			s.SelfLeadership = obs
+			// TODO: There's a subtle but important bug here. We're currently conflating
```
dweitzman, line 690: There's a significant TODO here. To keep PR size manageable, I'm going to address that in a follow-up

PR #1185, opened 2026-06-24
fix(multigateway): preserve SQL EXECUTE argument expressions

haritabh17:
## Summary

- Keep SQL PREPARE gateway/consolidator-managed so prepared-statement consolidation is preserved.
- Rewrite top-level SQL EXECUTE to the canonical prepared-statement name while preserving argument expressions for PostgreSQL to evaluate.
- Add pgregress patches documenting accepted pg_prepared_statements and plan-cache observability differences from consolidation.

## Problem

SQL EXECUTE previously converted arguments into portal Bind values at the gateway. That failed for valid PostgreSQL expressions such as casts and arrays in EXECUTE arguments.

## Solution

Route SQL EXECUTE as SQL after rewriting only the prepared-statement name to the canonical consolidated name and attaching the prepared-statement metadata so the multipooler ensures the backend statement exists. This keeps consolidation while letting PostgreSQL handle EXECUTE argument expression semantics.

## Validation

- go test -short ./go/services/multigateway/... ./go/services/multipooler/internal/executor/... ./go/common/preparedstatement/...
- make build
- RUN_VARS="RUN_PGREGRESS=1" PGREGRESS_PATCH_MODE=verify docker/pgregress-generate.sh

### DIFF
--- a/go/services/multigateway/engine/prepared_statement_primitive.go
+++ b/go/services/multigateway/engine/prepared_statement_primitive.go
@@ -17,7 +17,6 @@ package engine
 import (
 	"context"
 	"fmt"
-	"strconv"
 
 	"github.com/multigres/multigres/go/common/constants"
 	"github.com/multigres/multigres/go/common/mterrors"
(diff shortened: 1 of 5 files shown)


shedges (approved):
LGTM

PR #1187, opened 2026-06-24
feat(multipooler): wait on pg_rewind until the leader is rewind-ready

dweitzman:
A diverged follower that runs pg_rewind against a freshly promoted leader copies the leader's control-file checkpoint timeline into its own minRecoveryPoint. PostgreSQL's post-promotion checkpoint is lazy, so the leader can still advertise the old timeline at that moment; the follower then stamps a stale timeline and FATALs on startup ("requested timeline N does not contain minimum recovery point ... on timeline M"), wedging until a full re-bootstrap.

Make a node advertise rewind-readiness (it has checkpointed onto its current timeline) and defer a diverged follower's rewind until the leader is ready:

- A freshly promoted primary runs an async CHECKPOINT; the postgres monitor marks ConsensusStatus.ReplicationPrimary.rewind_ready once pg_control_checkpoint().timeline_id matches the running timeline (the async checkpoint also marks it directly, term-guarded). The checkpoint runs off the failover path and needs no replicas, since CHECKPOINT does not wait on synchronous replication.
- multiorch relays the leader's rewind_ready into SetPrimary. setPrimaryLocked and the monitor's stale-primary demote defer the restart-as-standby until it is true, leaving the node up as a queryable deposed primary (no downtime); fix_replication defers its pg_rewind likewise. We never restart-without-rewind, which would FATAL on the node's own un-replicated WAL. The gating is the correctness fix; the explicit checkpoint only makes readiness prompt.

- Unify SetPrimaryRequest to carry a clustermetadata.ReplicationPrimary (rule + primary + rewind_ready) instead of duplicating the rule and leader.

- Add the multipooler.rewind.checkpoint_wait.duration histogram to measure how long a rewind waited for the leader's checkpoint (~0 when already ready), emitted once per leader change.


rafael (approved):
LGTM

PR #1188, opened 2026-06-24
docs(ha): get some documentation in place

dweitzman:
This isn't meant to be great documentation, but something is probably better than nothing


rafael (approved):
LGTM

PR #1189, opened 2026-06-24
fix(query): preserve reserved state and notices

haritabh17:
_Last updated: 2026-07-06 (rebased onto upstream/main)_

## Summary
- preserve reserved-connection state on PostgreSQL-level statement/portal errors while unwinding statement-local pins
- preserve backend notices through commit/rollback, unary results, streaming, and COPY finalize paths
- track dynamic set_config state only after backend success, with prevalidated gateway-managed variable actions
- keep pg_regress isolation lock-wait detection on client-visible application_name while retaining explicit VPID stamping compatibility

## Review feedback addressed
- unwind newly-added statement-local reservation reasons when PostgreSQL rejects a statement (no leaked pins)
- narrow dynamic set_config to the pg_dump-safe shape (`pg_settings.name` with static value/is_local), so the resolve phase is side-effect-free and mistyped arguments are rejected

## Validation
- `go build ./go/...` and `make proto` (no generated-file drift after rebase)
- `go test -short ./go/services/multigateway/... ./go/services/multipooler/...`

### DIFF
--- a/go/services/multigateway/engine/resolve_set_config.go
+++ b/go/services/multigateway/engine/resolve_set_config.go
@@ -38,7 +38,7 @@ import (
 // A session-scoped (is_local=false) set_config must be tracked in
 // SessionSettings so it survives pool rotation, which needs the concrete GUC
 // name — but here the name is a column reference resolved per row. Rather than
-// reject it (or pin the connection), this primitive runs in three steps:
+// reject it (or pin the connection), this primitive runs in four steps:
 //
 //  1. Resolve: execute the "unroll" projection — the original SELECT with each
 //     set_config(a, b, c) target replaced by its three arguments a, b, c — once,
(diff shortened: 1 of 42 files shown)


go/services/multipooler/internal/executor/executor.go
```diff
@@ -633,6 +690,15 @@ func (e *Executor) streamExecuteOnReservedConn(
 		return e.buildReservedStateFromAPI(rc), wrapQueryError(err)
```
shedges, line 687: `streamExecuteOnReservedConnWithPostState` adds nonBeginReasons (let's say for temp table) before the query at line 653 but doesn't remove them on failure. Assume we perform `CREATE TEMP TABLE`on already pinned connection and it fails then this path will leaks a stale reservation reason on connection?

I could see we are explicitly calling `RemoveReservationReason` inside `reserveAndStreamExecute` i.e.  new reservation line 556

go/services/multigateway/engine/resolve_set_config.go
```diff
@@ -166,6 +170,15 @@ func (s *ResolveTrackSetConfig) execute(
 		return callback(ctx, s.emptyResult())
 	}
 
+	// Prepare tracking before backend apply so gateway-managed validation errors
+	// (for example an unparsable statement_timeout) are returned before any
+	// client-visible result is streamed. The returned closures are intentionally
+	// not executed until after PostgreSQL accepts the synthesized apply query.
+	trackActions, err := s.prepareTrackActions(conn, state, rows)
```
shedges, line 175: Resolve and Apply are 2 separate backend commands? If yes, then in auto commit they run as two implicit transactions.  so this is not atomic the way PostgreSQL's single `set_config(fn())` statement is. 

If the value expression/function has side effects (`INSERT`, `nextval`, NOTIFY), those commit during resolve even if the apply later fails. PostgreSQL would roll the whole statement back.

go/services/multigateway/engine/resolve_set_config.go
shedges, line 160: Here value will be evaluated as ordinary and not as `set_config`text argument. Then it applied as text literal. Is this correct understanding?
If yes, then it skips `set_config's` own argument type check. Whereas PostgreSQL will reject it.
e.g.  
`set_config('x', now(), false) without ::text.`

 Native PostgreSQL errors with:
```
  function set_config(unknown, timestamp with time zone, boolean) does not exist
```

go/services/multipooler/internal/executor/executor.go
haritabh17: Good point - updated `streamExecuteOnReservedConnWithPostState` to track which statement-local reservation reasons were newly added before the query and unwind only those if PG rejects the statement. 

Pre-existing reasons are preserved, ReasonTransaction still survives PG statement errors, and connection-level errors still release/taint the backend.

go/services/multigateway/engine/resolve_set_config.go
haritabh17: Ack. Fixed this together with the atomicity issue by narrowing the dynamic path and adding stricter argument validation. The dynamic path now requires text-like name/value args and a bool `is_local`, and rejects expressions like `set_config('x', now(), false)` rather than stringifying the resolved value and applying it as text.

go/services/multigateway/engine/resolve_set_config.go
haritabh17: Agree. Narrowed this path to the pg_dump-safe shape we need: dynamic `pg_settings.name` with static value/is_local args. Arbitrary dynamic expressions/functions are now rejected, so side-effectful expressions can’t run during resolve before apply.

shedges (approved):
LGTM 🚀

PR #1190, opened 2026-06-24
feat(multipooler): keep retrying pg_rewind until the standby rejoins

dweitzman:
A diverged standby that needs pg_rewind could get permanently stuck. Make the
rewind keep retrying until the node actually rejoins as a standby, closing two
gaps:

**1) Don't get stuck on a crash.**
pgctld `Start` gains `allow_crash_recovery`: a standby that can't start (e.g. an
early pg_rewind stamped `minRecoveryPoint` onto the wrong timeline and postgres
FATALs on startup) is crash-recovered first. `runCrashRecovery` is now
`standby.signal`-aware, since `postgres --single` refuses to run with a
`standby.signal` present. When crash recovery runs and a *different* node is the
known leader, the node marks `suspectedDivergence` so the next standby
transition routes through pg_rewind.

**2) Don't forget to retry just because the rewind path is idempotent.**
- `suspectedDivergence` is cleared only after postgres restarts and is verified
  in recovery mode, not as soon as pg_rewind returns. Clearing early left an
  un-startable (wrong-timeline) node with the flag already false, so the next
  attempt skipped the rewind and merely retried the doomed start.
- A new monitor remedial action re-runs the rewind against the recorded leader
  when a standby cannot come up and divergence is suspected, rate-limited with
  exponential backoff so it does not thrash disruptive restarts.

The rewind bookkeeping (target selection + backoff + `suspectedDivergence`)
lives on the `ConsensusManager`; the monitor decides, then composes those
queries. The backoff is incident-scoped — cleared on any healthy monitor tick
(postgres running, no rewind pending) — so an unrelated future divergence starts
with a fresh window.

### DIFF
--- a/go/cmd/pgctld/command/crash_recovery.go
+++ b/go/cmd/pgctld/command/crash_recovery.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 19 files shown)


go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -28,6 +29,11 @@ import (
 	"github.com/multigres/multigres/go/tools/retry"
 )
 
+// standbySignalFilename is the marker file whose presence puts PostgreSQL into
+// standby mode. It must be absent for single-user mode (postgres --single),
+// which is why crash recovery removes and recreates it.
+const standbySignalFilename = "standby.signal"
+
```
mkindahl: Would probably be good to store all these constants in a single place, e.g., `constants/postgres.go`.

go/cmd/pgctld/command/crash_recovery_test.go
mkindahl, line 1: You might want to add a test that failure to remove the signal file does not run the recovery.

mkindahl (approved):
Would suggest adding an additional test, but otherwise looks fine.

go/cmd/pgctld/command/crash_recovery.go
dweitzman: Done

go/cmd/pgctld/command/crash_recovery_test.go
dweitzman: Done

PR #1191, opened 2026-06-25
fix(prepared-statements): materialize SQL EXECUTE via pooler

haritabh17:
## Summary

- Add `ExecuteSqlPreparedStatement` metadata for SQL-level `EXECUTE` wrappers so multigateway can pass prefix/suffix templates instead of gateway canonical backend names.
- Materialize top-level and wrapped `EXECUTE` statements in multipooler using pooler-level prepared statement consolidation (`ppstmt*`).
- Update planner, routing, scatter connection, unit tests, and pgregress expectations for backend-visible prepared statements.

## Problem

SQL-level `EXECUTE` wrappers had to reference gateway-managed canonical names directly, which bypassed the multipooler's consolidator and could expose unstable names across gateway/pooler boundaries.

## Solution

The gateway now sends the visible SQL shell plus a structured `EXECUTE` template. The multipooler ensures the underlying query via its prepared-statement consolidator, substitutes the pooler canonical name into the template, and executes the materialized SQL on the selected backend connection.


shedges (approved):
LGTM 🚀

PR #1192, opened 2026-06-25
fix(parser): match syntax error format to PostgreSQL

shedges:
## What

The parser emitted `parse error at position N: syntax error`, which does not match PostgreSQL's wire format, so clients and ORMs that key off the SQLSTATE and error position received an unfamiliar message with no source cursor.

This shapes parser and lexer errors like PostgreSQL's `scanner_yyerror`:

- `syntax error at or near "<token>"`, or `syntax error at end of input`
- scanner errors (trailing junk, invalid integer, zero-length delimited identifier) carry the same `at or near` / `at end of input` suffix
- the error position is the start of the offending lexeme, reported as a 1-based character offset in the `ErrorResponse` `P` field, so psql renders the `LINE: ... ^` cursor

The standalone parser stays free of the `mterrors` vocabulary: it returns a `ParseSyntaxError` carrying the message and cursor position, and the serving boundary maps it to a `42601` diagnostic.

## Testing

- Unit-level checks confirm both the goyacc path (`syntax error at or near "SELCT"`, `... at end of input`) and the lexer path (`trailing junk after numeric literal at or near "123abc"`, `invalid binary integer at or near "0b"`, `zero-length delimited identifier at or near """"`) match PostgreSQL exactly, including multi-byte cursor positions.
- Parser corpus testdata regenerated; `parser`, `mterrors`, `preparedstatement`, and `multigateway/handler` packages pass; full build is clean.
- Verified against real PostgreSQL via the `pgregresstest` harness (generate mode): the syntax/lexer divergence hunks across the affected regression tests collapse to matching context, i.e. output is now identical to PostgreSQL.

## pgregress patches

- Removed patches that existed only to translate the old format where output now matches PostgreSQL (`collate`, `copyselect`, `vacuum`), and trimmed the syntax hunks from `create_table`, `event_trigger`, `sqljson`, `sqljson_jsontable`.
- The remaining per-test patches still carry the old format and need regeneration once an unrelated `SET SESSION AUTHORIZATION` divergence (separate from this change) is resolved. Regenerating them now would fold that unrelated divergence into this PR.


GuptaManan100 (approved):
Love it!

PR #1196, opened 2026-06-26
fix(pgregress): refresh stale regression patches

haritabh17:
## Summary
- refresh stale pg17 pgregress patches after parser error output now matches PostgreSQL
- remove obsolete syntax-error patches for tests that now match upstream output
- keep only current accepted divergences (unlogged warnings, FDW/LOAD blocking, minor cursor/token diffs)

## Validation
- rebased on `upstream/main`
- `git diff --check`
- reconstructed and verified the 12 reported CI regressions against the downloaded pgregress artifact from run 28217084225

## Follow-up
- Requesting the PostgreSQL Compatibility Tests workflow via label to run the full regress suite on this PR.


shedges (approved):
LGTM 🚀

Issue #1128, opened 2026-06-09
multigateway: COPY FROM STDIN errors lose PostgreSQL COPY context


Verolop:
## Summary

Errors raised during `COPY FROM STDIN` are wrapped by `multigateway` and lose PostgreSQL's COPY context, including line and column details.

The data load fails correctly, but the error returned to the client is less useful than PostgreSQL's native error.

## Reproduction

Start a local Multigres cluster, run an invalid `COPY FROM STDIN` through the local `multigateway` PostgreSQL endpoint:

```bash
export PGURI='postgresql://postgres:postgres@localhost:15432/postgres?sslmode=disable'

psql "$PGURI" -X -v ON_ERROR_STOP=0 <<'SQL'
DROP TABLE IF EXISTS copy_from_error_repro;
CREATE TABLE copy_from_error_repro (id int);

COPY copy_from_error_repro FROM STDIN;
not_an_int
\.

DROP TABLE copy_from_error_repro;
SQL
```

### Expected behaviour

PostgreSQL reports the original error with COPY context:

```text
ERROR: invalid input syntax for type integer: "not_an_int"
CONTEXT: COPY copy_from_error_repro, line 1, column id: "not_an_int"
```

## Actual behaviour

Through `multigateway`, the error is wrapped and the COPY context is lost:

```text
ERROR: failed to finalize COPY: COPY finalization failed: COPY operation failed: ERROR: invalid input syntax for type integer: "not_an_int"
```

## Impact

For bulk loads, the `CONTEXT` line is important because it tells users which COPY line and column failed. Without it, debugging bad input files is significantly harder.

### Additional examples

The same wrapping/context-loss pattern appears for other `COPY FROM STDIN` failures, including missing data, extra data, constraint violations, and invalid COPY options.

pratikscfr:
Hi @Verolop , can I take this up?

PR #1201, opened 2026-06-29
test(copy): cover COPY FROM STDIN error context preservation

shedges:
## Summary

Adds `TestErrorFormat_CopyFromStdinContext`, a regression test locking in that errors raised during `COPY FROM STDIN` preserve PostgreSQL's COPY context (the diagnostic `Where` field, e.g. `COPY t, line 1, column id: ...`) end to end through multigateway.

This is the regression guard that was missing for MUL-621 / #1128. The underlying fix already shipped (#1022, on top of #963 and #1019, which stopped wrapping the PG error into a flat Go string and round-trip a structured `PgDiagnostic` over the COPY bidi stream). No test asserted on the COPY `CONTEXT` line, so a refactor that re-wrapped COPY errors would not have been caught.

## What the test does

Modeled on the existing `TestErrorFormat_ConstraintViolation` harness: each case runs against **both direct PostgreSQL and multigateway** and asserts the structured `pgconn.PgError`. It uses pgx's low-level `PgConn().CopyFrom` so the text-format COPY parse errors reproduce exactly (the high-level `CopyFrom` converts client-side and would miss them).

| Subcase | SQLSTATE | Asserts |
| --- | --- | --- |
| `type_error` | `22P02` | `Where` names table + `line 1` + `column id` |
| `missing_data` | `22P04` | `Where` names table + `line 1` |
| `extra_data` | `22P04` | `Where` names table + `line 1` |
| `not_null_violation` | `23502` | `Detail` has the failing row **and** `Where` carries the COPY context |

## Verification

The original issue repro through multigateway now matches native PostgreSQL byte for byte:

```
ERROR:  invalid input syntax for type integer: "not_an_int"
CONTEXT:  COPY copy_from_error_repro, line 1, column id: "not_an_int"
```

The `Where` field is identical between the two targets (modulo the unique table name) across all four cases. Test result:

```
--- PASS: TestErrorFormat_CopyFromStdinContext (20.77s)   [10/10 subtests]
ok  github.com/multigres/multigres/go/test/endtoend/queryserving
```

Also updates the executor comment that justified not wrapping the PG error so it points at this test instead of a fixture that did not exist.

Refs MUL-621, #1128


GuptaManan100 (approved):
Love it!

PR #1203, opened 2026-06-29
feat(docker-compose): support custom PostgreSQL base image

rafael:
## Summary

- Make the PostgreSQL base image the all-in-one cluster builds on overridable via `MULTIGRES_POSTGRES_IMAGE` (build arg `POSTGRES_IMAGE`, default `postgres:17.7`), so the compose cluster can run on a newer PostgreSQL or a prebuilt image without editing the Dockerfile.
- Add `MULTIGRES_PROVISION_PG_PACKAGES` (build arg `PROVISION_PG_PACKAGES`, default `true`) to skip the Debian/apt provisioning when the base image already bundles pgBackRest, pgvector, and procps — required for non-Debian bases where apt won't work.
- Honor the base image's `POSTGRES_USER` end-to-end so an image that ships post-`initdb` SQL for a custom superuser bootstraps correctly.
- The official `postgres` path is behavior-preserving: defaults reproduce the previous build exactly.

## Description

Previously `Dockerfile.cluster` hardcoded `FROM postgres:17.7` and always installed pgBackRest/pgvector/procps from the PostgreSQL APT repo, so the all-in-one compose cluster could only run on the official Debian base. This change parameterizes that. The mechanism is fully generic — any base image that bundles the required dependencies works — and this PR was validated end-to-end against the `supabase/postgres` `*-multigres` image as the motivating example.

**Base image + provisioning.** `POSTGRES_IMAGE` is now a global build ARG used by the final stage's `FROM`, and the apt provisioning step is wrapped so it only runs when `PROVISION_PG_PACKAGES=true`. When skipped, the build asserts the base actually provides the skipped dependencies (`pgbackrest`, `pgrep`, pgvector) so a missing one fails the build with a clear message instead of surfacing as a confusing runtime failure mid-bootstrap. The pgvector apt package now follows the base image's `PG_MAJOR` instead of hardcoding `17`. Build steps run as `USER root` since some base images switch to a non-root user, which can't install packages or write to `/multigres`.

**Superuser.** The cluster's PostgreSQL superuser now follows the base image's `POSTGRES_USER` (default `postgres`). The entrypoint flows it into the generated config's `pg-user` (driving pgctld's `initdb` and the cluster-start probe), the multipooler inherits it, and both healthchecks connect as it — so `initdb`, any post-`initdb` init SQL, the pooler, and the healthcheck all agree on one superuser. With the default unset, everything resolves to `postgres` as before.

**Docs.** `docker/README.md` gains a short "Custom base image" section covering the two env vars and the custom-superuser login.

### Official base image (default)

```bash
docker compose up --build -d
# wait for healthy, then connect as the default `postgres` superuser
PGPASSWORD=postgres psql -h 127.0.0.1 -p 15432 -U postgres -d postgres -c 'select version();'
docker compose down
```

### Custom base image (validated with `supabase/postgres`, but generic)

```bash
MULTIGRES_POSTGRES_IMAGE=ghcr.io/supabase/postgres:17.6.1.129-multigres \
MULTIGRES_PROVISION_PG_PACKAGES=false \
docker compose up --build -d
# the superuser follows the image's POSTGRES_USER (here: supabase_admin)
PGPASSWORD=postgres psql -h 127.0.0.1 -p 15432 -U supabase_admin -d postgres -c 'select current_user;'
```

Check the extensions the image ships (installed and available):

```bash
PGPASSWORD=postgres psql -h 127.0.0.1 -p 15432 -U supabase_admin -d postgres \
  -c 'select extname, extversion from pg_extension order by extname;' \
  -c 'select count(*) as available from pg_available_extensions;'
```

Confirm the image's bundled migrations ran during bootstrap — the supabase `initial-schema.sql` creates its roles, schemas, and publications:

```bash
PGPASSWORD=postgres psql -h 127.0.0.1 -p 15432 -U supabase_admin -d postgres \
  -c "select rolname from pg_roles where rolname like 'supabase%' or rolname in ('anon','authenticated','authenticator','service_role') order by 1;" \
  -c '\dn' \
  -c 'select pubname from pg_publication;'
```

Exercise a loaded extension end-to-end through the gateway:

```bash
PGPASSWORD=postgres psql -h 127.0.0.1 -p 15432 -U supabase_admin -d postgres <<'SQL'
CREATE EXTENSION IF NOT EXISTS vector;
CREATE TABLE items (id bigserial PRIMARY KEY, embedding vector(3));
INSERT INTO items (embedding) VALUES ('[1,0,0]'), ('[0,1,0]'), ('[0.9,0.1,0]');
SELECT id, embedding FROM items ORDER BY embedding <-> '[1,0,0]' LIMIT 3;
SQL
```

Also verified the skip-path assertion fails the build on the official base (missing pgbackrest/pgvector) and passes on a base that bundles them.

### DIFF
--- a/docker/cluster-entrypoint.sh
+++ b/docker/cluster-entrypoint.sh
@@ -22,6 +22,14 @@ set -euo pipefail
 
 CONFIG_PATH="${MULTIGRES_CONFIG_PATH:-/multigres/cluster}"
 
+# PostgreSQL superuser for the cluster. Defaults to `postgres`. A base image can
+# customize it by exporting POSTGRES_USER (e.g. an image that ships post-initdb
+# SQL assuming a specific superuser role); we honor it end-to-end: it flows into
+# the generated config's pg-user (driving pgctld's initdb and the cluster-start
+# probe) and is inherited by the multipooler's admin connection, so initdb, any
+# init SQL, the pooler, and the healthcheck all agree on one superuser.
+PG_SUPERUSER="${POSTGRES_USER:-postgres}"
+
 # Number of cells to run. Each cell is a full stack: one PostgreSQL + pgctld,
 # one multipooler, one multiorch, and one multigateway. The local provisioner
 # bootstraps the shard with an AtLeastN(2) durability policy, so the cluster
@@ -99,6 +107,16 @@ set_gateway_ports() {
   done
 }
 
+# set_pg_user rewrites every cell's pgctld pg-user (a sibling of pg-port, same
+# indentation) to ${PG_SUPERUSER}. `cluster init` always generates `postgres`;
+# overriding it makes pgctld's initdb and the cluster-start probe use the base
+# image's superuser so they agree with the multipooler (which inherits
+# POSTGRES_USER) and any post-initdb SQL the image ships.
+set_pg_user() {
+  local file="$1" user="$2"
+  sed -i "s/^\(                pg-user: \).*\$/\1${user}/" "${file}"
+}
+
 shutdown() {
   echo "==> Received shutdown signal, stopping Multigres cluster..."
   multigres cluster stop --config-path "${CONFIG_PATH}" || true
@@ -143,6 +161,9 @@ if [ ! -f "${CONFIG_PATH}/multigres.yaml" ]; then
   multigres cluster init --config-path "${CONFIG_PATH}"
   trim_cells "${CONFIG_PATH}/multigres.yaml" "${NUM_CELLS}"
   set_gateway_ports "${CONFIG_PATH}/multigres.yaml" "${GATEWAY_PG_PORT}" "${NUM_CELLS}"
+  if [ "${PG_SUPERUSER}" != "postgres" ]; then
+    set_pg_user "${CONFIG_PATH}/multigres.yaml" "${PG_SUPERUSER}"
+  fi
 fi
 
 # Assemble extra PostgreSQL config from the max_connections knob and any raw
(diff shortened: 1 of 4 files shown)


docker/cluster-entrypoint.sh
```diff
@@ -99,6 +107,16 @@ set_gateway_ports() {
...
 }
 
+# set_pg_user rewrites every cell's pgctld pg-user (a sibling of pg-port, same
+# indentation) to ${PG_SUPERUSER}. `cluster init` always generates `postgres`;
+# overriding it makes pgctld's initdb and the cluster-start probe use the base
+# image's superuser so they agree with the multipooler (which inherits
+# POSTGRES_USER) and any post-initdb SQL the image ships.
+set_pg_user() {
+  local file="$1" user="$2"
+  sed -i "s/^\(                pg-user: \).*\$/\1${user}/" "${file}"
```
cuongdo: Matching on a specific number of leading spaces seem fragile. Would it be better to match against `[[:space:]]+` instead?

cuongdo (commented):
Looks reasonable. I left one comment. Is there a test that sets up a Docker cluster with the `supabase/postgres` image?

rafael:
> Looks reasonable. I left one comment. Is there a test that sets up a Docker cluster with the `supabase/postgres` image?

I did not want to add a test that is specific to the supabase image in the multigres repo. I did some manual verification. Now if this works, the plan is that supabase will run integration tests using this method, so I think that should give us coverage in case something regress.

GuptaManan100 (approved):
LGTM!

PR #1204, opened 2026-06-29
refactor(multipooler): derive writable from recovery + committed leadership

dweitzman:
## What

Make write-safety a *derived* value rather than a single physical-recovery bit,
and gate write admission on it.

A pooler is **writable** iff: postgres is out of recovery **AND** this pooler is
the highest **non-revoked committed** leader. This is delivered to components on
a new `servingstate.State`, which carries two deliberately distinct leadership
notions:

- **`IsHighestKnownLeader`** — highest-*known* rule, revocation-ignorant. Drives
  routing (the `PoolerType` label, gateway staleness) and can still route reads
  to a deposed-but-running leader.
- **`Writable`** — committed, non-revoked leadership + out of recovery. Gates
  writes.

The query server then splits admission by mode:

- **WRITABLE** queries require `Writable`.
- **CONSISTENT** reads keep gating on `IsHighestKnownLeader`, so a
  deposed-but-running leader can still serve them.

## Why

Gating on the *committed* (not highest-known) rule and excluding revoked terms
closes two windows where a node would otherwise advertise/accept writes
incorrectly:

1. The gap between `pg_promote()` and the new rule actually committing — the node
   is physically primary but not yet the committed leader.
2. A deposed-but-not-yet-demoted primary still running at a revoked term.

When a committed leader is briefly not writable it returns `MTF01`, so the
gateway buffers and holds the write (via its own writable-gating) until the
pooler publishes writable — rather than admitting a write postgres would reject.

## How it's derived

The `StateManager` stores only `inRecovery` (the async postgres observation it
can't compute itself) and recomputes `writable` live via an injected
`committedLeader` query — so a revocation or a new committed-rule observation
flips writability with no stored copy to keep in sync. Set-sites poke the
recovery fact, never `writable` directly. Terminology is made explicit:
`highestKnownRule` vs `highestCommittedRule`, with
`isHighestNonRevokedCommittedLeader` as the write-safety input.

## Behavior change

- WRITABLE admission now follows committed, non-revoked leadership + recovery
  state instead of highest-known leadership. CONSISTENT/INCONSISTENT routing is
  unchanged.
- Heartbeat/LISTEN and the health publisher react to `Writable`; health labels
  still come from `IsHighestKnownLeader`.

### DIFF
--- a/go/services/multipooler/internal/poolerserver/pooler.go
+++ b/go/services/multipooler/internal/poolerserver/pooler.go
@@ -54,12 +54,18 @@ type QueryPoolerServer struct {
 	shard      string
 
 	mu sync.Mutex
-	// isHighestKnownLeader reports whether this pooler is the consensus-elected
-	// leader (the shard's write target). Set from OnStateChange; replaces the
-	// PoolerType label, which the gateway no longer routes on.
+	// isHighestKnownLeader reports highest-known leadership (the routing signal).
+	// Set from OnStateChange. CONSISTENT reads gate on this — a deposed-but-
+	// running leader can still serve read-consistent queries.
 	isHighestKnownLeader bool
-	servingStatus        clustermetadatapb.PoolerServingStatus
-	healthProvider       HealthProvider
+	// writable is true iff this pooler can accept a write: postgres is out of
+	// recovery AND the last-written committed WAL rule names this pooler as the
+	// (non-revoked) leader. WRITABLE requests gate on this, so a write is never
+	// admitted on the strength of merely highest-known leadership (a mid-promote
+	// or deposed-but-running node).
+	writable       bool
+	servingStatus  clustermetadatapb.PoolerServingStatus
+	healthProvider HealthProvider
 
 	// pubsubListener is the shared LISTEN/NOTIFY listener, set by MultiPoolerManager.
 	pubsubListener *pubsub.Listener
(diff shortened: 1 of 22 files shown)


go/services/multipooler/internal/poolerserver/pooler.go
```diff
@@ -347,15 +358,25 @@ func (s *QueryPoolerServer) checkTargetLocked(target *query.Target, existingRese
...
+	// the gateway has stale topology (a demotion happened). Return MTF01 to trigger
+	// buffering and re-discovery. The required leadership differs by mode:
+	//
+	//   - WRITABLE gates on writability: postgres out of recovery AND we are the
+	//     last-written committed (non-revoked) leader. Highest-known leadership is
+	//     NOT sufficient — a mid-promote or deposed-but-running node could otherwise
+	//     admit a write before consensus committed.
+	//   - CONSISTENT (a read) gates on highest-known leadership: a deposed-but-
+	//     running leader can still serve read-consistent queries.
+	switch target.GetMode() {
```
dweitzman, line 371: The previous "isConsensusLeader" gate for when to allow user queries to the pooler's query server here wasn't as crisp as this. Now we're saying you can write to the pooler if the pooler's state is writable

Elsewhere, writable is defined as:
- not in recovery mode
- the most recent committed rule names this pooler as leader
- the most recent committed rule is not revoked

PR #1206, opened 2026-06-30
fix(queryserving): manage idle session timeout at gateway

haritabh17:
## Summary
- manage `idle_session_timeout` in multigateway instead of replaying it to pooled PostgreSQL backends
- enforce idle-session expiry at the pgwire layer with PostgreSQL-compatible FATAL 57P05
- keep multipooler stale-backend retry behavior for safe first-use connection errors

## Tests
- `go test -short ./go/common/mterrors ./go/common/pgprotocol/server ./go/services/multigateway/handler ./go/services/multigateway/engine ./go/services/multigateway/planner ./go/services/multipooler/internal/executor -count=1`
- `go test -short ./go/services/multigateway/... ./go/services/multipooler/internal/... -count=1`

## Manual validation
- rebuilt and restarted a local Multigres cluster from this branch
- verified `SET idle_session_timeout = '2s'` terminates the idle client session with FATAL 57P05\n- verified fresh connections continue to work afterward

### DIFF
--- a/go/common/pgprotocol/server/conn.go
+++ b/go/common/pgprotocol/server/conn.go
@@ -703,16 +703,38 @@ func (c *Conn) serve() error {
 		}
 	}
 
-	// Main command loop.
+	// Main command loop. Startup already sent ReadyForQuery, so the first
+	// frontend message is the start of a new command cycle. For extended query
+	// protocol, only Sync completes that cycle and returns to ReadyForQuery;
+	// Parse/Bind/Execute/Flush/etc. remain within the same cycle.
+	waitingForCommand := true
 	for {
 		// Check if connection is closed.
 		if c.closed.Load() {
 			return nil
 		}
 
-		// Read the message type (1 byte).
-		msgType, err := c.ReadMessageType()
+		// Read the message type (1 byte). If the handler manages
+		// idle_session_timeout, arm a read deadline only while waiting for the
+		// first client message of a new command cycle in an idle transaction state.
+		// Clear it as soon as a byte arrives so message-body reads and query
+		// execution are not accidentally governed by the idle-session clock.
+		idleDeadlineArmed, err := c.armIdleSessionTimeout(waitingForCommand)
 		if err != nil {
+			c.logger.Error("failed to arm idle_session_timeout", "error", err)
+			return err
+		}
+		msgType, err := c.ReadMessageType()
+		if err == nil && idleDeadlineArmed {
+			if clearErr := c.conn.SetReadDeadline(time.Time{}); clearErr != nil {
+				c.logger.Error("failed to clear idle_session_timeout read deadline", "error", clearErr)
+				return clearErr
+			}
+		}
+		if err != nil {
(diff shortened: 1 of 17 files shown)


go/common/pgprotocol/server/conn.go
```diff
@@ -710,9 +710,27 @@ func (c *Conn) serve() error {
 			return nil
 		}
 
-		// Read the message type (1 byte).
+		// Read the message type (1 byte). If the handler manages
+		// idle_session_timeout, arm a read deadline only while waiting for the
+		// next top-level client message in an idle transaction state. Clear it as
+		// soon as a byte arrives so message-body reads and query execution are not
+		// accidentally governed by the idle-session clock.
+		idleDeadlineArmed, err := c.armIdleSessionTimeout()
```
shedges: Optimization: Here we are performing check per message. In case of extended query protocol, transaction status will be in `TxnStatusIdle` throughout Parse,Bind, Execute, Sync. Generally JDBC driver or other connector uses extended query protocol heavily. PostgreSQL only enforces `idle_session_timeout` at the outer command boundary i.e. after `ReadyForQuery is sent and the server is waiting for the first byte of a new cycle.  We can follow same here?

GuptaManan100: Yep, I think this is wasteful too

go/common/pgprotocol/server/conn.go
```diff
@@ -768,6 +786,50 @@ func (c *Conn) serve() error {
...
+// idle_session_timeout applies only when the session is idle outside a
+// transaction, so non-idle transaction states clear any previous deadline.
+// Returns true when a timeout deadline was armed for the next ReadMessageType.
+func (c *Conn) armIdleSessionTimeout() (bool, error) {
+	provider, ok := c.handler.(IdleSessionTimeoutProvider)
+	if !ok {
+		return false, nil
+	}
+	if c.txnStatus != protocol.TxnStatusIdle {
+		return false, c.conn.SetReadDeadline(time.Time{})
```
shedges: we will clear deadline on every message even if we have not set alarm? same for line 805. As per cloude it looks to be expensive call. Please verify

```
1. Go runtime — timer heap
Go's network poller maintains a timer heap. Setting a deadline adds a timer entry; clearing it cancels one. Even SetReadDeadline(time.Time{}) — the "clear" call — still goes into the runtime, acquires a lock on the poll descriptor, and checks whether there's a timer to cancel. It doesn't short-circuit on "already zero."

2. OS interaction
On Linux this involves epoll, on macOS kqueue. The runtime may need to re-register the file descriptor with updated flags or deadlines in the OS poller. This is a kernel transition — user space → kernel space → back.


```

haritabh17: Ack, fixed only clearing when its armed.

go/common/pgprotocol/server/conn.go
haritabh17: Updated — the loop only does the timeout work at command-cycle boundaries.

shedges (approved):
LGTM 🚀

PR #1211, opened 2026-06-30
refactor(consensus): introduce ConsensusRole and adopt it over NamesSelfAsLeader/PoolerType

dweitzman:
The idea here is to start splitting apart the types for consensus vs routing

## Changes

**1. Introduce `ConsensusRole` and adopt it (`refactor(consensus)`)**
- Add `ConsensusRole` (`leader`/`follower`/`observer`) with `SelfConsensusRole(cs)`
  over a `ConsensusStatus`, plus the `RuleNamesLeader` and
  `IsNonRevokedCommittedLeader` helpers.
- Replace `NamesSelfAsLeader` (a leader-only boolean) with `SelfConsensusRole`
  everywhere; the wrapper is removed. `SelfConsensusRole` keeps follower and
  observer distinct, which `NamesSelfAsLeader` collapsed.
- Retype `PoolerAnalysis.NamesSelfAsLeader bool` -> `SelfConsensusRole ConsensusRole`
  in the multiorch recovery analyzers.

**2. Drop cached `primaryTerm` from `postgresState` (`refactor(multipooler)`)**
- The monitor recomputed `primaryTerm` every tick (an extra consensus read +
  change-detection churn) for a single consumer (the resign path). Compute the
  term on demand from the highest-known rule at that one site.
- Rename `latestRule()` -> `highestKnownRule()` to make the highest-known vs.
  committed distinction explicit; TODO left for a `highestCommittedRule`
  companion that write-safety decisions should use.

**3. Drive monitor role decisions from `SelfConsensusRole` (`refactor(multipooler)`)**
- `determineRoleAction` / `shouldMarkRewindReady` branch on the consensus role
  instead of the `PoolerType` label routed through `intendedRole`/`deriveTypeAndObs`.
- Behavior-preserving: the no-rule wait stays (`highestKnownRule() == nil`), leader
  -> PRIMARY and follower/observer -> REPLICA, and the published-label drift check is
  computed locally. `deriveTypeAndObs` survives only to build the
  `LeaderObservation` for publishing.

## Behavior

No intended behavior change. Existing monitor/recovery/consensus unit tests pass
unchanged (the equivalences `highestKnownRule()==nil <=> UNKNOWN`,
`SelfConsensusRole==Leader <=> rule names self` hold).

### DIFF
--- a/go/common/consensus/compare.go
+++ b/go/common/consensus/compare.go
@@ -206,9 +206,15 @@ func idsEqual(a, b *clustermetadatapb.ID) bool {
 }
 
 // RuleNamesLeader reports whether rule names id as its leader. Returns false
-// when rule or its leader ID is absent.
+// when rule, its leader ID, or id is absent — two absent IDs must not be treated
+// as a match (idsEqual(nil, nil) is true), or a pooler with no ID would be
+// classified as the leader of a leaderless bootstrap rule.
 func RuleNamesLeader(rule *clustermetadatapb.ShardRule, id *clustermetadatapb.ID) bool {
-	return rule != nil && idsEqual(rule.GetLeaderId(), id)
+	leader := rule.GetLeaderId()
+	if leader == nil || id == nil {
+		return false
+	}
+	return idsEqual(leader, id)
 }
 
 // ComparePosition returns negative, zero, or positive based on whether a is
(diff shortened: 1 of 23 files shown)


rafael (approved):
LGTM

PR #1215, opened 2026-07-01
fix(multipooler): bump current_rule read timeout to reduce CI flakes

dweitzman:
500ms was too tight under CI contention, causing intermittent TestFixReplication failures.


rafael (approved):
LGTM

PR #1222, opened 2026-07-02
fix(query-serving): preserve listen notify ordering

haritabh17:
## Summary

- Adds a bidirectional multipooler `NotificationStream` RPC for LISTEN/NOTIFY delivery.
- Uses one ordered notification stream per gateway client session, with subscribe/unsubscribe updates sent on that stream.
- Buffers notifications while a client session is inside a transaction and drains them after transaction end without allowing newer notifications to overtake buffered ones.
- Keeps accepted PostgreSQL compatibility divergences recorded in pg_regress patches instead of virtualizing them.

## Current behavior

- LISTEN/UNLISTEN state remains owned by the multigateway client session.
- Notifications are delivered through one session-scoped stream, preserving PostgreSQL cross-channel delivery order.
- Subscription updates are acknowledged only after the multipooler listener has applied them.
- Notification source PIDs are normalized in regression verification because Multigres reports physical PostgreSQL backend PIDs.
- `pg_listening_channels()` and `pg_notification_queue_usage()` divergences are captured in PostgreSQL regression patch files.
- Isolation test results use the same patch-verification pipeline as regression tests.

## Validation

- `go test -short ./go/services/multigateway/... -count=1`
- `go test -short ./go/services/multigateway/... ./go/services/multipooler/grpcpoolerservice -count=1`
- `go test -run TestListenNotify ./go/test/endtoend/queryserving -count=1 -timeout=20m`
- `RUN_PGREGRESS=1 PGREGRESS_TESTS=guc PGREGRESS_PATCH_MODE=verify go test -v -run 'TestPostgreSQLRegression/regress' ./go/test/endtoend/pgregresstest/... -timeout 60m`
- `RUN_PGISOLATION=1 PGISOLATION_TESTS=async-notify PGREGRESS_PATCH_MODE=verify go test -v -run 'TestPostgreSQLRegression/isolation' ./go/test/endtoend/pgregresstest/... -timeout 60m`


go/services/multigateway/engine/pg_listening_channels.go
```diff
@@ -0,0 +1,100 @@
...
+// path. The supported pg_listening_channels() forms carry no parameters, so the
+// result is identical to simple-protocol execution.
+func (p *PgListeningChannels) PortalStreamExecute(
+	ctx context.Context,
+	exec IExecute,
+	conn *server.Conn,
+	state *handler.MultiGatewayConnectionState,
+	_ *preparedstatement.PortalInfo,
+	_ int32,
+	_ bool,
```
GuptaManan100: We can't ignore the bool here 😄, this field represents whether we need to also return Describe results. So, if it is a portal execute, it needs describe information to be returned too.

We should add a test for extended query protocol sending this query also wanting describe information and a case where we don't want it

From what I can tell `SELECT pg_listening_channels()` via extended protocol without a Describe emits a spurious RowDescription, a protocol divergence from stock PostgreSQL.

GuptaManan100 (commented):
Went through half the PR right now, regarding the virtualisation of pg_listening_channels. We need ot fix 👇 and let's add e2e tests to make sure all combinations of protocol work

go/services/multigateway/engine/pg_listening_channels.go
haritabh17: After much back and forth, I have pivoted the implementation 😄 . There were 3 things to address:

1. Preserving delivery of notifications across xact (and savepoint) boundaries across channels
2. `pg_listening_channels` behaviour
3. `pg_notification_queue_usage` behaviour

Out of this, fixed only (1) by moving to one session-scoped stream across different notification channels and accepted the function output as accepted diff -- as those fall under catalog table divergence ...

go/services/multigateway/poolergateway/notification_adapter.go
```diff
@@ -58,160 +52,231 @@ func NewGRPCNotificationManager(
...
-		if firstAttempt {
-			close(ready)
-		}
 		return errNoClient
 	}
-
-	stream, err := client.StreamNotifications(ctx, &multipoolerpb.StreamNotificationsRequest{
-		Channels: []string{pgChannel},
-	})
+	ctx, cancel := context.WithCancel(context.TODO())
```
GuptaManan100: We should use context.Background() since we want a detached context. Let's use no lint like we had before cause we are sure of the reasoning why we're adding this - 

```
//nolint:gocritic // Metric recording at unsubscribe time, no request context available.
				m.metrics.StreamRemove(context.Background())
```

go/services/multigateway/handler/connection_state.go
```diff
@@ -138,6 +138,17 @@ type MultigatewayConnectionState struct {
 	// AsyncNotifCh is the channel for the server.Conn async notification pusher.
 	AsyncNotifCh chan<- *sqltypes.Notification
 
+	// PendingNotifications holds notifications received while this session is
+	// inside a transaction. PostgreSQL delivers LISTEN notifications only between
+	// transactions, so these are drained after COMMIT/ROLLBACK.
+	PendingNotifications []*sqltypes.Notification
```
GuptaManan100, line 146: This slice is unbounded, a long-open transaction with high NOTIFY volume grows it without limit (PG's async queue is bounded). Maybe we should add a bound to this.

GuptaManan100 (commented):
Just one more nit: the title needs updating, we aren't virtualising listening channels and pids anymore. The title shows up in commit history of the repo, so we should fix it.

GuptaManan100 (approved):

go/services/multigateway/poolergateway/notification_adapter.go
haritabh17: Ack, updated

go/services/multigateway/handler/connection_state.go
haritabh17: Ack, updated

PR #1223, opened 2026-07-02
refactor(multigateway): route by writable routing role, not consensus role

dweitzman:
## Summary

The multigateway is a **gate**: for routing and buffering, all that matters is *is there a writable primary for this shard, and which pooler is it*. This PR leans fully into that — the gateway routes by a **routing role** that each pooler derives and self-reports on its live health stream, rather than inferring leadership from consensus role or the etcd topology record.

The key shift: **a pooler advertises a leader observation only when it is the writable routing primary** (postgres out of recovery AND the active, non-revoked, highest-known committed leader). So the *presence* of the observation now implies writability — the gateway deals only in "writable primary present → route writes here; absent → buffer," with the rule number as a tiebreaker during a brief failover overlap.

## What changed

- **Route by a live routing-primary set, not etcd / highest-known** — the gateway's per-shard summary is the set of poolers currently claiming to be the writable primary, fed purely by health-stream observations. Dropped `mergeTopologyLeader`: etcd is not consulted for leader identity. We need a gRPC connection to send queries anyway, so that seems like the authoritative place to learn about pooler health
- **Derive the leader observation from routing state** — it is now a pure function of the fanned serving state (non-nil, naming self, iff routing PRIMARY; nil otherwise). Deleted the imperative `UpdateLeaderObservation` and all its callers; followers and demoted poolers advertise nil automatically. Added `StateManager.Recalc` so a revocation flips routing to REPLICA (and clears the advertisement) immediately instead of on the next monitor tick.
- **Dropped the `writable` health field** — presence of the observation implies writability, so the separate bool is redundant (proto + health state + gateway-facing conversion).
- **Shard summary is the sole authority for primary claims** — dropped `believesSelfLeader`; both stale-leader reporting and replica-read exclusion read the summary. Any primary claimant is excluded from replica reads: a claimant headed for demotion will restart / `pg_rewind` postgres, interrupting reads that outlast the drain grace period.

## Behavior changes worth calling out

- **CONSISTENT reads buffer until a writable leader exists**, rather than being served by a not-yet-writable appointed leader — more correct for read-your-writes, at the cost of buffering during the promotion window.
- The gateway discovers a new primary from its **health stream**, not etcd — slightly later than an etcd write, but live and authoritative.
- A demoted / stale primary is excluded from replica reads until it stops claiming primary.

## Potential follow-ups

- Evolve `LeaderObservation` → `RoutingState {role, rule}` with a prefixed `RoutingRole` proto3 enum. Kept out of this PR to stay reviewable.
- Lock-free recalc so `ObservePosition` can trigger a re-fan on a superseding committed rule (today: monitor tick).
- Multi-phase fan-out to retire the `AwaitStateChange` rendezvous.

### DIFF
--- a/go/services/multigateway/poolergateway/load_balancer.go
+++ b/go/services/multigateway/poolergateway/load_balancer.go
@@ -144,6 +144,17 @@ func (s *shardSummary) clearPrimary(poolerID topoclient.ComponentID) bool {
 	return !sameObservation(before, s.primaries.highest())
 }
 
+// hasPrimary reports whether poolerID currently claims to be a routing primary
+// for this shard — i.e. its latest health observation named itself. This is the
+// single authority for "believes itself the writable primary": both stale-leader
+// detection and replica-read exclusion read it, rather than re-deriving the same
+// signal from each connection's live health.
+func (s *shardSummary) hasPrimary(poolerID topoclient.ComponentID) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.primaries.has(poolerID)
+}
+
 // sameObservation reports whether two leader observations name the same elected
 // leader at the same rule (both nil counts as same).
 func sameObservation(a, b *clustermetadatapb.LeaderObservation) bool {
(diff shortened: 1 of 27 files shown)


go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -52,51 +52,113 @@ func shardKeyOf(sk *clustermetadatapb.ShardKey) shardKey {
...
 
-	mu sync.Mutex
-	// leaderObs is the most-authoritative LeaderObservation seen across this
-	// shard's poolers — merged across self_leadership topology records
-	// (folded in at mergeTopologyLeader time) and LeaderObservation
-	// messages on health streams (folded in via onPoolerHealthUpdate).
-	// May be nil if no leader has been observed yet.
-	leaderObs *clustermetadatapb.LeaderObservation
+	mu        sync.Mutex
+	primaries primarySet
```
dweitzman, line 78: In the load balancer, there's now a struct for keeping track of all poolers that currently claim to be a routing primary

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -52,51 +52,113 @@ func shardKeyOf(sk *clustermetadatapb.ShardKey) shardKey {
...
+	}
+	p.byID[id] = obs
+}
+
+func (p *primarySet) clear(id topoclient.ComponentID) { delete(p.byID, id) }
+
+func (p *primarySet) has(id topoclient.ComponentID) bool { _, ok := p.byID[id]; return ok }
+
+// highest returns the member with the highest rule, or nil if the set is empty.
+func (p *primarySet) highest() *clustermetadatapb.LeaderObservation {
```
dweitzman, line 106: You can ask the set of self-proclaimed primaries for either:
- what's the most authoritative claim? (useful for routing)
- is a particular pooler claiming to be a primary? (useful for displaying stale leaders in some dashboards)

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -227,33 +285,6 @@ func newLoadBalancer(opts loadBalancerOpts) *loadBalancer {
...
-// topology record) into the shard's shardSummary, keeping whichever
-// observation has the higher rule number. A pooler that is not the leader
-// carries no self_leadership, so this is a no-op for it — it never clears an
-// existing entry, since a higher observation from any source is what
-// supersedes a leader.
-//
-// Called by the pooler cache's OnLive and OnUpdate hooks. Acquires lb.mu and
-// the summary's mu internally (in that order, non-overlapping); must not be
-// called while holding either.
-func (lb *loadBalancer) mergeTopologyLeader(pooler *clustermetadatapb.MultiPooler) {
```
dweitzman, line 240: etcd doesn't factor in at all anymore. You need a gRPC connection to send traffic to a pooler anyway, so it seems reasonable to use that gRPC connection for health checks

go/services/multigateway/poolergateway/load_balancer.go
```diff
@@ -315,19 +346,14 @@ func (lb *loadBalancer) notifyLeaderServingFromSummary(summary *shardSummary, co
...
+	// is safe without a separate writability check.
 	if !proto.Equal(health.LeaderObservation.GetLeaderId(), connID) {
 		return
 	}
-	// Gate write-resume on writability, not just leadership. A leader can answer
-	// reads (CONSISTENT) and report SERVING before it has finished promoting,
-	// while postgres is still in recovery. Draining the failover buffer then would
-	// route buffered writes to a read-only node ("cannot ... during recovery" /
-	// read-only transaction). isWritable() reflects postgres being out of recovery.
-	if !health.isWritable() {
```
dweitzman, line 328: A leader observation now implies writability

rafael (approved):
Rubber stamp.

PR #1225, opened 2026-07-02
refactor(heartbeat): clarify that ReplTracker is based on routing role

dweitzman:
`ReplTracker.isPrimary` doesn't mean "primary" — it means "am I running the heartbeat writer," which is a function of the routing role. Two cleanups:

1. **Rename to match:** `isPrimary` → `writingHeartbeats`, `makePrimary`/`makeNonPrimary` → `startWriting`/`stopWriting`, and the test-only `IsPrimary()` → unexported `isWritingHeartbeats()`.

2. **Drop the `SERVING` gate:** the writer now runs iff `RoutingRole.Writable()`. Heartbeats are an internal signal (write-path proof + replica lag) written via the internal query service, so they should keep flowing on a writable primary even while user serving is paused — freezing them feeds replicas a false lag signal. Two gating test cases flip accordingly.


go/services/multipooler/internal/heartbeat/repltracker.go
```diff
@@ -62,36 +65,46 @@ func (rt *ReplTracker) HeartbeatReader() *Reader {
...
-// PRIMARY — out of recovery AND the active consensus leader) AND serving;
-// otherwise the reader runs. Writability already folds in the out-of-recovery
-// requirement, so heartbeats are never written to a standby.
+// OnStateChange transitions the heartbeat tracker based on the routing role: the
+// writer runs whenever this pooler is the writable leader (routing role PRIMARY —
+// out of recovery AND the active consensus leader); otherwise the reader runs.
+// Writability folds in the out-of-recovery requirement, so heartbeats are never
+// written to a standby.
+//
+// It intentionally does NOT gate on ServingStatus. Heartbeats are an internal
```
dweitzman, line 96: This is a behavior change, but it feels reasonable?...

Verolop (commented):
LGTM,

Verolop (approved):

PR #1226, opened 2026-07-02
refactor(multipooler): replace recovery-mode bools with a pgmode.Mode enum

dweitzman:
De-conflates postgres's physical recovery mode from the routing-role
(writability) and consensus-leadership axes it was being confused with.

- Introduce a dedicated `pgmode.Mode` enum (`Unknown`/`Primary`/`InRecovery`)
  in `internal/pgmode`, replacing the recovery-mode bools (`PostgresPrimary`,
  `postgresState.isPrimary`, `promotionState.isPrimaryInPostgres`, and the
  backup `RoleFunc`). The zero value `Unknown` is safe: `OutOfRecovery()` is
  false, so an unobserved node never derives a writable role.
- Consolidate to a single recovery-mode accessor: fold the bool-returning
  `isInRecovery()` into `postgresMode()` (returns `pgmode.Mode`) and delete the
  redundant `isPrimary()` wrapper. Guardrails read `Mode.OutOfRecovery()`, so no
  recovery-mode bool is left to misuse — a local `isPrimary`/`isInRecovery` bool
  still read as topology/consensus "primary".
- Left as-is (different axes, already-disambiguated names): topology predicates
  over `PoolerType` (`isPrimaryInTopology` / `isReplicaInTopology`) and the
  read-only GUC bool.

TODOs added for follow-ups: consider renaming the `SetPrimary` RPC to
`SetReplicationPrimary`, and whether backup readiness should key off recovery
mode or the routing role.


rafael (approved):
Rubber stamp

go/services/multipooler/internal/manager/backup/backup_health.go
```diff
@@ -272,29 +273,37 @@ func (e *Engine) refreshHealth(ctx context.Context) {
 		return
 	}
 
-	isPrimary := e.resolveRole(ctx)
+	// pgMode is the physical recovery mode: it gates the archive_command /
+	// archive_mode readiness checks and the WAL-archive-lag gauge, since only a
+	// primary (out of recovery) archives WAL.
+	//
+	// TODO: consider whether these checks should instead key off the routing role
```
dweitzman, line 280: FYI @cuongdo 

I haven't changed behavior here, but I'm wondering if this backup logic should depend purely on recovery mode or if it should also factor in whether a non-recovery pooler is qualified to accept user transactions or not. I haven't looked too closely at what the mode is being used for here yet

PR #1227, opened 2026-07-02
refactor: replace LeaderObservation with RoutingState{role, rule}

dweitzman:
De-conflates the gateway-facing leadership signal into a single
routing/writability type, continuing the axis-separation work (see the
pgmode.Mode and RoutingRole efforts).

## Proto
- Rename `clustermetadata.LeaderObservation` → `RoutingState {role, rule}` and add
  a prefixed `RoutingRole` enum. `leader_id` is dropped — a pooler only reports
  its own routing state, so identity is contextual.
- Rename `MultiPooler.self_leadership` → `routing_state` and
  `StreamPoolerHealthResponse.leader_observation` → `routing_state`.

## multipooler (producer)
- `servingstate.State` collapses the `RoutingRole` + `Leadership` pair into one
  full `RoutingState` (role + qualifying rule). `deriveRoutingState` is the
  low-level derivation; REPLICA carries the highest-known rule (advisory).
- `PoolerType` is dropped from the record's mutable state and derived at publish
  (`typeForState`: SHUTDOWN→UNKNOWN, else PRIMARY iff routing_state PRIMARY), so
  it can never drift and callers never set it.
- The record holds the full routing state; the publisher swaps `nil` for
  non-primaries to avoid etcd churn on replica rule bumps. The health stream
  always publishes the full state so the gateway sees REPLICA reports too.

## multigateway (consumer)
- Keys the routing primary off `role == PRIMARY` and the reporting pooler's id
  (the map key) rather than a `leader_id` in the message; ranks competing
  primaries by rule.

## multiadmin / multiorch / ruleguard
- Backup leader selection, orch's discovery log, and the `.Type`-ban messages
  updated to the renamed field.

Behavior-preserving rename + internal collapse; no new buffering/consensus
logic.

### DIFF
--- a/go/common/consensus/compare.go
+++ b/go/common/consensus/compare.go
@@ -66,24 +66,6 @@ func FormatRuleNumber(rn *clustermetadatapb.RuleNumber) string {
 	return fmt.Sprintf("%d.%d", rn.GetCoordinatorTerm(), rn.GetLeaderSubterm())
 }
 
-// MostAuthoritativeObservation returns the LeaderObservation with the highest
-// rule number among those given (nil entries are skipped). Used to merge leader
-// observations from different sources — a topology record's self_leadership and
-// a health-stream report — choosing the one made under the newest rule. Returns
-// nil if all inputs are nil.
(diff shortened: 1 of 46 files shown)


rafael (approved):
Rubber Stamp

PR #1229, opened 2026-07-02
refactor: enforce routing/consensus axis separation via ruleguard

dweitzman:
Enforces the routing/consensus axis separation: the gateway routes on
`RoutingState` (writability) and multiorch derives leadership from consensus, so
neither should read the other's signal.

## Fixes
- Drop the `routing_state` read from orch's pooler-discovery log line — orch
  derives leadership from consensus (`SelfConsensusRole`), never the topology
  routing label; at discovery time there's no consensus snapshot, so the log
  omits it.
- Drop `promotionState.isPrimaryInTopology`: it was write-only (set from the
  `PoolerType` label, only ever logged).

## ruleguards
- `disallowRoutingStateInOrch`: multiorch must not read `routing_state` /
  `GetRoutingState()` on a topology record — the routing-axis sibling of the
  existing `MultiPooler.Type` ban.
- `disallowConsensusStateInGateway`: multigateway must not consult consensus
  (`SelfConsensusRole` / `IsActiveLeader` / `NamesSelfAsLeader`) — it routes on
  `RoutingState` only.

## Proto
Add a TODO on `MultiPooler.type` noting it is now derived at publish and no
longer consulted for routing/leader identity, so it can be removed once external
consumers move to `routing_state.role`.


haritabh17 (approved):
🚀

PR #1231, opened 2026-07-03
refactor(multipooler): deprecate PoolerType, confine derivation to pooler_record

dweitzman:
Marks the topology `PoolerType` label deprecated so staticcheck surfaces every
reader, and migrates the internal decision readers off it — leaving it a
publish-only projection for external consumers.

## Derive, don't store
- `MultiPooler.type` is `[deprecated = true]`; SA1019 now flags any use.
- `poolerRecord.Type()` computes the label via `typeForState` (from lifecycle +
  routing_state) instead of reading the stored proto field, and nothing writes it
  internally. It is stamped only on the published wire copy in
  `routingStateForPublish` — a single documented `//nolint:staticcheck`. Test
  fixtures that construct/assert the still-published label are excluded in
  `.golangci.yml`.

## Migrate internal readers off the label
- Drop the redundant `checkPoolerType(REPLICA)` guardrail: `checkReplicaGuardrails`
  already checks `postgresMode()` (must be in recovery), the authoritative
  physical precondition, which strictly implies "not the writable primary."
  Deletes the now-dead `checkPoolerType`.
- Demotion idempotency reads `routing_state` instead of the `Type` label.

## Deprecate DRAINED, add LIFECYCLE_QUARANTINED
- The `DRAINED` `PoolerType` value is `[deprecated = true]` — it is no longer
  produced (Type derives to UNKNOWN on shutdown, never DRAINED).
- `LIFECYCLE_QUARANTINED` replaces its one real semantic: a still-running pooler
  that cannot recover to a healthy replica (failed rewind, failed
  restore-from-backup, irrecoverable replication lag). The operator should treat
  it as absent from the shard's healthy replica count — retaining the pod for
  forensics and provisioning a replacement. Self-declared only, since poolers
  continuously republish their own topology record.

The only remaining reader is the external multigres-operator (finding the primary
for backups/drain/status); once it moves to `routing_state.role == PRIMARY`, the
field can stop being published and be removed.


proto/clustermetadata.proto
```diff
@@ -297,8 +301,14 @@ enum PoolerType {
...
-  // DRAINED is used for poolers that are temporarily removed from serving traffic
-  DRAINED = 3;
+  // DRAINED was used for poolers temporarily removed from serving traffic.
+  //
+  // Deprecated: no longer produced — the multipooler derives Type from
+  // routing_state + lifecycle and never emits DRAINED (draining is now expressed
+  // via serving_status / lifecycle). Marked deprecated to surface any remaining
+  // reader via staticcheck SA1019; slated for removal together with the Type
+  // field once the external multigres-operator stops reading it.
+  DRAINED = 3 [deprecated = true];
```
dweitzman, line 316: DRAIN shouldn't be here, since this is a proxy for routing state not provisioning state

proto/clustermetadata.proto
```diff
@@ -327,11 +337,31 @@ enum PoolerLifecycleStatus {
...
+  // overwritten. (A future gRPC call will let the orchestrator ask a pooler to
+  // quarantine itself, e.g. when orch observes it cannot catch up while its
+  // peers replicate fine; the pooler still owns the decision to write it.)
+  //
+  // Unlike SHUTDOWN, the pod is deliberately kept alive for forensics. The
+  // operator/provisioner should treat a QUARANTINED pooler as absent from the
+  // shard's healthy replica count — a shard expecting 4 replicas with one
+  // QUARANTINED is 3 healthy + 1 retained-for-investigation — and provision a
+  // replacement rather than tearing the quarantined pod down.
+  LIFECYCLE_QUARANTINED = 5;
```
dweitzman, line 369: Lifecycle stage is related to provisioning, so this new quarantined stage can serve the role DRAINED was meant to

go/services/multipooler/internal/manager/pooler_record_test.go
```diff
@@ -236,7 +236,7 @@ func TestPoolerRecord_Snapshot_ReturnsClone(t *testing.T) {
 	r := mustNewPoolerRecord(t, ts, newTestPoolerProto(clustermetadatapb.PoolerType_PRIMARY, clustermetadatapb.PoolerServingStatus_SERVING))
 
 	snap := r.Snapshot()
-	snap.Type = clustermetadatapb.PoolerType_DRAINED
+	snap.Type = clustermetadatapb.PoolerType_REPLICA
```
mkindahl, line 239: Why this change? Seems like it is an accidental change.

go/services/multipooler/init.go
```diff
@@ -315,8 +315,6 @@ func (mp *Multipooler) Init(startCtx context.Context) error {
 	multipooler.ServingStatus = clustermetadatapb.PoolerServingStatus_DISABLED
 	multipooler.PoolerDir = mp.poolerDir.Get()
 	multipooler.PgDataDir = os.Getenv(constants.PgDataDirEnvVar)
-	// For now, all poolers start as REPLICA
-	multipooler.Type = clustermetadatapb.PoolerType_REPLICA
```
mkindahl, line 319: Shouldn't the lifecycle status be set when starting the pooler, and maybe the routing role?

mkindahl (commented):
Just a few questions, but otherwise looks good.

go/services/multipooler/internal/manager/pooler_record_test.go
dweitzman: The .proto change in this PR deprecates the DRAINED, so for the purposes of this test REPLICA is fine

go/services/multipooler/init.go
dweitzman: The lifecycle stage is actually set earlier in topoclient.NewMultipooler()

mkindahl (approved):

PR #1233, opened 2026-07-03
fix(query-serving): close protocol and PostgreSQL fidelity gaps

haritabh17:
## Summary

- Align pgwire behavior with PostgreSQL for backend FATAL/PANIC relay and close semantics, unsupported fast-path FunctionCall handling, COPY notices before errors, transaction warnings/hints, and pooled login event trigger warnings.
- Split retry classification from backend-session discard handling so non-retryable PostgreSQL FATAL diagnostics evict dead backend sockets without widening reconnect retries.
- Preserve routed/planned SQL execution while tightening parser fidelity for conflicting `WITHIN GROUP` aggregate syntax.
- Refresh pg_regress, pgproto, and isolation verification patches for accepted compatibility divergences, including replay-safe `25006 read_only_sql_transaction` failover buffering.

## Problem

Query-serving compatibility gaps caused client-visible behavior to diverge from PostgreSQL across protocol errors, COPY diagnostics, transaction warnings, parser validation, failover retries, dead backend handling, and regression verification.

## Solution

Relay PostgreSQL diagnostics and close semantics through pgwire, reject unsupported protocol features without killing the session, preserve notice ordering, warn and document known pooling semantics, classify read-only transaction errors by replay safety, discard FATAL/PANIC backend sessions without treating every FATAL as retryable, and verify accepted regression/isolation diffs with committed patches.


docs/query_serving/connection_pooling.md
```diff
@@ -866,6 +866,47 @@ Future improvements could add support for:
...
+
+- **Login triggers fire once per pooled-backend creation, as the pooler's
+  connection — never per client connection.** A client connecting to
+  multigateway is attached to an already-running backend, so no login event
+  fires for it. Backends created before the trigger existed never fire it at
+  all.
+- **Trigger output is not delivered to clients.** A `RAISE NOTICE` in a login
+  trigger arrives during the pooler's connection startup, where notices are
+  discarded (`pgprotocol/client/startup.go`). No client ever sees it.
+- **The gateway warns at CREATE time.** `CREATE EVENT TRIGGER ... ON login`
```
shedges, line 883: Nice

GuptaManan100, line 883: The alternate would be to store triggers at the pooler, make the gateway query the pooler every time a connection opens to see whether triggers exist, and then spoof the notices, or open a real connection to postgres to get the notices. Either way, not worth doing right now.

go/services/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -218,6 +224,33 @@ func isSingleQuery(reservedConnID uint64, willReserve bool) bool {
...
+		}
+		switch strings.Trim(fields[i+1], ";,") {
+		case "ONLY":
+			return true
+		case "WRITE":
+			return false
+		}
+	}
+	return false
+}
```
GuptaManan100: I see, this change is for transactions that are read only getting an error when they do a write, and we shouldn't buffer those? 

Should we parse the begin and check for the appropriate field instead of string matching?

GuptaManan100: If a session has `default_transaction_read_only = on`, a plain `BEGIN` (or a single autocommit write) produces a *genuine* 25006 that isn't a failover, but`beginQueryIsReadOnly` returns false, so we'd classify it replay-safe and buffer + retry until `MaxBufferingRetries` before surfacing it. So we should check the connection settings too for this field.

go/common/pgprotocol/client/query.go
```diff
@@ -198,6 +199,22 @@ func (c *Conn) writeQueryMessage(queryStr string) error {
...
+// socket without a ReadyForQuery. The diagnostic is the real error and must
+// not be masked by the I/O failure. The read error stays in the chain so
+// mterrors.IsConnectionError classifies the connection as dead regardless of
+// the diagnostic's SQLSTATE.
+func responseReadError(captured, readErr error) error {
+	var diag *mterrors.PgDiagnostic
+	if errors.As(captured, &diag) {
+		return fmt.Errorf("%w (connection lost before ReadyForQuery: %w)", captured, readErr)
+	}
+	return fmt.Errorf("failed to read message: %w", readErr)
```
GuptaManan100: Shouldn't we use mterrors.Wrapf?

go/common/parser/context.go
```diff
@@ -123,9 +123,16 @@ type ParseError struct {
 
 	// Context and hints
 	Context     string // Additional context information (e.g., "in quoted string")
-	HintText    string // Helpful hint for user
+	HintText    string // Helpful hint for user (internal/tooling; never sent to clients)
```
GuptaManan100: We can actually remove this field altogether, its never read, only written to

go/common/parser/context.go
```diff
@@ -695,7 +708,14 @@ func (ctx *ParseContext) AddSyntaxError(message string) *ParseError {
...
 	}
 
-	return ctx.AddError(message, location)
+	pe := ctx.AddError(message, location)
+	if hint != "" {
+		// The stored slice holds a copy of *pe; set the wire hint on both so
+		// GetErrors-based readers and the returned pointer agree.
+		pe.WireHint = hint
+		ctx.errors[len(ctx.errors)-1].WireHint = hint
+	}
```
GuptaManan100: Are hints only generated for errors, cause sometimes addErrors adds warning and stores them instead, in which case we'd add the hint to the wrong ctx.errors field.
Also wouldn't it be better to just store *PgErrors in ctx.errors and ctx.warnings, and that way we don't need to double write?
Alternately we can pass the hint to the AddError call, and write it to pe, before we add it to the context.

GuptaManan100 (approved):
Rest LGTM!

go/common/pgprotocol/client/query.go
GuptaManan100: Instead of this, we can also change the `IsConnectionError` itself ot realise that if the severity of the error is FATAL, then it is a dead connection. `diag.IsFatal() ⇒ connection dead`

go/services/multigateway/poolergateway/pooler_gateway.go
haritabh17: Ack, updating with parser field

haritabh17: Ack, updating

go/common/parser/context.go
haritabh17: Yes actually, removing.

go/common/pgprotocol/client/query.go
haritabh17: Ack, updating

go/common/parser/context.go
haritabh17: Tried out various shapes and for almost all of them we are able to send back the warning + hints because that query is sent to PG and it returns the warning and we fwd it.

Only found the following cases where we are not able to send out the warning:

1.
```
      CREATE GLOBAL TEMP TABLE t AS EXECUTE p;
      EXPLAIN CREATE GLOBAL TEMP TABLE t AS EXECUTE p;                                                  
```
This because deparse logic drops the GLOBAL keyword because its deprecated, and the warning is about the deprecation so the query that hits PG does not contain GLOBAL 😄 


2.
```
      SET application_name = 'a\b';                                                                     
```
^^ This because we dont run SETs as is on PG.

Nevertheless, good idea to refactor, updating.

go/common/pgprotocol/client/query.go
haritabh17: Yupp, makes sense - adding the `diag.IsFatal()` check

GuptaManan100 (approved):
Love it!

PR #1234, opened 2026-07-03
feat(telemetry): add CPU and memory usage metrics to all components

GuptaManan100:
## Summary

- Every Multigres component (multigateway, multipooler, pgctld, multiorch, multiadmin, multigres CLI) now exports process CPU/memory and Go runtime metrics, wired once in the shared telemetry bootstrap.
- `process.cpu.time`, `process.memory.usage` (RSS), and `process.memory.virtual` are the intended replacement for `kubectl top` in nightly benchmarking; the Go runtime metrics (heap, GC, goroutines) come along as a debugging complement.
- Metrics are collected via gopsutil (cross-platform, works in local clusters too) plus the OpenTelemetry Go runtime instrumentation, and appear in the generated metric catalog / keep-list for all binaries.

## Description

The metrics are registered centrally in `InitTelemetry` on the shared `MeterProvider`, so all components pick them up with no per-component code. Registration failure at init is fatal (observability is a first-class requirement); a transient per-metric read failure during collection is logged and skipped rather than crashing a live process.

The Go runtime instruments are created inside the `contrib/instrumentation/runtime` dependency, so the static `metricsgen` scanner cannot see them. They are hard-added to the generator as external metrics (attributed to the telemetry package, so per-binary keep-list attribution is computed correctly) with names verified against a live `/metrics` scrape.

Dependencies: adds `github.com/shirou/gopsutil/v4` and `go.opentelemetry.io/contrib/instrumentation/runtime` (pinned to keep OTel at the existing 1.43.0 — no OTel version bump).

## Testing

- Unit (`processmetrics_test.go`): asserts the process instruments' units, RSS > 0, VMS >= RSS, and the `process.cpu.time` user/system split; asserts the exact set of 7 Go runtime instruments is emitted.
- Unit (`metricsgen_test.go`): locks the OTel-to-Prometheus name transform for the hard-added runtime metrics.
- Unit (`catalog_test.go`): asserts all 10 resource series are cataloged and admitted by every per-binary keep-list.
- E2E (`metrics_test.go`): scrapes the live Prometheus endpoints of a multipooler and the multigateway, asserting all 10 series are present with RSS > 0.


haritabh17 (approved):
Looks good 🚀. Can we add these panels in the Grafana dashboard as well?

GuptaManan100:
Done!

PR #1235, opened 2026-07-03
feat: `cluster status` implementation

piyushsingariya:
## Summary

Implements `multigres cluster status` — replaces the TODO stub with a live health readout.

**What changed:**

- **`cluster status` command** (`go/cmd/multigres/command/cluster/status.go`): Full implementation. Fetches topology (cells, poolers, gateways, orchestrators) from multiadmin in parallel, then fans out per-pooler `GetPoolerStatus` RPCs (failures tolerated — shown as "status unavailable"). Renders a tree-formatted text view or `--output json`. Flags: `--admin-server`, `--cell`, `--database`, `--output`.

- **`pg_version` proto field** (`proto/multipoolermanagerdata.proto`, `go/services/multipooler/internal/manager/pg_replication.go`): Adds `pg_version` (field 15) to `Status`. Multipooler populates it via `current_setting('server_version')` with a 200ms timeout; empty when postgres is unreachable. Stripped to bare semver (drops platform suffixes like `(Homebrew)`).

- **Config helper refactor** (`go/cmd/multigres/command/admin/client.go`): Extracts `parseLocalConfig` so both `getServerFromConfig` and the new `GetLocalConfig` export share one config-parsing path. Status command uses `GetLocalConfig` to read the etcd port and probe `/readyz` (HTTP) to report topology connectivity.

**Sample output (text):**

```
Multigres Cluster Status
========================

Cells: zone1, zone2, zone3

Database: postgres
├── PRIMARY: bw29flg9 (healthy)
│   └── PostgreSQL: 17.10 (running, LSN: 0/517BC88)
├── REPLICA: z54hbrv8 (healthy)
│   └── PostgreSQL: 17.10 (streaming, LSN: 0/517BC88)
└── REPLICA: dl2nkhdm (healthy)
    └── PostgreSQL: 17.10 (streaming, LSN: 0/517BC88)

Gateways:
├── localhost:15432 (healthy, connections: 0)
├── localhost:15433 (healthy, connections: 0)
└── localhost:15434 (healthy, connections: 0)

Orchestrators:
├── h7ctbhs9 (healthy)
├── 4kbhwp8p (healthy)
└── phspg276 (healthy)

Topology: etcd://localhost:2379 (connected)
```


piyushsingariya:
Hi @GuptaManan100, can you help with a little feedback, I thought of adding unit test, but codechanges ain't entirely fit for that. (will resolve the merge conflict after feedback)

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+	}
+	var state string
+	switch poolerType {
+	case clustermetadatapb.PoolerType_PRIMARY:
+		state = labelPostgresRunning
+	case clustermetadatapb.PoolerType_REPLICA:
+		state = labelPostgresStream
+	default:
+		state = strings.ToLower(strings.TrimPrefix(s.GetPostgresStatus().String(), "POSTGRES_STATUS_"))
+	}
```
mkindahl, line 365: Just using `PoolerType` will probably show up a little strange. For example, a primary that is not yet fully running might show up as "running" and a replica that is disconnected hence not streaming data will show up as "streaming".

The `PoolerType` is the "assigned role", so it I suggest to either be clear that this is the assigned role, or use the actual physical status to decide what to show since the names "running" and "streaming" suggest that they are doing precisely that.

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+		metricsURL := fmt.Sprintf("http://localhost:%d/readyz", localCfg.Etcd.Port+2)
+		if req, err := http.NewRequestWithContext(ctx, http.MethodGet, metricsURL, nil); err == nil {
+			if resp, err := http.DefaultClient.Do(req); err == nil {
+				resp.Body.Close()
+				etcdConnected = resp.StatusCode == http.StatusOK
+			}
+		}
+	}
+
+	if outputFormat == "json" {
```
mkindahl, line 241: This means that a format like "jabberwocky" will emit text, which seems a little silly. Suggest to validate the output format early in this function and throw an error if it is not a valid format.

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+		if gs.connCountOK {
+			label = fmt.Sprintf("%s, connections: %d)", strings.TrimSuffix(label, ")"), gs.connCount)
+		}
+		cmd.Printf("%s%s:%d %s\n", branch, gs.component.GetHostname(), pgPort, label)
 	}
+	cmd.Println()
 
-	clusterCmd.AddCommand(statusCmd)
+	cmd.Println("Orchestrators:")
+	for i, os := range orchs {
```
mkindahl, line 305: Nit: this is using "os" (which is a package name) as a local variable, which is a little confusing. Suggest to pick a different name.

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+	labelPostgresStream  = "streaming"
+)
+
+var (
+	// flags
+	adminServer  string
+	cell         string
+	database     string
+	outputFormat string
 )
```
mkindahl, line 62: It's a little strange to store these as global variables given that other commands do not use global variables. Suggest to use the same approach as for the other commands (that is, `cmd.Flags.GetString(...)`), for consistency.

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+	labelConnected       = "connected"
+	labelUnreachable     = "unreachable"
+	labelPostgresStopped = "(stopped)"
+	labelPostgresRunning = "running"
+	labelPostgresStream  = "streaming"
+)
+
+var (
+	// flags
+	adminServer  string
```
mkindahl, line 58: Nit: This variable does not seem to be used, so suggest to remove it.

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
-		RunE: func(cmd *cobra.Command, args []string) error {
-			fmt.Println("Checking Multigres cluster status...")
-			// TODO: Implement cluster status logic
-			fmt.Println("Cluster status: Running")
+		RunE:  runStatus,
+	}
+	cmd.Flags().StringVar(&adminServer, "admin-server", "", "gRPC address of the multiadmin server (overrides config)")
+	cmd.Flags().StringVar(&cell, "cell", "", "Filter results to a specific cell")
+	cmd.Flags().StringVar(&database, "database", "", "Filter poolers to a specific database")
+	cmd.Flags().StringVar(&outputFormat, "output", "text", "Output format: text or json")
```
mkindahl, line 75: The other commands use `String` rather than `StringVar`, so suggest to follow the convention used by the other commands.

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+
+	// Fetch topology in parallel.
+	var (
+		cellNames []string
+		poolers   []*clustermetadatapb.MultiPooler
+		gateways  []*clustermetadatapb.MultiGateway
+		orchs     []*clustermetadatapb.MultiOrch
+	)
+
+	g, _ := errgroup.WithContext(ctx)
```
mkindahl, line 115: Since you're aborting on an error after these goroutines, I suggest to use the context here so that you can cancel the other goroutines on a failure. Otherwise you will wait for them to time out rather than aborting them if one of the routines fail. See the additional suggestions below.

```suggestion
	g, gctx := errgroup.WithContext(ctx)
```

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+
+	g.Go(func() error {
+		resp, err := client.GetCellNames(ctx, &multiadminpb.GetCellNamesRequest{})
```
mkindahl, line 118:
```suggestion
		resp, err := client.GetCellNames(gctx, &multiadminpb.GetCellNamesRequest{})
```

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+		if err != nil {
+			return fmt.Errorf("GetCellNames: %w", err)
+		}
+		cellNames = resp.GetNames()
+		return nil
+	})
+
+	g.Go(func() error {
+		resp, err := client.GetPoolers(ctx, &multiadminpb.GetPoolersRequest{Cells: cells, Database: database})
```
mkindahl, line 127: See above.
```suggestion
		resp, err := client.GetPoolers(gctx, &multiadminpb.GetPoolersRequest{Cells: cells, Database: database})
```

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+		if err != nil {
+			return fmt.Errorf("GetPoolers: %w", err)
+		}
+		poolers = resp.GetPoolers()
+		return nil
+	})
+
+	g.Go(func() error {
+		resp, err := client.GetGateways(ctx, &multiadminpb.GetGatewaysRequest{Cells: cells})
```
mkindahl, line 136:
```suggestion
		resp, err := client.GetGateways(gctx, &multiadminpb.GetGatewaysRequest{Cells: cells})
```

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+		if err != nil {
+			return fmt.Errorf("GetGateways: %w", err)
+		}
+		gateways = resp.GetGateways()
+		return nil
+	})
+
+	g.Go(func() error {
+		resp, err := client.GetOrchs(ctx, &multiadminpb.GetOrchsRequest{Cells: cells})
```
mkindahl, line 145:
```suggestion
		resp, err := client.GetOrchs(gctx, &multiadminpb.GetOrchsRequest{Cells: cells})
```

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+		if err != nil {
+			return fmt.Errorf("GetOrchs: %w", err)
+		}
+		orchs = resp.GetOrchs()
+		return nil
+	})
+
+	if err := g.Wait(); err != nil {
+		return err
+	}
```
mkindahl, line 155: Here you're discarding the results if any of them fails.

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+
+	// Fetch per-pooler live status in parallel; tolerate individual failures.
+	poolerStatuses := make([]*poolerStatusResult, len(poolers))
+	pg, _ := errgroup.WithContext(ctx)
```
mkindahl, line 159: Since you're discarding all errors after these goroutines have exited, it is unnecessary to create a context. Suggest to just create an error group for them all and just throw it away after.

```suggestion
	var pg errgroup.Group
```

go/cmd/multigres/command/cluster/status.go
```diff
@@ -15,24 +15,505 @@
...
+		}
+	}
+	slices.SortStableFunc(poolerStatuses, func(a, b *poolerStatusResult) int {
+		return cmp.Compare(priority(a.pooler.GetType()), priority(b.pooler.GetType()))
+	})
+
+	// Fetch health for gateways and orchs in parallel via gRPC health API.
+	gwStatuses := make([]*componentStatus[*clustermetadatapb.MultiGateway], len(gateways))
+	orchStatuses := make([]*componentStatus[*clustermetadatapb.MultiOrch], len(orchs))
+	hg, _ := errgroup.WithContext(ctx)
```
mkindahl, line 194: Same here. You're discarding the errors after, so there is no point in creating a context.

```suggestion
	var hg errgroup.Group
```

go/cmd/multigres/command/cluster/status.go
mkindahl, line 1: I'm missing tests for the status command, e.g., testing output flag with "json", "text", and "not-a-valid-format".

mkindahl, line 1: Also missing tests that show that the status command emits reasonable output.

mkindahl (commented):
Thank you for submitting this proposal. The changes look good, but I have a few minor comments below and suggest that you add tests for the command to show that it behaves in a sane manner.

PR #1236, opened 2026-07-04
feat(consensus): split PoolerPosition into decision/proposal

dweitzman:
## Summary

A first step toward propagating stuck rules: split `PoolerPosition.rule` into
`RulePosition{decision, proposal}` so a pooler's position can distinguish a
confirmed decision from an outstanding (WAL-backed but not yet decided)
proposal. This PR does not implement propagation, but in some places
where a rule change is undecided it can now return an error instead of
potentially silently doing the wrong thing. The behavior is still a little bit
wrong until rule store actually implements rule position, which will be in a
follow-up PR.

Then the last PR will be actually implementing propagation based on the
new state information.

## Changes

- `RulePosition{decision, proposal}` proto message, threaded through
  `PoolerPosition`, `ReplicationPrimary`, `CoordinatorProposal`, and the
  `ApplyCertifiedRuleChange` requests.
- `PossiblyUndecidedRule`/`IsRuleDecided` helpers: prefer the proposal for
  leadership/identity reads, the decision for write-safety-gate checks.
- Bootstrap's initial rule number moves from `{0,0}` to `{0,1}`;
  `ruleNumberIsUnset` (renamed from `ruleNumberIsZero`) is now an
  unambiguous check, not a "phantom zero" heuristic.
- `IsRuleRevoked` fixed to treat a nil/`{0,0}` `OutgoingRule` on a revocation
  consistently (both mean "no position constraint to compare").
- All test files across `common/consensus`, `multiorch`, `multipooler`, and
  the endtoend suites updated to compile and pass against the new shape,
  including a couple of real test bugs found along the way (checking the
  stale outgoing rule instead of the newly-proposed one).


proto/clustermetadata.proto
```diff
@@ -451,18 +451,38 @@ message ShardRule {
...
+// that has been written to local WAL but not yet marked decided. A proposal
+// may already have reached quorum durability without this pooler (or the
+// coordinator that wrote it) ever learning the outcome — for example if the
+// writer crashed between the synchronous write completing and the decision
+// being confirmed.
+//
+// Comparable independent of any specific point in physical time: prefer the
+// higher rule (proposal if present, else decision), compared by
+// (coordinator_term, leader_subterm).
+message RulePosition {
```
dweitzman, line 465: The .proto files are the first things to look at, and this is on of the important changes: making a clear distinction between a decided rule vs an proposed rule transition that we don't know for sure reached quorum yet

proto/clustermetadata.proto
```diff
@@ -451,18 +451,38 @@ message ShardRule {
...
 //
-// Comparison: prefer the pooler with the higher rule (coordinator_term first,
-// then leader_subterm). Break ties by LSN.
+// Comparison: prefer the pooler with the higher position (see RulePosition).
+// Break ties by LSN.
 message PoolerPosition {
-  // The highest shard rule this pooler has committed to local WAL.
-  ShardRule rule = 1;
+  // The highest rule position this pooler has committed to local WAL.
+  RulePosition position = 1;
```
dweitzman, line 485: In general now, whenever we're talking about either a proposed change or a position in the WAL, we can/should talk about the RulePosition that can represent undecided transitions vs decisions

That's true here for PoolerPosition, for ReplicationPrimary (used by SetPrimary RPC), in the RPCs that drive rule changes like Promote(), etc.

proto/multipoolermanagerdata.proto
```diff
@@ -208,12 +208,6 @@ message Status {
   // Only meaningful when postgres_action != UNSPECIFIED.
   google.protobuf.Duration postgres_action_duration = 12;
 
-  // Cohort members from the most recent multigres.rule_history record.
-  // Empty if the table has no records or the database is unreachable.
-  // A record with an empty list (written during first-backup creation) signals
-  // that the shard has been initialized but not yet had its cohort established.
-  repeated clustermetadata.ID cohort_members = 13;
```
dweitzman, line 215: This was unnecessary, the same data can and should be directly read from ConsensusStatus

go/services/multipooler/internal/manager/consensus/rule_store.go
```diff
@@ -353,7 +378,7 @@ func (rs *ruleStore) CreateRuleTables(ctx context.Context, policy *clustermetada
 		INSERT INTO multigres.current_rule
 		  (shard_id, coordinator_term, leader_subterm, coordinator_id, cohort_members,
 		   durability_policy_name, durability_quorum_type, durability_required_count, created_at)
-		VALUES ($1, 0, 0, $2, '{}', $3, $4, $5, now())`,
+		VALUES ($1, 0, 1, $2, '{}', $3, $4, $5, now())`,
```
dweitzman, line 387: As a convenience, the first rule in a new backup is now 0.1 instead of 0.0

That helps with proto3 not distinguishing between unset vs zero-valued fields: a RuleNumber of 0.0 will be clear evidence that something is missing/wrong

go/services/multiorch/recovery/actions/reconcile_cohort.go
```diff
@@ -108,10 +109,18 @@ func (a *ReconcileCohortAction) Execute(ctx context.Context, problem types.Probl
...
-	if leader == nil || members.HighestKnownRule == nil {
+	if leader == nil || members.HighestKnownPosition == nil {
 		return mterrors.Errorf(mtrpcpb.Code_FAILED_PRECONDITION,
 			"no consensus leader known for shard %s", problem.ShardKey)
 	}
+	// Propagation is not yet supported — wait rather than dispatch a cohort
+	// change against an outgoing rule that isn't decided yet.
+	// TODO: once propagation lands, this should be able to proceed using a
+	// quorum-verified proposal as the outgoing rule.
+	if !commonconsensus.IsRuleDecided(members.HighestKnownPosition) {
```
dweitzman, line 120: In orch, if a rule change is in progress then we shouldn't try to attempt something like a cohort change until the rule change is complete

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -86,18 +86,21 @@ func (pm *MultiPoolerManager) staleStandbyDemoteTarget() *clustermetadatapb.Pool
 	}
 	// The recorded leader is us: not a "superseded by another leader" case, so
 	// there is nothing to demote toward.
-	if leader := rp.GetRule().GetLeaderId(); leader != nil && pm.serviceID != nil &&
+	if leader := commonconsensus.PossiblyUndecidedRule(rp.GetPosition()).GetLeaderId(); leader != nil && pm.serviceID != nil &&
```
dweitzman, line 89: In most situations it's ok to deal with a potentially undecided rule. The times when we need to distinguish an undecided rule are usually when deciding if we can make consensus-related changes

go/common/consensus/proposals.go
```diff
@@ -294,6 +298,15 @@ func buildProposalCore(
 	if len(recruitedStatuses) == 0 {
 		return nil, errors.New("all recruited nodes reported an invalid or missing WAL position")
 	}
+	// Propagation (building a proposal from a quorum-verified but undecided
+	// rule) is not yet supported — every recruit must already be decided.
+	// TODO: Implement propagation and make sure externally certified changes can ignore stuck proposals.
+	for _, cs := range recruitedStatuses {
```
dweitzman: This protects against attempting any failovers when a rule change is stuck / in progress.

This protects against doing the wrong thing, but until we implement propagation we won't support doing the right thing

go/common/consensus/proposals_test.go
```diff
@@ -1602,34 +1632,37 @@ func TestBuildSafeProposal_CohortReplacementSplitBrain(t *testing.T) {
...
+			ProposalLeader:     &clustermetadatapb.PoolerAddress{Id: zone1.d},
+			ProposedTransition: &clustermetadatapb.RulePosition{Decision: &clustermetadatapb.ShardRule{RuleNumber: r.TermRevocation.GetOutgoingRule()}, Proposal: makeRule(ruleNum(6, 0), atLeast(2), zone1.d, zone1.e)},
 		}, nil
 	})
 
 	require.NoError(t, err1)
 	assert.NotNil(t, proposal1)
-	require.NoError(t, err2)
-	assert.NotNil(t, proposal2)
+	require.ErrorContains(t, err2, "propagation is not yet supported")
```
dweitzman, line 1732: This is an important change worth calling out: previously this test was documenting a situation where we could have split brain under full general consensus rule changes

By replacing Rule with RulePosition and making it clear what's decided vs proposed, it becomes possible for a stateless coordinator to recognize WAL entries that didn't reach quorum so we can return an error instead of having split brain

For now this is only implemented at the type level: a follow-up PR will make rule_store explicitly track whether a rule position is a proposal vs a durable decision so it'll be reflected correctly in the ConsensusStatuses that power the safe rule change process

rafael (approved):
Rubber stamp

PR #1237, opened 2026-07-04
refactor: enforce single-word component names with a CI linter

GuptaManan100:
## Summary

- Rename the camel-cased component names to their single-word forms — only the leading `M` is capitalized: `MultiPooler`→`Multipooler`, `MultiOrch`→`Multiorch`, `MultiGateway`→`Multigateway`, `MultiAdmin`→`Multiadmin` (and the leading-lowercase `multiPooler`/... variants). Applied across Go code, protos, docs, scripts, and the web UI.
- Add a grep-based CI linter (`tools/naming_linter.sh` + `make naming-lint` + a `Naming Lint` job) that rejects the camel-cased forms so the inconsistency can't creep back in. It also covers `Multigres`, which was already consistent (0 offending usages) but is worth guarding.

## Description

These names are single words, so only the first letter should be capitalized — yet they were inconsistently written as `MultiPooler`/`MultiAdmin`/etc. in some places and `Multipooler`/... in others. The rename covers hand-written Go identifiers, proto message/service names, comments, log/error strings, docs, and the web UI. Generated code under `go/pb/` and the web protobuf-es output were regenerated from the renamed protos rather than hand-edited, keeping the embedded descriptors consistent. It also renames the `PoolerHealthState.multi_pooler` proto field to `multipooler`.

All-caps constant forms such as `ID_MULTIPOOLER`/`MULTIADMIN` are intentionally left unchanged — SCREAMING_CASE has no internal-capitalization problem — and the linter allows them. The linter is `git grep '[Mm]ulti(Pooler|Orch|Gateway|Gres|Admin)'`, wired into a `make naming-lint` target and a `Naming Lint` job in the Lint workflow.

**Wire/API note:** gRPC service renames change method paths (e.g. `/…MultiPoolerService/…` → `/…MultipoolerService/…`) and the `multi_pooler` JSON field name becomes `multipooler`. Proto binary encoding (field numbers) is unaffected.


Verolop (approved):
lgtm!

PR #1238, opened 2026-07-04
fix(multigateway): stop gateway-managed set_config leaking backend GUC

GuptaManan100:
## Summary

- Fixes a cross-client leak where `SELECT set_config('statement_timeout', …)` ran the real `set_config` on the pooled backend, persisting the GUC and leaking it to later clients on that connection.
- Gateway-managed `set_config` calls are now applied to gateway state **and rewritten out** of the query routed to PostgreSQL, so the real GUC is never written to a pooled backend. Values that can't be rewritten fail closed instead of leaking.
- Unifies the scattered gateway-managed-variable (GMV) dispatch behind a single registry, so adding a GMV is one entry the compiler checks, and adds a startup-param safety net so a new GMV can't leak via connection parameters.
- Adds dual-target (postgres + multigateway) e2e coverage and unit tests, and updates `docs/query_serving/gateway_managed_variables.md`.

## Problem

`statement_timeout` (and `idle_session_timeout`) are gateway-managed: the gateway owns their value and enforces the timeout itself, and it never forwards `SET`/`RESET` of them to the backend. But `SELECT set_config('statement_timeout', v, …)` was planned as `Sequence[ApplySessionState, Route]`, and the trailing `Route` ran the real `set_config` on the pooled backend. Because the gateway never resets that backend GUC, it leaked across statements and clients sharing the connection — producing both over-cancellation and "the value doesn't take effect / gets reset" symptoms.

## Solution

For a SELECT whose target list contains gateway-managed `set_config` calls, the gateway still applies each call to gateway-local state (so `SHOW` and enforcement stay correct), but rewrites each gateway-managed call out of the query it sends to the backend, replacing it with the value it would have returned:

- **Literal value** → its canonical constant computed at plan time (`'1000'` → `'1s'`), inlined as a constant.
- **Bound value** (`$N`, extended protocol) → the projection keeps a bind slot that is canonicalized at execute time. If `$N` is also used elsewhere in the query, a fresh synthetic bind slot is allocated so the shared param's other uses keep the raw value untouched.

The client always sees PostgreSQL's canonical form, so return-value parity is preserved. Mixed batches (a GMV `set_config` alongside an ordinary one) rewrite only the GMV call; ordinary `set_config` calls still run on the backend. An expression-valued GMV `set_config` now fails closed (internal error) rather than routing to the backend.

The old per-variable `case "statement_timeout"` switches across the planner and engine are replaced by a single registry (`handler/gateway_managed_variables.go`), built via a constructor whose parameters are all the behaviors — so a registration that forgets one is a compile error.

## Follow-up

- **NULL `set_config` value is not handled and needs a follow-up.** In PostgreSQL `set_config(name, NULL, _)` resets the parameter to its default; multigres currently rejects a NULL value up front at the planner (for every variable, GMV or not) with "must be a literal constant or a bound parameter". This is a pre-existing limitation, not introduced here, but it's a real divergence from PostgreSQL — it's pinned by `TestStatementTimeoutSetConfigNullValue` so the behavior is intentional and visible, and should be addressed separately.

## Testing

- Dual-target e2e (`go/test/endtoend/queryserving/statement_timeout_setconfig_test.go`): session/local × in-txn/out-of-txn matrix with both `SHOW` and real enforcement; deterministic backend-leak pin; mixed simple + extended protocol; shared-bound-value; per-GMV + multi-GMV batch; and the NULL-divergence pin. Every assertion runs against native postgres and the multigateway, so any divergence fails the run.
- Unit tests for the plan shape (literal / bound / shared-bound / fail-closed) and the execute-time canonicalization and portal decode (including the shared-param synthetic slot).


haritabh17 (approved):
🚀

shedges (approved):

PR #1239, opened 2026-07-04
refactor(tools): drop unnecessary atomic from list

ferhatelmas:
it's misleading for thread-safety (it's not) and also aligns with [upstream more closely](https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/container/list/list.go;l=50)


ferhatelmas:
This is wrong because connlist is trying to make a fast path lock free read.

PR #1241, opened 2026-07-04
fix(multipooler): reconcile drifted primary_conninfo on SetPrimary no-op

dweitzman:
## Summary
- `SetPrimary`'s rule-position no-op gate treated "position unchanged" the same as "position went backward" — in both cases it returned early without checking whether `primary_conninfo` itself had drifted from the recorded primary.
- The only backstop for that drift was `MonitorPostgres`'s periodic tick, 5 seconds apart, which is why `TestFixReplication` flaked roughly one run in three (~3/10-13 in local repro): `RequireRecovery` reported success while the replica was still waiting on the next monitor tick to actually reconfigure.
- `SetPrimary` already holds the action lock and has everything needed to check for drift itself, so it now does that inline via the same `primaryConnInfoDiffersFromRecorded` check the monitor uses, closing the race instead of waiting up to 5s for it.
- Extracted the reconcile logic (previously inlined in `MonitorPostgres`'s `remedialActionFixPrimaryConnInfo`) into a shared `reconcilePrimaryConnInfoToRecorded` so both call sites can't drift apart.
- Added diagnostics to `TestFixReplication`'s polling loops (wal receiver state, recorded primary, rewind-readiness) so any future flake here is diagnosable from CI logs alone, without needing local repro.

## Test plan
- New unit test `TestSetPrimary_NoOpReconcilesDriftedConnInfo` exercises the fix directly (drifted conninfo + unchanged rule position → still reconciled).
- `TestFixReplication` run locally 15/15 times with the fix (previously observed ~70-80% pass rate before the fix, reproduced both locally and via linked CI failures).

### DIFF
--- a/go/services/multipooler/internal/manager/rpc_consensus.go
+++ b/go/services/multipooler/internal/manager/rpc_consensus.go
@@ -604,6 +604,20 @@ func (pm *MultiPoolerManager) SetPrimary(ctx context.Context, req *consensusdata
 		pm.logger.InfoContext(ctx, "SetPrimary: incoming rule not higher, no-op",
 			"incoming_rule", rule.GetRuleNumber(),
 			"self_rule", selfPos.GetRule().GetRuleNumber())
+		// The rule itself is a no-op, but primary_conninfo may have drifted
+		// from what we're recorded as following (e.g. an operator or test
+		// manually cleared it without changing consensus state). We already
+		// hold the action lock and have everything needed to check — fix it
+		// now rather than leaving it to MonitorPostgres's next periodic tick,
+		// up to monitorRetryInterval away.
+		//
+		// Read the target back from the just-updated record rather than
+		// trusting this call's own leader/port: RecordTermPrimary has its own
+		// staleness comparison against whatever was already recorded, so this
+		// request may not be what actually got persisted.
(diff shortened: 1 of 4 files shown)


go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -604,6 +604,20 @@ func (pm *MultiPoolerManager) SetPrimary(ctx context.Context, req *consensusdata
...
+		// manually cleared it without changing consensus state). We already
+		// hold the action lock and have everything needed to check — fix it
+		// now rather than leaving it to MonitorPostgres's next periodic tick,
+		// up to monitorRetryInterval away.
+		//
+		// Read the target back from the just-updated record rather than
+		// trusting this call's own leader/port: RecordTermPrimary has its own
+		// staleness comparison against whatever was already recorded, so this
+		// request may not be what actually got persisted.
+		if pm.primaryConnInfoDiffersFromRecorded(ctx) {
```
dweitzman: Without this postgres monitor would eventually resolve the issue they (the tests) synthetically created by directly mucking with primary conninfo, but it might take a few seconds so tests we were timing out waiting for those auto-fix to apply. This does the fix more aggressively/sooner to make the tests happier

Verolop (approved):

PR #1242, opened 2026-07-05
feat(docker): Supabase cluster image + forward MULTIGRES_PG_EXTRA_CONF

GuptaManan100:
## Summary

- Publish a `multigres-cluster-supabase` image alongside the stock `multigres-cluster`, built on the `supabase/postgres:17.6.1.142-multigres` base (Alpine, `PROVISION_PG_PACKAGES=false`), in both the push-to-main cluster build and the nightly build matrices. This lets consumer CI pull a prebuilt cluster image on the Supabase Postgres base instead of building it themselves.
- Forward `MULTIGRES_PG_EXTRA_CONF` from the host env in `docker-compose.yml`. The entrypoint already applied it, and `MULTIGRES_PG_MAX_CONNECTIONS` was already forwarded, but this knob was missed — so there was no way via compose to set startup-only GUCs like `shared_preload_libraries`, which pg_cron/pg_net need to load.

## Notes

- No `Dockerfile.cluster` change was needed: the `-multigres` base already bundles pgBackRest, pgvector, and pgrep, so its `PROVISION_PG_PACKAGES=false` guardrail passes. The base tag is pinned for reproducibility. Per-image GHA cache scope keeps the two cluster variants from clobbering each other's cache.
- Verified locally: the Supabase cluster image builds and serves queries through the gateway including `CREATE EXTENSION vector`; and with the compose fix, running compose with `MULTIGRES_PG_EXTRA_CONF=shared_preload_libraries='pg_cron,pg_net'` brings both extensions up (`SHOW shared_preload_libraries` → `pg_cron,pg_net`, `CREATE EXTENSION pg_cron`/`pg_net` succeed).

### DIFF
--- a/.github/workflows/docker-build-cluster.yml
+++ b/.github/workflows/docker-build-cluster.yml
@@ -26,8 +26,27 @@ permissions:
 
 jobs:
   docker-cluster:
-    name: Build and Push cluster Docker Image
+    name: Build and Push ${{ matrix.image }} Docker Image
     runs-on: ubuntu-24.04
+
+    strategy:
+      # Don't let one variant's failure cancel the other's build.
+      fail-fast: false
+      matrix:
+        include:
+          # Stock cluster image: official Debian PostgreSQL base. The Dockerfile
+          # default (PROVISION_PG_PACKAGES=true) installs pgBackRest, pgvector,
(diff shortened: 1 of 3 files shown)


rafael (approved):
LGTM

PR #1243, opened 2026-07-06
fix(servenv): reject empty mTLS allow-list in grpc auth

haritabh17:
## Summary

- Reject an empty `--grpc-auth-mtls-allowed-substrings` (and empty entries like `a::b` or a trailing `a:`) at server startup when `--grpc-auth-mode=mtls`, instead of silently authorizing every CA-valid client certificate.
- Align the flag help, variable comment, and `MtlsAuthPlugin` doc comment with the new validation.

## Problem

In mTLS mode the allow-list is parsed with `strings.Split(clientCertSubstrings, ":")`. An empty flag yields `[""]`, and `strings.Contains(subject, "")` is always true, so any certificate the configured CA ever issued is authorized. Empty entries in an otherwise-populated list (`a::b`, `a:`) silently turn a restrictive allow-list into allow-all. This behavior was inherited verbatim from Vitess.

## Solution

The plugin initializer now returns an error if the split list contains any empty entry (which also covers the fully-empty flag). The error propagates through the auth plugin initializer, so a misconfigured server fails fast at startup rather than accepting all clients. `MtlsAuthPlugin` is constructed in exactly one place and its field is unexported, so no other path can produce a permissive instance.

Added `grpc_server_auth_mtls_test.go` covering the empty flag, a mid-list empty token, a trailing empty token, and valid single/multiple entries.

### DIFF
--- a/go/common/servenv/grpc_server_auth_mtls.go
+++ b/go/common/servenv/grpc_server_auth_mtls.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"slices"
 	"strings"
 
 	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


shedges (approved):
LGTM 🚀

PR #1244, opened 2026-07-06
feat(queryserving): propagate caller identity to the multipooler

shedges:
## Summary

Populates the previously-empty `caller_id` on gateway queryservice requests and mirrors the client's identity into OpenTelemetry baggage, so the multipooler can attribute a query to the **app that issued it** rather than only the shared database user. Closes the four `// TODO: Add caller_id` markers in the gateway.

This is observability, not authorization — Postgres already enforces table access natively (the real user reaches the backend), so `caller_id` here is for attribution/tracing.

## What changed

- **`go/common/callerid`** (new): a small context helper that carries a typed `mtrpc.CallerID` (Layer 3, the request field) and mirrors `principal`/`component` into OTel **baggage** (Layer 2, observability propagation). Uses `NewMemberRaw` so an `application_name` with spaces/punctuation survives.
- **multigateway**: builds the `CallerID` from the client connection — `principal` = authenticated DB user, `component` = client `application_name` — at `HandleQuery`/`HandleExecute`/`HandleDescribe`, and sets it on `ExecuteQuery` / `StreamExecute` / `PortalStreamExecute` / `Describe`.
- **multipooler**: records `mg.caller.principal` / `mg.caller.component` as span attributes on those RPC handlers.

## Trust note

`principal` is the authenticated DB user and is trustworthy; `component` is the client-supplied `application_name` and is only an assertion. Any future policy decision must key on `principal`, never `component`.

## Flow

client → gateway builds caller_id from the connection → context + baggage → gRPC (typed field on the request, baggage on the wire) → pooler records it on its span. The query shares the existing OTel `trace_id`, so it's now traceable **and** attributable by app.

## Tests

- `go/common/callerid`: 6 unit tests (round-trip, baggage mirroring, arbitrary `application_name`, nil no-op, empty-component omission). Pass.
- Edited packages (handler, poolergateway, grpcpoolerservice) still pass.


go/common/callerid/callerid.go
```diff
@@ -0,0 +1,92 @@
...
+	BaggagePrincipal = "mg.principal"
+	BaggageComponent = "mg.component"
+)
+
+type callerIDKey struct{}
+
+// New builds a CallerID from a client's identity. principal is the
+// authenticated database user; component is the client's application_name.
+func New(principal, component string) *mtrpcpb.CallerID {
+	return &mtrpcpb.CallerID{Principal: principal, Component: component}
```
haritabh17: nit: Why not call them what they are? 

`principal` -> `authenticatedUser`
`component` -> `clientApplicationName`

Want to use it for other purposes as well, since its `common`?

go/services/multigateway/handler/handler.go
```diff
@@ -189,11 +190,20 @@ func (h *MultigatewayHandler) statementTimeoutCtx(ctx context.Context, state *Mu
 	return ctx, func() {}
 }
 
+// callerContext enriches ctx with the client's identity so it reaches the
+// multipooler: a typed CallerID for the queryservice request field and
+// OpenTelemetry baggage for observability propagation. principal is the
+// authenticated database user; component is the client's application_name.
+func (h *MultigatewayHandler) callerContext(ctx context.Context, conn *server.Conn) context.Context {
+	return callerid.NewContext(ctx, callerid.New(conn.User(), conn.GetStartupParams()["application_name"]))
```
haritabh17: This is immutable snapshot from the startup handshake. So a client that does `SET application_name = '...'` mid-session will keep being attributed to its startup name.

Is this intended??

Would it make more sense to read the current value from the merged session view (GetSessionSettings()["application_name"]) instead? That would match what a user sees for `application_name` in PG. Also note that `callerContext` currently runs before `getConnectionState`, so may need the connection state fetched first.

haritabh17 (approved):
Rest all looks good 🚀

go/common/callerid/callerid.go
shedges: Done. It's inherited from Vitess. Idea was that it will be part of common Proto and other component can pass identity which can be different from PostgreSQL (DB) style auth. Currently there is only single component (Multigateway) so we can revisit in future if needed.

go/services/multigateway/handler/handler.go
shedges: Fixed

PR #1247, opened 2026-07-07
feat(multipooler): propagate a stuck proposal via ordinary automatic failover

dweitzman:
Ordinary automatic failover now trusts a cohort's most-advanced undecided proposal as its outgoing baseline instead of rejecting it outright. Both quorum modes (ordinary failover and externally-certified writes) finalize the stuck proposal first -- deciding it exactly as it already stands, under its own synchronous-ack quorum proof -- before applying the caller's actual change from a clean decided baseline. This removes the need for an external cert in the common case where only the leader is changing.

UpdateRule is decomposed into resolveNewRuleValues/maybeFinalizeStuckProposal/writeRuleProposal, each caching its own write's position rather than leaving that to callers.

### DIFF
--- a/go/services/multipooler/internal/manager/consensus/rule_store.go
+++ b/go/services/multipooler/internal/manager/consensus/rule_store.go
@@ -566,6 +566,154 @@ func (rs *ruleStore) readCurrentRuleLocked(ctx context.Context, inRecovery bool)
 	return pos, lockedCtx, nil
 }
 
+// resolveNewRuleValues resolves the leader/cohort/durability-policy values
+// UpdateRule should write: caller-supplied values (via the builder) take
+// priority, nil retains the existing value from currentRule. Also validates
+// the resolved cohort can satisfy the resolved policy, and computes the
+// incoming PolicyWithCohort used both by propagation's finalize step (see
+// finalizeStuckProposal) and the GUC transition.
+func resolveNewRuleValues(
+	currentRule *clustermetadatapb.ShardRule,
+	update *RuleUpdateBuilder,
+	isPromotion bool,
+) (newLeader *clustermetadatapb.ID, newCohort []*clustermetadatapb.ID, newDP *clustermetadatapb.DurabilityPolicy, incomingPWC consensus.PolicyWithCohort, err error) {
+	// Outside of a promotion, an existing leader can't be replaced: the current leader
+	// is the one issuing this write, and it has no way to instantaneously stop accepting
+	// transactions the moment the row changes — only a promotion (which goes through
+	// revocation/fencing on the outgoing leader first) can safely hand off leadership.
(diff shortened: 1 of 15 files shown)


go/services/multipooler/internal/manager/consensus/rule_store.go
```diff
@@ -668,56 +826,29 @@ func (rs *ruleStore) UpdateRule(ctx context.Context, update *RuleUpdateBuilder)
...
+	// it exactly as it already stands — same leader, cohort, and policy the
+	// proposal names — before doing anything else with it. See
+	// maybeFinalizeStuckProposal's doc comment for the full design rationale.
+	// promoted tracks whether promotionHook already ran there, so the
+	// ordinary isPromotion branch further below doesn't call it a second
+	// time — pg_promote() only needs to happen once before either write.
+	// The returned position isn't needed here: currentRule/currentTerm/
+	// currentSubterm above already reflect it (finalize never changes a
+	// proposal's own content, only whether it's marked decided).
+	_, promoted, err := rs.maybeFinalizeStuckProposal(ctx, lockedCtx, current, update, isPromotion, incomingPWC)
```
dweitzman, line 844: Now the first step of a rule change is finalizing any stuck proposals, which means any of:
- do nothing if the previous rule was already decided / finalized
- if we're in skipOutgoingCohort mode, we can mark a stuck proposal as decided/finalized without reaching the outgoing cohort
- in normal mode, we verify that the stuck transition reaches quorum under the transition quorum rules from its outgoing decision, then mark it as decided/finalized, then proceed with the rest of the rule change below as normal

go/test/endtoend/multiorch/stuck_proposal_test.go
```diff
@@ -346,3 +352,142 @@ func TestApplyCertifiedRuleChange_FromStuckProposal(t *testing.T) {
...
+	// stuck-proposal scenario, so it doesn't race the seeding below.
+	setup.StartMultiorchs(t.Context(), t)
+	resumeRecovery := setup.DisableRecovery(t, "multiorch")
+
+	// Seed a stuck proposal directly via raw SQL on the current primary,
+	// bypassing UpdateRule entirely — see the sibling tests in this file for
+	// why. Same leader/cohort/policy as the current decision: the whole
+	// point of this test is that ordinary failover can recover it without a
+	// cert precisely because nothing but the leader is changing.
+	stuckTerm := currentTerm + 1
```
dweitzman, line 404: It'd be possible but awkward to cause a stuck proposal by simulating network partitions or disconnecting replicas or something like that, so this e2e test just inserts a synthetic stuck proposal directly into the rule store, kills the primary, and asserts that orch was successfully able to do a failover even though the outgoing rule wasn't finalized

rafael (approved):
LGTM

PR #1248, opened 2026-07-07
feat(multipooler): make the action lock preemptable via UrgentAcquire

dweitzman:
- Add `ActionLock.UrgentAcquire`: an urgent caller cancels the current holder's context and takes priority over any already-queued ordinary `Acquire` callers, without an active-cancellation race (leans on `golang.org/x/sync/semaphore`'s own atomic hand-off).
- Wire `Recruit` (drives failover) and `GracefulShutdown` (SIGTERM, bounded by `--onterm-timeout`) to use `UrgentAcquire` instead of `Acquire`, so they preempt whatever currently holds the lock rather than queue behind it.
- Add an e2e test (`TestGracefulShutdownPreemptsStuckRuleWrite`) proving `GracefulShutdown` preempts a genuinely stuck synchronous-commit rule write end to end. The stall fix it exercises landed separately in #1400.


go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -121,8 +121,11 @@ func (pm *MultipoolerManager) Recruit(ctx context.Context, req *consensusdatapb.
 	ctx, span := telemetry.Tracer().Start(ctx, "consensus/recruit")
 	defer span.End()
 
+	// Urgent: Recruit drives failover, so it should preempt whatever
+	// currently holds the action lock (and take priority over queued
+	// ordinary work) rather than wait behind it — see actionlock.UrgentAcquire.
 	var err error
-	ctx, err = pm.actionLock.Acquire(ctx, "Recruit")
+	ctx, err = pm.actionLock.UrgentAcquire(ctx, "Recruit")
```
dweitzman, line 229: The main goal of all this change is that Recruit() is presumed to be more urgent than most other operations (backup, cohort changes, even potentially a stuck promotion, etc.)

We don't want to delay failover because a backup is in progress, for example

PR #1249, opened 2026-07-07
refactor(consensus): drop per-policy CheckSufficientRecruitment, derive revocation from SatisfiedBy

dweitzman:
## Summary
- `AtLeastNPolicy` and `MultiCellPolicy` each had a bespoke `CheckSufficientRecruitment` implementing the same majority + revocation invariants. The revocation check was structurally identical to that same policy's own achievability check, just inverted and applied to the un-recruited leftover set — so both are replaced with a single free function (`CheckSufficientRecruitment`) that asks the policy itself, removing `CheckSufficientRecruitment` from the `DurabilityPolicy` interface entirely.
- Renamed `CheckAchievable` to `SatisfiedBy` (and its `proposedCohort` parameter to `poolers`), since it's now called with proposed cohorts, recruited sets, and un-recruited leftovers alike — the old name/param only fit the first case.
- Revocation-failure error messages now name the actual un-recruited poolers instead of just a count.

### DIFF
--- a/docs/ha/rule-change.md
+++ b/docs/ha/rule-change.md
@@ -105,7 +105,7 @@ pooler that crashes and restarts still honors it.
 ### Why recruitment is enough: majority + revocation
 
 The coordinator only proceeds once the recruited set is large enough, checked by
-[`CheckSufficientRecruitment`](../../go/common/consensus/policy_at_least_n.go)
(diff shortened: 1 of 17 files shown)


Verolop (approved):
neat!

PR #1251, opened 2026-07-08
feat(multipooler): never let a cohort member trust archive-sourced WAL

dweitzman:
## Summary

`restore_command` is written implicitly by pgbackrest's initial `restore --type=standby` and never touched again today. A cohort member must only ever advance via streaming from the current leader — left enabled, a subsequent restart-as-standby can resolve `recovery_target_timeline=latest` via `restore_command`'s archive-get to a timeline the local checkpoint doesn't descend from and FATAL at startup, crash-looping with no way for the RPC-driven self-heal (`FixReplicationAction`) to even reach the node.

- `pgctld` gains a `restore-wrapper` subcommand and a `StopRestoreCommand` RPC: postgres cannot cancel an in-flight `restore_command` invocation on its own, so disabling the GUC alone only stops future fetches. The wrapper records its own PID so `StopRestoreCommand` can check liveness and terminate it (SIGTERM, then SIGKILL after a bounded grace period) if still running.
- `ConsensusManager.IsPotentialCohortMember` checks both the decided rule and an outstanding proposal, not just whichever one is "current" — a confirmed member must not read as "no longer a member" just because a newer, undecided proposal excludes them.
- `Recruit` clears `restore_command` and confirms it has actually stopped for both branches (a demoted primary restarting as standby, and an already-standby being recruited), as a hard error — this is a correctness invariant, not best-effort.
- `SetPrimary` also clears/confirms it whenever the incoming rule names this pooler a cohort member, best-effort.
- Restore-from-backup asserts the pooler is not a cohort member, and wraps the `restore_command` pgbackrest just wrote so it's confirmable/stoppable later — this is the one case `restore_command` is actually expected to be in active use.
- The postgres monitor disables `restore_command` as an ongoing backstop whenever it's set on a cohort member or an already-successfully-streaming standby.

Re-enabling `restore_command` for a stuck, non-cohort observer that was previously caught up is deliberately not implemented: a replica lagging enough to need archive catch-up shouldn't still be in the cohort, and there's no existing structured signal for "stuck because the primary lacks the WAL" to trigger on.

### DIFF
--- a/go/cmd/pgctld/command/restore_wrapper.go
+++ b/go/cmd/pgctld/command/restore_wrapper.go
@@ -1,0 +1,111 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package command
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"os/signal"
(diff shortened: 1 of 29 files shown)


rafael (approved):
LGTM

PR #1253, opened 2026-07-09
refactor(multipooler): remove the explicit RestoreFromBackup RPC

dweitzman:
## Summary
- `RestoreFromBackup` (multipooler RPC, multiadmin API, proto, rpcclient) is unreachable in practice: multiadmin's caller never stops postgres or wipes PGDATA first, so the precondition always rejected it.
- The postgres monitor's autonomous restore-on-missing-PGDATA self-heal already covers the real use case, so the explicit RPC is removed end-to-end instead of fixed.
- The monitor's autonomous restore now respects the same `postgresRestartsDisabled` flag the auto-start path already honors, so tests/demos that disable restarts aren't undermined by an autonomous restore kicking in anyway.
- The endtoend restore test now exercises the autonomous self-heal path directly and asserts the restored row count matches the pre-backup snapshot, rather than only logging it.


go/test/endtoend/multipooler/backup_test.go
```diff
@@ -220,7 +254,7 @@ func TestBackup_CreateListAndRestore(t *testing.T) {
 						}
 						restoredTerm = statusResp.ConsensusStatus.GetTermRevocation().GetRevokedBelowTerm()
 						return statusResp.Status.PostgresReady
-					}, 10*time.Second, 100*time.Millisecond, "PostgreSQL should be running after restore")
+					}, 30*time.Second, 100*time.Millisecond, "PostgreSQL should be running after restore")
```
cuongdo: I'm a little worried that 30 seconds is not enough for very slow CI runners.

cuongdo (approved):
Thanks for doing this!

cuongdo:
codecov is complaining. You'll probably want to fix that.

go/test/endtoend/multipooler/backup_test.go
```diff
@@ -145,7 +175,33 @@ func TestBackup_CreateListAndRestore(t *testing.T) {
 						foundBackup.BackupId, foundBackup.Status, foundBackup.StopLsn)
 				})
 
-				t.Run("RestoreAndVerify", func(t *testing.T) {
+				t.Run("RestoreAndVerify_ArchivingOff", func(t *testing.T) {
```
dweitzman, line 178: I split out two tests here, one with WAL archiving on and one with it off

The test without WAL archiving makes sure we load the correct backup: it shouldn't have rows that were added after the backup was taking

The one with WAL archiving on makes sure the WAL restore brings back rows that we added after the backup was taken, without primary conninfo so the rows must have come from the restore command

PR #1255, opened 2026-07-10
feat(multigateway): add SHOW multigres_version command

GuptaManan100:
## Summary

- Adds a `SHOW multigres_version` SQL command so a client connected to multigateway can read the running build's version without shell access to run `--version` — handy for bug reports.
- The gateway answers it locally and never forwards it to postgres, in both the simple and extended (prepared-statement) query protocols.
- The version string is derived from the existing build snapshot in `servenv` (VCS revision, modified flag, commit date, Go version); no new source of truth is introduced.

## Description

`SHOW multigres_version` mirrors `SHOW server_version` and is intercepted in the existing `planVariableShowStmt` path, so it reuses the gateway's synthetic-result machinery. It is intentionally **not** added to the `gatewayManagedVariables` registry, which mandates SET/RESET semantics that don't fit a read-only value.

The extended protocol needed two extra touches beyond the simple path:

- **Describe** forwards straight to the backend with no planning, and postgres has no `multigres_version` GUC to describe — so `executor.Describe` now answers it locally with a synthetic `StatementDescription` (empty parameters, one text column).
- **Execute** must not re-send `RowDescription`: the wire callback emits it whenever `result.Fields` is set, so `PortalStreamExecute` attaches column metadata only when a `Describe('P')` was folded into the Execute (`includeDescribe`). Otherwise it would send an illegal second `RowDescription` mid-Execute.

A new exported `servenv.AppVersion()` formats the build snapshot, with a `version` seam reserved for a future ldflags-stamped release tag.

Covered by unit tests (servenv formatting, the primitive's protocol contract, planner routing) and an end-to-end test exercising both protocols against a real cluster.

Note: `MultigresVersionVariable` lives in `go/common/constants`. In a linked git worktree Go does not stamp VCS info, so local worktree dev builds show `(unknown revision)`; normal checkouts and CI stamp correctly — this affects the whole existing version stack, not just this command.

### DIFF
--- a/docs/query_serving/gateway_managed_variables.md
+++ b/docs/query_serving/gateway_managed_variables.md
@@ -16,10 +16,11 @@ Both are registered in the single source of truth,
 `gatewayManagedVariables` in `handler/gateway_managed_variables.go` (see
 [Adding or changing a GMV](#adding-or-changing-a-gmv)).
 
-> **Not a GMV:** `SHOW multigres_version` is also answered by the gateway, but
-> it is a read-only pseudo-variable, not a GMV — it has no `SET`/`RESET`/
-> `set_config`/transaction behavior and is not in this registry. See
-> [`multigres_version.md`](./multigres_version.md). Don't add it here.
+> **Not a GMV:** `SHOW multigres_version` (and `SELECT multigres.version()`) is
+> also answered by the gateway, but it is a read-only pseudo-variable, not a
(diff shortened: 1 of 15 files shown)


haritabh17 (approved):
Lets have the GUC named like `multigres.server_version` to match PG's `server_version` . Rest all good 🚀

PR #1256, opened 2026-07-10
feat(multipooler): record a recruit-position floor before pg_rewind

dweitzman:
- `pg_rewind` rewinds to the last shared checkpoint, not the last common WAL position, so it could silently discard acknowledged transactions that are supposed to be written to local storage. Before a rewind, record the pooler's pre-rewind position as a durable floor.
- Surface the floor on `ConsensusStatus` (`recruit_blocked_until`), omitted once the pooler catches back up — no explicit clearing needed.
- Enforce the floor in `ValidateRevocation`, the predicate shared by both the real `Recruit()` RPC and the coordinator's pre-flight "check" path, so both reject a pooler that hasn't caught up. Also excludes such a pooler from cohort-addition candidacy.
- Persist the floor to disk alongside the term revocation in a single `ConsensusPromises` file, so a process restart doesn't silently drop it.


rafael (approved):
LGTM

PR #1257, opened 2026-07-12
feat(scram): negotiate SCRAM channel binding to postgres

Snehil-Shah:
Follow up to #986, for the pooler-to-Postgres leg.

If PG offers `SCRAM-SHA-256-PLUS` mechanism, it can now negotiate the same using server's TLS cert. This can be controlled by a new libpq-style `channel_binding` knob (`disable`/`prefer`/`require`), exposed as the `--pg-client-channel-binding` flag in the multipooler CLI.

### DIFF
--- a/go/services/multipooler/internal/connpoolmanager/config.go
+++ b/go/services/multipooler/internal/connpoolmanager/config.go
@@ -539,7 +539,11 @@ func (c *Config) PgChannelBinding() (client.ChannelBindingMode, error) {
 // (which cannot be satisfied without TLS) is validated in that case.
 func (c *Config) ValidatePGSSL(host string) error {
 	if host == "" {
-		if cb, err := c.PgChannelBinding(); err == nil && cb == client.ChannelBindingRequire {
+		cb, err := c.PgChannelBinding()
+		if err != nil {
+			return fmt.Errorf("--pg-client-channel-binding: %w", err)
+		}
+		if cb == client.ChannelBindingRequire {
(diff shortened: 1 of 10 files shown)


go/services/multipooler/internal/connpoolmanager/config.go
```diff
@@ -510,16 +522,26 @@ func (c *Config) PgSSLNegotiation() (client.SSLNegotiation, error) {
...
 //
 // host is the address the multipooler will dial postgres on (only used to
 // validate verify-full). Pass an empty host when the multipooler is configured
-// for a Unix socket; this function only runs the SSL validation when host is
-// non-empty.
+// for a Unix socket; a socket never runs TLS, so only channel_binding=require
+// (which cannot be satisfied without TLS) is validated in that case.
 func (c *Config) ValidatePGSSL(host string) error {
 	if host == "" {
+		if cb, err := c.PgChannelBinding(); err == nil && cb == client.ChannelBindingRequire {
```
haritabh17: Will an invalid channel-binding value (e.g., "bogus") be silently accepted during startup validation?? Considering `if cb, err := c.PgChannelBinding(); err == nil && cb == ... ` when the host is empty, so the operator gets no feedback about their misconfiguration.

Snehil-Shah: Thanks for the catch, fixed in https://github.com/multigres/multigres/pull/1257/changes/a292cbe4358696bb07da0c932b0164eed9e11d3c.

PR #1258, opened 2026-07-13
feat(multigateway): return gateway value for current_setting() of GMVs

GuptaManan100:
## Summary

- `current_setting('statement_timeout')` (and `idle_session_timeout`) now returns the gateway-owned value, agreeing with `SHOW` — matching native PostgreSQL, where the two always agree.
- Gateway-managed variables (GMVs) are never applied to pooled backends, so the call evaluated on the backend saw the backend default and diverged. The gateway now rewrites the call out of the query and fills it from gateway state at execute time, mirroring the existing `set_config` rewrite.
- Detection rides on the pass `analyzeStatement` already does, so the common case (no such call) costs no extra tree walk.
- Adds a small reusable `ast.MaxParamRef` helper.

## Problem

After the recent `statement_timeout` fix, `SHOW statement_timeout` reflected the value correctly but `current_setting('statement_timeout', true)` did not. GMVs are owned by the gateway and stripped from backend connections, so any `current_setting('<gmv>')` that runs on a pooled backend observes that backend's local GUC (the default), not the client's session value.

## Solution

Rewrite gateway-managed `current_setting('<gmv>'[, missing_ok])` calls whose name is a literal, then resolve the value at execute time from gateway connection state (the same string `SHOW` returns):

- **Normalizer** keeps `current_setting`'s literal name (like `set_config`) so the planner can recognize a GMV at plan time.
- **`analyzeStatement`** sets `NeedsCurrentSettingRewrite` on its existing FuncCall walk; the mutating rewrite runs only when that flag is set — no second walk in the common case.
- **`GatewayManagedValueRoute`** gained a `reads` list: each rewritten call becomes a synthetic bind slot filled from `ShowGatewayManaged` at execute time (simple + extended protocol). Resolving per-execution keeps the plan cacheable across value changes. Repeated reads of the same variable share one slot (one gateway lookup); the output column keeps its `current_setting` name.
- **Scope** is statements that *evaluate* the call for a result: `SELECT` (incl. `SELECT INTO`), `INSERT`/`UPDATE`/`DELETE`, and `CREATE TABLE AS`. `CREATE VIEW` and `CREATE MATERIALIZED VIEW` are excluded — they *store* a re-evaluable/refreshable query, so rewriting would freeze the value. Materialized view shares the `CREATE TABLE AS` AST node and is caught by its object type.

A **bound** variable name (`current_setting($1)`) is not recognized at plan time and routes to the backend, the same documented gap as a bound `set_config` name.

## Tests

- Planner + engine unit tests for the rewrite, slot reuse, column-name preservation, and the extended-protocol portal decode.
- `TestCurrentSettingGatewayManaged` (dual-target vs native PG): simple + extended protocol and a `SET LOCAL` override.
- `TestCurrentSettingMaterialized`: `CREATE TABLE AS` / `SELECT INTO` store the gateway value and keep the column name.
- The leak-detection e2e tests move from `current_setting` (now spoofed) to `SELECT setting FROM pg_settings WHERE name = '<gmv>'`, which reads the real backend GUC and is never rewritten — so they still genuinely detect a leaked GUC.


haritabh17 (approved):
Looks good

PR #1260, opened 2026-07-13
fix(multipooler): stop unit tests racing the background postgres monitor

dweitzman:
## Summary
- `openLocked` unconditionally starts the postgres monitor loop (`pm.pgMonitor.StartWithOptions`), which polls `SELECT pg_is_in_recovery()` via `postgresMode()`.
- Every RPC/consensus unit test in `go/services/multipooler/internal/manager` calls `Start()` against a strict, order-sensitive mock query service. The background poll can occasionally consume a `consumeOnce` mock pattern meant for the RPC handler's own synchronous call, producing `no matching query pattern for: SELECT pg_is_in_recovery()` — a flake seen across `TestPromote`, `TestRecruit`, and any other RPC path calling `postgresMode()`.
- Fix: abstract `pm.pgMonitor` behind a small `periodicRunner` interface so `NewMultipoolerManagerForTesting` can inject a no-op implementation that never schedules its callback, instead of the real ticking `timer.PeriodicRunner`. Production behavior is unchanged (real constructor path still gets the real ticking runner). Tests that need to exercise monitor iterations already call `monitorPostgresIteration` directly (see `postgres_monitor_test.go`).

### DIFF
--- a/go/services/multipooler/internal/manager/consensus_manager_test.go
+++ b/go/services/multipooler/internal/manager/consensus_manager_test.go
@@ -15,6 +15,7 @@
 package manager
 
 import (
+	"context"
 	"log/slog"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


rafael (approved):
LGTM

PR #1262, opened 2026-07-13
test(multipooler): drop fragile term assertion

dweitzman:
## Summary
- `TestManagerStatus_NodeIdentityAndConsensus` asserted the shared fixture's consensus term equals exactly 1, but term only ever increases and is never reset between tests sharing that fixture — several other tests bump it, and Go doesn't guarantee this test runs before them.
- Term revocation is also an internal consensus-protocol detail. Dropped the assertion entirely: the very next line already checks that the committed rule names this pooler as leader (`consensus.SelfConsensusRole`), which is the externally-meaningful proof that bootstrap completed.


rafael (approved):
LGTM

PR #1259, opened 2026-07-13
fix(pgregress): close query-serving isolation gaps

haritabh17:
## Summary

- preserve PostgreSQL transaction-time PREPARE validation and relation-lock semantics
- propagate SQL prepared-statement session metadata and keep same-query statement plan caches independent
- avoid early SERIALIZABLE snapshots for in-transaction SET and preserve deferred BEGIN options
- improve isolation safe-snapshot detection and verify accepted isolation output divergences
- update the pg_regress image to the Go toolchain required by `go.mod`

## Validation

- `go test -short -count=1 ./go/common/preparedstatement ./go/services/multigateway/engine ./go/services/multigateway/handler ./go/services/multigateway/planner ./go/services/multigateway/executor ./go/services/multipooler/internal/executor ./go/test/endtoend/pgregresstest ./go/test/endtoend/queryserving`
- commit-time lint: 0 issues

## Status

Draft for CI and a full extended pg_regress/isolation verify run.

Refs MUL-665

PR #1265, opened 2026-07-14
fix(query-serving): preserve transactional SET and notification order

haritabh17:
## Summary

- defer backend transaction startup for `SET`, retaining the pending `BEGIN` so SERIALIZABLE snapshots begin at the first data statement
- serialize asynchronous notifications through one pgwire writer and coordinate forwarding shutdown to preserve FIFO delivery
- include only the associated pg_regress/isolation harness updates and accepted-output patches

This is a scoped extraction from #1259. It excludes the prepared-statement, relation-lock, broader snapshot, and container/toolchain changes from that PR.

## Validation

- `go test -short -count=1 ./go/services/multigateway/...`
- `go test -short -count=1 ./go/common/pgprotocol/server ./go/services/multigateway/handler`
- `go test -race -count=20 -run 'TestAsyncNotificationFlushPreservesFIFO|TestStopAsyncNotificationsDrainsAndWaits' ./go/common/pgprotocol/server`
- `go test -short -count=1 ./go/test/endtoend/pgregresstest`
- `make build`
- `MULTIGRES_PORT_POOL_ADDR=/tmp/multigres-port-pool.sock go test -count=1 -run '^(TestListenNotify|TestMultigateway_SessionSettings)$' ./go/test/endtoend/queryserving/...`

## Status

Draft for CI and the full compatibility rerun.

Refs MUL-665


go/services/multigateway/handler/transaction_helpers.go
```diff
@@ -242,11 +253,14 @@ func (h *MultigatewayHandler) executeWithImplicitTransaction(
...
-		// options cannot incorrectly "adopt" a transaction that already ran work.
-		state.PendingBeginQuery = ""
+		// A backend statement has now been attempted in the transaction. In production,
+		// ScatterConn consumes PendingBeginQuery when it starts the backend transaction;
+		// mirror that here for mocked executors. Gateway-local statements (e.g.
+		// statement_timeout) must not clear it, or the first real backend statement
+		// loses BEGIN options such as ISOLATION LEVEL SERIALIZABLE.
+		if state.PendingBeginQuery != "" && !runsWithoutBackend(stmt) {
+			state.PendingBeginQuery = ""
+		}
```
GuptaManan100: I don't like this. scatter_conn is already clearing this whenever it is consumed. We should just remove this, and fix the mocking code to consume PendingBeginQUery too

GuptaManan100 (approved):

go/services/multigateway/handler/transaction_helpers.go
haritabh17: Ack, updated.

PR #1266, opened 2026-07-14
fix(multipooler): confirm WAL replay completion before recruit proceeds

mkindahl:
## Problem

During standby revoke (`Recruit`) and the pre-`pg_rewind` position measurement, the node stops its WAL receiver and waits for replay to "settle" before reporting its WAL position. The old heuristic waited for `pg_last_wal_replay_lsn()` to **stop advancing** (3 identical readings ~30 ms apart).

That is unsafe: anything that freezes the replay LSN *below* the received position — a recovery conflict, an I/O stall, `recovery_min_apply_delay` — looks identical to "replay finished" over the stability window. So `Recruit` could proceed before the received WAL was fully replayed and report a **stale position**, which feeds the coordinator's leader selection and the audit-table-derived cohort/term.

## Fix

Replace the stability heuristic with two **positive** completion signals, decided per poll now that the receiver is stopped and `receive_lsn` is frozen:

1. **`replay_lsn >= receive_lsn`** — replay has applied every received byte. The two LSNs are read by the query and compared numerically in Go via `pgutil.ParseLSN` (their `X/Y` hex text does not sort lexicographically).
2. **Startup process idle at end-of-WAL** — `pg_stat_activity.wait_event` is `RecoveryRetrieveRetryInterval` / `RecoveryWalStream` for `backend_type = 'startup'`, observed on **2 consecutive polls**. That state is reached only after recovery has exhausted every WAL source, so every *complete* received record is applied; any trailing bytes are an incomplete record it cannot replay (the normal way streamed WAL ends, where `replay_lsn` legitimately settles below `receive_lsn`).

Every other observed state — actively applying, an I/O or recovery-conflict stall, an apply delay, a `NULL`/uninstrumented wait, or an unrecognized wait — is treated as **"not done, keep waiting"**. A genuine stall therefore rides out the 10 s deadline and surfaces as a `DEADLINE_EXCEEDED` error instead of a premature, unsafe "done".

## Preconditions

Signal 2 is only sound if the startup process cannot leave the end-of-WAL wait by obtaining more WAL. Two sources could do that, and `checkNoWALSource` requires **both** to be unset up front, failing loudly otherwise:

1. **`primary_conninfo`** — a WAL receiver could stream more. Both callers stop the receiver first; the check confirms it is empty.
2. **`restore_command`** — recovery could fetch more segments from the archive. The check confirms it is empty.

Rejecting `restore_command` upholds a consensus invariant, not just signal-2 timing: a **cohort member** (an active consensus participant) must advance only by streaming from the current leader, never by replaying a possibly-foreign archive. Left enabled, a subsequent restart-as-standby can resolve `recovery_target_timeline=latest` through the archive to a timeline this node's own checkpoint doesn't descend from and `FATAL` at startup. So `restore_command` is now cleared on **every** path that makes a node a cohort member:

- **Recruit** and the cohort-member **`SetPrimary`** clear it (pre-existing).
- **`restartAsStandbyLocked`** (the `pg_rewind` path) clears it before the replay-completion wait, and strips it from the `postgresql.auto.conf` that `pg_rewind` copies from the source.
- **Promotion** clears it on the new primary — the root-cause reset. `restore_command` is set implicitly by pgBackRest's initial `restore --type=standby` and was never cleared on promotion, so a promoted primary carried it in its `auto.conf` and `pg_rewind` copied it back onto every rejoining follower. A clean primary `auto.conf` stops that at the source.

## Changes

- Rename `waitForReplayStabilize` → `waitForReplayComplete`; remove the LSN-stability completion heuristic.
- Add `queryReplayProgress` returning a `replayProgress` struct — reads the replay/receive LSNs, pause state, and startup wait state in a single query. Zero rows means no startup process (not in recovery) and is rejected; a NULL replay LSN is guarded defensively.
- Add `checkNoWALSource` (enforces `primary_conninfo` **and** `restore_command` empty) and the end-of-WAL wait-event recognizer.
- Clear `restore_command` on the `pg_rewind` path (`restartAsStandbyLocked`, including stripping it from the copied `auto.conf`) and on promotion.
- Move the trace span from `Recruit` into `waitForReplayComplete` (`consensus/replay_completion_wait`) so both the recruit and rewind callers are instrumented.
- Extend the HA decision log with the positive-signal design, the wait-event taxonomy and version-drift notes (names are stable across PG 13+; unknown states bias toward "keep waiting"), and a section on what can move the startup process out of `RecoveryRetrieveRetryInterval`.

## Testing

- **Unit** (`TestWaitForReplayComplete`): caught-up, end-of-WAL-below-receive, I/O stall → timeout, recovery-conflict → timeout, NULL-receive/NULL-wait, paused → error, `primary_conninfo` set → error, `restore_command` set → error, not-in-recovery → error. Promote-path mock asserts the post-promotion `restore_command` reset.
- **End-to-end** against real PostgreSQL 18:
  - `multipooler/backupfaults` `TestPrimaryCrashWithUnarchivedWAL_NoDataLoss` — the load-bearing check. `RestoreFromBackup` genuinely **sets** `restore_command` on the freshly provisioned pooler; a new `verifyRestoreCommandCleared` then asserts every cohort member (including that restored node, the `pg_rewind`-demoted old primary, and the promoted new primary) has it **cleared** once the cluster settles.
  - `multiorch` `TestRewindDivergedReplica` (rewind) and `TestDeadPrimaryRecovery` (promote) — assert the rewound / promoted node reports an empty `restore_command`.

## Notes for reviewers

- A recovery-conflict stall is bounded by `max_standby_streaming_delay` (30 s default), which exceeds the 10 s wait, so such a stall fails `Recruit` and is retried rather than waited out. If these prove common, neutralizing conflicts at the source (stopping read serving before the wait) would be a follow-up.
- The `pg_rewind`-path `auto.conf` strip is defense-in-depth alongside the promote-side reset: it still cleans up a source primary whose `auto.conf` predates this fix. `auto.conf` is stripped surgically (only `restore_command`) rather than deleted, because it also carries `archive_command`/`archive_mode` (path-fixed by `fixPgBackRestPaths`) and other `ALTER SYSTEM` settings that must survive the rewind.

### DIFF
--- a/docs/ha/decision-log/2026-02-12-wait-for-replay-stabilize-during-revoke.md
+++ b/docs/ha/decision-log/2026-02-12-wait-for-replay-stabilize-during-revoke.md
@@ -29,43 +29,151 @@ This matters during revoke because we need to query the audit table and rely on 
 
 ## Decision
 
-After disconnecting the WAL receiver during standby revoke, poll `pg_last_wal_replay_lsn()` until it stabilizes (same value for 3 consecutive polls at 10ms intervals), then report the final position. The full sequence is:
+**Updated:** 2026-07-21
 
-1. Reset `primary_conninfo` and reload config (breaks the connection to the old primary)
+After disconnecting the WAL receiver during standby revoke, wait until replay has applied all received WAL, then report the final position. The full sequence is:
+
+1. Reset `primary_conninfo` and `restore_command` and reload config (breaks the connection to the old primary)
 2. Wait for the WAL receiver process to fully disconnect (`pg_stat_wal_receiver` shows no rows)
-3. Poll `pg_last_wal_replay_lsn()` until it stops advancing (3 consecutive identical readings, approximately 30ms of stability)
+3. Poll until replay has applied all available WAL, decided by two positive signals only (see the update below).
 4. Query and report the final replication status
 
-The overall timeout for the stabilization wait is 10 seconds.
+The overall timeout for the wait is 10 seconds.
 
-This is a heuristic, not a guarantee. A stable reading can also occur if replay is stalled. We have not found a better approach in PostgreSQL. Alternatives considered:
+### Update: positive-signal completion (`waitForReplayComplete`)
 
-- **Wait for receive LSN to equal replay LSN:** does not work because they track different units (byte position vs record boundary) and may never match exactly.
+The original implementation waited for `pg_last_wal_replay_lsn()` to stop
+advancing (3 identical readings ≈30ms apart). Reported evidence showed this can
+declare completion prematurely: anything that freezes the replay LSN below the
+received position — a recovery conflict, an I/O stall, `recovery_min_apply_delay`
+— looks identical to "replay finished" over the stability window, so `Recruit()`
+could proceed before WAL was fully replayed and report a stale position.
+
+The **stability heuristic was removed as a completion criterion.** Completion is
+now decided by two positive signals, and every other state means "not done, keep
+waiting" until the deadline:
+
+1. **`pg_last_wal_replay_lsn() >= pg_last_wal_receive_lsn()`** — replay has applied
+   every received byte. The receiver is already stopped so `receive_lsn` is
+   frozen; the comparison is done in SQL for a single consistent snapshot and
+   PostgreSQL's native `pg_lsn` ordering (LSN strings do not compare
+   lexicographically).
+
+2. **The startup (recovery) process is parked in an end-of-WAL wait** —
+   `wait_event` is `RecoveryRetrieveRetryInterval` or `RecoveryWalStream` in
+   `pg_stat_activity` for `backend_type = 'startup'`. This means recovery looked
+   for more WAL and found none, so every _complete_ received record has been
+   applied. This is necessary because streamed WAL normally ends in an _incomplete_
+   record (a partial cross-page record): `pg_last_wal_receive_lsn()` counts those
+   trailing bytes but replay stops at the last complete record, so `replay_lsn`
+   legitimately settles below `receive_lsn`. Signal 1 alone would never fire in
+   that (common) case. To avoid acting on a transient sample, the end-of-WAL wait
+   must be observed on 2 consecutive polls (the startup process passes briefly
+   through I/O / running states while cycling WAL sources).
+
+Any other observed state — replay actively applying, a stall on I/O
+(`DataFileRead` / `WalRead` / `WalSync`), a recovery conflict (`Lock` /
+`BufferPin` / `RecoveryConflict*`), an apply delay, a `NULL`/uninstrumented wait,
+or an unrecognized wait — is treated as "not done". A genuine stall therefore
+rides out the 10s deadline and surfaces as a `DEADLINE_EXCEEDED` Recruit error
+(the safe, loud failure) instead of a premature, unsafe "done".
+
+Alternatives considered and rejected:
+
+- **Stop when the replay LSN stops advancing (the original heuristic):** a stalled
+  I/O or a recovery conflict freezes the LSN below `receive_lsn` while WAL is still
+  pending, and is indistinguishable from completion. This is the bug this update fixes.
+- **Require exact `replay_lsn == receive_lsn`:** rejected — they track different
+  units (byte position vs record boundary), so with a trailing incomplete record
+  they never match and Recruit would always time out.
+
+Version note: the recovery wait-event names are stable across PostgreSQL 13+
+(well below the multigres floor). Before 13 they were renamed and `RecoveryWalStream`
+meant something different, and in a future major the `BufferPin` wait-event _class_
+is renamed to `Buffer`; the completion logic keys on a small allow-list of
+end-of-WAL events and biases every unrecognized state toward "keep waiting", so
+name drift can only cost latency (up to the deadline), never a premature "done".
+
+### What can move the startup process out of `RecoveryRetrieveRetryInterval`
+
+Signal 2's soundness rests on the fact that, in our context, once the startup
+process reaches the end-of-WAL retry wait it cannot resume applying WAL. That wait
+is a `WaitLatch` (`xlogrecovery.c`, `WaitForWALToBecomeAvailable`) that returns on:
+
+- **timeout** — every `wal_retrieve_retry_interval` (5s default);
+- **latch set** (`WakeupRecovery()`) — a running walreceiver started/flushed WAL,
+  or a `SIGHUP` (config reload) / `SIGUSR2` (promote) / `SIGTERM` (shutdown) /
+  `pg_wal_replay_pause()`;
+- **postmaster death**.
+
+Crucially, waking is not the same as resuming replay: after any wake, recovery
+just retries the sources (archive → pg_wal → stream) and re-enters the wait unless
+a source now actually has the next record. Replay therefore resumes only via one
+of two paths:
+
+1. **Streaming (re)starts** — a walreceiver connects and delivers WAL. Requires
+   `primary_conninfo` to be set.
+2. **The archive grows** — `restore_command` succeeds on a retry because a new
+   segment appeared. Requires a live primary still archiving to the repo.
+
+Both callers stop the receiver before the wait, so path 1 is closed — and
+`waitForReplayComplete` additionally verifies `primary_conninfo` is empty
+(`checkNoWALSource`) as a defensive guard. Path 2 is closed in the failover and
+rewind scenarios because the source primary is dead or demoted, leaving a static
+archive; a set `restore_command` is therefore harmless there (recovery replays the
+archive to exhaustion and then parks). This is why the recruit path clears
+`restore_command` for a _consensus_ reason (a cohort member must not chase a
+possibly-foreign archive) rather than for signal-2 timing, and why the rewind path
+may legitimately run with `restore_command` set.
+
+Residual limitation: if the archive were _growing_ (a live primary archiving to the
+same repo) while `restore_command` is set, a new segment could be fetched between
+two polls and replay would resume, making signal 2 premature. This does not arise
+in the failover/rewind scenarios (dead/demoted primary), but it is the reason the
+recruit path does not rely on it.
 
 ## Consensus Implications and What Could Go Wrong
 
-In the worst case, replay is stalled but the standby has already received WAL that includes commits affecting the audit table. In that situation, the standby may report stale cohort or primary term information.
+With positive-signal completion (see the update above), neither a stalled
+replay nor a trailing incomplete record can be mistaken for completion: the former
+keeps us waiting until the deadline (then errors), and the latter is recognized
+explicitly via the end-of-WAL wait event. The residual failure modes are now:
 
-- A stale cohort could yield an election that assumes the wrong quorum and triggers failover without discovering a more advanced replica.
-- A stale primary term could prevent multiorch from detecting conflicting timelines across multiple failovers.
+- **A false end-of-WAL reading.** If the startup process reported an end-of-WAL
+  wait on 2 consecutive polls but WAL was in fact still pending, we could report a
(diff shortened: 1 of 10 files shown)


go/services/multipooler/internal/manager/pg_replication.go
```diff
@@ -544,42 +544,75 @@ func (pm *MultipoolerManager) stopRestoreCommand(ctx context.Context) error {
...
+//
+// See decision: 2026-02-12-wait-for-replay-stabilize-during-revoke.md.
+func (pm *MultipoolerManager) waitForReplayComplete(ctx context.Context) (*multipoolermanagerdatapb.StandbyReplicationStatus, error) {
 	waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
 
-	ticker := time.NewTicker(10 * time.Millisecond)
+	// Poll every 100ms: frequent enough to keep revoke latency low, but not so
+	// tight that we hammer postgres while waiting out a slow replay.
+	ticker := time.NewTicker(100 * time.Millisecond)
```
mkindahl, line 610: This is still a very short timeout for the work. We might consider increasing it to at least something like 500ms. I increased it to 100 ms since 10ms seemed unnecessary.

docs/ha/decision-log/2026-02-12-wait-for-replay-stabilize-during-revoke.md
```diff
@@ -29,43 +29,151 @@ This matters during revoke because we need to query the audit table and rely on
 
 ## Decision
```
dweitzman, line 30: I don't have strong feelings about this, but I'm not sure in this decision log model how we want to deal with decisions that have changed.

I wonder if the date in the filename should be updated?... cc @rafael for any thoughts

go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -168,11 +168,11 @@ func (pm *MultipoolerManager) Recruit(ctx context.Context, req *consensusdatapb.
 
 	if !pgMode.OutOfRecovery() {
 		stabilizeCtx, stabilizeSpan := telemetry.Tracer().Start(ctx, "consensus/stabilize")
-		_, err = pm.waitForReplayStabilize(stabilizeCtx)
+		_, err = pm.waitForReplayComplete(stabilizeCtx)
```
dweitzman: Nits:
- I wonder if the trace span here should be capture within `waitForReplayComplete()` instead of around it?
- The span name is `stabilize` but the method is called `waitForReplayComplete()` now. I wonder if they should have matching naming, like `consensus/replay_complete`, etc.?

go/services/multipooler/internal/manager/pg_replication.go
```diff
@@ -544,91 +545,247 @@ func (pm *MultipoolerManager) stopRestoreCommand(ctx context.Context) error {
 	return nil
 }
 
-// waitForReplayStabilize waits, best effort, for WAL replay to stop making
-// observable progress. The intent is to approximate replay is idle given the WAL
-// that is currently available to this standby.
+// waitForReplayComplete waits until WAL replay has provably applied every WAL
```
dweitzman: Nits:
- "provably" feels strong here?
- maybe instead of "every WAL record available to this standby" the wording could be more clear that it's every complete WAL entry, ignoring partial WAL entries that were received but can't be applied?

go/services/multipooler/internal/manager/pg_replication.go
```diff
@@ -544,91 +545,247 @@ func (pm *MultipoolerManager) stopRestoreCommand(ctx context.Context) error {
...
+// is static (the source primary is dead/demoted in the scenarios this runs in).
+// See the decision log, "What can move the startup process out of
+// RecoveryRetrieveRetryInterval". The recruit path additionally clears
+// restore_command, but for a consensus reason (a cohort member must not replay a
+// possibly-foreign archive), not for signal-2 timing.
+func (pm *MultipoolerManager) checkNoWALSource(ctx context.Context) error {
 	queryCtx, cancel := context.WithTimeout(ctx, 500*time.Millisecond)
 	defer cancel()
-	result, err := pm.query(queryCtx, "SELECT pg_last_wal_replay_lsn(), pg_is_wal_replay_paused()")
+	result, err := pm.query(queryCtx, "SELECT current_setting('primary_conninfo', true)")
```
dweitzman: Should this method also check that restore_command isn't set?

go/services/multipooler/internal/manager/pg_replication.go
```diff
@@ -544,91 +545,247 @@ func (pm *MultipoolerManager) stopRestoreCommand(ctx context.Context) error {
...
-func (pm *MultipoolerManager) queryReplayState(ctx context.Context) (replayLsn string, isPaused bool, err error) {
+// isEndOfWalWait reports whether the given startup-process wait_event means the
+// recovery process is idle waiting for WAL that will not arrive (the receiver is
+// stopped), i.e. it has applied every complete record it has.
+//
+// These names are stable across PostgreSQL 13+ (the multigres floor is well
+// above that). Before 13 the recovery wait events were renamed and the string
+// "RecoveryWalStream" meant something else, so this must not be relied upon on
+// older majors. See the WAL-replay-completion decision log.
+func isEndOfWalWait(waitEvent string) bool {
```
dweitzman: Here's an idea for maybe making this a little easier for people without context to read, but feel free to ignore if you disagree:

```
// WaitEvent is a wait event value from pg_stat_wal.
// See https://www.postgresql.org/docs/current/monitoring-stats.html#WAIT-EVENT-ACTIVITY-TABLE for
// possible values.
type waitEvent string

func isNoMoreWalToReplay(evt waitEvent) bool {
  return evt == "RecoveryRetrieveRetryInterval" || evt == "RecoveryWalStream"
} 
```

go/services/multipooler/internal/manager/rpc_consensus_test.go
```diff
@@ -1231,3 +1240,150 @@ func TestSetResignedLeaderAtTerm_BroadcastsOnChange(t *testing.T) {
...
+		m.AddQueryPattern(`^SELECT current_setting\('primary_conninfo', true\)`, mock.MakeQueryResult(
+			[]string{"primary_conninfo"}, [][]any{{""}}))
+		m.AddQueryPattern("pg_stat_activity", mock.MakeQueryResult(progressCols, [][]any{{"0/4000000", "0/4000000", true, nil, nil}}))
+
+		_, err := pm.waitForReplayComplete(t.Context())
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "paused")
+	})
+
+	// Note: restore_command tolerance (the pg_rewind catch-up path) is not unit
```
dweitzman: This comment seems to be referencing in intermediate state of the PR related to my earlier comment. Seems like there's a reason `checkNoWALSource` isn't looking at restore_command?

dweitzman (approved):

docs/ha/decision-log/2026-02-12-wait-for-replay-stabilize-during-revoke.md
mkindahl: No strong opinion either, I'm game for anything.

It is quite straightforward to use `git blame` and `git show` to find out where things changed though. It is reasonable to change the documentation in a PR though to ensure that it is correct and reflect what is in the code.

go/services/multipooler/internal/manager/pg_replication.go
mkindahl: There are some discussions later in the comment about that, but I clarified the initial lines to make it clear that only complete records are applied.

go/services/multipooler/internal/manager/pg_replication.go
mkindahl: No opinion on names, so used this name.

go/services/multipooler/internal/manager/pg_replication.go
mkindahl: Yes, this is what I initially did but we cannot do this since this is called indirectly from two locations:

1. The rewind path. Here `restore_command` is could be set.
2. The recruit path. Here `restore_command` is reset (not set to any value).

For the recruit path, having restore command clear is proper, but for the rewind path, we need it since we can have WAL on the primary that has been archived and we need to read it from the archive.

The steps are roughly this:

1. For some reason, a `pg_rewind` is needed: a new node was added and as a standby, the current primary was demoted, etc.
2. `pg_rewind` is executed, which usually results in the standby being behind the primary (it can be caught up exactly, but not more). This means that there is WAL that needs to be processed, potentially WAL that is not present on the primary.
3. Since there potentially is WAL that needs to be processed, we need to set `restore_command` so that WAL is read from the shared archive if it is not found in the primary.
4. We start the server and wait for it to catch up using `waitForReplayComplete`, which calls `checkNoWALSource`. Not allowing `restore_command` here would not work.
5. Once it has caught up, we can set `primary_conninfo` and start streaming from the primary, which (should) also reset `restore_command`.

It is still sound since the state is set to `RecoveryRetrieveRetryInterval` only when `restore_command` fails to fetch any more WAL from the archive.

There is a potential problem here that we cannot do anything about though: if the command fails for any _other_ reason than that the WAL segment does not exist, it will be treated as if the WAL segment does not exist. We need to deal with this, but it is out of scope for this fix.

go/services/multipooler/internal/manager/rpc_consensus.go
mkindahl: Seems reasonable. Fixed. (I used `consensus/replay_completion_wait` though.)

go/services/multipooler/internal/manager/rpc_consensus_test.go
mkindahl: Yes, it was something I tested initially since I reasoned the same way you did. See above. Updated the comment though.

go/services/multipooler/internal/manager/pg_replication.go
mkindahl: We discussed adding a check that `restore_command` was reset at this point, but this creates problems when you restore from backup in the following scenario:

1. A node has no `PGDATA` (fresh/rebuilt/lost data). Monitor picks `remedialActionRestoreFromBackup`.
2. `restoreFromBackupLocked` runs `pgbackrest restore --type=standby`, which writes `restore_command` into `postgresql.auto.conf`, and Postgres is (re)started in standby/recovery mode.
3. The node then catches up by replaying WAL, including from the archive via `restore_command`.
4. In a later monitor cycle (or via `SetPrimary` / `RewindToSource`), the node is told to follow the current leader and divergence is suspected and restartAsStandbyLocked is called.
6. `restartAsStandbyLocked` sees the node is still `!OutOfRecovery()` and calls `waitForReplayComplete` — with `restore_command` still set from step 2 since nothing on this self-heal path has cleared it.

The problems are multiple here:

- Ideally, a member of the cohort should not have `restore_command` set (as we discussed) since it is either the primary, hence do not need the `restore_command`, or a standby and should keep up to date only by streaming.
- When a server is promoted, `restore_command` is not reset, so it leaves the setting in `postgresql.auto.conf`. This needs to be fixed.
- `pg_rewind` copies the setting from the primary through the file `postgresql.auto.conf`, which means that the standby has `restore_command` set. This also needs to be fixed.

mkindahl: Fixed the above issues. Now restore_command should be reset when the pooler becomes a member of the cohort either as a primary or as a standby.

go/services/multipooler/internal/manager/rpc_consensus_test.go
mkindahl: This was removed as outlined above.

docs/ha/decision-log/2026-02-12-wait-for-replay-stabilize-during-revoke.md
mkindahl: Added a "Update: <date>" as agreed.

PR #1269, opened 2026-07-15
fix(multipooler): preserve backend vpid table during promotion

haritabh17:
## Summary

- preserve `multigres.backend_vpid` during the asynchronous post-promotion unlogged-table sweep
- continue dropping user-created unlogged tables asynchronously
- verify the sidecar table is not dropped while preserving dependency-blocked drop coverage

## Testing

- `/mt-dev unit ./go/services/multipooler/internal/manager TestPromoteDropsUnloggedTables`
- `/mt-dev unit ./go/services/multipooler/internal/manager/...`


shedges (approved):
LGTM

GuptaManan100 (approved):

PR #1270, opened 2026-07-15
feat(query-serving): opaque row passthrough

shedges:
## What

Opaque row passthrough on the gateway-to-pooler execution path. The multipooler carries raw PostgreSQL DataRow frames straight to the client instead of parsing each row into columns and rebuilding and re-serializing it on the multigateway. Default-on, covering regular and reserved connections. Gateway primitives that read result rows themselves (for example resolve_set_config) opt out via PlanExecInfo.KeepStructured.

The second commit adds the benchmarking used to justify and measure it: a result-row micro-benchmark (go/test/microbench/serialization), a large-result mode for the pgbench harness, and a slow-client backpressure test.

## Why

Large-result queries are bound by result-row deserialization and GC in the proxy layer, with the gateway spending far more CPU than Postgres itself. Opaque removes that path: no per-row ToProto or RowFromProto, no per-row protobuf framing, and near-constant per-row allocations. Micro-benchmark: at 10000 rows the columnar path does ~80053 allocations versus 7 for opaque.

## Impact

Large-result A/B (10000-row, about 1 MB result, extended protocol, off vs on), Postgres max_connections=1500:

| clients | TPS off | TPS on | dTPS | avg off ms | avg on ms | P99 off ms | P99 on ms |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 1    | 196 | 310  | +58% | 5.08    | 3.21    | 6.27    | 4.11    |
| 10   | 581 | 842  | +45% | 17.16   | 11.85   | 30.00   | 18.74   |
| 100  | 646 | 1001 | +55% | 152.76  | 98.46   | 260.37  | 187.54  |
| 1000 | 583 | 812  | +39% | 1442.00 | 1032.33 | 2446.74 | 1801.16 |

The win holds across the whole range: throughput +39 to +58 percent, average latency -28 to -37 percent, P99 -26 to -37 percent. Gateway CPU is down about 57 percent at 10 clients; profiles confirm writeRawDataBlock replaces the per-row decode path and ToProto, RowFromProto, and the per-row protobuf framing disappear. Tiny-row OLTP is neutral (that regime is syscall and scheduler bound, not row-path bound), so no regression.

Reaching 1000 clients cleanly needed one benchmark fix (unrelated to opaque): pgbench ran with its default single worker thread, which could not drain 1000 one-MB results fast enough, so data backed up into the proxy. Driving pgbench with -j = min(clients, NumCPU) fixes the client-side drain.

## Correctness

- Full top-level queryserving integration suite green with opaque default-on.
- pgparity (Multigres output compared against direct Postgres) green.
- Value-asserting tests confirm data survives end to end, not just row counts: the 8 MiB single-value large-result test and SET then SHOW returning the exact value.
- Slow-client backpressure test: one client reading a ~600 MB result slowly grows proxy memory only ~30 MB (streaming plus backpressure holds), not toward the result size.

## Scope and notes

- Unary ExecuteQuery stays structured (no gateway client-passthrough path uses it, and its single-result return would truncate a multi-batch opaque result).
- Opaque is unconditional (no runtime kill-switch); a startup flag defaulting on can be added if wanted.
- Bounding total in-flight buffered memory across many concurrent slow clients (admission or a memory budget) is a separate hardening that applies to the structured path too.


go/common/sqltypes/sqltypes.go
```diff
@@ -183,19 +195,44 @@ type Result struct {
 	Notifications []*Notification
 }
 
+// RowCount returns the number of data rows in the result, accounting for opaque
+// passthrough mode where the rows live in RawData rather than Rows.
+func (r *Result) RowCount() int {
+	if r == nil {
+		return 0
+	}
+	if r.RawData != nil {
```
GuptaManan100: nit: Small consistency thing: this branches on `r.RawData != nil`, while the call sites added elsewhere use `len(result.RawData) > 0`. But proto3 removes the nil-vs-empty distinction, so a non-nil-empty `RawData` would round-trip to nil and silently take the structured branch. Let's just stick to one predicate (`len(...) > 0`) everywhere.

go/common/pgprotocol/client/extended.go
haritabh17, line 442: Does it make sense to refactor `processExecuteResponses` , `processQueryResponses` to use `resultBatcher` - similar to `processBindAndExecuteResponses`, `processPrepareAndExecuteResponses`, `processBindDescribeAndExecuteResponses`

go/common/pgprotocol/client/query.go
```diff
@@ -239,6 +327,9 @@ func (c *Conn) processQueryResponses(ctx context.Context, callback func(ctx cont
 	var currentFields []*query.Field
 	var batchedRows []*sqltypes.Row
 	var batchedSize int
+	// Opaque passthrough accumulators (used only when c.RawRowPassthrough).
+	var rawBlock []byte
+	var rawCount int
```
GuptaManan100: `resultBatcher` was introduced to keep opaque handling in one place, but only 3 of the 5 response loops use it - this one and `processExecuteResponses` still hand-roll the `rawBlock`/`rawCount`/`batchedSize` accumulators inline. Any specific reason for this?

go/services/multipooler/internal/executor/executor.go
haritabh17: Would report zero rows for every passthrough query?

Does it make sense to add a helper like `Result.StructuredRows()` that panics/errors if `RawData != nil`, and have internal consumers (`resolve`, any future reader) call it instead of `.Rows` directly? Or maybe invert the default?

go/test/endtoend/queryserving/benchmarking/pgbench_runner.go
```diff
@@ -155,18 +159,37 @@ func (r *PgBenchRunner) RunScenario(ctx context.Context, t *testing.T, target su
...
+	if n := runtime.NumCPU(); threads > n {
+		threads = n
+	}
+
 	args := []string{
 		"-h", target.Host,
 		"-p", strconv.Itoa(target.Port),
 		"-U", target.User,
 		"-c", strconv.Itoa(scenario.Clients),
+		"-j", strconv.Itoa(threads),
```
haritabh17, line 176: Lets mention this in the skill that is containing details on how to run pgbench.

haritabh17 (approved):
Rest LGTM 🚀

go/services/multigateway/engine/engine.go
```diff
@@ -56,6 +56,13 @@ type PlanExecInfo struct {
 	// per-statement hot path.
 	RecheckAdvisoryLocks bool
 
+	// KeepStructured opts this statement out of opaque row passthrough, forcing
+	// the multipooler to return structured Rows even when passthrough is enabled.
+	// Set by gateway paths that read the result rows themselves (for example the
+	// resolve step of ResolveTrackSetConfig) rather than streaming them straight
+	// to the client. Test-only, tied to the raw-row-passthrough experiment.
+	KeepStructured bool
+
```
GuptaManan100: I think this field is in the wrong place, and PlanExecInfo's own doc comment makes the case: 'These signals are scoped to a single StreamExecute / PortalStreamExecute call... the intent must be a per-call value, never a field on the primitive.' `KeepStructured` fails both halves of that contract.

And it isn't derived at execution time, `KeepStructured` has no runtime input at all.

The precedent I'd point at is `Route.ExecuteSQLPreparedStatement`: same shape (a static, plan-build-time property that ends up in the pooler's ExecuteOptions), and it lives as a field on the primitive with `Route.StreamExecute` forwarding it as its own argument — deliberately not through PlanExecInfo. Could `KeepStructured` follow that? A field on `Route`, or set on `ResolveTrackSetConfig`'s `ResolveRoute` at construction, with `Route.StreamExecute` doing the fold.

go/services/multigateway/engine/engine.go
```diff
@@ -56,6 +56,13 @@ type PlanExecInfo struct {
 	// per-statement hot path.
 	RecheckAdvisoryLocks bool
 
+	// KeepStructured opts this statement out of opaque row passthrough, forcing
+	// the multipooler to return structured Rows even when passthrough is enabled.
+	// Set by gateway paths that read the result rows themselves (for example the
+	// resolve step of ResolveTrackSetConfig) rather than streaming them straight
+	// to the client. Test-only, tied to the raw-row-passthrough experiment.
```
GuptaManan100: Is it still test-only? I thought we want to start using it, right away?

go/services/multigateway/engine/resolve_set_config.go
```diff
@@ -209,6 +209,9 @@ func (s *ResolveTrackSetConfig) resolve(
 	bindVars []*ast.A_Const,
 	info PlanExecInfo,
 ) ([]*sqltypes.Row, error) {
+	// This callback reads res.Rows directly, so it must receive structured rows.
+	// Opt out of opaque row passthrough for the resolve query.
+	info.KeepStructured = true
```
GuptaManan100: Also I'm thinking instead of this, what if we decode too on the Gateway side? That way we can make the pooler code not worry about decoding at all?

go/services/multipooler/internal/executor/executor.go
GuptaManan100, line 298: `rowsStreamed += int64(len(r.Rows))` — with opaque passthrough default-on, `Rows` is always empty on this path, so `mg.pooler.query.rows` will record 0 for essentially every streamed query in production. This is the exact case `Result.RowCount()` was added for, and the gateway's equivalent callbacks (`handler.go:280`, `handler.go:531`) were migrated, this one got skipped. Should be `rowsStreamed += int64(r.RowCount())`

go/services/multipooler/internal/pools/regular/regular_conn.go
```diff
@@ -242,6 +242,21 @@ func (c *Conn) Query(ctx context.Context, sql string) ([]*sqltypes.Result, error
 
 // QueryStreaming executes a query with streaming results via callback.
 // If the context is cancelled, the backend query is cancelled via adminPool.
```
GuptaManan100, line 244: The function landed between the comment of the next function 😆, please reorder.

go/services/multipooler/internal/executor/executor.go
```diff
@@ -649,7 +662,13 @@ func (e *Executor) streamExecuteOnReservedConnWithPostState(
 		rc.ReserveForPortal(name)
 	}
 
-	if err := rc.QueryStreaming(ctx, sql, callback); err != nil {
+	// Opaque row passthrough for this statement; set and reset here where the
+	// reserved connection is still owned (not yet released by the error paths
+	// below), so a failover-time release cannot race a deferred reset.
+	rc.Conn().SetRawRowPassthrough(rawRows)
```
GuptaManan100: If every call is going to own setting the RawPassthrough value, then we don't need to reset it at all, right? Otherwise the reset can be on the connection code itself when it is released. WDYT?

go/services/multigateway/engine/resolve_set_config.go
GuptaManan100: Maybe not, because the pooler is itself a major row-reading consumer. InternalQueryService.QueryArgs and friends back heartbeat (reader.go:153), consensus rule_store.go (~6 sites), pg_replication.go (~10), manager.go (~3), admin_conn.go.

go/test/microbench/serialization/rowpath_bench_test.go
```diff
@@ -0,0 +1,197 @@
...
+//     DataRow frame (that is what the backend hands it). It copies the frames
+//     into one batch blob, marshals a single carrier message, the gateway
+//     unmarshals it and writes the blob straight to the client socket. No
+//     per-column re-framing on either side.
+//
+// The opaque side is a faithful model of the proposal, not the eventual
+// production code: it reuses the generated query.Row message as the carrier
+// (Lengths = per-frame byte lengths, Values = the concatenated frames) so the
+// benchmark exercises real protobuf marshalling rather than a toy. Compare the
+// two with:
```
GuptaManan100: The package doc calls the opaque side 'a faithful model of the proposal, not the eventual production code' and carries the blob in `querypb.Row` (Lengths/Values). Now that `QueryResult.raw_data_block` exists in the same PR, the benchmark could use the real field and stop being a model.

go/test/microbench/serialization/rowpath_bench_test.go
```diff
@@ -0,0 +1,197 @@
...
+	dst = binary.BigEndian.AppendUint16(dst, uint16(len(values)))
+	for _, v := range values {
+		if v == nil {
+			dst = binary.BigEndian.AppendUint32(dst, ^uint32(0)) // -1
+			continue
+		}
+		dst = binary.BigEndian.AppendUint32(dst, uint32(len(v)))
+		dst = append(dst, v...)
+	}
+	_ = start
```
GuptaManan100: Leftover `start := len(dst)` / `_ = start` that can go.
Both lines can be removed.

go/common/pgprotocol/client/extended.go
shedges: Make Sense. Will do it

go/common/pgprotocol/client/query.go
shedges: there is no specific reason. will fix it

go/services/multigateway/engine/engine.go
shedges: removed

go/services/multipooler/internal/executor/executor.go
shedges: Sure. Let me handle reset part in connection release path as follow up PR (will create sub-issue in same issue). Changes will touch pool release code which will bloat this PR. Does that make sense?

go/services/multipooler/internal/executor/executor.go
shedges: Good Catch!! switching that counter to Result.RowCount()

shedges: Right!!

go/services/multigateway/engine/engine.go
shedges: Make sense!! following same pattern as `ExecuteSQLPreparedStatement`. Need changes in multiple files as function signature changed. We can revisit and see if we can pass structure itself.

go/services/multigateway/engine/resolve_set_config.go
shedges: Agreed with your follow-up.

go/services/multipooler/internal/executor/executor.go
shedges: Added Result.StructuredRows() safeguard. It touch multiple files but I think it's fine. One think to note that currently function will Panic if gets raw rows as expectation is to invoke function where we expect only structured rows.

haritabh17:
LGTM  🚀

PR #1275, opened 2026-07-16
fix(tests): stabilize failover and replication flakes

haritabh17:
## Summary

- Make coordinator recruitment tolerate transient PostgreSQL restarts and stop completed elections from triggering the competing-coordinator backoff.
- Give PostgreSQL configuration writes a bounded two-second budget and keep post-disconnect test writes asynchronous on the exact backend executing them.
- Preserve queue-depth metrics during eviction slot transfer and align end-to-end waits with real convergence and safety windows.

## Problem

Coverage and integration runs intermittently failed when 500ms PostgreSQL configuration contexts expired, an in-flight `restore_command` briefly restarted a standby during Recruit, a completed election was mistaken for a competing coordinator, or tests raced pooled session state and protocol backoff. Buffer eviction also exposed a temporary queue-depth decrement before the replacement entry was counted.

## Solution

Use an operation-specific PostgreSQL configuration timeout, retry only transient connectivity failures while replay stabilizes, ignore recent revocations whose term is already installed, and make eviction replacement metric-neutral. The affected end-to-end tests now use transaction-local asynchronous commit and readiness/backoff-aware waits instead of racing manager reconciliation.

## Testing

- `go test -short ./go/services/multigateway/buffer ./go/services/multiorch/consensus ./go/services/multipooler/internal/manager`
- `TestBootstrap_ViaExternalAPI` passed 5 repeated runs before rebase and once after rebase.
- Replication disconnect paths and `TestUpdateConsensusRule` passed 5 repeated runs before rebase and once after rebase.
- `TestBufferMultipleFailovers` passed 3 repeated runs before rebase and once after rebase.
- `TestPrimaryGracefulShutdownTriggersFailover` passed 3 repeated runs before rebase and once after rebase.
- `TestQueueDepthMetricOnEviction` passed 100 repeated runs.

### DIFF
--- a/go/services/multiorch/consensus/rule_change.go
+++ b/go/services/multiorch/consensus/rule_change.go
@@ -459,8 +459,14 @@ func checkRecentAcceptance(ctx context.Context, logger *slog.Logger, cohort []*m
 	const backoffWindow = 4 * time.Second
 	now := time.Now()
 	for _, pooler := range cohort {
-		rev := pooler.GetConsensusStatus().GetTermRevocation()
+		status := pooler.GetConsensusStatus()
+		rev := status.GetTermRevocation()
 		if rev == nil || rev.CoordinatorInitiatedAt == nil {
+			continue
+		}
+		// Once this term is installed as a decision, its recruitment round is
+		// complete and must not delay a later failover.
+		if status.GetCurrentPosition().GetPosition().GetDecision().GetRuleNumber().GetCoordinatorTerm() >= rev.GetRevokedBelowTerm() {
 			continue
 		}
 		timeSince := now.Sub(rev.CoordinatorInitiatedAt.AsTime())
--- a/go/services/multiorch/consensus/rule_change.go
+++ b/go/services/multiorch/consensus/rule_change.go
@@ -93,17 +93,17 @@ func (r *coordinatorLedRuleChange) Run(
 		"outgoing_rule", commonconsensus.FormatRuleNumber(revocation.GetOutgoingRule()),
 		"cohort_size", len(cohort))
 
-	// Back off if any node recently accepted a revocation — another coordinator
-	// may be running an election. A backoff is a decision not to start this
-	// appointment, so it precedes the Started event below rather than failing one.
-	if err := checkRecentAcceptance(ctx, r.coordinator.logger, cohort); err != nil {
-		return mterrors.Errorf(mtrpcpb.Code_UNAVAILABLE, "%v", err)
-	}
-
 	// Pre-validate that a proposal would be feasible with current statuses before
-	// committing to a recruitment round.
+	// interpreting previous acceptance timestamps or committing to a recruitment round.
 	if err := r.checkProposalPossible(revocation, initialStatuses); err != nil {
 		return mterrors.Errorf(mtrpcpb.Code_UNAVAILABLE, "pre-vote failed: %v", err)
+	}
+
+	// Back off if any node recently accepted an unresolved revocation — another
+	// coordinator may be running an election. A backoff is a decision not to start
+	// this appointment, so it precedes the Started event below rather than failing one.
+	if err := checkRecentAcceptance(ctx, r.coordinator.logger, cohort); err != nil {
+		return mterrors.Errorf(mtrpcpb.Code_UNAVAILABLE, "%v", err)
 	}
 
 	// Mark the start of the appointment before any RPCs go out. The leader is not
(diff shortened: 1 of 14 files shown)


go/services/multiorch/consensus/rule_change.go
```diff
@@ -459,10 +459,16 @@ func checkRecentAcceptance(ctx context.Context, logger *slog.Logger, cohort []*m
...
 	for _, pooler := range cohort {
-		rev := pooler.GetConsensusStatus().GetTermRevocation()
+		status := pooler.GetConsensusStatus()
+		rev := status.GetTermRevocation()
 		if rev == nil || rev.CoordinatorInitiatedAt == nil {
 			continue
 		}
+		// Once this term is installed as a decision, its recruitment round is
+		// complete and must not delay a later failover.
+		if status.GetCurrentPosition().GetPosition().GetDecision().GetRuleNumber().GetCoordinatorTerm() >= rev.GetRevokedBelowTerm() {
```
dweitzman: Hmm, something feels off here in the sense that this is a newly created rule change run, so if we already know that a position was decided we should have chosen a higher term revocation or, if the revocation was provided by and RPC, maybe the thing to do is validate the externally-provided revocation isn't out-of-date before doing anything else?

go/services/multipooler/internal/manager/pg_replication.go
```diff
@@ -581,6 +583,12 @@ func (pm *MultipoolerManager) waitForReplayStabilize(ctx context.Context) (*mult
 		case <-ticker.C:
 			replayLsn, isPaused, err := pm.queryReplayState(waitCtx)
 			if err != nil {
+				// Stopping an in-flight restore_command can briefly restart a
```
dweitzman: Can you say more about this comment, what about stopping an in-flight restore_command would cause a standby to restart?

The stopping happens through a wrapper we control, so if need be we could have the wrapper kill the pgbackrest command and then return successfully, if postgres has a problem with restore command returning an error or something like that

go/services/multiorch/consensus/rule_change.go
haritabh17: Ack. `rev` here is the previously persisted acceptance, not the new run’s revocation. In case of the flaky failure - the proposed term was 2 while term 1 was already accepted and decided. 

Moved pre-vote validation before this check so stale external revocations fail first, and clarified that backoff applies only to unresolved prior acceptances.

go/services/multipooler/internal/manager/pg_replication.go
haritabh17: PG runs `restore_command` synchronously from its recovery startup process. If the command is terminated by `SIGTERM`, PG interprets that as a shutdown request, exits the startup process, and the postmaster shuts down. Our monitor then restarts it. `StopRestoreCommand` signalled our wrapper, but the wrapper published its PID before installing its signal handler, and `pgctld`’s `200ms` grace was shorter than the wrapper’s `500ms` grace, so PG could observe the wrapper as signal-terminated. The CI logs show `Terminated`, followed by `startup process ... exited with exit code 1` and `shutting down due to startup process failure`. 

Fixed the wrapper to catch the signal before publishing its PID, terminate/reap `pgBackRest`, and exit successfully, and aligned the grace periods.

go/test/endtoend/queryserving/buffer_test.go
```diff
@@ -409,6 +409,13 @@ func triggerFailover(t *testing.T, setup *shardsetup.ShardSetup) {
 	require.NotNil(t, recruitResp.GetConsensusStatus(), "Recruit response should carry consensus status")
 	t.Logf("Recruit accepted, emergency demotion triggered")
 
+	// This Recruit names a different coordinator, so multiorch must honor the
+	// four-second competing-coordinator backoff before starting its own round.
+	acceptedAt := recruitResp.GetConsensusStatus().GetTermRevocation().GetCoordinatorInitiatedAt().AsTime()
```
dweitzman, line 416: Hmm, I feel like a 4-second backoff shouldn't be necessary if the previous failover is fully decided. Is this needed? If so, maybe but a TODO(dweitzman) here to remind me to look at it at some point

dweitzman (approved):

go/test/endtoend/queryserving/buffer_test.go
haritabh17: Yes, a fully decided failover shouldn’t need the backoff, and we now skip that case. This wait is for the synthetic `Recruit` just above, which records an accepted term `N+1` under test-coordinator but never completes the round. The four seconds matches the contention window in `checkRecentAcceptance`. I added the `TODO` to replace this with a proper planned-failover API.

PR #1278, opened 2026-07-17
fix(pgprotocol): parser diagnostics and scs=off string-literal fidelity

GuptaManan100:
## Summary

- Lex string literals with the session's `standard_conforming_strings`, and re-emit any backslash-bearing value as `E'...'`, so a reconstructed query keeps PostgreSQL's meaning under `scs=off`.
- Match PostgreSQL's Unicode-escape diagnostics — message, HINT, SQLSTATE, and error cursor — for both `U&'...'` and `E'...'` literals.
- Let a parse error carry a SQLSTATE other than 42601, which PostgreSQL requires for a malformed `E'...'` escape (22025).
- Make the `ast` package the single source of truth for keyword classification, and use it to quote function names the grammar would otherwise re-parse as keyword syntax.

## Problem

The multigateway parses and re-emits every single-statement query, so deparser and diagnostic infidelity is visible to clients.

With `standard_conforming_strings=off` the gateway still lexed with the default `scs=on`, mis-lexing `'a\\b'`, and a re-emitted bare `'...'` containing a backslash was then double-decoded by the backend.

Unicode-escape errors did not match PostgreSQL. `U&'...'` failures carried neither the hint nor an error cursor, so psql could not draw its `LINE`/`^` marker. `E'...'` failures were further off: invented wording ("Unicode escape sequence out of range"), always SQLSTATE 42601, and a cursor at the start of the literal instead of on the offending escape.

And `"normalize"(a, b)` round-tripped unquoted, letting the backend re-parse an explicit function call as `NORMALIZE` keyword syntax.

## Solution

**scs:** the gateway reads the session GUC and lexes with it via a new `ParseSQLWithStandardConformingStrings`. `QuoteStringLiteral` emits backslash-bearing values as `E'...'`, which decodes identically under either setting.

**Diagnostics:** `U&'...'` decoding returns a structured error mirroring `str_udeescape`'s message, hint, and cursor. `E'...'` mirrors `scan.l`'s `xeu` rules — a malformed escape is an `ereport` (22025 + hint, no `at or near`), while surrogate and value failures go through `yyerror` (42601, suffixed with only the text the rule matched); both put the cursor on the escape. `ParseSyntaxError` now carries an optional `SQLState` and `NewParseErrorAt` takes it at construction, which also corrects the extended-protocol `Parse` path (it reported 42601 where simple query reports 22025).

**Keywords:** categories move to `go/common/parser/ast/keywords.go` with a test guarding against drift. `QuoteFunctionName` quotes the col_name keywords whose bare function-call form the grammar rejects, and `normalize`/`is_normalized` deparse as keyword syntax only under `COERCE_SQL_SYNTAX`.

`TestErrorFormat_ParserDiagnostics` runs each case against live PostgreSQL and the gateway and compares the full diagnostic (SQLSTATE, message, detail, hint, position), so nothing is hardcoded.

## Behavior change

An `E'...'` escape for code point zero is now rejected as `invalid Unicode escape value` rather than embedding a NUL. This matches PostgreSQL's `is_valid_unicode_codepoint` (`c > 0 && c <= 0x10FFFF`) and the `U&'...'` path, which already rejected it.


go/services/multigateway/handler/handler.go
```diff
@@ -203,6 +203,19 @@ func (h *MultigatewayHandler) callerContext(ctx context.Context, conn *server.Co
 	return callerid.NewContext(ctx, callerid.New(conn.User(), appName))
 }
 
+// standardConformingStringsFor returns the session's effective
+// standard_conforming_strings. It defaults to true — the PostgreSQL default and
+// the value a fresh backend session uses — when the client has not changed it or
+// stored an unparseable value.
+func standardConformingStringsFor(st *MultigatewayConnectionState) bool {
+	if v, ok := st.GetSessionVariable("standard_conforming_strings"); ok {
```
haritabh17: Not using the merged startup+session view e.g. `st.GetSessionSettings()["standard_conforming_strings"];` ??

haritabh17 (approved):
Rest LGTM 🚀

go/services/multigateway/handler/handler.go
GuptaManan100: Good catch!

PR #1279, opened 2026-07-17
feat(pooler): client-side pgBackRest backup encryption

rafael:
## TL;DR

This PR tackles two main problems:

1. It adds support for client-side backup encryption using pgBackRest.
2. It adds the metadata Multigres will need to rotate encryption secrets.

The high-level design is to rotate secrets by adding new repositories rather than re-encrypting an existing repository in place.

This PR does not implement key rotation yet, but it adds the metadata and lifecycle states required to support it in a follow-up PR. 

## Description

Backup encryption is implemented through two main mechanisms:

1. **Cipher key file.** Poolers now accept a `--pgbackrest-cipher-key-file` flag. The file contains a JSON document that maps repository generations to passphrases and is used to render the pgBackRest configuration. In our deployment, this file will be provided through a Kubernetes secret.

   The repository generation is what will allow Multigres to manage multiple repositories and rotate their encryption secrets. This is needed because pgBackRest does not currently support re-encrypting an existing repository in place. Rotating a secret therefore requires creating a new repository. There is an open issue tracking support for this in [[pgBackRest#2814](https://github.com/pgbackrest/pgbackrest/issues/2814)](https://github.com/pgbackrest/pgbackrest/issues/2814). If this gets implemented in pgbackrest, we don't have to implement rotation by adding a new repo, but having the metadata in the table will still be useful to mark if a repo is being re-encrypted and understand the lifecycle. So regardless of the rotation implementation detail, tracking this metadata will be useful. 

2. **Topology flag.** When creating the topology entry for a database, it is now possible to specify whether its backups must be encrypted.

   When encryption is required, the pooler makes sure that the correct secret is present in the pgBackRest configuration during bootstrap and restore. If the secret is missing, the pooler refuses to start rather than silently creating or using an unencrypted repository.

These are the main functional changes introduced by this PR.

### Key rotation

The rest of the changes establish the metadata and lifecycle that key rotation will use in a follow-up PR.

Each repository has a `state` and a separate `authoritative` flag. The state represents where the repository is in its lifecycle, while `authoritative` identifies the canonical repository for new backups and the preferred source for restores.

A repository moves through the following states:

* **staged**: The repository has been created, recorded in the metadata table, and rendered into the pgBackRest configuration of every pooler in the cohort.

  This fleet-wide convergence is an important invariant. Every potential primary must know about the repository before it can be used. Otherwise, a newly promoted primary could fail to archive WAL to it, leaving a gap in the backup history.

* **seeded**: A full backup has been created in the repository and verified.

  A repository can only transition to `seeded` after the `staged` invariant holds.

* **active**: The repository is receiving WAL.

  During a rotation, both the old and new repositories can be active at the same time. Exactly one repository is marked as `authoritative`.

  The authoritative repository is the canonical target for new backups and the preferred source for restores. Switching authority to the new repository also requires updating the topology restore hint so that future restores resolve to it.

  Temporary configuration staleness on other poolers is acceptable during this transition because WAL is already being archived to both active repositories. This is guaranteed by the earlier `staged` step.

* **retiring**: The old repository is being removed from service.

  A repository can only begin retiring after the new authoritative repository has been proven durable. This includes creating and verifying a fresh full backup and confirming WAL continuity.

  Until those checks pass, the old repository remains configured and can still be used for restores. This ensures that there is never a point during rotation where Multigres does not have a complete and restorable backup lineage.

This lifecycle provides the ordering and invariants required to rotate encryption secrets safely.

## Testing

- Unit: cipher key-file loading (including control-character and malformed-generation rejection), per-repo config/retention rendering, `0600` atomic conf writes, fingerprint verification, the require-encryption startup checks, and the monitor's refuse-to-act-on-unreadable-repo behavior.
- E2E (`go/test/endtoend/multipooler`): encrypted backup + restore round-trip on the s3mock and filesystem harnesses, and a fault test proving a pooler with the wrong cipher key refuses to bootstrap over the encrypted repository rather than clobbering it.
- Tested local setup and validated that backups files are encrypted (e.g archive.info).


go/common/backup/config_test.go
```diff
@@ -21,31 +21,36 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/multigres/multigres/go/common/backup"
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	"github.com/multigres/multigres/go/test/utils"
 )
 
-func TestDefaultRetentionConfig(t *testing.T) {
```
rafael, line 29: The equivalent of this test is now at the end of the file

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -331,9 +331,16 @@ func (pm *MultipoolerManager) discoverPostgresState(ctx context.Context) (postgr
 		}
 	}
 
-	// Check if backups are available (only if directory not initialized)
+	// Check if backups are available (only if directory not initialized). A
```
rafael, line 334: This was in a way a small bug. We were treating error when reading backups, as proof that backups don't exist.  This created an error when trying to start a pooler with the wrong cypher (it would incorrectly tried to re-boostrap when failing to read backups).

config/pgbackrest/pgbackrest_template.conf
```diff
@@ -2,7 +2,7 @@
 log-path={{.LogPath}}
 spool-path={{.SpoolPath}}
 lock-path={{.LockPath}}
-archive-async=n
+archive-async={{if .ArchiveAsync}}y{{else}}n{{end}}
```
cuongdo, line 8: We have to be careful to provision enough disk space for `spool-path` when enabling `archive-async`. Also, longer-term, we'll want to put `spool-path` on fast storage to maximize WAL archiving throughput.

go/services/multipooler/internal/manager/pgbackrest_repos.go
```diff
@@ -0,0 +1,104 @@
...
+// this point; ON CONFLICT DO NOTHING keeps a crash-retry harmless.
+func (pm *MultipoolerManager) insertInitialPgBackRestRepo(ctx context.Context) error {
+	repo := backup.InitialPgBackRestRepo(pm.config.BackupCipherKeys)
+	pm.logger.InfoContext(ctx, "Seeding pgbackrest_repos",
+		"generation", repo.Generation, "repo_number", repo.RepoNumber, "key_fingerprint", repo.KeyFingerprint, "encrypted", repo.Encrypted)
+	execCtx, cancel := context.WithTimeout(ctx, 500*time.Millisecond)
+	defer cancel()
+	err := pm.execArgs(execCtx, `INSERT INTO multigres.pgbackrest_repos (generation, repo_number, encrypted, key_fingerprint, state, authoritative)
+		VALUES ($1, $2, $3, $4, $5, TRUE)
+		ON CONFLICT (generation) DO NOTHING`,
```
cuongdo: I don't think this is currently an issue, but why `DO NOTHING` on conflict?

go/services/multipooler/internal/manager/postgres_monitor.go
cuongdo: Yes, nice catch!

go/test/endtoend/multipooler/backup_test.go
```diff
@@ -1056,3 +1058,83 @@ func leaderTerm(cs *clustermetadatapb.ConsensusStatus) int64 {
...
+
+// TestBackup_Encrypted verifies client-side backup encryption end to end.
+// The cluster bootstraps with require_initial_repo_encryption and a mounted
+// cipher key file — bootstrap itself already proves the encrypted path works
+// (stanza-create with cipher, first full backup, and every pooler restoring
+// from the encrypted repo). This test then asserts the observable properties:
+// cipher settings in each pooler's conf (0600), the seeded pgbackrest_repos
+// row with the key's fingerprint, repo files that are actually unreadable as
+// plaintext, and an on-demand backup under the cipher.
+func TestBackup_Encrypted(t *testing.T) {
```
cuongdo, line 1069: What do you think about making the existing tests in this file all cover encrypted backups? This test would then become `TestBackup_Unencrypted`. It seems beneficial to perform more testing on the more production-like setup (encrypted backups)

cuongdo (approved):
Overall, this looks good. I left some comments.

go/services/multipooler/internal/manager/pgbackrest_repos.go
rafael: The idea for this one is to make this insert idempotent. But after second thought, on any failure during bootstrap we delete everything and start from scratch. So maybe safer and clearer to rely on that, than adding this.

Going to remove it .

config/pgbackrest/pgbackrest_template.conf
rafael: I'll add comment for this, so it's captured. Right now this is ok as with a single key is never async

go/test/endtoend/multipooler/backup_test.go
rafael: Good idea. Done

PR #1280, opened 2026-07-17
fix(pgctld): derive WAL limits from volume size

Verolop:
## Description

this PR derives WAL sizing and retention limits from the filesystem backing PGDATA instead of applying fixed values. It preserves the existing limits on volumes of 16 GiB or larger while balancing disk safety, physical replica reconnect headroom, and logical replication slot retention.

## Main changes

- Derives `min_wal_size`, `max_wal_size`, `wal_keep_size`, and `max_slot_wal_keep_size` from total volume capacity with bounded floors and caps.
- Accounts for the cluster's WAL segment size so generated values satisfy Postgres startup constraints, including non-default `initdb --wal-segsize` values.
- Reads volume capacity from the nearest existing filesystem ancestor and renders the derived values through the Postgres configuration template.
- Tests and documents the trade-off that a lagging logical replication slot can require resynchronization after exceeding its disk-safe retention cap.
- Preserves the previous WAL limits on volumes of at least 16 GiB.

## Testing

Validated with pgctld unit tests, the full binary build, and a real Postgres integration test using 64 MiB WAL segments. The integration test verifies that scaled settings are accepted at startup and that Postgres marks a logical slot as lost after it exceeds the configured retention cap.

### DIFF
--- a/config/postgres/template.cnf
+++ b/config/postgres/template.cnf
@@ -87,7 +87,7 @@ wal_buffers = {{.WalBuffers}}			# min 32kB, -1 sets based on shared_buffers
 					# (change requires restart)
 min_wal_size = {{.MinWalSize}}
 max_wal_size = {{.MaxWalSize}}
-wal_keep_size = 1000
+wal_keep_size = {{.WalKeepSize}}
 
 # - Checkpoints -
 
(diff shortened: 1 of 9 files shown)


cuongdo (approved):
This looks reasonable. We're going to have to try this with different workloads to fully gauge the effect.

PR #1281, opened 2026-07-19
fix(query-serving): preserve non-transaction reservation reasons after a failed COMMIT/ROLLBACK

shedges:
## Summary

`Executor.ConcludeTransaction` unconditionally tainted and released the reserved connection on any COMMIT/ROLLBACK error even when another independent reason (e.g. an open temp table) still legitimately held the reservation. Concretely:

```sql
CREATE TEMP TABLE p(id int PRIMARY KEY);
CREATE TEMP TABLE f(id int, fk int REFERENCES p DEFERRABLE INITIALLY DEFERRED);

BEGIN;
INSERT INTO f VALUES (1, 99);
COMMIT;          -- fails: deferred FK violation
SELECT * FROM f; -- temp table `f` should still exist on the same backend
```

A COMMIT failing on a deferred constraint is a clean, healthy-connection outcome. PostgreSQL treats it exactly like ROLLBACK. The bug conflated that with a genuine dead-connection error and released/tainted the backend regardless, orphaning the temp-table reservation onto a different physical connection for the next statement.

## Changes

- `CommitResult`/`CommitAndChainResult` now distinguish a dead connection (`mterrors.IsConnectionDead`) from a clean SQL-level failure, mirroring `RollbackResult`'s bookkeeping (restore the pre-transaction connstate snapshot, clear the transaction reason) in the latter case.
- `Executor.ConcludeTransaction` only releases/taints on a dead connection; otherwise it checks `RemainingReasons()` before releasing, matching the success path's existing behavior.
- Since `ConcludeTransaction` is a unary RPC, a returned error normally drops the entire response body. The surviving `ReservedState` now rides along as an additional gRPC status detail. The same `status.WithDetails` mechanism `mterrors.ToGRPC` already uses for `PgDiagnostic` and the gateway unpacks it so `ScatterConn` keeps tracking the reservation instead of clearing it.

New unit tests cover both the fixed path (temp-table reason survives a clean COMMIT failure, connection not released) and the existing correct behavior (a genuine dead connection still releases/taints).

### DIFF
--- a/go/services/multipooler/grpcpoolerservice/service.go
+++ b/go/services/multipooler/grpcpoolerservice/service.go
@@ -821,13 +821,36 @@ func (s *poolerService) ConcludeTransaction(ctx context.Context, req *multipoole
 		req.GetReleasePortalNames(), req.GetReleaseAllPortals(), req.GetChain(),
 	)
 	if err != nil {
-		return nil, mterrors.ToGRPC(err)
+		return nil, withReservedStateDetail(mterrors.ToGRPC(err), reservedState)
 	}
 
 	return &multipoolerpb.ConcludeTransactionResponse{
 		Result:        result.ToProto(),
 		ReservedState: reservedState,
 	}, nil
+}
+
+// withReservedStateDetail attaches state as an additional gRPC status detail
+// alongside whatever mterrors.ToGRPC already attached (e.g. a PgDiagnostic),
+// so a unary RPC that must return an error can still tell the caller the
+// authoritative post-failure reservation state — e.g. a temp-table reason
+// that survives a COMMIT which failed on a deferred constraint. A unary
+// handler that returns an error never sends its response message at all, so
+// this is the only way to carry state alongside grpcErr. Falls back to the
+// plain error if state is nil or attaching the detail fails.
+func withReservedStateDetail(grpcErr error, state *query.ReservedState) error {
+	if grpcErr == nil || state == nil {
+		return grpcErr
+	}
+	st, ok := status.FromError(grpcErr)
+	if !ok {
+		return grpcErr
+	}
+	stWithState, detailErr := st.WithDetails(state)
+	if detailErr != nil {
+		return grpcErr
+	}
+	return stWithState.Err()
 }
 
 // DiscardTempTables sends DISCARD TEMP on a reserved connection and removes the temp table reason.
--- a/go/services/multipooler/internal/executor/executor.go
+++ b/go/services/multipooler/internal/executor/executor.go
@@ -1235,6 +1235,29 @@ func (e *Executor) reservedConnError(rc reservedConnAPI, wrap string, err error)
 	return e.buildReservedStateFromAPI(rc), mterrors.Wrap(err, wrap)
 }
 
+// concludeTransactionError classifies a CommitResult/RollbackResult failure
+// for ConcludeTransaction. On a dead connection, release/taint it, matching
+// reservedConnError's handling elsewhere.
+//
+// Otherwise the backend is still healthy: CommitResult/RollbackResult already
+// reconciled the reservation bitmask and connstate with what PostgreSQL
+// actually did (e.g. a COMMIT that failed on a deferred constraint behaves
+// like ROLLBACK and already cleared the transaction reason). So release only
+// if no other reason (temp table, portal, ...) still holds the connection —
+// and never taint a connection we've confirmed is fine — otherwise return the
+// authoritative surviving state so the caller can propagate it.
+func (e *Executor) concludeTransactionError(reservedConn *reserved.Conn, releaseReason reserved.ReleaseReason, options *query.ExecuteOptions, err error) (*query.ReservedState, error) {
+	if mterrors.IsConnectionDead(err) {
+		reservedConn.Release(reserved.ReleaseError, nil)
+		return nil, err
+	}
+	if reservedConn.RemainingReasons() == 0 {
+		reservedConn.Release(releaseReason, e.sessionSettingsFromOptions(options))
+		return nil, err
+	}
+	return e.buildReservedState(reservedConn), err
+}
+
 // Describe returns metadata about a prepared statement or portal.
 func (e *Executor) Describe(
 	ctx context.Context,
@@ -2063,8 +2086,8 @@ func (e *Executor) ConcludeTransaction(
 			result, err = reservedConn.CommitResult(ctx)
 		}
 		if err != nil {
-			reservedConn.Release(reserved.ReleaseError, nil)
-			return nil, nil, err
+			state, rerr := e.concludeTransactionError(reservedConn, releaseReason, options, err)
+			return nil, state, rerr
 		}
 	case multipoolerpb.TransactionConclusion_TRANSACTION_CONCLUSION_ROLLBACK:
 		commandTag = "ROLLBACK"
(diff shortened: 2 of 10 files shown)


go/services/multipooler/internal/pools/reserved/reserved_conn.go
```diff
@@ -203,6 +204,22 @@ func (c *Conn) CommitResult(ctx context.Context) (*sqltypes.Result, error) {
 
 	results, err := c.pooled.Conn.Query(ctx, "COMMIT")
 	if err != nil {
+		if !mterrors.IsConnectionDead(err) {
```
GuptaManan100: I don't think `!IsConnectionDead(err)` is sufficient evidence that PG answered. On context cancel/deadline, `regular.execOnce` / `execWithContextCancel` return `context.Cause(ctx)`. Plain `context.DeadlineExceeded`/`Canceled`, which `IsConnectionDead` reports false for and `drainOpAfterCancel` may already have `ForceClose`d the socket.

Could we classify *positively* instead, do this only when the error carries a `*PgDiagnostic` (i.e. PG actually replied with an ERROR), and treat everything else as indeterminate -> release/taint, as before? That keeps the deferred-constraint fix while leaving the cancel path fail-safe.

go/services/multipooler/internal/executor/executor.go
```diff
@@ -2075,8 +2098,8 @@ func (e *Executor) ConcludeTransaction(
 			result, err = reservedConn.RollbackResult(ctx)
 		}
 		if err != nil {
-			reservedConn.Release(reserved.ReleaseError, nil)
-			return nil, nil, err
+			state, rerr := e.concludeTransactionError(reservedConn, releaseReason, options, err)
+			return nil, state, rerr
```
GuptaManan100, line 2114: The ROLLBACK branch changes behavior in a way I think is unintended. `RollbackResult` and `RollbackAndChainResult` return early on error without any bookkeeping, so `ReasonTransaction` is still set when we get here, `RemainingReasons()` is therefore never 0 on that path, and we never release or taint.

The deeper issue is that on an error we don't actually *know* the backend's state, did PG roll back, is it still in a failed block, did it ever answer? Leaving `ReasonTransaction` set is asserting 'still in a transaction', which is just as much a guess as clearing it would be. Could we determine it instead of inferring?

Maybe we can use the `reserved.Conn.TxnStatus()` that already exposes the backend's own ReadyForQuery status, and `client.Conn` updates it on the error path too. We can do this for ☝️ too.

go/services/multipooler/grpcpoolerservice/service.go
```diff
@@ -814,7 +814,7 @@ func (s *poolerService) ConcludeTransaction(ctx context.Context, req *multipoole
 		req.GetReleasePortalNames(), req.GetReleaseAllPortals(), req.GetChain(),
 	)
 	if err != nil {
-		return nil, mterrors.ToGRPC(err)
+		return nil, withReservedStateDetail(mterrors.ToGRPC(err), reservedState)
```
GuptaManan100, line 824: Hmmm... We solve this reserved state information passing on error problem in StreamExecute by sending the state first and then the error. 
This is a unary RPC, so we don't have that option, but putting it in the status feels weird. Maybe we have no better option. I don't know. WDYT @haritabh17 @shedges?
We could put the reserved state and error both in the response, in which case RPC errors will mean we treat them as before, but for actual postgres errors, we send the error and the new state in the main response packet.

go/services/multipooler/internal/executor/executor.go
shedges: Ack. Updated

go/services/multipooler/internal/pools/reserved/reserved_conn.go
shedges: Valid point.

go/services/multipooler/grpcpoolerservice/service.go
shedges: Fair concern but I think gRPC status is standard mechanism to pass structured data alongside of an error on unary RPC

go/services/multipooler/internal/pools/reserved/reserved_conn.go
```diff
@@ -261,6 +306,23 @@ func (c *Conn) RollbackResult(ctx context.Context) (*sqltypes.Result, error) {
...
+			// RemainingReasons() never reaches 0 for a ROLLBACK that errors
+			// this way, and concludeTransactionError can never release the
+			// connection even though nothing else holds it reserved.
+			if c.txnSnapshot != nil {
+				c.pooled.Conn.State().RestoreFromTxn(c.txnSnapshot)
+				c.txnSnapshot = nil
+			}
+			c.ClearSessionStateUntrusted()
+			c.recordTxnOutcome(ctx, txnOutcomeRollback)
+			c.RemoveReservationReason(protoutil.ReasonTransaction)
```
haritabh17: Does it make sense to extract these blocks into separate helper function?

go/services/multipooler/internal/pools/reserved/reserved_conn.go
```diff
@@ -307,6 +369,19 @@ func (c *Conn) RollbackAndChainResult(ctx context.Context) (*sqltypes.Result, er
 
 	results, err := c.pooled.Conn.Query(ctx, "ROLLBACK AND CHAIN")
 	if err != nil {
+		if isRecoverableSQLError(err) {
```
haritabh17: In cases when its non-recoverable -- i.e. connection died, we are forwarding the fatal / panic to upstream client so that the client conn to gateway also dies right?

I see that `concludeTransactionError ` does `reservedConn.Release(reserved.ReleaseError, nil)` when the conn is dead, just confirming that it does end up killing the upstream conn

go/services/multipooler/internal/executor/executor.go
```diff
@@ -1234,6 +1234,29 @@ func (e *Executor) reservedConnError(rc reservedConnAPI, wrap string, err error)
...
+// like ROLLBACK and already cleared the transaction reason). So release only
+// if no other reason (temp table, portal, ...) still holds the connection —
+// and never taint a connection we've confirmed is fine — otherwise return the
+// authoritative surviving state so the caller can propagate it.
+func (e *Executor) concludeTransactionError(reservedConn *reserved.Conn, releaseReason reserved.ReleaseReason, options *query.ExecuteOptions, err error) (*query.ReservedState, error) {
+	if mterrors.IsConnectionDead(err) {
+		reservedConn.Release(reserved.ReleaseError, nil)
+		return nil, err
+	}
+	if reservedConn.RemainingReasons() == 0 {
```
haritabh17, line 1265: What do you think about having a branch that `taints` on anything that isn't provably a clean SQL error, so the indeterminate case joins the dead case.

haritabh17 (approved):
Rest looks good. 🚀

go/services/multipooler/internal/pools/reserved/reserved_conn.go
shedges: Done

go/services/multipooler/internal/executor/executor.go
shedges: Handled and added comment for same

go/services/multipooler/internal/pools/reserved/reserved_conn.go
shedges: Right!! we are not killing client connection in case of Panc/FATAL. Created separate Linear issue under MVP.

PR #1282, opened 2026-07-20
fix(query-serving): close COPY, streaming, and ParameterStatus gaps

GuptaManan100:
## Summary

- `COPY ... TO STDOUT` delivered trailing notices twice: the multipooler sent them both inside the embedded `QueryResult` and in the response's dedicated `Notices` field, and the gateway folded both into the final result.
- Rows PostgreSQL streamed before a mid-statement error were discarded, so a client received the error but none of the rows that preceded it. Both the simple-query and extended-protocol client read loops were affected.
- A `SET` that changed a `GUC_REPORT` parameter was never reported to the client, so after `SET standard_conforming_strings = off` the client still believed `on` and psql kept parsing input under the wrong rule.

## Solution

Notice duplication: clear the notices from the embedded `QueryResult` on the COPY-out path so the dedicated `Notices` field is the single transport, matching what the `CopyFinalize` path already did.

Partial rows: the client read loops only flushed batched rows on `CommandComplete` or a 2MB threshold, but a mid-statement failure sends `ErrorResponse` with no `CommandComplete`. Flush on `ErrorResponse` too — before the error is recorded, since `flushBatch` is gated on `firstErr`.

ParameterStatus: `ValidateSetting` already runs `set_config(name, value, true)` to validate a `SET`, and its return value is PostgreSQL's canonicalized result (`DateStyle 'ISO'` → `'ISO, MDY'`), so capture that instead of echoing the client's raw string. Handing it to the trailing `ApplySessionState` needs a per-execution channel because plans are cached across connections, so `Sequence` creates a `SequenceExchange` per execution and threads it via `PlanExecInfo`. The wire server emits the message after `CommandComplete`, and suppresses it when the value did not actually change — mirroring PostgreSQL, which reports on change only.


go/common/pgprotocol/server/conn.go
```diff
@@ -1039,6 +1048,17 @@ func (c *Conn) handleQuery() error {
 				return fmt.Errorf("writing command complete: %w", err)
 			}
 
+			// Report changed GUC_REPORT parameters after CommandComplete and
+			// before ReadyForQuery, matching PostgreSQL's order (postgres.c
+			// reports them just before ReadyForQuery). Carried on the result for
+			// a SET that named a reportable GUC; reportParameterStatus drops the
+			// ones whose value did not actually change.
+			for name, value := range result.ParameterStatus {
```
shedges, line 1056: We need same changes under  `handleExecute` to handle extended query protocol ?

shedges (commented):
We need different mechanism to handle case where GUC get modified as part of pg_reload_conf. may be separate issue for same

go/common/pgprotocol/server/startup.go
```diff
@@ -1147,13 +1147,34 @@ func (c *Conn) sendParameterStatuses() error {
...
+	c.reportedParams[name] = value
+	return nil
+}
+
+// reportParameterStatus sends a ParameterStatus only when value differs from
+// what this connection last told the client, mirroring PostgreSQL: a GUC_REPORT
+// parameter is re-reported on change, not on every SET that names it. A
+// parameter with no recorded value (never sent at startup, e.g. application_name)
+// is always reported the first time.
+func (c *Conn) reportParameterStatus(name, value string) error {
```
haritabh17, line 1173: Is this used in the RESET and in-xact paths??

go/common/pgprotocol/server/conn.go
GuptaManan100: Good catch! Yes, I fixed it

GuptaManan100:
> We need different mechanism to handle case where GUC get modified as part of pg_reload_conf. may be separate issue for same

Yes it should be an issue of its own, cause that requires sending info to all connected sessions 🤔

haritabh17 (approved):
Rest looks fine, approved 🚀

go/common/pgprotocol/server/startup.go
GuptaManan100: I've fixed RESET path now. Working on transactions, but that's more involved 😢

PR #1285, opened 2026-07-20
ci(flaky-tests): dedupe and enrich Trunk-created Linear tickets

rafael:
> [!NOTE]
> The implementation in this PR was written by AI, with me providing the requirements and reviewing the approach. I validated the script locally, and the code looks reasonable to me.

## Summary

* Adds `go/tools/trunkticketsync`, a reconciler for Linear tickets created by the Trunk Flaky Tests ticketing integration. It removes duplicate tickets, keeps one canonical ticket per test updated with the latest status from Trunk, closes tickets when tests become healthy again, and reopens completed tickets when tests regress.
* Adds `.github/workflows/trunk-ticket-sync.yml`, which runs the tool every 4 hours. It can also be triggered manually.
* All organization-specific identifiers, including the Linear team, label, project, and Trunk organization slug, come from `ci` environment secrets. No internal identifiers or tokens are stored in the repository, and the test fixtures use scrubbed placeholder values.

## Problem

Trunk's ticketing integration creates a new Linear ticket every time a test transitions to flaky, but it does not update that ticket afterward. There is no deduplication, status synchronization, or automatic closing, so the backlog ends up accumulating stale and duplicate tickets.

Trunk confirmed that these capabilities are not currently supported and recommended that we periodically reconcile the tickets ourselves using their `get-test-details` API.

## Solution

* Tickets are grouped using the Trunk test UUID embedded in each ticket description. The oldest open ticket becomes the canonical ticket, and newer duplicates are deleted. Linear keeps deleted tickets in the trash for 30 days, so they remain recoverable if something goes wrong.
* The canonical ticket gets a managed markdown section generated from `get-test-details`. It includes:
  * Current status and transition date
  * First seen date
  * Latest failure
  * 7-day failure rate
  * Number of PRs impacted
  * Quarantine status
  * Most common failure reasons
* Updates are gated using a content hash embedded in the managed section marker. This is needed because Linear normalizes stored markdown, which means direct text comparison does not reliably converge.
* Healthy and unquarantined tests automatically have their tickets closed with a comment.
* Flaky or broken tests reopen tickets that were previously marked as Done.
* When the Trunk state cannot be determined, either because of a fetch error or a 404, the ticket is left unchanged.
* Retries use the shared `go/tools/retry` backoff implementation.
* Unit and end-to-end tests run against `httptest` fakes for both the Linear GraphQL API and the Trunk REST API.

## Required secrets

The following secrets are required in the `ci` environment:

* `LINEAR_API_TOKEN` with write access
* `LINEAR_TEAM_KEY`
* `LINEAR_LABEL`
* `LINEAR_PROJECT`
* `TRUNK_ORG_SLUG`

The existing `TRUNK_API_TOKEN` secret is reused.

### DIFF
--- a/.github/workflows/trunk-ticket-sync.yml
+++ b/.github/workflows/trunk-ticket-sync.yml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Flaky Ticket Sync
+name: Trunk Ticket Sync
 
 on:
   schedule:
@@ -66,4 +66,4 @@ jobs:
           if [ "$DRY_RUN" = "true" ]; then
             args+=(--dry-run)
           fi
-          go run ./go/tools/flakyticketsync "${args[@]}"
+          go run ./go/tools/trunkticketsync "${args[@]}"
--- a/.github/workflows/trunk-ticket-sync.yml
+++ b/.github/workflows/trunk-ticket-sync.yml
@@ -18,16 +18,7 @@ on:
   schedule:
     # Every 4 hours.
     - cron: "0 */4 * * *"
-  # checkov:skip=CKV_GHA_7: the input is a read-only flag that switches the
-  # sync tool to dry-run mode; it does not influence any build artifact, so
-  # the "build output cannot be affected by user parameters" rule does not
-  # apply.
   workflow_dispatch:
-    inputs:
-      dry_run:
-        description: "Log ticket mutations instead of performing them"
-        required: false
-        default: "false"
 
 permissions: {}
 
@@ -60,10 +51,4 @@ jobs:
           LINEAR_LABEL: ${{ secrets.LINEAR_LABEL }}
           LINEAR_PROJECT: ${{ secrets.LINEAR_PROJECT }}
           GITHUB_REPOSITORY: ${{ github.repository }}
-          DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}
-        run: |
-          args=()
-          if [ "$DRY_RUN" = "true" ]; then
-            args+=(--dry-run)
-          fi
-          go run ./go/tools/trunkticketsync "${args[@]}"
+        run: go run ./go/tools/trunkticketsync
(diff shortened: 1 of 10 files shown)


rafael:
Secrets are already configured, so this should work once we merge it

go/tools/trunkticketsync/main.go
```diff
@@ -0,0 +1,650 @@
...
+// edit stripped the marker comments, the section heading anchors the
+// replacement (to end of description, since the section is always appended
+// last); with no trace of the section it is appended.
+func replaceManagedSection(desc, section string) string {
+	if begin := strings.Index(desc, managedBeginPrefix); begin >= 0 {
+		if end := strings.Index(desc[begin:], managedEnd); end >= 0 {
+			return desc[:begin] + section + desc[begin+end+len(managedEnd):]
+		}
+	}
+	prefix, _, _ := strings.Cut(desc, sectionHeading)
```
haritabh17: If a UI edit strips the marker comments but someone has added notes below the managed section, this fallback truncates from the heading to the end of the description and silently deletes those notes. 

Possible to preserve any trailing content after the old section instead of assuming the section is always last?

.github/workflows/trunk-ticket-sync.yml
```diff
@@ -0,0 +1,54 @@
...
+      # secrets, which GitHub also masks in the public run logs.
+      - name: Run ticket sync
+        env:
+          TRUNK_API_TOKEN: ${{ secrets.TRUNK_API_TOKEN }}
+          TRUNK_ORG_SLUG: ${{ secrets.TRUNK_ORG_SLUG }}
+          LINEAR_API_TOKEN: ${{ secrets.LINEAR_API_TOKEN }}
+          LINEAR_TEAM_KEY: ${{ secrets.LINEAR_TEAM_KEY }}
+          LINEAR_LABEL: ${{ secrets.LINEAR_LABEL }}
+          LINEAR_PROJECT: ${{ secrets.LINEAR_PROJECT }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
```
haritabh17: Redundant? This is already set by the runner by default.

haritabh17 (approved):

PR #1286, opened 2026-07-20
fix(multigateway): improve error; report credential lookup outages as 57P03

Verolop:
## Description
this PR addresses a misleading error that frequently appears during failover and topo recovery issues. When Multigateway could not reach a writable primary to look up credentials, clients received `28P01 password authentication failed`, which made a number of scenarios look like credential problems, sending the wrong signal.

With this change, failures that occur before the credential lookup now return `57P03 cannot_connect_now`, while genuine credential failures remain as 28P01 responses.
## Main Changes
- Preserve RPC error codes alongside structured PostgreSQL diagnostics through wrapped errors.
- Return `57P03` when the primary, pooler, Postgres backend, or failover buffer is unavailable.
- Preserve internal failure details while exposing a stable, retryable message to clients.
## Testing
Added coverage for missing and disconnected primaries, unavailable backends, planned failover, buffer expiry, and the complete client-facing authentication path.

### DIFF
--- a/go/test/endtoend/queryserving/pgbouncertests/nonexistent_db_test.go
+++ b/go/test/endtoend/queryserving/pgbouncertests/nonexistent_db_test.go
@@ -41,11 +41,10 @@ import (
 // and the auth-vs-database error ordering.
 //
 // **Known divergence**: the multigateway rejects a nonexistent connection
-// database with 28P01 (auth failure) instead of 3D000 / "no such database" —
-// which is what both PostgreSQL and pgbouncer return. See
+// database with 57P03 (cannot_connect_now) instead of 3D000 / "no such
+// database", which is what both PostgreSQL and pgbouncer return. See
 // TestNonexistentDatabaseConnectBehavior; the eventual fix is to align with
-// pgbouncer. The auth-ordering case (28P01 when both password is wrong AND
-// database is missing) matches between the two.
+// pgbouncer.
 
 // nonexistentDB is a database name that is not registered in the test topology
 // nor created in postgres.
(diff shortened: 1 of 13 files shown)


go/test/endtoend/queryserving/pgbouncertests/nonexistent_db_test.go
```diff
@@ -120,10 +113,13 @@ func TestAuthErrorPrecedesDatabaseError(t *testing.T) {
 			code, stage, err := probeConnect(t, ctx, target.Port, nonexistentDB, "definitely_the_wrong_password")
 			require.Error(t, err, "a wrong password must fail the connection")
 			t.Logf("%s: failed at %s stage with SQLSTATE %q", target.Name, stage, code)
-			assert.Equalf(t, "28P01", code,
-				"auth failure (28P01) must win over the nonexistent-database error, got %q", code)
+			expectedCode := "28P01"
+			if target.Name == "multigateway" {
+				expectedCode = "57P03"
+			}
```
GuptaManan100, line 119: Isn't this a difference from Postgres behaviour for multigres?

GuptaManan100 (approved):
Just that one comment

go/test/endtoend/queryserving/pgbouncertests/nonexistent_db_test.go
Verolop: yeah, but I discovered that the previous behaviour wasn't really matching Postgres either 😅 . Postgres returns `28P01` in that case because it checks the password before it resolves the db, so the auth failure genuinely wins, but the gateway never gets to that point, because when the db doesn't exist there's no pooler registered for it and the credential lookup can't be routed anywhere. The `28P01` it used to return was just masking that routing failure as an auth error, so the two only agreed on the wire code by coincidence, not because they were doing the same thing.

Now, what we actually want for the case where a db that doesn't exist is `3D000`, i.e. the same "no such database" error that Postgres and pgbouncer both return, I've opened #1290 to track it. Until that's in, 57P03 is at least honest about what broke, since the connection fails before we ever consult any credentials, instead of spamming us with fake credential errors. 

Thank you!

PR #1287, opened 2026-07-21
feat(pgctld,multipooler): start postgres as a standby by default

mkindahl:
## Summary

Make PostgreSQL **start in standby (recovery) mode by default** when starting an already-initialized data directory, so a recovering node can never come up as a writable primary on its own. This closes a dual-primary window: previously the pgctld `Start` RPC brought postgres up in whatever mode its on-disk state dictated, so a former primary with no `standby.signal` came back **physically writable** before consensus was consulted — and if a new leader had been elected while it was down and the recovering node could not reach it, the cluster had two writable servers.

## Approach

- `StartRequest` gains an `as_primary` bool (default `false` = standby). The default writes `standby.signal` so postgres comes up in recovery mode; `as_primary=true` removes it for a writable start. (Uses the shared `standby.signal` create/remove helpers landed separately as `refactor(pgctld): consolidate standby.signal handling into shared helpers`, now in `main`.)
- Because `Start` now defaults to standby, the multipooler **monitor's restart** of a stopped-but-initialized node comes up **read-only automatically** — no monitor change required. Promotion to a writable primary happens **only** through the consensus-gated `Promote` path (`pg_promote`).
- The one-time **first-backup bootstrap** (`rpc_first_backup.go`) and the **standalone pgctld test helpers** pass `as_primary=true` — the only places that legitimately need a writable start with no consensus layer to promote them (bootstrap runs at term 0 and is torn down immediately, so there is no split-brain window).

## Behavior change / trade-off

A former primary that restarts now comes back **read-only** and is re-promoted only by the coordinator (multiorch) — a crash of the primary's postgres becomes a fast failover rather than an in-place restart. This is the intended safety trade: never risk two writable servers, even when the recovering node is isolated.

## Tests

- Reframed `TestBackendCrashRecoveryUnderLoad` to assert transparent **read** recovery + durability of pre-crash writes, since a crashed primary no longer self-heals to writable without a coordinator (`TestBackendCrashTransparentReconnect` already only reads post-crash).
- Made the shared end-to-end cluster fixture tolerate a leadership change: `SetupTest` re-discovers the current primary and re-derives clean-state baselines, and the destructive primary-killing tests rebuild a clean cluster — clearing the cascade this behavior change caused across the `multipooler` and `queryserving` suites.
- Added unit coverage for the `Start` handler's `standby.signal` step: `TestPgCtldServiceStart_AsPrimaryCrashRecoveryMatrix` (all four combinations of `as_primary` × `allow_crash_recovery`, asserting `standby.signal` presence and whether crash recovery ran) and `TestPgCtldServiceStart_StandbySignalError` (both error-return branches — failing to create and to remove `standby.signal`).
- Added `TestCreateFirstBackupAndInitialize_StartFails`, covering the writable first-backup bootstrap `Start(as_primary=true)` failure path in `createFirstBackupAndInitializeLocked`, including its data-directory cleanup.

## ⚠️ Intermediate fix — real issue tracked separately

`TestPrimaryCrashWithUnarchivedWAL_NoDataLoss` began failing under this change and is currently addressed with an **intermediate, test-only fix** (`test(backupfaults): poll for restore_command clear on cohort join`): `verifyRestoreCommandCleared` polls until it converges instead of reading once. This does **not** fix the underlying product issue.

**Root cause (needs a real fix):** with start-as-standby, a primary crash becomes a full failover, so a freshly-restored standby joins the *already-established* cohort **out-of-band** — via multiorch's `PoolerNotInCohort` reconcile (`UpdateConsensusRule`, which runs on the primary) rather than the promotion-time `Recruit`. `Recruit` is the only path that clears a new member's `restore_command` **synchronously**; the out-of-band join does not, and `SetPrimary` correctly skips it (the node is still an observer at that point, and an observer catching up may legitimately need the archive). The clear is therefore delivered only by the node's **async** monitor backstop (`remedialActionDisableRestoreCommand`), ~9s later. During that window a restart-as-standby (now the default) could resolve `recovery_target_timeline=latest` via archive `archive-get` to a divergent timeline and FATAL at startup — exactly the divergence hazard this feature exists to prevent.

**Real fix (follow-up):** clear `restore_command` synchronously at the observer→cohort-member transition (e.g. re-issue `SetPrimary` to the new member once membership is recorded, so the existing cohort-gated clear fires), **without** regressing an observer still legitimately catching up from the archive. It must not be done unconditionally in `SetPrimary` (would strand a lagging observer) nor in `UpdateConsensusRule` (runs on the primary, cannot reach the joining member).

Closes MUL-1004

### DIFF
--- a/go/cmd/pgctld/command/server_test.go
+++ b/go/cmd/pgctld/command/server_test.go
@@ -174,6 +174,123 @@ func TestPgCtldServiceStart(t *testing.T) {
 	}
 }
 
+// TestPgCtldServiceStart_AsPrimaryCrashRecoveryMatrix covers all four
+// combinations of the two StartRequest controls on an already-initialized data
+// directory that was NOT cleanly shut down (so allow_crash_recovery has an
+// effect):
+//   - as_primary selects the start mode: false (default) writes standby.signal
+//     (recovery/standby mode); true removes it (writable primary).
+//   - allow_crash_recovery selects whether an unclean node is crash-recovered
+//     before the start; the response reports it via CrashRecoveryRan.
+func TestPgCtldServiceStart_AsPrimaryCrashRecoveryMatrix(t *testing.T) {
+	cases := []struct {
+		name                 string
+		asPrimary            bool
+		allowCrashRecovery   bool
+		wantStandbySignal    bool
+		wantCrashRecoveryRan bool
+	}{
+		{
+			name:      "standby, no crash recovery",
+			asPrimary: false, allowCrashRecovery: false, wantStandbySignal: true, wantCrashRecoveryRan: false,
+		},
+		{name: "standby, allow crash recovery", asPrimary: false, allowCrashRecovery: true, wantStandbySignal: true, wantCrashRecoveryRan: true},
+		{name: "primary, no crash recovery", asPrimary: true, allowCrashRecovery: false, wantStandbySignal: false, wantCrashRecoveryRan: false},
+		{name: "primary, allow crash recovery", asPrimary: true, allowCrashRecovery: true, wantStandbySignal: false, wantCrashRecoveryRan: true},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			baseDir := t.TempDir()
+			dataDir := filepath.Join(baseDir, "pg_data")
+			t.Setenv(constants.PgDataDirEnvVar, dataDir)
+
+			binDir := filepath.Join(baseDir, "bin")
+			require.NoError(t, os.MkdirAll(binDir, 0o755))
+			testutil.CreateMockPostgreSQLBinaries(t, binDir)
+			// Report an UNCLEAN shutdown so allow_crash_recovery actually triggers
+			// recovery (the default mock reports "shut down").
+			testutil.MockBinary(t, binDir, "pg_controldata",
+				`echo "Database cluster state:               in production"`)
+			t.Setenv("PATH", binDir+":"+os.Getenv("PATH"))
+
+			// Initialized data dir, plus a pre-existing standby.signal so we can
+			// observe as_primary=true removing it and as_primary=false keeping it.
+			require.NoError(t, os.MkdirAll(dataDir, 0o755))
+			require.NoError(t, os.WriteFile(filepath.Join(dataDir, "PG_VERSION"), []byte("16\n"), 0o644))
+			require.NoError(t, os.WriteFile(filepath.Join(dataDir, constants.StandbySignalFile), []byte(""), 0o644))
+
+			service, err := NewPgCtldService(testLogger(), testServiceConfig, 30, baseDir, "localhost", 0, "")
+			require.NoError(t, err)
+			defer service.Close()
+
+			ctx := context.Background()
+			resp, err := service.Start(ctx, &pb.StartRequest{
+				AsPrimary:          tc.asPrimary,
+				AllowCrashRecovery: tc.allowCrashRecovery,
+			})
+			// Stop the mock postgres (pg_ctl start backgrounds a sleep) on cleanup.
+			defer func() { _, _ = service.Stop(ctx, &pb.StopRequest{Mode: "fast"}) }()
+			require.NoError(t, err)
+			require.NotNil(t, resp)
+
+			_, statErr := os.Stat(filepath.Join(dataDir, constants.StandbySignalFile))
+			if tc.wantStandbySignal {
+				assert.NoError(t, statErr, "standby.signal must be present for as_primary=%v", tc.asPrimary)
+			} else {
+				assert.True(t, os.IsNotExist(statErr), "standby.signal must be absent for as_primary=%v", tc.asPrimary)
+			}
+			assert.Equal(t, tc.wantCrashRecoveryRan, resp.GetCrashRecoveryRan(),
+				"CrashRecoveryRan for allow_crash_recovery=%v on an unclean node", tc.allowCrashRecovery)
+		})
+	}
+}
+
+// TestPgCtldServiceStart_StandbySignalError exercises the two error-return
+// branches of the Start handler's standby.signal step: as_primary=false must
+// surface a createStandbySignal failure and as_primary=true a removeStandbySignal
+// failure, in both cases returning an error instead of starting postgres. A
+// standby.signal path that is a NON-EMPTY directory triggers both: os.WriteFile
+// fails (the path is a directory) and os.Remove fails (directory not empty), and
+// neither error is IsNotExist.
+func TestPgCtldServiceStart_StandbySignalError(t *testing.T) {
+	cases := []struct {
+		name      string
+		asPrimary bool
+	}{
+		{name: "standby start, createStandbySignal fails", asPrimary: false},
+		{name: "primary start, removeStandbySignal fails", asPrimary: true},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			baseDir := t.TempDir()
+			dataDir := filepath.Join(baseDir, "pg_data")
+			t.Setenv(constants.PgDataDirEnvVar, dataDir)
+
+			// Initialized data dir (PG_VERSION present) so Start proceeds past the
+			// not-initialized guard to the standby.signal step.
+			require.NoError(t, os.MkdirAll(dataDir, 0o755))
+			require.NoError(t, os.WriteFile(filepath.Join(dataDir, "PG_VERSION"), []byte("16\n"), 0o644))
+
+			// Make standby.signal a non-empty directory so the file operation
+			// as_primary selects fails before postgres is ever started.
+			signalDir := filepath.Join(dataDir, constants.StandbySignalFile)
+			require.NoError(t, os.MkdirAll(signalDir, 0o755))
+			require.NoError(t, os.WriteFile(filepath.Join(signalDir, "child"), []byte(""), 0o644))
+
+			service, err := NewPgCtldService(testLogger(), testServiceConfig, 30, baseDir, "localhost", 0, "")
+			require.NoError(t, err)
+			defer service.Close()
+
+			resp, err := service.Start(context.Background(), &pb.StartRequest{AsPrimary: tc.asPrimary})
+			require.Error(t, err, "Start must fail when the standby.signal operation fails")
+			assert.Nil(t, resp)
+			assert.Contains(t, err.Error(), "standby.signal")
+		})
+	}
+}
+
 func TestPgCtldServiceStart_MissingPoolerDir(t *testing.T) {
 	t.Run("missing pooler-dir", func(t *testing.T) {
 		_, err := NewPgCtldService(testLogger(), PgCtldServiceConfig{}, 0, "", "", 0, "")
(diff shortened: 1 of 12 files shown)


proto/pgctldservice.proto
```diff
@@ -103,8 +116,12 @@ message RestartRequest {
...
   repeated string extra_args = 4;
 
-  // If true, creates standby.signal before restart (for demotion to standby)
-  bool as_standby = 5;
+  // If true or unset, creates standby.signal before restart so postgres comes
+  // up in recovery (standby) mode rather than as a writable primary. Set to
+  // false explicitly to restart while staying (or becoming) a writable
+  // primary. Declared `optional` for the same explicit-presence reason as
+  // StartRequest.as_standby, which this mirrors.
+  optional bool as_standby = 5;
```
dweitzman: Maybe be simpler to make this `as_primary` or `writable` or even an enum like `RoutingRole` instead of using proto3 `optional` semantics? I don't have super strong feelings, but personally I have a soft preference for avoiding `optional` semantics and for enums over bools for type safety.

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -862,35 +862,49 @@ func (pm *MultipoolerManager) hasCompleteBackups(ctx context.Context) (bool, err
...
 	}
 
-	_, err := pm.pgctldClient.Start(ctx, &pgctldpb.StartRequest{})
-	if err != nil {
+	// Never come back up as a writable primary on our own. Flag suspected
+	// divergence so a subsequent follow of the current leader rewinds if needed.
+	// consensusMgr is always set in production; the guard keeps minimal unit
+	// tests (which construct a bare manager) working.
+	if pm.consensusMgr != nil {
+		if _, err := pm.consensusMgr.SetSuspectedDivergence(ctx, true); err != nil {
```
dweitzman: It feels aggressive to set suspected divergence here? Couldn't this trigger in a scenario like bootstrapping a new cluster where there's no reason to suspect divergence?

Normally we only mark suspected divergence in response to something that appears to indicate a failover happened and leadership may have been stolen away, which is consensus context this method doesn't have

dweitzman (commented):
Still looking at the tests, just sending out comments on the non-test code now

proto/pgctldservice.proto
mkindahl: I avoided that because it would mean we need to change the name of the field and deprecate the old field to avoid protocol mismatches. If we are fine with doing a strict rename (hence change the protocol-level meaning), `as_primary` (with default false) is definitely cleaner and easier to understand.

go/services/multipooler/internal/manager/postgres_monitor.go
mkindahl: > It feels aggressive to set suspected divergence here? Couldn't this trigger in a scenario like bootstrapping a new cluster where there's no reason to suspect divergence?

For the bootstrapping case, it could indirectly be triggered, but only if there is a crash after the bootstrapping.

The bootstrapping path uses `startPostgreSQLAfterRestore` (I renamed it here as well since it breaks the convention to write `Postgres` rather than `PostgreSQL` rather than `startPostgres`. The function `startPostgres` is called when a server restarts and the data directory is already initialized, not when bootstrapping a new server by restoring a backup. 

The key point is not in what scenarios it _can_ trigger but rather what scenarios that would break if it _didn't_ trigger. Also note that a suspected divergence is safe since it will check if rewind can work before doing anything and for that reason should be the default.

The only case where we can skip the rewind is when we _know_ it is not needed. All other cases should suspect divergence.

(Maybe a better name for `startPostgres` would be `restartPostgresWithDataDir` to indicate that it is a restarting Postgres with an existing data directory, not a start of Postgres bootstrapping when bootstrapping a new pod.)

> Normally we only mark suspected divergence in response to something that appears to indicate a failover happened and leadership may have been stolen away, which is consensus context this method doesn't have

Yes, but in this particular case, we have this scenario:

1. Primary A plus two standbys B and C are running.
2. Primary A crash-stop, and has a transaction T1' that is unacknowledged.
3. B is elected as the new primary and C streams from it.
4. A comes back. Since the volume is intact, the transaction T1' is already there.
5. Postgres starts and if we do not set `SuspectDivergence`, it will try to connect to the primary without calling `pg_rewind`, which will break.

dweitzman: In the scenario you're describing, I would expect today that `suspectedDivergence` is already set without any other changes needed when `SetPrimary` informs pooler A that it's no longer the leader. Is that not the case?

mkindahl: This is when the pooler is starting on its own accord, before any `SetPrimary` has been sent.

mkindahl: This code is not necessary after you pushed your fix: it now detects if it was an actual crash that brought down the server and sets suspected divergence when the node crashes.

Unfortunately, this is not sufficient. Consider this scenario:

1. Primary A executes transaction T'. Under `synchronous_commit`, T' is written and durably flushed to A's local WAL *before* A waits for a standby's acknowledgment.
2. A is shut down via SIGTERM (smart/fast shutdown). This performs a normal shutdown checkpoint; `pg_control` records cluster state `"shut down"`.
3. Standby B is promoted and becomes the new primary on a new timeline that never received T'.
4. A restarts. The multipooler monitor's `startPostgres` (`postgres_monitor.go:966`) calls pgctld's `Start` with `AllowCrashRecovery: true`.
5. pgctld's crash-recovery check (`isPostgresCleanlyStopped`, `crash_recovery.go:42`) only treats a non-`"shut down"`/`"shut down in recovery"` cluster state as needing recovery. Since step 2 left a clean state, `crashRecoveryNeeded` returns `false`, so `StartResponse.crash_recovery_ran` is `false`.
6. Back in `startPostgres`, the divergence-marking gate (`if resp.GetCrashRecoveryRan()`, `postgres_monitor.go:1012`) never fires. `SetSuspectedDivergence` is not called from this path.
7. `startPostgres` always brings A back up as a standby by default so A is immediately `InRecovery`.
8. When consensus/orch later calls `SetPrimary` on A to point it at B, `setPrimaryLocked`'s other safety check — `if pgMode.OutOfRecovery() { SetSuspectedDivergence(true) }` (`rpc_consensus.go:688`) — no longer trips, because A is already a standby by the time this RPC arrives.
9. `setPrimaryLocked` therefore takes the non-divergent branch (`rpc_consensus.go:736`), which just calls `setPrimaryConnInfoLocked` to swap `primary_conninfo` to B and resume streaming, with no rewind and no timeline compatibility check.
10. A streams from B without ever being rewound. Depending on where the fork point falls relative to A's WAL, this either FATALs the WAL receiver, or leaves T' still visible in reads against A even though B (the authoritative primary) never has it.

However, I want to investigate this as a separate PR since this would complicate this PR to handle that case here. This is not a problem in bringing the server up as a standby, this is a problem introduced in the previous commit 2ff20aa4.

proto/pgctldservice.proto
mkindahl: This is now changed. The code uses `as_primary` for all the requests, both `Start` and `Restart`.

go/services/multipooler/internal/manager/postgres_monitor.go
dweitzman: I sent out https://github.com/multigres/multigres/pull/1305 with a fix for step 6 of the scenario, broadening `if pgMode.OutOfRecovery() { SetSuspectedDivergence(true) }` to trigger even in recovery mode if the last-written rule in WAL names this pooler as leader

go/test/endtoend/multipooler/backupfaults/backup_archive_lag_test.go
```diff
@@ -405,23 +405,37 @@ func logWALState(t *testing.T, setup *shardsetup.ShardSetup, primaryName string)
...
-// has fully settled (all recovery scenarios resolved).
+// promotion clears it on the new primary.
+//
+// A newly-restored standby that joins the cohort via multiorch's lightweight
+// "PoolerNotInCohort" reconcile (UpdateConsensusRule on the primary), rather
+// than a full Recruit, does not get restore_command cleared synchronously — the
+// clear then comes from that node's own monitor backstop
+// (remedialActionDisableRestoreCommand), which is asynchronous. Poll until it
+// converges rather than reading once, so we tolerate that settling delay
+// instead of racing it.
```
mkindahl, line 416: This is a test-only fix to avoid a test failure. There is a real fix the the problem in #1307 that is discussed separately.

go/services/multipooler/internal/manager/postgres_monitor.go
mkindahl: But I think it can be a problem even for the case where a standby restarts as a standby and it has not been a leader before.

Consider this scenario:

1. One primary A. Three standbys B, C, D. Rule is {Primary: A, Cohort: [A,B,C,D], Ack: ≥3}.
2. Primary writes transaction T' and waits for two ACKs.
3. T' arrives to B, but not C or D.
4. A and B shuts down cleanly.
5. C is elected primary. Note that T' was not acked, so it is OK to discard it.
6. A and B comes back.
7. A sets SuspectDivergence, since it was previously a leader.
8. B does not set SuspectDivergence, since it was _not_ previously a leader.
9. A rewinds and connect to C (correct).
10. B does _not_ rewind, but connect to C (_not_ correct). B now has a transaction T' that did not exist before.

go/cmd/pgctld/command/server_test.go
```diff
@@ -174,6 +174,120 @@ func TestPgCtldServiceStart(t *testing.T) {
...
+//     before the start; the response reports it via CrashRecoveryRan.
+func TestPgCtldServiceStart_AsPrimaryCrashRecoveryMatrix(t *testing.T) {
+	cases := []struct {
+		name                 string
+		asPrimary            bool
+		allowCrashRecovery   bool
+		wantStandbySignal    bool
+		wantCrashRecoveryRan bool
+	}{
+		{"standby, no crash recovery", false, false, true, false},
```
dweitzman: Having these be named fields in the struct would be easier to read

go/test/endtoend/queryserving/pgbouncertests/backend_restart_test.go
```diff
@@ -98,20 +103,24 @@ func TestBackendCrashRecoveryUnderLoad(t *testing.T) {
...
-	// count to climb well past where it was at recovery time, with no client
-	// reconnect or intervention.
-	atRecovery, _ := validator.Stats()
-	require.Eventually(t, func() bool {
-		s, _ := validator.Stats()
-		return s >= atRecovery+30
-	}, 30*time.Second, 200*time.Millisecond,
-		"writes should resume on their own after the backend is back (transparent reconnect)")
-
+	// Without an orchestrator in this isolated cluster, the crashed primary comes
```
dweitzman, line 106: Something worth considering outside of this PR is whether we want to let pooler self-promote if they recognize they should be the leader according to the information they have in postgres_monitor

Currently we choose to resign rather than re-promote, but it would be technically possible to let a pooler re-promote itself if it's the named leader in the rule store

dweitzman (approved):

go/cmd/pgctld/command/server_test.go
mkindahl: Added.

go/test/endtoend/queryserving/pgbouncertests/backend_restart_test.go
mkindahl: Hmm... and idea, but I suspect there are some risks in accidentally get two primaries if one server replicate a rule change that was not acknowledged, temporarily gets partitioned, triggering another primary, and then coming back again (a little hand-wavy, I agree).

PR #1288, opened 2026-07-21
feat(query-serving): report ParameterStatus for in-transaction SET/RESET and ROLLBACK

GuptaManan100:
## Summary

- Reports `ParameterStatus` for a `SET`/`RESET` of a `GUC_REPORT` parameter issued **inside a transaction**, and for the reverts on **`ROLLBACK`** — the in-transaction follow-up to the base ParameterStatus support, matching PostgreSQL's behavior and timing.
- Treats PostgreSQL as authoritative for reserved (transaction) connections: routes the real statement to the pinned backend and forwards the backend's own `ParameterStatus` messages, rather than reconstructing them gateway-side.
- Adds a `parameter_status` field to `QueryResult` so the multipooler forwards the backend's `ParameterStatus` across gRPC; the multigateway relays it with the existing change-detection dedup, over both the simple and extended protocols. `ROLLBACK`/`COMMIT` reverts ride the same field on `ConcludeTransactionResponse.result`, so no extra proto surface is needed.

## Problem

Outside a transaction, SET/RESET already report `ParameterStatus`. Inside a transaction the report was missing: an in-transaction RESET only updated the gateway's local tracking, and the backend caught up lazily on the *next* query via internal settings reconciliation — whose `ParameterStatus` is never surfaced. So the client was never told, and the report (when it eventually happened) was at the wrong time relative to PostgreSQL.

## Solution

For reserved connections, route the real SET/RESET to the pinned backend so PostgreSQL emits the correct `ParameterStatus` immediately — including the reverts on `ROLLBACK` (and `ROLLBACK TO SAVEPOINT`, which already routes). The pgprotocol client surfaces those updates as results, the multipooler forwards them on the query result, and the wire server emits them after `CommandComplete` and before `ReadyForQuery`, matching PostgreSQL's order.

Routing an in-transaction RESET reverts the real backend GUC, so the multipooler must record the reverted state or its per-connection settings tracker drifts from the backend — a later checkout would then trust a stale value and hand back a connection whose actual GUC state does not match. Fixed by giving the RESET path the same `previewPostSessionSettings` the in-transaction SET path already has (dropping the variable from the recorded settings), keeping the pooler's connstate in sync with the backend.

New end-to-end tests compare the reported values on the wire against a direct PostgreSQL connection for in-transaction SET, in-transaction RESET, and ROLLBACK (revert-to-prior-value and revert-to-default).

### DIFF
--- a/go/common/pgprotocol/client/parameter_status_test.go
+++ b/go/common/pgprotocol/client/parameter_status_test.go
@@ -1,0 +1,72 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+import (
+	"bytes"
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/multigres/multigres/go/common/pgprotocol/protocol"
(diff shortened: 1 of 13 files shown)


haritabh17 (approved):
Looks good 🚀

PR #1289, opened 2026-07-21
feat(multiorch): collective failover recruitment backoff

dweitzman:
Implements **collective recruitment backoff**: each orchestrator derives the same escalating, staggered retry
schedule from shared, observed consensus state — no orch-to-orch coordination.

The backoff is based on the outgoing decision, so successive failed attempts to establish a new leader will keep backing off exponentially until some leader establishment succeeds.

This allows us to make a first failover attempt aggressively but keep backing off in case, for example, failover is possible but establishment needs more time than we might typically expect. Maybe progress depends on a pooler doing something potentially slow like taking a checkpoint, restoring from backup, a pg_rewind, a flaky network-mounted disk, etc., and if we keep interrupting failovers too frequently it could prevent the slow operation from eventually succeeding.

- **`TermRevocation.recruit_intent`** (`RecruitIntent`) carries `{replace_decision,
  attempt}`: the decided baseline the recruit is moving past, and how many
  consecutive recruits have targeted it.
- **`NewTermRevocation`** carries `attempt` forward while `replace_decision` is
  unchanged (successive *undecided* attempts against the same decision escalate
  the backoff instead of rapid-firing stuck failovers), and resets it when the
  decision advances or the prior recruit is stale. `replace_decision` is the
  marked decision, distinct from `outgoing_rule`, so a stuck proposal can't reset
  the count.
- **`go/common/ha`** — `BackoffSchedule.NextAttempt(rev, orchID)` =
  `coordinator_initiated_at + backoff(attempt) + jitter(orchID, replace_decision,
  attempt)`. Pure/deterministic, so orchs converge and it doesn't drift across
  recovery ticks or reset on restart.
- **Recovery loop** — failover gates on this backoff instead of the local grace
  period; a shard with no observed revocation acts immediately (aggressive-first).


simonbtomlinson (approved):
This looks good to me! I think the collective backoff approach is quite elegant.

PR #1292, opened 2026-07-22
test(e2e): scale wait budgets under coverage; quarantine TestDeadPrimaryRecovery

mkindahl:
## Problem

The **Code Coverage** workflow is flaky. Specifically the **Direct test coverage** job runs `go test -cover -covermode=atomic -coverpkg=./...`, which instruments the *test process itself* and makes the end-to-end suite run ~2-3x slower. Timing-sensitive bootstrap/failover tests then intermittently breach wait/deadline bounds that a normal run clears comfortably — failing with connection-refused / timeout / context-deadline errors, not on logic. On `main` the workflow fails intermittently among successes; the same tests pass in the integration job (which retries per-test) and in the Subprocess coverage job.

The codebase already compensated for coverage slowness in *one* spot (`process.go`, gated on `GOCOVERDIR`), but `GOCOVERDIR` is set only for the **Subprocess** job. The **Direct** job uses `-coverpkg` and sets no `GOCOVERDIR`, so its slowdown was never compensated — which is exactly the job that flakes.

This continues the effort started in #1272 (MUL-998), which stabilized two other Code Coverage-job flakes.

## Fix — coverage-aware timeout scaling (test-timing hardening, not masking)

New `go/test/utils/coverage.go`:

- `RunningUnderCoverage()` = `testing.CoverMode() != "" || os.Getenv("GOCOVERDIR") != ""` — detects **both** coverage jobs (the Direct job via `CoverMode`, which the old `GOCOVERDIR`-only check missed).
- `ScaleTimeout(d)` = `d` normally, `d * 3` under coverage.

Routed through the shared end-to-end wait/deadline helpers so every bound widens automatically under coverage with **zero effect on normal runs**:

- `utils.WithTimeout` / `WithShortDeadline`
- `shardsetup.EventuallyPoolerCondition` / `EventuallyPoolersCondition` (covers `WaitForNewPrimary`, `WaitForHigherTermPrimary`, `waitForShardReady`, …)
- `RequireRecovery`, `WaitForManagerReady`, `WaitForEvent`, `WaitForPortReady`, `WaitForHealthStreamsEstablished`
- the failover-family inline `require.Eventually` timeouts (skipping the `assert.Never` negative checks)
- unified the `process.go` verify-replication bump to fire for both coverage jobs

Widening a ceiling never slows the happy path — it only defers when a genuinely stuck operation is declared failed. Negative-path timeout tests still time out (their operations never complete), just later.

## Verification

Run under the Direct-coverage configuration locally (`-cover -covermode=atomic -coverpkg=./...`, real Postgres), `-count=10`:

- `TestBootstrapInitialization` — **10/10**
- `TestBootstrap_ViaExternalAPI` — **10/10**
- `TestEncryptedClusterBootstrap` — **10/10**

The other failover tests (`fix_replication`, `demote_stale_primary`, `rewind_diverged_replica`, `graceful_shutdown`, `cohort_rotation`, `stuck_proposal`) receive the same timeout-scaling hardening as a bonus; they were not in the reported flaky set and were not individually stress-verified here.

## Quarantine: `TestDeadPrimaryRecovery` (under coverage only)

Its residual coverage failure is **not** a timing flake and is **not** fixable by widening timeouts. Under the wider coverage timing:

1. A killed primary's postgres restarts as a **stale primary** (the monitor's normal auto-restart) before a multiorch coordinator that still holds a **stale leader view** has adopted the new term.
2. That coordinator issues an **un-termed `RewindToSource`** demoting the *current* (higher-term) primary to follow the resurrected killed node.
3. The current primary resigns → a re-election **re-promotes the killed node** instead of it rejoining as a standby, so the test (which asserts the killed node rejoins as a standby) hangs.

This is a real coordinator term-awareness bug (cascade re-election, MUL-557 family), and notably **not** a regression of the WAL-replay-completion fix (that gate runs and passes correctly, because by term-4 the killed node has been rewound and caught up). Masking it by loosening the test would hide the bug, so the test is **skipped under coverage with a logged reason** and tracked for a separate fix. It continues to run — with per-test retries — in the non-coverage **Run full integration test suite** job.

## Notes

- No CI workflow changes; no changes to the server-log-derived regression bounds (`maxRecruit`/`maxPromote`), which measure the non-instrumented service binaries in the Direct job.
- The concurrent start-postgres-as-standby change (#1287) likely *eliminates* the `TestDeadPrimaryRecovery` trigger (a killed node restarting as a standby rather than a stale primary), rather than amplifying it.

### DIFF
--- a/go/test/endtoend/shardsetup/pooler_diagnostics.go
+++ b/go/test/endtoend/shardsetup/pooler_diagnostics.go
@@ -26,6 +26,7 @@ import (
 	clustermetadatapb "github.com/multigres/multigres/go/pb/clustermetadata"
 	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 
+	"github.com/multigres/multigres/go/test/utils"
 	"github.com/multigres/multigres/go/tools/testtiming"
 )
 
@@ -95,6 +96,9 @@ func EventuallyPoolersCondition[T any](
 	msgAndArgs ...any,
 ) T {
 	t.Helper()
+	// Widen the wait budget under coverage instrumentation (see ScaleTimeout);
+	// the poll interval is left unscaled.
+	timeout = utils.ScaleTimeout(timeout)
 	var result T
 	require.Eventually(t, func() bool {
 		statuses := fetchPoolerStatuses(t, poolers)
@@ -133,6 +137,10 @@ func EventuallyPoolerCondition(
 	msgAndArgs ...any,
 ) {
 	t.Helper()
+	// Widen the wait budget under coverage instrumentation, where the cluster
+	// settles more slowly. The poll interval is left unscaled so diagnostics
+	// still log at the same cadence.
+	timeout = utils.ScaleTimeout(timeout)
 	require.Eventually(t, func() bool {
 		failures := checkPoolerCondition(t, poolers, condition)
 		for _, f := range failures {
@@ -230,7 +238,9 @@ func WaitForNewPrimary(t *testing.T, setup *ShardSetup, oldPrimaryName string, t
 		},
 		"new primary not elected within %v", timeout,
 	)
-	testtiming.Record(t, fmt.Sprintf("failover: %s → new primary", oldPrimaryName), time.Since(start), timeout)
+	// EventuallyPoolersCondition scales the wait budget under coverage, so record the
+	// scaled limit to keep the timing report's elapsed/limit ratio meaningful.
+	testtiming.Record(t, fmt.Sprintf("failover: %s → new primary", oldPrimaryName), time.Since(start), utils.ScaleTimeout(timeout))
 	return name
 }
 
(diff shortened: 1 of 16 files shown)


go/test/endtoend/shardsetup/pooler_diagnostics.go
```diff
@@ -95,6 +96,9 @@ func EventuallyPoolersCondition[T any](
 	msgAndArgs ...any,
 ) T {
 	t.Helper()
+	// Widen the wait budget under coverage instrumentation (see ScaleTimeout);
+	// the poll interval is left unscaled.
+	timeout = utils.ScaleTimeout(timeout)
```
Verolop, line 101: I think the timeout scaling here causes the reported timing limit to drift from the actual one, e.g. `WaitForNewPrimary` may wait up to 90s under coverage, but its failure message and `testtiming.Record` still use 30s. That could make a successful 45s failover appear as 150% of its limit in the timing summary, right?. Could you make sure the scaled timeout is also used for diagnostics and timing records?

Verolop (approved):
non blocking comment; lgtm!

go/test/endtoend/shardsetup/pooler_diagnostics.go
mkindahl: Good idea. Fixed.

PR #1293, opened 2026-07-22
fix(multipooler): reject stale-source RewindToSource with a term-aware guard

mkindahl:
## What

Make `RewindToSource` independently safe with a term-aware staleness gate, mirroring the existing `SetPrimary` gate.

`RewindToSource` restarts the target as a standby of the source, but the RPC carried **no consensus position and performed no validation** — its safety relied entirely on the caller always passing `SetPrimary`'s staleness gate first. A future or refactored caller that invoked `RewindToSource` directly, or against a live-but-stale source, could demote a **higher-term primary onto an older node** (the failure mode tracked as MUL-1091).

## Changes

- **proto**: add optional `source_position` (`RulePosition`) to `RewindToSourceRequest`.
- **multipooler** (`internal/manager/rpc_manager.go`): `RewindToSource` reads its own **fresh** position via `consensusMgr.Rules().ObservePosition` and refuses with `FAILED_PRECONDITION` when the source rule is revoked by this node's promise, or when `source_position` does not **strictly dominate** this node's position. Compared by rule position only (LSN is not part of the gate, same as `SetPrimary`). `source_position` is optional; when unset the gate is skipped, so existing callers are unaffected.
- **multiorch** (`recovery/actions/fix_replication.go`): `tryPgRewind` passes the shard's highest-known rule position as `source_position`; on a term-based refusal the RPC errors and the next recovery cycle retries rather than demoting the node.

## Why defense-in-depth (not a live bug on main)

MUL-1091 is **already prevented on `main`**: `FixReplication` calls `SetPrimary` first, whose `ObservePosition` + revocation gate no-ops a stale demote, and `recruit_blocked_until` blocks re-promotion of a rewound node. This change removes the reliance on caller ordering so `RewindToSource` is safe on its own.

(The original scalar-`consensus_term` fix was developed against the pre-reorg consensus model on the `start-postgres-as-standby` line, where the bug is live; on `main`'s rule-position/revocation model the guard is expressed in terms of `RulePosition` instead.)

## Testing

- New `TestRewindToSource_StalenessGate` (table): a source position older-than / equal-to this node's fresh position is refused before any postgres interaction.
- Existing `RewindToSource` tests pass unchanged (they pass `nil` source_position → gate skipped).
- `go build ./...`, `go vet`, and the `internal/manager`, `recovery/actions`, `grpcconsensusservice`, and `rpcclient` suites pass.

Refs: MUL-1091

### DIFF
--- a/go/pb/multipoolermanagerdata/multipoolermanagerdata.pb.go
+++ b/go/pb/multipoolermanagerdata/multipoolermanagerdata.pb.go
@@ -2615,9 +2615,18 @@ func (x *BackupMetadata) GetPgVersion() string {
 type RewindToSourceRequest struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
 	// Source multipooler (the primary) to rewind to
-	Source        *clustermetadata.Multipooler `protobuf:"bytes,1,opt,name=source,proto3" json:"source,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	Source *clustermetadata.Multipooler `protobuf:"bytes,1,opt,name=source,proto3" json:"source,omitempty"`
+	// source_position is the rule position the caller believes the source holds
+	// (the shard's highest-known rule). RewindToSource restarts this node as a
+	// standby of source, so the target refuses the rewind when its own fresh rule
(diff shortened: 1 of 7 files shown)


mkindahl:
Seems to be redundant for the time. Reopen if needed.

PR #1294, opened 2026-07-22
feat(multiorch): split leader-failure detection into explicit cause codes

dweitzman:
Replaces the heuristic in `LeaderNeedsReplacementAnalyzer` with an evidence-based, two-axis model.

**Cause** — why the leader must be replaced:
- `LeaderResigned`, `LeaderUnhealthy` — first-hand from the leader; no quorum gate.
- `LeaderUnreachableByCohort` — observer-derived (leader not directly observed live). Convicted only on *positive, conclusive* evidence: a durability-sufficient set of followers that (a) know this leader — by their highest-known rule, which may come from their replication primary, not just their own WAL — (b) have had time to connect (a connect grace), and (c) are configured to follow it yet are not streaming. This is the dual of `CheckSufficientRecruitment`'s revocation check, so detection requires the same conclusive revocation the failover itself would — orch never convicts on absence of evidence (e.g. a freshly (re)started orch that hasn't yet observed the cohort streaming). A follower's relation to the leader is a single classification: connected / cut-off / adapting (learned within the grace) / unaware.
- `LeaderUnspecified` — a cohort with no named leader.

**Feasibility** — whether a failover could actually succeed (`CheckSufficientRecruitment`: a strict majority of the cohort reachable, the remainder unable to satisfy the policy). Non-actionable outcomes are alert-only:
- `ShardStuck` — the leader needs replacing but no sufficient recruitment quorum is reachable.
- `ShardAtRisk` — leader healthy, but could not be recovered if lost (warning).
- `NoHealthyCohortMembers` — no fresh, valid health from any cohort member, so orch is blind and can't judge the leader; surfaced instead of convicting on stale evidence. Often a transient orch-side gap that clears on its own. (Renamed from `ShardHealthUnknown`.)
- `LeaderHealthUnknown` — a reachable cohort, but orch can neither confirm the leader is serving a quorum nor that it is cut off from one. A `PriorityNormal` warning (the leader may be fine): transient at cold start, persistent means orch has lost sight of the leader with an ambiguous cohort.

Also: detection keys on observation freshness rather than `IsLastCheckValid` (a brief stream blip no longer reads as failure); a single streaming follower is treated as proof the leader is alive; the promotion-suppression window is bounded by rule age. Non-actionable alerts use a no-op `AlertOnlyAction`.

`StaleLeader` moves to the replica problem group (a deposed leader is fixed by demoting it to a replica).


go/services/multiorch/recovery/analysis/leader_needs_replacement_analyzer.go
dweitzman, line 1: The significant changes are in this file

go/services/multiorch/recovery/analysis/leader_needs_replacement_analyzer.go
```diff
@@ -209,177 +193,438 @@ func leaderPromoting(sa *ShardAnalysis) bool {
...
-				"leader_postgres_running", leaderPGRunning,
-			)
+	excludeKey := topoclient.ComponentIDString(exclude)
+	var recruited []*clustermetadatapb.ID
+	for _, m := range cohort {
+		if exclude != nil && topoclient.ComponentIDString(m) == excludeKey {
+			continue
+		}
+		pa, ok := byID[topoclient.ComponentIDString(m)]
+		// observationFresh, not IsLastCheckValid: the latter flips false the instant a
```
Verolop: `observationFresh` allows a short stream disconnect, but I saw that`pa.IsInitialized()` requires `IsLastCheckValid`, which becomes false immediately. Is it possible to check the init state here so the freshness window still works?

go/services/multiorch/recovery/analysis/leader_needs_replacement_analyzer.go
```diff
@@ -209,177 +193,438 @@ func leaderPromoting(sa *ShardAnalysis) bool {
...
+		// health stream drops, so a momentary blip would wrongly drop a reachable
+		// member from the recruitment set (and could falsely trip ShardStuck/AtRisk).
+		// Freshness ages out only when the observation is genuinely stale.
+		if ok && observationFresh(pa, sa.Now, sa.Policy.LeaderLivenessFreshness) && pa.IsInitialized() {
```
Verolop: what happens with the membersthat are `INELIGIBLE` or have an pending `RecruitBlockedUntil`?

Verolop (commented):
looks great! left two small comments. Let me know if those are addressed somewhere else/ not matter.

go/services/multiorch/recovery/analysis/leader_needs_replacement_analyzer.go
dweitzman: Merged in #1374 to move from IsLastCheckValid/IsInitialized() to HealthWithin(now, maxAge) elsewhere and here

go/services/multiorch/recovery/analysis/leader_needs_replacement_analyzer.go
dweitzman: Factored out a shared `recruitable()` predicate and renamed this method from `reachableCohort` to `recruitableCohort`

go/test/endtoend/multiorch/rewind_diverged_replica_test.go
```diff
@@ -164,6 +165,28 @@ func TestRewindDivergedReplica(t *testing.T) {
 	// Re-enable postgres restarts so multipooler manages R1 going forward
 	resumeRestarts()
 
+	// Confirm P is the writable primary.
+	var inRecovery bool
+	err = primaryDB.QueryRow("SELECT pg_is_in_recovery()").Scan(&inRecovery)
+	require.NoError(t, err, "should query primary recovery status")
+	require.False(t, inRecovery, "primary should be writable, not in recovery")
+
+	// R1's pre-rewind rule position becomes the floor it must advance past to
```
dweitzman, line 174: Drive-by fix: now that `ShardAtRisk` is a detected problem, this rewind test needs to actually bump the rule number when it creates a divergence to make sure the rewound replica passes the recruitment floor and becomes recruitable again

It wasn't happening here because this wasn't a real failover, it was manually-created divergence

(Even manually-created divergence would heal in practice as long as the LSN gets high enough after the rewind)

Verolop (approved):
lgtm! Thank you 💚

PR #1295, opened 2026-07-22
fix(multigateway): match statement_timeout out-of-range error to PostgreSQL

Ketan-Karki:
## Summary
- Align `outOfRangeParamError` range bounds with live PostgreSQL: `(0 .. max)` instead of `(0 ms .. max ms)`
- Update unit expectations; verified by `TestStatementTimeoutErrorParity`

## Test plan
- [x] `go test -short ./go/services/multigateway/handler/ -run TestParsePostgresInterval`
- [x] `TestStatementTimeoutErrorParity` (pgregresstest e2e)

### DIFF
--- a/go/services/multigateway/handler/statement_timeout.go
+++ b/go/services/multigateway/handler/statement_timeout.go
@@ -142,13 +142,13 @@ func invalidParamError(paramName, value, hint string) *mterrors.PgDiagnostic {
 
 // outOfRangeParamError returns a PgDiagnostic for an out-of-range statement_timeout
 // value (SQLSTATE 22023). ms is the value in the base unit (milliseconds). The
-// message matches PostgreSQL's guc.c exactly: the "ms" unit is appended to the
-// value and to both range bounds, e.g.
+// message matches PostgreSQL: "ms" is appended to the value, but range bounds
+// are bare integers, e.g.
 //
-//	-1 ms is outside the valid range for parameter "statement_timeout" (0 ms .. 2147483647 ms)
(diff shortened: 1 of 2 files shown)


Ketan-Karki:
Hi - gentle bump. First-time contributor, CI is still waiting on workflow approval for this fork PR. Happy to address review feedback once checks can run. Thanks!

Ketan-Karki:
Following up on this - still waiting on maintainer approval so CI can run. Happy to rebase or address review comments if anything is needed. Thanks for taking a look when you have time.

GuptaManan100:
I am not exactly sure why you think Postgres produces `(0 .. max)` instead of `(0 ms .. max ms)`. I checked Postgres code too where the error is produced - https://github.com/postgres/postgres/blob/REL_17_5/src/backend/utils/misc/guc.c#L3183-L3190.

Additionally, we already have `TestStatementTimeoutErrorParity` test to verify our error matches exactly what Postgres produces. This change is incorrect.

Ketan-Karki:
> I am not exactly sure why you think Postgres produces `(0 .. max)` instead of `(0 ms .. max ms)`. I checked Postgres code too where the error is produced - https://github.com/postgres/postgres/blob/REL_17_5/src/backend/utils/misc/guc.c#L3183-L3190.
> 
> Additionally, we already have `TestStatementTimeoutErrorParity` test to verify our error matches exactly what Postgres produces. This change is incorrect.

Thanks for checking the Postgres source and clarifying. You're right, I misread the original error. Will be more careful next time.

PR #1298, opened 2026-07-22
fix(multipooler): detect unreadable pgbackrest repos via status code

rafael:
## Summary

- `ListBackups`, `FindByJobID`, and the backup health poller's `refreshFromRepo` silently misread an unreadable pgBackRest repository (e.g. a cipher-key mismatch) as "zero backups" instead of surfacing an error.
- Added `repoStatusError`, which checks pgBackRest's own `repo[].status.code` and treats any code other than the known-benign ones (ok, missing stanza path, no valid backups) as an error.
- Added regression tests for `ListBackups` and `refreshFromRepo` covering this case directly, without depending on real pgbackrest binaries.
- After this change, this test passes consistently in my local:  `TestJoin_FailsWithWrongCipherKey`. I was able to repro the failure before the change.

## Problem

`pgbackrest info` always exits 0, even when it cannot decrypt/read a repository — the failure is reported via `repo[].status.code` inside the JSON body, not through a nonzero exit code or command error. The Go struct decoding that output never captured `repo[].status`, so an unreadable repo decoded as an empty `backup: []` list, which was read as "no backups yet" rather than "repo unreadable."

Whether this got masked depended on an unrelated coincidence: some of pgbackrest's own error messages for this case splice a raw, unescaped snippet of the repo's undecrypted ciphertext directly into the JSON message string. That snippet sometimes contains bytes illegal inside a JSON string, which breaks decoding of the whole document — in that case the pre-existing decode-error path happened to surface an error anyway. This made `TestJoin_FailsWithWrongCipherKey` flaky: it passed most of the time by accident (the decode happened to fail), and occasionally failed when the decode happened to succeed, letting the monitor read the unreadable repo as empty and attempt to bootstrap over it.

## Solution

Added `pgBackRestRepoStatus`/`pgBackRestRepoState` to actually capture `repo[].status`, and `repoStatusError` to treat any status code other than the known-benign set (0 ok, 1 missing stanza path, 2 no valid backups) as a real error. Wired into `ListBackups`, `FindByJobID`, and `refreshFromRepo`, so an unreadable repo is never silently treated as an empty one, regardless of whether pgbackrest's diagnostic message happens to decode as JSON.

## Test
* Added a few more tests to capture this.
*  I validated this change by running:
```
MULTIGRES_PORT_POOL_ADDR=/tmp/multigres-port-pool.sock go test -run TestJoin_FailsWithWrongCipherKey -count=10 -v -timeout=15m ./go/test/endtoend/multipooler/backupfaults/...
```

### DIFF
--- a/go/services/multipooler/internal/manager/backup/backup_health.go
+++ b/go/services/multipooler/internal/manager/backup/backup_health.go
@@ -334,6 +334,16 @@ func (e *Engine) refreshFromRepo(ctx context.Context) (reachable bool, reason st
 		e.logger.DebugContext(ctx, "backup health: failed to parse pgbackrest info; keeping cached age/count", "error", err)
 		return false, ReadyReasonRepoUnreachable
 	}
+	// `info` always exits 0 and always attaches the correct status code, even
+	// when a repository can't be evaluated (e.g. a cipher-key mismatch) — but
+	// getting here means the JSON above happened to decode (see decodeInfo for
+	// when it doesn't). The failure still shows up as backup:[] with no decode
+	// error, so it must not be read as "zero backups" — that would report
+	// readiness as reachable for a repo we actually can't read.
+	if err := repoStatusError(infoData); err != nil {
+		e.logger.DebugContext(ctx, "backup health: repo status error; keeping cached age/count", "error", err)
+		return false, ReadyReasonRepoUnreachable
+	}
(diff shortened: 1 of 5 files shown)


cuongdo (approved):
Thanks for plugging up this hole in backup error handling!

PR #1300, opened 2026-07-22
feat(consensus): pooler-driven divergence recovery + reconnect over-recruited followers

dweitzman:
## Summary

Two related consensus-recovery changes: multiorch no longer drives destructive `pg_rewind` on replicas, and a spuriously over-recruited follower can rejoin without a failover.

### Pooler-driven divergence recovery
`FixReplication` no longer escalates to an orch-driven `pg_rewind`. When `SetPrimary` doesn't bring a standby to streaming, it leaves divergence recovery to the pooler. The multipooler monitor detects a standby that is up with a correct `primary_conninfo` but not streaming past a short threshold, marks suspected divergence, and self-rewinds against the recorded leader (revocation-gated). An actively streaming standby clears any lingering suspicion, which also breaks any rewind loop. `discoverPostgresState` now surfaces a WAL-receiver read failure instead of defaulting to "not streaming".

### Reconnect an over-recruited follower
A follower `Recruit()`'d at a higher term raises its `TermRevocation` above the leader's committed rule, so it rejects the leader's `SetPrimary` and sits disconnected while the leader keeps serving — an abandoned partial failover.

multiorch now detects this (a self-revoked follower whose revocation still revokes the highest-known rule, once past the failover grace window so a legit in-flight failover is left to finish) and repairs it with a leader-led no-op rule advance: `UpdateConsensusRule(RULE_OPERATION_ADVANCE)` rewrites the rule with the same leader and cohort at a fresh `leader_subterm`, moving the committed decision past the rule the revocation was authored to transition away from. That defeats the revocation via the runaway-recruit override in `IsRuleRevoked`, without changing the coordinator term or any pooler's revocation. The action then relays the advanced rule via `SetPrimary` and the follower rejoins.

### Supporting changes
- `consensus.IsSelfRevoked`, keyed on the highest-known rule so a replica streaming toward a higher accepted rule while its WAL lags is not misread as stranded.
- `RULE_OPERATION_ADVANCE` on `UpdateConsensusRule`; renamed `CohortUpdateOperation` → `RuleOperation` now that it carries a non-cohort operation.
- `ReplicaNotReplicating` defers a stranded follower to the new analyzer instead of firing a `SetPrimary` it would ignore.

### DIFF
--- a/go/common/consensus/revocation.go
+++ b/go/common/consensus/revocation.go
@@ -138,6 +138,35 @@ func IsRuleRevoked(position *clustermetadatapb.RulePosition, revocation *cluster
 	return true
 }
 
+// IsSelfRevoked reports whether a node has no rule it can act on under the
+// revocation it has accepted: its highest known rule is still revoked by its own
+// promise (revoked_below_term with the recorded outgoing_rule). This is the
+// "recruited but not caught up and not yet re-ruled" state — after a Recruit
+// raises the promise floor, a node reports this until it accepts a rule at or
+// beyond that term.
+//
+// It keys on the highest known rule — the max of the node's own WAL position and
+// the rule last delivered via SetPrimary/Promote (ReplicationPrimary) — not just
+// the WAL position: a replica that has accepted a higher rule is following it and
+// is not stranded even while its WAL position lags behind. A rejected SetPrimary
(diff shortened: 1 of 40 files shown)


dweitzman:
Closing

PRs #1303 and #1322 cover all of this

PR #1301, opened 2026-07-23
feat(sharding): draft logical/physical sharding data model

GuptaManan100:
## Summary

- Proposes a sharding data model that separates the **logical view** (`database → schema → table`) from the **physical representation** (storage clusters), joined by an explicit map — replacing the `ShardKey{database, table_group, shard}` conflation of what the user sees with where it physically lives.
- Adds `proto/sharding.proto`: `LogicalObject`, `StorageCluster`, `ShardingFunction` (a `oneof` interface where each kind stores only what it needs), `ShardingGroup` + `TableBinding`, `DirectPlacement`, and `ShardingSchema`.
- Adds a draft reference resolver (`go/common/sharding`) that turns a logical object + shard-key values into target storage clusters, with table-driven tests that double as worked examples.
- Adds `docs/sharding/design.md`: the model, worked examples, and open questions.

## Description

This is a **draft for team discussion**, not wired into query routing yet. The resolver — and its hash function in particular — are placeholders that exist to validate the proto model with runnable examples: unsharded databases, hash-sharded tables, co-located tables (local joins), reference tables, resharding as a layout edit that preserves coverage, and placement precedence (most-specific-wins).

The representation is deliberately maximally expressive: placement can be attached at database, schema, or table granularity, so any sharding scheme is representable and restrictions become a policy/UI concern rather than something baked into the data model. Open questions — naming of the physical unit (it collides with "cluster"), where the map is stored, precedence semantics, and the co-location test for join pushdown — are enumerated in the design doc.


docs/sharding/design.md
```diff
@@ -0,0 +1,423 @@
+# Sharding: Logical view, physical representation, and the map between them
+
+**Status: Draft for team discussion.** This proposes the data model for how
+Multigres describes where a user's data lives and how it is split. It defines
+the vocabulary and proto; it deliberately does _not_ specify DDL syntax, the
+resharding execution mechanism, or cross-shard 2PC (see [Out of scope](#out-of-scope)).
```
mkindahl, line 6: Suggestion: You have a number of goals with this design when it comes to what you want to support, for example:

- Supporting "reference tables" / "global tables". This affects the design, so it is good for the reviewer to know.
- Supporting sharding keys for all tables in a schema, including tables that are created after. This is the reason to why you have the "schema" level in the key as well.
-

docs/sharding/design.md
```diff
@@ -0,0 +1,423 @@
...
+
+## TL;DR
+
+Today `ShardKey{database, table_group, shard}` conflates two different things:
+_what the user sees_ and _where it physically lives_. This proposal splits them
+into three independent layers:
+
+1. **Logical view** - plain Postgres namespacing the user already knows:
+   `database → schema → table`. No sharding concepts leak in.
```
mkindahl, line 15: For general sharding, the schema is not needed since all table names can be qualified with the schema name. It might be good to clarify why you have this extra level and not just database + qualified table name.

docs/sharding/design.md
```diff
@@ -0,0 +1,423 @@
...
+2. **Physical representation** - named Postgres HA units ("storage clusters").
+   Each is one failover domain and can hold fragments of many logical objects.
```
mkindahl, line 17: Probably good to define "storage cluster" early. It is not entirely clear what it is. Is it a cell? A primary-standby "cluster"? Something else?

docs/sharding/design.md
```diff
@@ -0,0 +1,423 @@
...
+  repeated DirectPlacement  direct_placements = 3;
+  repeated TableBinding     table_bindings    = 4;
+  repeated TableSharding    table_shardings   = 5;
+}
+```
+
+## Worked examples
+
+Assume storage clusters **A, B, C, D**. Keyranges use Vitess notation: `-80` is
+`[0x00, 0x80)`, `80-` is `[0x80, +inf)`.
```
mkindahl, line 223: These are good, but one of the key issues is to take a query, pick out the necessary information, and use the metadata to decide where to send the query. Could you have some examples using queries as well?

I would argue that just because you can represent a concept does not mean it is meaningful, but if you have queries that are reasonable and common where you can demonstrate the benefits of the schema, it is very clear what the benefits of the schema is.

If would be particularly good to see:

- Point lookups with a simple shard key. These are the basics, but could introduction to understand the schema.
- Joins including a reference table. This is one of the goal you had with this schema.
- Joins between tables that have a foreign key relationship (the normal case). This we expect to work.
- Joins between tables that are not an eqjoin (for example the one below).
- Queries with outer joins, which could require you to send queries wider, so good to include.

If you support mapping multiple "sharding hierarchies" (unrelated schema of tables that serve different purposes), and want to do a join between tables in these two hierarchies, it would be good include such a query as well since this is also one of the goal of the metadata schema.

Example of a query that does not use eqjoin is a query that computes the running total of a table.
```sql
SELECT d1.date, d1.agent_fee, SUM(d2.agent_fee) AS total_agent_fee
FROM deals d1
JOIN deals d2
ON d1.date >= d2.date
GROUP BY d1.agent_fee, d1.date
ORDER BY d1.date;
```

docs/sharding/design.md
```diff
@@ -0,0 +1,423 @@
...
+are a policy/UI concern layered on top, not baked into the representation.
+
+**Why now:** `ShardKey` does not yet live in many places. Replacing it as the
+routing key is cheap today and gets more expensive every week we wait.
+
+## The model
+
+### 1. Logical view
+
+What the user reasons about is ordinary Postgres:
```
mkindahl, line 34: I would assume that information about sharding keys used for each table should be at the logical level as well.

mkindahl (commented):
A few first comments.

PR #1302, opened 2026-07-23
feat(query-serving): resolve SQL PREPARE names from server-side dynamic SQL

haritabh17:
## Summary

Makes a client's SQL `PREPARE` namespace resolvable from server-side dynamic
SQL by materializing it on whichever pooled backend serves the session.

An `EXECUTE` the gateway can parse is rewritten to the pooler's consolidated
`ppstmt*` name. But an `EXECUTE` issued *inside* a PL/pgSQL body — e.g.
`EXECUTE format('execute %s', ...)` in a function — runs entirely on the
backend: the gateway only ever sees the outer `SELECT`, so there is nothing to
rewrite, and PostgreSQL resolves the name against the backend session's own
prepared-statement namespace. This PR ships the session's logical namespace to
the backend so that resolution succeeds.

## Design

- Names created by **SQL-level `PREPARE`** become *aliases*: validated at
  PREPARE time (Describe round trip, backend resolution of parameter types the
  gateway cannot map statically), then attached to every routed query and
  portal execute via `ExecuteOptions.prepared_statement_aliases`.
- The multipooler's `reconcilePreparedAliases` makes the selected backend's
  client-visible namespace exactly match the session before the query runs:
  stale entries are closed, missing ones parsed, and a `ppstmt*` collision
  yields to the client-visible name. Protocol-level (wire `Parse`) names stay
  gateway-local and are never materialized.
- The handler exposes this through `server.PreparedStatementAliasProvider`, a
  named optional-capability interface with a compile-time conformance
  assertion.

## Lifecycle and safety guarantees

- **Session isolation.** The pool purges materialized aliases at borrow
  (`PurgePreparedAliases` on the pool `Connection` interface), and every
  backend-acquisition path — including both COPY directions — runs the
  reconciling validate. A recycled backend cannot expose a previous session's
  `PREPARE` names. Consolidated `ppstmt*` statements survive checkin by design.
- **Dormant aliases are harmless**, matching vanilla PostgreSQL where an
  invalidated `PREPARE` only errors at `EXECUTE`. A definition that fails to
  parse on a backend is recorded in a per-backend negative cache and not
  retried until the client re-`PREPARE`s it; inside an explicit transaction
  the parse is savepoint-guarded so a failure cannot abort the client's
  transaction.
- **Zero-alias sessions pay nothing**: reconcile returns before any clone or
  I/O when both alias sets are empty.
- **No silent retry.** Queries carrying aliases (or a materialized EXECUTE) do
  not use the reconnect-retry streaming variant, since a reconnect wipes the
  backend's PREPARE namespace.
- **Name collisions resolve safely.** A client alias legally named `ppstmtN`
  is evicted by `ensurePrepared` before parsing the consolidated statement
  (no 42P05), and re-materialized by the next reconcile if still wanted.

## Diagnostics

SQL `EXECUTE` errors are translated back to the client's statement name and
command text, with parameter positions computed against the client's own SQL.
PREPARE-time validation errors are position-shifted by the body's offset so
the caret lands on the full `PREPARE … AS …` statement, on both the Describe
and in-transaction Parse paths. Known edge: positions assume the deparsed body
matches the source byte-for-byte and are byte-based; multibyte or
deparse-divergent bodies can still misplace the caret.

## Parameter types

`PREPARE p (type, ...)` maps unqualified and `pg_catalog`-qualified names to
static OIDs; any other qualifier (which can shadow a builtin via a domain) and
unknown names are resolved through the backend via `regtype`.

## pg_regress baseline

- `prepare` and `partition_prune` pass with **no patch**.
- `plancache.patch` covers only the `pg_prepared_statements` plan-counter
  columns: executions run against the shared `ppstmt*` entry, so counters
  accrue there while the alias row reports zeros; results and plan-mode
  behavior match upstream.
- `xml.patch` covers only cursor-position normalization divergences.
- `explain.patch` records the `pg_temp`-qualified `CREATE FUNCTION` rejection
  (a temp-isolation policy predating this PR, #1360) and its cascade; not a
  prepared-statement divergence.

The design is documented in `docs/query_serving/prepared_statements_design.md`.

## Testing

- Unit coverage across `preparedstatement`, `multigateway`, `multipooler`:
  reconcile idempotency and negative cache, borrow-time purge, alias/physical
  name collisions, connection-locality and SQL-alias-only filtering, qualified
  type deferral, position arithmetic, and translation guards.
- End-to-end: dynamic `EXECUTE` through a PL/pgSQL `DO` block; queryserving and
  pgbouncer prepared-statement suites.
- CI-container `pg_regress` (Linux): `test_setup`, `prepare`, `plancache`,
  `xml`, `partition_prune` all pass with the patches as committed.

Remaining red regress tests (`brin`×3, `stats`) are pre-existing
backend-local-state issues outside this PR's scope. Known follow-up: avoiding
the gateway-wide consolidator lock for zero-alias sessions (needs a benchmark
to justify the bookkeeping).


haritabh17:
We have decided to postpone this for now. Hence closing this PR currently.

PR #1303, opened 2026-07-23
refactor(multipooler): remove RewindToSource RPC; self-heal diverged standbys locally

mkindahl:
## What

Removes the `RewindToSource` consensus RPC and moves diverged-standby repair into the pooler's own monitor.

A diverged standby used to be repaired by multiorch's `FixReplication` action issuing the `RewindToSource` RPC. With the monitor's automatic rewind-until-successful loop already in place, that cross-node RPC is redundant for every case except one: a standby that stays up, in recovery, with correct `primary_conninfo`, whose WAL receiver FATALs on a diverged timeline (phantom transactions). This PR closes that gap so the RPC can go.

## Changes

- **Local self-detection of a stuck diverged standby** (`standbyStuckDiverged`): in recovery, `primary_conninfo` points at the correctly-recorded rewind target, the WAL receiver has not streamed past a debounce threshold, **and the recorded leader is reachable**. It sets `suspectedDivergence`, routing the next monitor tick through the existing local rewind path (`restartAsStandbyLocked`) — gated on the leader being rewind-ready and rate-limited by backoff. The rewind reports `POSTGRES_ACTION_STARTING`.
- **Leader-liveness gate (load-bearing).** "Not streaming" happens both when the WAL diverged (leader alive, receiver FATALs) *and* when the leader is dead (a failover in progress). Only the first warrants a rewind. Marking divergence against a dead leader would stall the coordinator's election — its subsequent `SetPrimary` would take the stale-primary-demote path and defer. So the monitor concludes divergence only when a TCP dial confirms the leader is up, mirroring the old orch guard that refused `pg_rewind` unless the primary was running. The dial is injectable (`leaderReachableFn`) for tests.
- **Debounce window**: `Config.StandbyStuckDivergenceThreshold` (default `10s`) — only needs to outlast a transient reconnect now that the liveness gate excludes failover. It's an internal, programmatic override (no CLI flag); production uses the default. The e2e `FixReplication` recovery budget is widened to 45s to reflect that pooler-local self-heal legitimately takes longer than the old orch-immediate `RewindToSource` (measured ~20s locally vs the old ~5s).
- **`FixReplication` no longer escalates to `pg_rewind`**: it delivers the leader identity via `SetPrimary` and returns an error for retry when streaming doesn't start; the pooler self-heals. `tryPgRewind` is removed.
- **`RewindToSource` removed end to end**: proto (`consensusservice.proto`, `multipoolermanagerdata.proto` messages), regenerated pb, the manager method, the gRPC handler, and the `rpcclient` interface/impl/fake. `restartAsStandbyLocked` is kept (shared with `SetPrimary`'s stale-primary branch and the monitor).
- **`POSTGRES_ACTION_REWIND` enum removed**; `reserved 4` / `reserved "POSTGRES_ACTION_REWIND"` so the number/name can't be silently reused.

## Tests

- Added `TestStandbyStuckDiverged` (unit), including `UnreachableLeaderIsNotStuck` (pins the failover-vs-divergence distinction) and a `Config`-override case.
- Retargeted the `RewindToSource` manager regression tests to `restartAsStandbyLocked`; dropped the RPC-boundary `InvalidArgs` test.
- Reworked the `FixReplication` tests for the no-orch-rewind behavior.
- Updated the multiorch e2e (`TestRewindDivergedReplica`) and the backupfaults e2e (`TestPrimaryCrashWithUnarchivedWAL_NoDataLoss`) — comment + shortened threshold flag — so a diverged restore rejoins via the local self-heal within the recovery budget.

**Verification:** `go build ./go/...` clean, `make proto` produces no diff, gofumpt-clean. Unit suites pass. Integration (local): `TestRewindDivergedReplica`, `TestFixReplication`, both `TestDemoteStalePrimary_*`, `TestPrimaryGracefulShutdownTriggersFailover`, and the backupfaults regression test (which exercises *both* a failover and a diverged rejoin) all pass; backupfaults ran 3/3 clean.

## Note on the liveness gate

The gate resolves a subtlety the first cut missed: a pure time threshold cannot tell divergence (needs a *short* threshold to rejoin in budget) apart from failover (needs a threshold *longer* than failover so the old rule is revoked before a standby is mis-flagged). The distinguishing signal is leader liveness, not time — hence the dial. A false "alive" only costs a gated, rate-limited `pg_rewind` dry-run; a false "dead" just retries next tick.

## Follow-up (not in this PR)

Fully removing `FixReplication` remains gated on giving the pooler a topology-driven way to learn a new leader's **address** when it missed the consensus fan-out (a pooler learns the leader's host:port only from a pushed `SetPrimary`; the WAL-replicated rule carries only the leader ID). Tracked separately.

### DIFF
--- a/go/services/multipooler/internal/manager/postgres_monitor.go
+++ b/go/services/multipooler/internal/manager/postgres_monitor.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
+	"strconv"
 	"time"
 
 	commonconsensus "github.com/multigres/multigres/go/common/consensus"
@@ -106,10 +108,15 @@ func (pm *MultipoolerManager) staleStandbyDemoteTarget() *clustermetadatapb.Pool
 
 // postgresState represents the state of PostgreSQL for monitoring
 type postgresState struct {
-	pgctldAvailable          bool
-	dirInitialized           bool
-	postgresRunning          bool
-	backupsAvailable         bool
+	pgctldAvailable  bool
+	dirInitialized   bool
+	postgresRunning  bool
+	backupsAvailable bool
+	// connInfo is the standby's primary_conninfo (parsed + redacted), read once
+	// per tick in discoverPostgresState so the monitor's decision paths can reuse
+	// it without re-querying. Nil when not running as a standby or when the read
+	// or parse failed this tick.
+	connInfo                 *multipoolermanagerdatapb.PrimaryConnInfo
 	pgMode                   pgmode.Mode
 	bootstrapSentinelPresent bool
 	// rewindSourceReady is true when this pooler is a primary whose last completed
@@ -193,7 +200,31 @@ const (
 	// is restore-from-backup leaving it enabled for a freshly-restored
 	// observer (see WrapRestoreCommand).
 	remedialActionDisableRestoreCommand
+
+	// remedialActionMarkStandbyDiverged means postgres is up as a standby whose
+	// primary_conninfo already points at the correctly-recorded leader, yet it has
+	// been unable to stream from that leader for longer than
+	// standbyStuckDivergenceThreshold. The most likely cause is timeline
+	// divergence: the WAL receiver connects, gets FATAL, and exits, leaving
+	// postgres up but not streaming. We conclude the WAL diverged and set
+	// suspectedDivergence, which routes the next tick through
+	// remedialActionRewindToLeader (a pg_rewind dry-run — cheap when there is no
+	// divergence). This replaces orch's old explicit RewindToSource RPC: the
+	// standby now self-heals a stuck-replica scenario without orch in the loop.
+	remedialActionMarkStandbyDiverged
 )
+
+// standbyStuckDivergenceThreshold is the default for how long a standby must stay
+// unable to stream from its correctly-recorded (and reachable) leader before the
+// monitor concludes its WAL diverged and rewinds. It debounces transient
+// reconnects (a brief WAL-receiver drop, a leader still finishing its
+// post-promotion checkpoint) so a needless stop+rewind does not fire on a hiccup.
+// The leader-liveness gate (see standbyStuckDiverged) already excludes the
+// failover case, so this threshold need only outlast a transient reconnect, not a
+// whole failover. The subsequent rewind is itself gated on the leader being
+// rewind-ready and rate-limited by exponential backoff. Overridable via
+// Config.StandbyStuckDivergenceThreshold (see stuckDivergenceThreshold).
+const standbyStuckDivergenceThreshold = 10 * time.Second
 
 // monitorPostgresIteration performs one iteration of PostgreSQL monitoring.
 // Returns the discovered postgres state on success, or an error if the state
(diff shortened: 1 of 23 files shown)


go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -656,6 +682,121 @@ func (pm *MultipoolerManager) determineRemedialAction(ctx context.Context, curre
...
+		pm.standbyStuckSince.Store(0)
+		return false
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, 500*time.Millisecond)
+	defer cancel()
+
+	// Confirm primary_conninfo actually points at the recorded leader. If it
+	// doesn't, this is conninfo drift (handled elsewhere), not divergence.
+	connInfoStr, err := pm.readPrimaryConnInfo(ctx)
```
dweitzman: I have a soft preference to generally do things like reading primary conninfo or querying replication status earlier and put them into the `postgresState` struct, where practical, so the `determineXYZ` methods and their helpers are more deterministic and do fewer dynamic postgres queries

But I feel like there are tradeoffs either way, so I'll defer to your judgement

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -656,6 +682,121 @@ func (pm *MultipoolerManager) determineRemedialAction(ctx context.Context, curre
...
+		return false
+	}
+
+	// If the WAL receiver is streaming, we are not stuck.
+	status, err := pm.queryReplicationStatus(ctx)
+	if err != nil {
+		return false // can't verify; leave the debounce timer as-is
+	}
+	if status.GetWalReceiverStatus() == "streaming" {
+		pm.standbyStuckSince.Store(0)
```
dweitzman, line 770: How would you feel about moving `standbyStuckSince`, the threshold, etc into `ConsensusManager`? `MultipoolerManager` has a tendency to accumulate unlimited unrelated state fields, and I think we'll be happier if we keep consensus-related state under the `/consensus/` subpackage as much as practical

go/test/endtoend/shardsetup/setup.go
```diff
@@ -908,8 +908,12 @@ const (
 var recoveryScenarioTimeouts = map[RecoveryScenario]time.Duration{
 	RecoveryScenarioInitialSettle:      5 * time.Second,
 	RecoveryScenarioStalePrimaryDemote: 30 * time.Second,
-	RecoveryScenarioFixReplication:     30 * time.Second,
-	RecoveryScenarioEmergencyDemotion:  30 * time.Second,
+	// Repairing a diverged standby now runs through the pooler's local self-heal
+	// (a divergence-debounce window plus pg_rewind + restart) rather than orch's
+	// old immediate RewindToSource RPC, so it legitimately takes longer. Budget
+	// for that: the debounce default plus a rewind that runs slower under CI I/O.
+	RecoveryScenarioFixReplication:    45 * time.Second,
```
dweitzman: Just to push back gently on this: seems like there's no reason moving the detection from one place to another should make it slower? Should we reduce the window before stuck replication is detected by a pooler?

Seems like there are tradeoffs either way, but increasing the recovery time feels like it's not part of the intended scope of this PR

dweitzman (approved):

go/test/endtoend/shardsetup/setup.go
mkindahl: No, it does not make sense to change this unless we know it actually is needed. Reverted.

go/services/multipooler/internal/manager/postgres_monitor.go
mkindahl: I have no strong opinion regarding this except that we should be consistent. Since  the callers of `standbyStuckDiverged` has a `postgresState`, I added `connInfo` to the state and fetched it in `discoverPostgresState`, to be used here.

To reduce the number of queries, I think it would make sense to have a single query to fetch all information in `discoverPostgresState` and not, as it is now, multiple queries, but this is a separate refactoring out of scope for this PR.

go/services/multipooler/internal/manager/postgres_monitor.go
mkindahl: Hmm... not sure.

Looking at `ConsensusManager` it seems to deal mostly with the consensus state (rules, cohort, etc.), but this is a "replication/cluster status", i.e., more along the line of the state of Postgres replication configuration and status. I do notice that we have `suspectedDivergence` in the state though, which is kind of "replication state", but also notice that `primaryConnInfoDiffersFromRecorded` is already in `MultipoolerManager`, and this information is similar to `standbyStuckSince` in functionality.

If we move these, I think we should move all in a refactor, so right now I am inclined to keep it in `MultipoolerManager`.

I do like the idea of having a separate subpackage for "postgres/cluster configuration and status" things: it given a uniform interface to the information, which in turn will allow us to add caching and other logic to improve the system.

PR #1304, opened 2026-07-23
refactor(pgctld): consolidate standby.signal handling into shared helpers

mkindahl:
The Restart path and single-user crash recovery each wrote or removed standby.signal inline with duplicated os.WriteFile / os.Remove logic.

Introduce standbySignalPath, createStandbySignal, and removeStandbySignal and route those existing operations through them: hasStandbySignal, restart-as-standby, and the temporary remove/recreate around single-user crash recovery now share the helpers, which log consistently and return the resolved path. No behavior change.

Also mark the unused validateGlobalFlags parameters as _ to satisfy the linter.

### DIFF
--- a/go/cmd/pgctld/command/crash_recovery.go
+++ b/go/cmd/pgctld/command/crash_recovery.go
@@ -80,11 +80,43 @@ func crashRecoveryNeeded(ctx context.Context) (bool, error) {
 	return !cleanlyStopped, nil
 }
 
+// standbySignalPath returns the path to the standby.signal marker file inside dataDir.
+func standbySignalPath(dataDir string) string {
+	return filepath.Join(dataDir, constants.StandbySignalFile)
+}
+
 // hasStandbySignal reports whether a standby.signal marker file is present, i.e.
 // PostgreSQL is configured to start in standby mode.
 func hasStandbySignal() bool {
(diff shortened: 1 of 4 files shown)


go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -80,13 +80,45 @@ func crashRecoveryNeeded(ctx context.Context) (bool, error) {
 	return !cleanlyStopped, nil
 }
 
+// standbySignalPath returns the path to the standby.signal marker file inside dataDir.
+func standbySignalPath(dataDir string) string {
```
cuongdo, line 84: why does `dataDir` need to be a parameter? couldn't this function use `pgctld.PostgresDataDir()` as the existing code does?

mkindahl: I thought about that, and it works for all code except the tests, but the tests are creating a temporary directory and using that in the tests without setting up the `PostgresDataDir()`. It seemed easiest to just add it as a parameter. If you look at `TestRunCrashRecoveryInDir_RemovesAndRestoresStandbySignal`, for example, it doesn't use this function and does not set the environment variable, so it would not work there.

These are free-standing functions, which is kind of odd. It would be better to have them all as methods on, for example, `PostgresCtlConfig` and store the Postgres data directory in the instance and use that in the tests as well, but this is not how it is currently structured.

cuongdo (approved):

PR #1305, opened 2026-07-23
fix(multipooler): detect a stale leader from the rule store, not just pgMode

dweitzman:
`SetPrimary` decides whether to rewind before following a new primary by flagging suspected divergence when this pooler looks like a (possibly stale) leader whose WAL may have diverged. It previously keyed only on `pgMode.OutOfRecovery()` — postgres itself reporting primary.

A stale leader whose postgres was **already demoted to standby**, but whose committed rule still names it the leader, slipped through: `pgMode` reads standby, so no divergence was suspected, and it followed the new primary without rewinding its possibly-diverged WAL.

This also trips suspected divergence when the pooler's **own current-position rule names itself leader**. It keys on `CurrentPosition` (the rule replicated through this node's WAL), not the highest-known rule — a stale leader that has already been told about the new leader carries a `ReplicationPrimary` naming that leader, so a highest-known-rule role check would report not-leader and miss exactly this case.

### DIFF
--- a/go/services/multipooler/internal/manager/rpc_consensus.go
+++ b/go/services/multipooler/internal/manager/rpc_consensus.go
@@ -662,7 +662,7 @@ func (pm *MultipoolerManager) SetPrimary(ctx context.Context, req *consensusdata
 	}
 	eventlog.Emit(ctx, pm.logger, eventlog.Started, setPrimaryEvent)
 
-	resp, err := pm.setPrimaryLocked(ctx, req)
+	resp, err := pm.setPrimaryLocked(ctx, req, selfPos)
 	if err != nil {
 		eventlog.Emit(ctx, pm.logger, eventlog.Failed, setPrimaryEvent, "error", err)
 	} else {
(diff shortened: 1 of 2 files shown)


go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -685,7 +685,8 @@ func (pm *MultipoolerManager) setPrimaryLocked(ctx context.Context, req *consens
 		return nil, mterrors.Wrap(err, "failed to check recovery status")
 	}
 
-	if pgMode.OutOfRecovery() {
+	walRule := commonconsensus.PossiblyUndecidedRule(selfPos.GetPosition())
+	if pgMode.OutOfRecovery() || commonconsensus.RuleNamesLeader(walRule, pm.serviceID) {
```
mkindahl, line 689: It solves part of the problem, but if a previous standby is cleanly shut down and restarts, this would not have been a former leader but can still have transactions that do not exist on the current primary, so it would then not be rewound.

mkindahl (approved):
Approving since it solves part of the problem, but a restarting standby can also need rewind.

PR #1312, opened 2026-07-24
fix(query-serving): preserve DateStyle and savepoint state

haritabh17:
## Summary

- store PostgreSQL's complete reported `DateStyle` value so partial settings replay consistently on fresh backends
- retain original SET input for reportable GUCs whose values are not context-sensitive
- after `ROLLBACK TO SAVEPOINT`, synchronize reserved-connection metadata without issuing RESET/SET SQL that would overwrite surviving outer `SET LOCAL` values

Extracted from #1306 so reservation-aware RESET planning can be reviewed independently.

## Testing

- `go test -short -count=1 ./go/services/multigateway/engine ./go/services/multipooler/internal/executor`
- PostgreSQL compatibility suite triggered via PR label

### DIFF
--- a/go/common/pgsettings/reportable.go
+++ b/go/common/pgsettings/reportable.go
@@ -43,3 +43,11 @@ func ReportableGUCName(name string) (string, bool) {
 	display, ok := reportableGUCs[CanonicalGUCName(name)]
 	return display, ok
 }
+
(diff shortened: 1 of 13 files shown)


haritabh17:
Will do a refactor of the SET path in - https://github.com/multigres/multigres/pull/1306

PR #1313, opened 2026-07-24
feat(multipooler): track WAL-receive advance time on standbys

dweitzman:
## What

Adds a per-standby signal: **when `pg_last_wal_receive_lsn()` last increased**, surfaced on `StandbyReplicationStatus.last_receive_lsn_advance_time`.

The heartbeat reader already runs on standbys at ~1s. It now samples the WAL receiver's streamed position in the same query it uses for lag, and records the timestamp of each genuine *increase* (idle keepalives, which don't advance the LSN, leave it to age out).

## Why

This plumbs a signal that **LeaderStuck** failover detection will consume. `receive_lsn` advances **only** via streaming replication from the primary — `restore_command`/archive replay advances `replay_lsn`, not `receive_lsn` — so a fresh value is direct evidence the leader is actively streaming new WAL to this standby. A future analyzer will treat a durability-sufficient set of standbys with fresh advance times as proof the leader is making progress, and only fail over a reachable-but-stuck leader when no such progress exists.

No behavior change yet: this PR only plumbs the signal (reader → `queryReplicationStatus` → proto → orch health stream). The analyzer that consumes it lands separately.

## Why report the advance *time*, not just the LSN?

Orch could instead diff `last_receive_lsn` across its own health snapshots — but those arrive on a ~5s cadence (`DefaultSnapshotInterval`), so orch could only place the last advance to within ±5s and would learn of a stall a full cadence late. The standby samples at ~1s, so stamping the advance time at the source lets a consumer know the *age* of the last increase at that resolution, independent of the snapshot cadence. That matters because the LeaderStuck staleness threshold is only a small multiple of the snapshot cadence, where ±5s of slop would blur the decision. (Trade-off: it's a pooler-clock timestamp compared against orch's clock — negligible NTP skew versus the threshold, same as the existing lag fields. If we later want it skew-immune at coarser resolution, orch-side diffing is a drop-in swap since the consumer just reads a timestamp.)

## Notes

- Populated in `queryReplicationStatus` (single build site) so no caller sees a half-populated struct.
- Row reads use the bounds-checked `executor.Get*` helpers; a NULL LSN (streaming not started) is treated as absent, never an error on the lag path.

### DIFF
--- a/go/pb/multipoolermanagerdata/multipoolermanagerdata.pb.go
+++ b/go/pb/multipoolermanagerdata/multipoolermanagerdata.pb.go
@@ -610,8 +610,16 @@ type StandbyReplicationStatus struct {
 	// before terminating the WAL receiver connection.
 	// From current_setting('wal_receiver_timeout').
 	WalReceiverTimeout *durationpb.Duration `protobuf:"bytes,11,opt,name=wal_receiver_timeout,json=walReceiverTimeout,proto3" json:"wal_receiver_timeout,omitempty"`
-	unknownFields      protoimpl.UnknownFields
-	sizeCache          protoimpl.SizeCache
+	// When last_receive_lsn (pg_last_wal_receive_lsn — WAL streamed from the
+	// primary) last increased, as observed by the standby's heartbeat reader on
+	// its regular tick. This is a WAL-*progress* signal, distinct from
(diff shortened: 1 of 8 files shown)


GuptaManan100 (approved):
LGTM

PR #1316, opened 2026-07-27
fix(multigateway): multi-statement batches on temp-table sessions use implicit transaction handling

shedges:
## Problem

PostgreSQL's simple Query protocol requires that all statements in one `Q` message execute inside a single implicit transaction. A failure in any statement must roll back all preceding statements in that batch.

Sessions with a temp table reservation bypassed `executeWithImplicitTransaction` and executed each statement individually in a bare loop. This diverged from native PostgreSQL in four ways:

1. **Autocommit on failure** - a failing later statement did not roll back earlier statements; data was silently committed when it should have been rolled back.
2. **COMMIT/ROLLBACK AND CHAIN** - the error was not returned or the implicit transaction was not rolled back before surfacing it.
3. **VACUUM in batch** - `VACUUM` inside a multi-statement batch was not rejected as "cannot run inside a transaction block."
4. **Aborted state not set** - a batch failure inside an explicit transaction did not set `TxnStatusFailed`, so subsequent queries were not rejected with `25P02` (`in_failed_sql_transaction`).

## Fix

All multi-statement batches, regardless of session state now route unconditionally through `executeWithImplicitTransaction`, which correctly emulates PostgreSQL's implicit transaction semantics.

`executeWithImplicitTransaction` already handles the temp-table reserved connection path correctly. It uses `ScatterConn` with the existing reserved connection, so session state (temp tables, `SET` variables, etc.) is preserved.

## Tests

| Test | Scenario | Assertion |
|------|----------|-----------|
| `TmpTableSession_SuccessfulBatchCommits` | Two INSERTs in one batch | Both rows committed (implicit tx committed atomically) |
| `TmpTableSession_FailureRollsBackBatch` | `INSERT; SELECT 1/0` | Table empty. INSERT must not autocommit when batch fails |
| `TmpTableSession_ExplicitTxAbortedStateAfterBatchFailure` | `BEGIN; INSERT; SELECT 1/0` | Next query returns `25P02`; ROLLBACK recovers; table empty |

Added regression guards for `COMMIT AND CHAIN` / `ROLLBACK AND CHAIN` in implicit transactions. PostgreSQL rejects both, and the implicit transaction must be rolled back before the error is surfaced so preceding statements do not commit. These run against both targets too.


shedges:
Superseded by a clean branch; the handler.go fix landed in #1311. Reopening as test-only PR.

PR #1319, opened 2026-07-27
fix(query-serving): capture set_config's confirmed value for SET/RESET GUC replay

shedges:
## Summary

Fixes GUC value drift between the gateway's tracked session state and PostgreSQL's real backend state for `SET`, in two steps -

- Outside an explicit transaction, `SessionSettings` (replayed onto a backend on pool rotation) now records `set_config`'s canonical confirmed value instead of the client's literal (e.g. `DateStyle 'ISO'` resolves to `'ISO, MDY'`).
- Inside an explicit transaction, once an earlier statement has already run (`state.PendingBeginQuery` empty), `SET` now goes through the same confirmed value capture instead of tracking the client's literal. Fixing the `ROLLBACK TO SAVEPOINT` drift where PostgreSQL's transactional GUC stack reverts correctly but the gateway's tracked literal didn't follow. The transaction's own first statement stays on the existing literal-based path, since routing it through a `SELECT set_config(...)` would prematurely fix a `REPEATABLE READ`/`SERIALIZABLE` snapshot one statement early.

`RESET`'s in-transaction path is unchanged: its tracking clears the override rather than recording a value, so there's nothing for confirmed-value capture to correct there. A `pg_settings` scan for `DO`/function-call GUC changes remains a separate follow-up.

### DIFF
--- a/go/services/multigateway/planner/variable_set_stmt.go
+++ b/go/services/multigateway/planner/variable_set_stmt.go
@@ -36,6 +36,15 @@ import (
 //     name/value errors at SET time (matching PostgreSQL) without persisting on
 //     the pooled backend, and ApplySessionState records it for pool-rotation
 //     replay only if validation succeeded.
+//   - Inside an explicit transaction, once an earlier statement has already
+//     run (state.PendingBeginQuery is empty), SET var = value is instead
+//     planned as Sequence[ValidateSetting(Persist), ApplySessionState]:
+//     with Persist set, ValidateSetting runs set_config(name, value,
+//     is_local := false), which really persists the change (like a plain
+//     SET) and captures PostgreSQL's confirmed value for ApplySessionState
+//     to record. For the transaction's own first statement, this instead
+//     stays on Sequence[Route, ApplySessionStateSilent]; see the
+//     is-in-transaction branch below for why.
 //   - RESET / RESET ALL update local tracking only (no backend round-trip).
 //   - Gateway-managed variables, SET LOCAL, and SET TRANSACTION / FROM CURRENT
 //     are handled by their dedicated paths below.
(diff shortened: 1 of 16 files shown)


go/services/multigateway/planner/variable_set_stmt.go
```diff
@@ -137,7 +147,26 @@ func (p *Planner) planVariableSetStmt(
...
+		// already fixed and set_config's confirmed value can be captured safely,
+		// the same way the outside-transaction path does, instead of tracking
+		// the client's literal (the source of item 8's ROLLBACK TO SAVEPOINT
+		// drift).
 		if conn != nil && conn.IsInTransaction() {
+			if state != nil && state.PendingBeginQuery == "" {
+				value := extractVariableValue(stmt.Args)
+				apply := engine.NewValidateSettingPersist(p.defaultTableGroup, constants.DefaultShard, stmt.Name, value, sql)
+				track := engine.NewApplySessionState(sql, stmt)
+				plan := engine.NewPlan(sql, engine.NewSequence([]engine.Primitive{apply, track}))
```
haritabh17, line 173: If this SET is the final backend statement before COMMIT, ConcludeTransaction releases a trusted connection whose connstate still describes the pre-SET state. A later checkout can then skip the required reset.

With physical pool capacity 1:                                                                       

```sql                                                                                               
-- client A: SELECT 1 consumes PendingBeginQuery and selects this path                             
BEGIN;                                                                                             
SELECT 1;                                                                                          
SET work_mem = '64MB';                                                                             
COMMIT;                                                                                            

-- client B, reusing the backend                                                                   
SHOW work_mem; -- can observe 64MB instead of its default                                          
```

Maybe we can even remove this logic and defer to when needed later??

haritabh17 (approved):
Rest looks good. 🚀

PR #1320, opened 2026-07-27
fix(multipooler,multiorch): remove inner timeouts from Recruit and Promote RPC handlers

mkindahl:
# fix(multipooler,multiorch): remove inner timeouts from Recruit and Promote RPC handlers

## Problem

The Recruit and Promote RPCs are intentionally long-running. Their callers already set appropriate deadlines — `RuleWriteTimeout` (30 s) for Recruit RPCs and the `AppointLeaderAction` timeout (65 s = 2×`RuleWriteTimeout` + 5 s) for the full Recruit + Promote cycle. Three hardcoded inner timeouts inside the handlers overrode those caller-set deadlines and caused premature failures that triggered cascade re-elections.

**Bug 1 — `promote()` applied `RemoteOperationTimeout` (15 s) to the leader's Promote RPC.**

The leader's Promote RPC is a long-running blocking call. After `pg_promote()` fires, postgres runs an end-of-recovery checkpoint to flush all pages dirtied by WAL replay before leaving recovery mode. That checkpoint is proportional to replay volume and can take tens of seconds on a large or recently-restored node. Applying a 15 s deadline designed for fast RPCs caused gRPC to cancel the stream mid-checkpoint, propagating context cancellation into the Promote handler. Cancelling mid-promotion abandons a promotion whose timeline is already forked.

**Bug 2 — `waitForPromotionComplete` (Promote handler) had a hardcoded 30 s inner deadline.**

A 30 s inner `context.WithTimeout` on `waitForPromotionComplete` would override the caller's context. When it fired, `promotionInProgress` was cleared before `PostgresReady=true`, causing `LeaderNeedsReplacementAnalyzer` to see a dead leader and fire a new election — a cascade.

**Bug 3 — `waitForReplayComplete` (Recruit handler) had a hardcoded 10 s inner deadline.**

During Recruit, `waitForReplayComplete` waits for the node to confirm WAL replay is complete (`replay_lsn >= receive_lsn`) before reporting its position to the coordinator. A 10 s inner `context.WithTimeout` overrode the caller's 30 s `RuleWriteTimeout`. If replay took longer than 10 s — or if the timeout fired spuriously under I/O pressure while replay was still making progress — the Recruit RPC failed.

### Why Promote is long-running

Recruit already ensures WAL replay is complete on the winning candidate (`replay_lsn >= receive_lsn`) before the coordinator commits to a leader. Promote therefore does not wait for WAL replay. The latency comes from two subsequent steps:

1. **Post-promotion checkpoint.** `pg_promote()` triggers an end-of-recovery checkpoint that flushes all dirty pages produced by WAL replay. This runs synchronously before postgres leaves recovery mode and is proportional to the volume of WAL that was replayed — it can take tens of seconds on a recently-restored or large node.

2. **Rule commit.** After `waitForPromotionComplete` returns and `promotionInProgress` is cleared, the Promote handler calls `configureReplicationAfterPromotion` and then commits the new rule to the rule store. Under `AT_LEAST_2` durability that commit blocks in `SyncRepWaitForLSN` until standbys reconnect to the new primary and acknowledge the write. This is a reconnection delay, not a WAL replay delay.

### Observed in the 2026-05-29 AZ-outage (mg-scale12, zone-a pod deletion)

The cascade is visible in the zone-d multiorch log. After earlier elections at terms 7 and 8, zone-d elected p-28c00303 as leader at term 9. The Promote RPC carried the 15 s `RemoteOperationTimeout` deadline. Exactly 15 seconds later:

```
13:58:13  primary.promotion started  leader=p-28c00303  term=9
           ...
13:58:28  primary.promotion failed   leader=p-28c00303  term=9
          rpc error: code = DeadlineExceeded
                     desc = stream terminated by RST_STREAM with error code: CANCEL
13:58:28  recovery action failed  problem=LeaderIsDead
          failed to appoint leader: leader p-28c00303 failed to accept proposal
13:58:29  executing appoint leader action  ← term 12 starts immediately
```

`RST_STREAM with error code: CANCEL` is the gRPC wire signal for a client-side deadline cancellation. The 15 s deadline fired, cancelling the stream; this propagated as context cancellation into the multipooler's Promote handler. With `promotionInProgress` cleared, `LeaderNeedsReplacementAnalyzer` re-fired immediately, and zone-d started term 12 one second later.

The mg-scale12 cluster had just been scaled up from 3 to 12 nodes. The nine new nodes were restored from a pgbackrest backup and had been streaming WAL to catch up. p-28c00303 had received and replayed a large volume of WAL, so the post-promotion checkpoint — flushing all those dirty pages — took longer than the 15 s window.

### Cascade mechanism (before fix)

```mermaid
sequenceDiagram
    participant ORCH as multiorch<br/>(zone-d)
    participant ANA as LeaderNeedsReplacementAnalyzer
    participant MP as p-28c00303<br/>(multipooler)
    participant PG as postgres<br/>(p-28c00303)

    Note over ORCH,PG: t=0 (13:58:12) — LeaderIsDead fires · term 9 starts

    ORCH->>MP: Recruit(term=9) [RuleWriteTimeout=30s]
    MP->>MP: pause WAL receiver · wait for replay_lsn >= receive_lsn
    MP-->>ORCH: ok (WAL position reported · replay complete)
    Note over ORCH,MP: t ≈ 1s (13:58:13) — Recruit quorum reached<br/>p-28c00303 elected leader<br/>Promote RPC sent with RemoteOperationTimeout (15s) deadline

    ORCH->>MP: Promote(term=9) [deadline: t+15s]
    MP->>MP: promotionInProgress = true
    MP-->>ORCH: health: POSTGRES_STATUS_PROMOTING · PostgresReady=false
    Note over ANA: POSTGRES_STATUS_PROMOTING → leaderPromoting() → suppress ✓
    MP->>PG: SELECT pg_promote()
    PG-->>MP: ok
    Note over MP,PG: end-of-recovery checkpoint<br/>(flushing pages dirtied by WAL replay)

    Note over ORCH,MP: t=15s (13:58:28) — RemoteOperationTimeout fires
    ORCH--xMP: gRPC RST_STREAM CANCEL
    MP->>MP: ctx cancelled → waitForPromotionComplete returns error
    MP->>MP: defer: promotionInProgress = false
    MP-->>ORCH: health: POSTGRES_STATUS_RUNNING · PostgresReady=false
    Note over ANA: POSTGRES_STATUS_PROMOTING cleared
    MP-->>ORCH: Promote → DeadlineExceeded ✗
    Note over ORCH: primary.promotion failed

    ORCH->>ANA: Analyze(shard)
    ANA-->>ORCH: LeaderIsDead 🔴<br/>(RUNNING · not ready · not PROMOTING)
    Note over ORCH,PG: t=16s (13:58:29) — term 12 starts immediately
```

## Fix

Remove all three inner timeouts. Each function now uses the caller's context directly:

- `waitForReplayComplete` uses the Recruit RPC's context (bounded by the `AppointLeaderAction` outer timeout)
- `waitForPromotionComplete` uses the Promote RPC handler's context
- `recruit()` and `promote()` pass the caller's context directly to all RPCs in the AppointLeader path

The `AppointLeaderAction` timeout (2×`RuleWriteTimeout` + 5 s = 65 s) is the single outer bound for the full Recruit + Promote + SetPrimary cycle — this value predates this PR and is unchanged.

### Fixed behavior

```mermaid
sequenceDiagram
    participant ORCH as multiorch
    participant ANA as LeaderNeedsReplacementAnalyzer
    participant MP as leader multipooler
    participant PG as postgres

    Note over ORCH,PG: LeaderIsDead fires — Promote RPC sent with AppointLeaderAction ctx (65s)

    ORCH->>MP: Promote(term=N) [AppointLeaderAction ctx]
    MP->>MP: promotionInProgress = true
    MP-->>ORCH: health: POSTGRES_STATUS_PROMOTING · PostgresReady=false
    Note over ANA: POSTGRES_STATUS_PROMOTING → leaderPromoting() → suppress ✓
    MP->>PG: SELECT pg_promote()
    PG-->>MP: ok
    Note over MP,PG: end-of-recovery checkpoint...

    loop every 100ms
        ORCH->>ANA: Analyze(shard)
        ANA-->>ORCH: no problem (PROMOTING suppresses) ✓
        MP->>PG: pg_is_in_recovery() / pg_isready
        PG-->>MP: not ready yet
    end

    PG-->>MP: pg_isready = true ✓
    MP->>MP: defer: promotionInProgress = false
    MP-->>ORCH: health: POSTGRES_STATUS_RUNNING · PostgresReady=true
    MP-->>ORCH: Promote → success (rule commit follows) ✅

    ORCH->>ANA: Analyze(shard)
    ANA-->>ORCH: no problem (leader reachable and ready) ✅
```

### Changes

**`go/services/multipooler/internal/manager/pg_replication.go`**

- Removed the hardcoded 10 s `context.WithTimeout` from `waitForReplayComplete`. The function now uses the caller's context (Recruit RPC, bounded by the `AppointLeaderAction` outer timeout).

**`go/services/multipooler/internal/manager/manager.go`**

- Removed the hardcoded 30 s `context.WithTimeout` from `waitForPromotionComplete`. The function now uses the caller's context so `POSTGRES_STATUS_PROMOTING` is held for the full window until `PostgresReady=true`.
- Added `promotion.wal_replay` and `promotion.postgres_ready` eventlog events (Started/Success/Failed) so each phase is observable through the structured event log.

**`go/services/multiorch/consensus/rule_change.go`**

- `recruit()` and `promote()` no longer apply per-call timeouts to any RPC in the AppointLeader path. All three RPCs — Recruit, leader Promote, and non-leader SetPrimary — now use the `AppointLeaderAction` outer context as their deadline.

**`go/common/eventlog/events.go`**

- Added `PromotionWalReplay` (`promotion.wal_replay`) and `PromotionPostgresReady` (`promotion.postgres_ready`) event types.

**`go/cmd/pgctld/testutil/grpc_helpers.go`**

- Added `StatusFunc func(*pb.StatusRequest) (*pb.StatusResponse, error)` to `MockPgCtldService` to allow sequential response control in tests.

## Observability

The two new eventlog event types let operators reconstruct the pg_promote() to ready window from structured logs:

| Event type | Outcome | When |
|---|---|---|
| `promotion.wal_replay` | `started` | `waitForPromotionComplete` begins polling `pg_is_in_recovery()` |
| `promotion.wal_replay` | `success` | `pg_is_in_recovery()` first returns `false` (checkpoint complete) |
| `promotion.wal_replay` | `failed` | Context cancelled or query error |
| `promotion.postgres_ready` | `started` | `pg_is_in_recovery()` returned false; now polling `pg_isready` |
| `promotion.postgres_ready` | `success` | `pg_isready` returns true |
| `promotion.postgres_ready` | `failed` | Context cancelled |

These pair with the existing `consensus.promote` Started/Success/Failed events (emitted in `rule_change.go`) to give end-to-end visibility across the Promote RPC boundary.

## Tests

| Test | Type | What it verifies |
|------|------|-----------------|
| `TestWaitForPromotionComplete_KeepsPollingUntilPostgresReady` | Unit | Regression: `waitForPromotionComplete` keeps polling when `pg_isready` returns false, catching a reintroduced inner timeout |
| `TestWaitForPromotionComplete_ContextCancellationReturnsError` | Unit | Context cancellation propagates cleanly so `promotionInProgress` is cleared |
| `TestLeaderNeedsReplacementAnalyzer_Analyze/suppresses_LeaderIsDead_while_pg_promote()_is_running` | Unit | `POSTGRES_STATUS_PROMOTING` on the leader suppresses re-appointment |
| `TestPromotionAppointsExactlyOneLeader` | E2E | After a primary kill, `rule_history` shows exactly one promotion (cascade would show two or more) |
| `TestPromotingStatusClearedOnlyWhenReady` | E2E | `POSTGRES_STATUS_PROMOTING` never clears before `PostgresReady=true` |


go/services/multiorch/consensus/rule_change.go
```diff
@@ -357,14 +357,23 @@ func (r *coordinatorLedRuleChange) promote(
...
 	isLeader bool,
 ) error {
-	rpcCtx, cancel := context.WithTimeout(ctx, timeouts.RuleWriteTimeout)
-	defer cancel()
 	if isLeader {
-		_, err := r.coordinator.rpcClient.Promote(rpcCtx, p.Multipooler, req)
+		// The leader's Promote RPC is a long-running blocking call. WAL replay
+		// is already complete after Recruit, but pg_promote() still runs an
+		// end-of-recovery checkpoint to flush pages dirtied by replay — that
+		// checkpoint can be slow in proportion to replay volume. After postgres
```
dweitzman, line 361: The checkpoint doesn't block promotion unless replicas need to rewind, so maybe the wording here can be more specific?

Was this a case where a checkpoint was needed so a replica could rewind, or was the machine so overloaded by I/O that all writes lagged significantly?

go/services/multiorch/consensus/rule_change.go
```diff
@@ -357,14 +357,23 @@ func (r *coordinatorLedRuleChange) promote(
...
+		// leaves recovery, the rule commit may block in SyncRepWaitForLSN until
+		// standbys reconnect to the new primary and acknowledge the write.
+		// The caller's context (from AppointLeaderAction) is the outer bound.
+		_, err := r.coordinator.rpcClient.Promote(ctx, p.Multipooler, req)
```
dweitzman, line 365: There's still a `2 * RuleWriteTimeout + 5s` timeout for the full recovery action. Is the idea here that doubling the timeout helps?

Longer term (as a bigger change), I think it might make sense to allow the pooler to keep making progress on a promotion regardless of RPC timeouts until it's either recruited or otherwise informed it's not supposed to be the leader, so that even if orch times out here it doesn't necessarily mean the promotion won't eventually succeed

go/services/multipooler/internal/manager/manager.go
```diff
@@ -1550,46 +1550,52 @@ func (pm *MultipoolerManager) dropUnloggedTablesAfterPromotion(ctx context.Conte
 // is accepting connections. Both conditions are required: pg_is_in_recovery()=false
 // confirms the WAL-level promotion, and postgres_ready=true confirms clients can
 // connect. Clearing promotionInProgress only when both are true ensures multiorch's
-// PrimaryIsDeadAnalyzer suppression window matches the full visibility gap.
+// LeaderNeedsReplacementAnalyzer suppression window covers the full gap — including
```
dweitzman: Wasn't the replay due to now-resolved issues with clearing the restore command and making sure replay is complete before Recruit() returns?

The premise of this comment feels, if true, concerning. I wouldn't expect Promote() to need to spend any time on replay.

go/services/multipooler/internal/manager/manager.go
```diff
@@ -1550,46 +1550,52 @@ func (pm *MultipoolerManager) dropUnloggedTablesAfterPromotion(ctx context.Conte
...
 	defer ticker.Stop()
 
-	promotionTimeout := 30 * time.Second
-	promotionCtx, cancel := context.WithTimeout(ctx, promotionTimeout)
-	defer cancel()
-
+	// Phase A: wait for postgres to leave recovery mode (WAL replay).
 	eventlog.Emit(ctx, pm.logger, eventlog.Started, eventlog.PromotionWalReplay{})
-	promotedFromRecovery := false
 	for {
```
dweitzman, line 1568: Maybe break this into two subroutines instead of using a `goto`?

Could be `waitUntilOutOfRecovery()` and `waitUntilPostgresReady()`, etc[

go/services/multipooler/internal/manager/pg_replication.go
```diff
@@ -627,12 +627,11 @@ func (pm *MultipoolerManager) stopRestoreCommand(ctx context.Context) error {
 //
 // See decision: 2026-02-12-wait-for-replay-stabilize-during-revoke.md.
 func (pm *MultipoolerManager) waitForReplayComplete(ctx context.Context) (*multipoolermanagerdatapb.StandbyReplicationStatus, error) {
-	completeCtx, completeSpan := telemetry.Tracer().Start(ctx, "consensus/replay_completion_wait")
+	// The caller's context (Recruit RPC, bounded by RuleWriteTimeout) is the
+	// intended deadline; no inner timeout is imposed here.
+	waitCtx, completeSpan := telemetry.Tracer().Start(ctx, "consensus/replay_completion_wait")
 	defer completeSpan.End()
 
-	waitCtx, cancel := context.WithTimeout(completeCtx, 10*time.Second)
```
dweitzman, line 633: I don't have strong feelings about this, but was this timeout actually harmful? If for whatever reason replay takes 10+ seconds and Recruit() times out and it blocks failovers, I would expect the orch to retry and by the time it does there's a higher chance replay is complete

If the time to replay is relatively short it seems reasonable to wait for it to complete, but as it gets longer at some point it might make sense to just let the first failover attempt fail and follow up with a second attempt

go/test/endtoend/multiorch/promotion_stability_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/multigres/multigres/go/test/endtoend/shardsetup"
+	"github.com/multigres/multigres/go/test/utils"
+
+	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
+)
+
+// TestPromotionElectsExactlyOneLeader verifies that after a primary failure
```
dweitzman: Terminology nit: this file frequently says things like "leader election" rather than "leader appointment"

`TestPromotionElectsExactlyOneLeader` feels like it should be `TestPromotionAppointsExactlyOneLeader`, for example

go/test/endtoend/multiorch/promotion_stability_test.go
```diff
@@ -0,0 +1,225 @@
...
+	// re-elections would produce two or more rows.
+	assert.Equal(t, promotionsBefore+1, promotionsAfter,
+		"exactly one promotion should fire per failover (cascade would produce more)")
+}
+
+// TestPromotingStatusClearedOnlyWhenReady verifies that a newly promoted node
+// reports POSTGRES_STATUS_PROMOTING until its postgres process is accepting
+// connections, and transitions to a non-PROMOTING status only after that.
+//
+// This is the direct observable invariant of the fix: if PROMOTING clears
```
dweitzman: This PR has some AI comments that are referring to the context of the PR, but won't make sense in the codebase within the context of the PR ("This is the direct observable invariant of the fix", etc.)

go/test/endtoend/queryserving/pgbouncertests/copy_fault_test.go
```diff
@@ -97,25 +97,16 @@ func TestCopyInterruptedByBackendCrash(t *testing.T) {
...
 	t.Log("postgres auto-restarted by the monitor")
 
-	// The post-recovery assertions get their own fresh context. The top-level
-	// ctx above shares a single 120s budget across a chain of internal waits
-	// (up to 30s for the killed COPY to fail, up to 60s for postgres to
-	// auto-restart, up to 30s in the recovery Eventually below). On a slow
-	// runner those waits can burn nearly the whole budget, leaving the final
-	// count check with an already-exhausted deadline. A fresh context ensures a
-	// slow-but-successful recovery still has time to verify the outcome.
-	recoveryCtx := utils.WithTimeout(t, 30*time.Second)
```
dweitzman, line 107: Is this change related?

go/common/eventlog/events.go
```diff
@@ -176,6 +176,7 @@ func (e ConsensusRulePropose) LogAttrs() []slog.Attr {
 	return []slog.Attr{slog.String("rule", e.Rule)}
 }
 
+
```
dweitzman: Unrelated whitespace change

dweitzman (approved):

go/services/multiorch/consensus/rule_change.go
mkindahl: > The checkpoint doesn't block promotion unless replicas need to rewind, so maybe the wording here can be more specific?
>
> Was this a case where a checkpoint was needed so a replica could rewind, or was the machine so overloaded by I/O that all writes lagged significantly?

The case was that it was blocking on the `CHECKPOINT` execution, which in some cases that we saw took (a lot) longer than necessary, probably because of running out of IOPS. I'm lacking logs to validate for sure, but the logs that are visible stops progressing when they should be hitting the logging the execution of the `CHECKPOINT`.

go/common/eventlog/events.go
mkindahl: Removed.

go/services/multiorch/consensus/rule_change.go
mkindahl: > There's still a `2 * RuleWriteTimeout + 5s` timeout for the full recovery action. Is the idea here that doubling the timeout helps?

No, the idea was to divide the full timeout into intermediate steps, because we had internal timeouts that were hard-coded and too short. For example, the `AppointLeader()` had a timeout of 60s, so this was divided into a 30s timeout for `Recruit()` and `Promote()`/`SetPrimary()` respectively, plus 5s for a health check before calling `AppointLeader()`.

I'm starting to question if this is correct. Maybe we should instead pass the full timeout to each step instead of slicing it up. This might allow more `AppointLeader()` to successfully finish instead of aborting `Recruit()` (for example) after 30s and declaring the entire RPC failed.

> Longer term (as a bigger change), I think it might make sense to allow the pooler to keep making progress on a promotion regardless of RPC timeouts until it's either recruited or otherwise informed it's not supposed to be the leader, so that even if orch times out here it doesn't necessarily mean the promotion won't eventually succeed

I'm not opposed to the idea, but the problem here is that this might lead to a split-brain, so we need to be careful. If a server continue with promotion "more than it should" (RPC timeout might not be the limit) we might end up with two primaries. This should not be impossible to avoid, but we should be careful.

go/services/multipooler/internal/manager/manager.go
mkindahl: Yes, the comment needs modification. WAL replay should not happen during `Promote()` but it can still take time because of the checkpoint.

go/services/multipooler/internal/manager/manager.go
mkindahl: Good idea. Done.

go/services/multipooler/internal/manager/pg_replication.go
mkindahl: The reason for removing this timeout is not mainly because it is harmful, but because it is unnecessary and makes the system hard to debug and maintain. (The timeout _could_ potentially be harmful though, if the replay for some reason is delayed, is within the bounds of the RPC timeout, but takes longer than 10s. Or if we change the timeout to shorter than 10s, because the hard-coded timeout will still be 10s.)

The function is called as part of the `Recruit()` RPC and there is a budget for that call (in this case 30s, but the time is not the point here). The function overrides this and installs a hard-coded 10s limit despite the fact that the caller has budgeted 30s for this. This makes the code hard to debug and control since if we, for any reason, changes this timeout, it will be ineffective because we have this hard-coded internal limit.

We should instead honor the timeout set by the caller. If we think this is too long, we change that in the caller.

go/test/endtoend/multiorch/promotion_stability_test.go
mkindahl: No opinion on the terminology here. Changed.

go/test/endtoend/multiorch/promotion_stability_test.go
mkindahl: Removed.

go/test/endtoend/queryserving/pgbouncertests/copy_fault_test.go
mkindahl: No, removed.

go/services/multiorch/consensus/rule_change.go
mkindahl: The is the call chain:

```
AppointLeaderAction.Execute (65s outer bound)
├── 5s healthy-leader check (the +5s)
└── consensus.AppointLeader (~60s remaining)
    ├── Recruit fan-out (each RPC: RuleWriteTimeout = 30s)
    └── Promote fan-out
        ├── leader:  Promote RPC — now uses the outer ctx (no inner cap)
        └── others:  SetPrimary — RuleWriteTimeout = 30s each
```

mkindahl: I modified this to use this call chain instead. Now the RPC timeout governs the entire process, allowing individual steps to fluctuate.
```
AppointLeaderAction ctx (65s)
  └─ Recruit RPCs: ctx passed directly
       └─ waitForReplayComplete: uses whatever remains
  └─ leader Promote RPC: ctx passed directly
       └─ waitForPromotionComplete: uses whatever remains
  └─ non-leader SetPrimary: ctx passed directly
```

PR #1322, opened 2026-07-27
feat(multiorch): reconnect a follower stranded by an abandoned recruit

dweitzman:
This allows inviting a pooler back into replication if it was disconnected due to a Recruit() from abandoned failover attempt. This shouldn't be that common in practice since orch tries to predict if it has the reach for a failover before Recruit(), but it's a theoretical possibility

A `TermRevocation` indicates both the outgoing decision and proposed new rule, and a revocation rooted at an out-of-date decision is less powerful than a `SetPrimary` based on a higher decision.

This PR adds detection and repair:

- **`RecruitAbandonedAnalyzer`** flags if a follower is recruited for something like a transition from decision 4.5 (coordinator term 4, leader subterm 5) to proposal 6.0 and the shard is still healthy at decision 4.5
- **`ReconnectRecruitAbandonedAction`** performs a leader-led no-op rule advance: `UpdateConsensusRule(ADVANCE)` advances a decision that only changes the leader subterm. Then the shard will be running rule 4.6, and SetPrimary(4.6, ...) is allowed against a pooler that was revoked for transitioning from 4.5 -> 6.0

`UpdateConsensusRule` now returns the resulting `current_position` in its response. The advance commits synchronously (the write blocks until quorum-acked), so the reconnect uses that returned rule directly instead of polling `Status` for it.


go/services/multiorch/recovery/analysis/recruit_abandoned_analyzer.go
```diff
@@ -0,0 +1,131 @@
...
+// recruit as abandoned. A revocation with no coordinator_initiated_at (which a
+// valid one always carries) is treated as not-yet-aged, erring toward waiting.
+func revocationAgedPastFailoverGrace(factory *RecoveryActionFactory, sa *ShardAnalysis, p *store.Pooler) bool {
+	initiatedAt := p.Health().GetConsensusStatus().GetTermRevocation().GetCoordinatorInitiatedAt()
+	if initiatedAt == nil {
+		return false
+	}
+	var grace time.Duration
+	if cfg := factory.Config(); cfg != nil {
+		grace = cfg.GetLeaderFailoverGracePeriodBase() + cfg.GetLeaderFailoverGracePeriodMaxJitter()
```
Verolop, line 75: the default grace period here is 12 seconds, but the failover RPCs can each take up to 30 seconds. Could this mark a recruit as abandoned while the failover is still running?

go/services/multiorch/recovery/actions/reconnect_recruit_abandoned.go
```diff
@@ -0,0 +1,164 @@
...
+// UpdateConsensusRule on the leader.
+func (a *ReconnectRecruitAbandonedAction) RequiresHealthyLeader() bool {
+	return true
+}
+
+func (a *ReconnectRecruitAbandonedAction) Metadata() types.RecoveryMetadata {
+	return types.RecoveryMetadata{
+		Name:        "ReconnectRecruitAbandoned",
+		Description: "Advance the leader rule to reconnect a follower stranded by an abandoned recruit",
+		Timeout:     30 * time.Second,
```
Verolop, line 161: is 30 secs enough for both RPCs?

Verolop (approved):
lgtm, just left 2 non-blocking questions.

go/services/multiorch/recovery/analysis/recruit_abandoned_analyzer.go
dweitzman: This shouldn't be competing with failovers, so I modified the logic so it only runs if the highest known rule is decided, preventing competition with failovers

The purpose of this recovery action is to avoid revocation, but if a failover happens it should accomplish the same goal in a more disruptive way (through a higher revocation + new establishment) and make this gentler recovery action unnecessary.

go/services/multiorch/recovery/actions/reconnect_recruit_abandoned.go
dweitzman: This isn't a high enough priority recovery action to justify waiting longer. If the shard is down we'll want a full failover and if it's operating and healthy then we don't expect to wait a huge amount of time to do a no-op leader-led rule change

PR #1323, opened 2026-07-28
ci(nightly): publish attested full-sha images

Verolop:
## Description

this pr publishes signed nightly images under full commit SHA tags. This feature will be used to help tighten the security of other Multigres and Multigres Operator automations.

## Main changes

- Builds the exact `main` commit associated with the workflow run.
- Publishes `nightly-sha-<full-sha>` only after signing the image digest.
- Reuses an existing full-SHA image only when its provenance matches the image, workflow, branch, and source commit.
- Preserves immutable tags across retries while safely refreshing the existing nightly aliases.
- Uses pinned actions and minimum job permissions.
- Documents the full-SHA tag in `RELEASE.md`.

## Testing

Validated the workflow with `actionlint`, `zizmor`, and `git diff --check`. The review covered first publication, same-SHA nightly runs, partial build retries, stale reruns, signing failures, registry failures, and compatibility with the operator’s digest and provenance verification.

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -24,16 +24,61 @@ concurrency:
   group: nightly-build-main
   cancel-in-progress: false
 
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
+  prepare-nightly:
+    name: Prepare Nightly
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    outputs:
+      promote_floating: ${{ steps.prepare.outputs.promote_floating }}
+    steps:
+      - name: Decide whether to update the floating tag
+        id: prepare
+        env:
+          CURRENT_RUN_NUMBER: ${{ github.run_number }}
+          SOURCE_REF: ${{ github.ref }}
+          SOURCE_SHA: ${{ github.sha }}
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -87,8 +132,55 @@ jobs:
...
+          SOURCE_SHA: ${{ steps.determine-sha.outputs.full_sha }}
+        run: |
+          image_ref="${IMAGE_NAME}:nightly-sha-${SOURCE_SHA}"
+          error_file="$(mktemp)"
+          if digest="$(
+            docker buildx imagetools inspect "$image_ref" \
+              --format '{{json .Manifest}}' 2> "$error_file" |
+              jq -er '.digest | select(test("^sha256:[0-9a-f]{64}$"))'
+          )"; then
+            gh attestation verify "oci://${IMAGE_NAME}@${digest}" \
```
niconosenzo, line 151: Q: does it work without setting the GH_TOKEN env variable ? Or should we explicitly set it on the env step above:
```yaml
GH_TOKEN: ${{ github.token }} 
```

niconosenzo (approved):
Looks good 🚀.  Left a comment, if you tested it and it works then you can ignore my comment!

PR #1324, opened 2026-07-28
fix(query-serving): pin backend across setseed()

shedges:
## Summary

- setseed() mutates one backend's PRNG state, invisible to gateway tracking; a later random()/random_normal() landing on a different pooled backend silently breaks the reproducible sequence the client expects.
- Detected via unsafe_funccall.go (CallsSetSeed), routed through PlanOptions/PlanExecInfo.SetSeed into a new PlanTypeSetSeedRoute, and pinned via a new ReasonSetSeed reservation reason (proto 128), same pinning path as AdvisoryLock and LogicalReplicationSlot.
- Added keep_sticky_reservations to ReleaseReservedConnectionRequest, threaded through the multipooler's ReleaseReservedConnection (now returns the still-reserved state instead of just an error) and the gateway's ReleaseAllReservedConnections: DiscardAllPrimitive passes true (stays reserved), real client disconnect (Executor.ReleaseAll) passes false (always fully releases).

## Coverage

- Planner-level tests confirming setseed() detection and routing.
- ScatterConn tests for reservation/promotion on both the simple and portal paths, and for the sticky-preserve-vs-clear behavior of ReleaseAllReservedConnections.
- Multipooler-level tests for ReleaseReservedConnection's sticky/non-sticky behavior, including the real-disconnect override.


go/services/multipooler/internal/executor/executor.go
```diff
@@ -2304,9 +2313,23 @@ func (e *Executor) ReleaseReservedConnection(
...
+	// regardless of the sticky check below: a portal pin has a PG-visible
+	// unpin event (CLOSE ALL, which DISCARD ALL implies), unlike a sticky
+	// reason.
 	reservedConn.ReleaseAllPortals()
 
+	// Step 3c: if only a sticky reason remains, leave the connection reserved
+	// instead of returning it to the pool. Only DISCARD ALL passes
+	// keepStickyReservations; real disconnect cleanup always releases fully,
+	// so a sticky reason never blocks a connection's actual teardown.
+	if !cleanupFailed && keepStickyReservations && protoutil.HasSetSeedReason(reservedConn.RemainingReasons()) {
```
haritabh17, line 2328: After DISCARD ALL on a session that both seeded the generator and used temporary tables, will the gateway keeps believing the pinned connection still holds temporary tables??

haritabh17 (approved):
Rest looks good

go/services/multipooler/internal/executor/executor.go
shedges: Good catch!! Previously we used to release connection so it was fine to simply ignore it. Fixed

PR #1325, opened 2026-07-28
refactor(pgctld): move PostgresCtlConfig helpers onto PgCtldService

mkindahl:
## Summary

- `standbySignalPath`, `hasStandbySignal`, `createStandbySignal`, `removeStandbySignal`, `isPostgresCleanlyStopped`, `crashRecoveryNeeded`, `runSingleUserPostgres`, `runCrashRecovery`, `runCrashRecoveryInDir`, `runCrashRecoveryAttempts`, `StartPostgreSQLWithResult`, `StartPostgreSQLWithConfig`, `StopPostgreSQLWithResult`, `StopPostgreSQLWithConfig`, `stopPostgreSQLWithConfig`, `stopWithPgCtlWithConfig`, `takeCheckpoint`, and `RestartPostgreSQLWithResult` all took a bare `dataDir`/`logger`/`config` parameter set even though every caller already had a `PgCtldService` in hand. Made them methods on `PgCtldService` so they read `pgConfig` and `logger` off the receiver instead of threading both through every call site.
- Fixes a latent inconsistency: `runSingleUserPostgres` read `PGDATA` from the environment directly while the rest of the crash-recovery path used `pgConfig.PostgresDataDir`. The two happened to agree in production only because `pgConfig.PostgresDataDir` is seeded from `PGDATA` once at startup — nothing enforced it, and a `PgCtldService` built with a different data dir (as tests now do) would have silently run recovery against the wrong directory.
- The CLI command handlers (`runStart`, `runStop`, `runRestart`) build their `PostgresCtlConfig` from flags rather than owning a `PgCtldService`, so each now constructs a throwaway `PgCtldService` to call these methods on.

## Test plan

- [x] `go build ./go/...`
- [x] `go vet ./go/cmd/pgctld/...`
- [x] `go test -short ./go/cmd/pgctld/...`
- [x] `golangci-lint run ./go/cmd/pgctld/command/...` (v2.12.2, matching CI)

### DIFF
--- a/go/cmd/pgctld/command/crash_recovery.go
+++ b/go/cmd/pgctld/command/crash_recovery.go
@@ -17,14 +17,11 @@ package command
 import (
 	"context"
 	"fmt"
-	"log/slog"
 	"os"
-	"path/filepath"
 	"regexp"
 	"strings"
 
 	"github.com/multigres/multigres/go/common/constants"
-	"github.com/multigres/multigres/go/services/pgctld"
 	"github.com/multigres/multigres/go/tools/executil"
 	"github.com/multigres/multigres/go/tools/retry"
 )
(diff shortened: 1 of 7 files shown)


cuongdo (approved):
Thanks for this nice cleanup!

PR #1326, opened 2026-07-28
fix(multigateway): buffer when no writable primary exists

haritabh17:
Title: fix(multigateway): buffer when no writable primary exists

## Summary

- Mark the load balancer's canonical no-writable-primary error so buffering can distinguish it from unrelated failures.
- Buffer WRITABLE and CONSISTENT requests when the gateway's routing view has no primary, using the existing bounded failover window and drain path.
- Add a quiet-gateway regression test that starts with no buffered traffic, submits a new write during the primary gap, and verifies it resumes on the newly serving primary.

## Problem

Failover buffering currently starts only after a request observes MTF01 or a replay-safe 25006 from the demoting primary. A gateway with no request in flight during that drain does not arm its buffer. Its first request in the 2–4s demote-to-promote gap instead receives `FATAL: no writable primary is currently available`.

## Solution

The load balancer already retracts a demoted primary from each gateway's routing view, and the existing health-stream callback calls `StopBuffering` only after a replacement reports itself PRIMARY and SERVING. The missing piece is classification: the routing-generated no-writable-primary error was UNAVAILABLE/57P03 but was not one of the errors accepted by `classifyError`.

This change gives only that canonical routing error an internal marker and classifies it as buffer-worthy for leader-routed traffic. The first arrival on an otherwise quiet gateway therefore creates the shard buffer, waits within the existing request window and maximum failover duration, then retries after the replacement is routable. Replica traffic and unrelated errors still fail as before.

An out-of-band failover flag in topology was considered but not chosen. Gateways already receive the two facts needed through pooler health streams—no writable primary now, and a PRIMARY/SERVING replacement later—so another distributed state transition would add ordering and cleanup cases without closing a larger gap. If maintainers want proactive buffering before the old primary disappears, that signal can be added separately; it is not required for the reported quiet-gateway window.

## Testing

- `go test -short -count=1 ./go/services/multigateway/poolergateway`
- `go test -short -count=10 -run '^TestPoolerGateway_QuietGatewayBuffersUntilPrimaryAppears$' ./go/services/multigateway/poolergateway`
- `go build ./...`

### DIFF
--- a/go/services/multigateway/poolergateway/availability_errors.go
+++ b/go/services/multigateway/poolergateway/availability_errors.go
@@ -15,6 +15,8 @@
 package poolergateway
 
 import (
+	"errors"
+
 	"github.com/multigres/multigres/go/common/mterrors"
 	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
 )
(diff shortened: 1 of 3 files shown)


shedges (approved):
LGTM

PR #1330, opened 2026-07-28
feat(multiorch): exclude cohort-ineligible poolers from failover

dweitzman:
Cohort-ineliglbe is set during shutdown, so we can basically treat it the same as an unreachable pooler except with higher confidence and faster reaction time

This fixes an issue where it was sometimes possible for a resigning leader on shutdown to be re-selected as the leader at the next term instead of being removed from leadership and then eventually the cohort

### DIFF
--- a/go/services/multipooler/internal/manager/rpc_consensus.go
+++ b/go/services/multipooler/internal/manager/rpc_consensus.go
@@ -139,6 +139,16 @@ func (pm *MultipoolerManager) Recruit(ctx context.Context, req *consensusdatapb.
 		"revoked_below_term", revokedBelowTerm,
 		"coordinator_id", coordinatorID.GetName())
 
+	// Refuse recruitment while cohort-ineligible. That signal is raised only on a
+	// deliberate exit (graceful shutdown / admin-stopped WAL receiver), so this node
+	// must not join the new term — otherwise it could be re-elected leader or counted
+	// for quorum while leaving. Enforcing here is synchronous and does not depend on
+	// the coordinator having observed the (health-stream) eligibility signal in time.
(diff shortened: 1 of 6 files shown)


go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -139,6 +139,16 @@ func (pm *MultipoolerManager) Recruit(ctx context.Context, req *consensusdatapb.
 		"revoked_below_term", revokedBelowTerm,
 		"coordinator_id", coordinatorID.GetName())
 
+	// Refuse recruitment while cohort-ineligible. That signal is raised only on a
+	// deliberate exit (graceful shutdown / admin-stopped WAL receiver), so this node
+	// must not join the new term — otherwise it could be re-elected leader or counted
+	// for quorum while leaving. Enforcing here is synchronous and does not depend on
+	// the coordinator having observed the (health-stream) eligibility signal in time.
+	if pm.buildCohortEligibilityStatus().GetSignal() == clustermetadatapb.CohortEligibilitySignal_COHORT_ELIGIBILITY_SIGNAL_INELIGIBLE {
```
dweitzman, line 147: For additional defense in depth, poolers will reject a Recruit() RPC if they're in a cohort ineligible state.

mkindahl (approved):
I have a suspicion that filtering the cohort before sending out Recruits might risk blocking rule change and is not needed, but since this will attempt a new rule later once the eligibility/ineligibility has stabilized, I think we are fine.

I tried to reproduce the scenario I had in mind, but failed to produce something that would be both sensible and lead to blockage.

PR #1333, opened 2026-07-29
feat(multipooler): access multigres schema via admin pool

GuptaManan100:
## Summary

- Create and access the `multigres` sidecar schema exclusively through the admin pool (the true PostgreSQL superuser), so the schema is owned by the superuser and customer roles cannot drop, alter, or read it.
- `PUBLIC` gets only `USAGE` on the schema plus `SELECT` on `multigres.backend_vpid`; every other sidecar table is owner-only.
- Route all `multigres.*` reads/writes (schema and table DDL, heartbeat, consensus rules, tablegroup/shard, `pgbackrest_repos`) through the admin pool; pure system queries stay on the regular pool.

## Description

`InternalQueryService` gains `QueryAdmin` / `QueryAdminArgs` / `QueryAdminMultiStatement` alongside the existing regular-pool methods. Both families share internal `run*` helpers and differ only in which pool they borrow from (`borrowRegular` vs `borrowAdmin`), so there is one interface and one `Executor` implementation — the pool is chosen by method name at the call site.

Previously every `multigres.*` operation went through the regular pool as `PgUser` (despite a doc comment claiming otherwise); only `backend_vpid` writes already used the admin pool. Now the schema DDL, heartbeat reader/writer, consensus rule store, tablegroup/shard inserts, and `pgbackrest_repos` provisioning all use the `QueryAdmin*` variants. Pure system queries (`pg_is_in_recovery`, `ALTER SYSTEM`, WAL LSN reads, `SELECT 1`, the sync-standby manager) stay on the regular pool so the intentionally small admin pool is not exhausted by high-volume polling.

The `USAGE` grant is what lets customer sessions reach `backend_vpid` at all (schema access in PostgreSQL requires both `USAGE` on the schema and a privilege on the object); without it the `SELECT` grant on `backend_vpid` would be dead and the isolation-test lock-wait probe, which runs as the client role, would fail.

### DIFF
--- a/go/services/multipooler/internal/executor/admin_query.go
+++ b/go/services/multipooler/internal/executor/admin_query.go
@@ -1,0 +1,55 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package executor
+
+import (
+	"context"
+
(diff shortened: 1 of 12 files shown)


haritabh17 (approved):
Looks good 🚀

PR #1334, opened 2026-07-29
fix(failover): harden buffering during primary crashes

haritabh17:
## Summary

- retract stale writable routing before PostgreSQL crash recovery, publish the observed post-restart mode immediately, and keep suspected-divergent nodes `DRAINING` until rewind or promotion makes them safe to serve
- buffer failures proven to occur before execution (connection acquisition, Describe, and transient credential lookup) using an internal RPC marker that preserves the original PostgreSQL diagnostic across gRPC
- expose the preserved PostgreSQL diagnostic—or a standard `57P03` fallback—to clients instead of the internal pre-execution marker
- narrowly buffer `25006` and the exact recovery `0A000` rejection only under the existing replay-safety guards
- drain only on fresh self-attested `PRIMARY/SERVING` health; topology-only updates cannot drain the buffer
- treat a failure after draining as a new incident subject to `MinTimeBetweenFailovers`, rather than re-entering buffering and extending capacity occupancy
- run crash-recovery fixtures with orchestration so tests restore a writable leader without weakening production routing safety
- document the expanded failover-buffer state machine and safety boundaries

## Safety

Failures carrying the internal pre-execution marker are retryable because the multipooler proves that query execution and reservation state changes have not started. The marker is removed at the gateway boundary and is never emitted to PostgreSQL clients.

Backend rejections (`25006` and the exact recovery `0A000`) are replayed only for autocommit requests or the first deferred statement of a read-write transaction. Established transactions, ambiguous socket failures, streamed requests after output, and `COMMIT` are never replayed. A node with suspected WAL divergence is not exposed as either writable or readable while rewind/demotion is pending.

Buffering follows `IDLE → BUFFERING → DRAINING → IDLE`. Once a new primary self-reports `PRIMARY/SERVING`, buffered requests are released; a subsequent failure does not re-arm that drain and remains subject to the configured timing guard.

## Validation

- affected unit packages pass after rebasing onto current `upstream/main`
- buffer and pooler-gateway race tests pass
- `make build`
- affected query-serving integration tests pass locally:
  - `TestBufferPlannedFailover`
  - `TestMultigateway_PostgresCrashRecovery`
  - `TestBackendCrashRecoveryUnderLoad`
  - `TestBackendCrashTransparentReconnect`
  - `TestCopyInterruptedByBackendCrash`


go/services/multipooler/internal/executor/executor.go
shedges: we need to invoke preExecutionUnavailableError for CopyReady and CopyOutReady also? Or see if we can modify GetRegularConn, GetRegularConnWithSettings, NewReservedConn to update error with correct flag/marker

go/services/multigateway/buffer/shard_buffer.go
```diff
@@ -93,24 +93,31 @@ func (sb *shardBuffer) waitIfAlreadyBuffering(ctx context.Context) (RetryDoneFun
...
+}
+
+// waitForFailoverEnd either starts buffering or joins an existing buffer.
+// Returns (nil, nil) if buffering is not applicable for this request.
 func (sb *shardBuffer) waitForFailoverEnd(ctx context.Context) (RetryDoneFunc, error) {
 	// Fast path: if draining or idle with recent failover, skip.
 	sb.mu.Lock()
 	switch sb.state {
 	case stateDraining:
-		// Already draining — the new PRIMARY is available. Signal the caller
```
shedges, line 104: any reason to remove these comments? Minor and feel free to defer considering corner case.

Scenario - back to back primary failure.

1/ Gateway has 1000 req in failover buffer
2/ new primary detected and state transition Buffering -> Drain.
3/ after sending 100 waiting requests, new primary again died.
4/ now all remaining request will drained out and fail again. But this case will simply return with no action and caller will perform retry on them.
5/ if max retry count is 4 then 900 * 3 request will flow over n/w even though we know there is no active primary.

go/services/multigateway/poolergateway/availability_errors.go
```diff
@@ -50,11 +50,35 @@ func isNoWritablePrimaryError(err error) bool {
...
+}
+
+// isReadWriteDuringRecoveryError recognizes PostgreSQL's rejection of BEGIN
+// READ WRITE on a standby. Match both SQLSTATE and exact message because 0A000
+// covers many unrelated feature-not-supported errors.
+func isReadWriteDuringRecoveryError(err error) bool {
+	var diagnostic *mterrors.PgDiagnostic
+	return errors.As(err, &diagnostic) &&
+		diagnostic.Code == mterrors.PgSSFeatureNotSupported &&
+		diagnostic.Message == "cannot set transaction read-write mode during recovery"
```
shedges, line 73: Nice to handle special case like BEGIN READ WRITE.  Only thing which can hurt in future is comparing literal string. 
Just to confirm : if we fail to identify then those query will be rejected is the only downside. Right?
If yes, then adding E2E (once framework available) for this scenario will help.

go/services/multigateway/poolergateway/load_balancer.go
shedges: notifyIfLeaderServing is removed. Need to update comment.

shedges (commented):
Rest LGTM

GuptaManan100 (approved):
I'm very happy with this!

go/services/multipooler/internal/executor/executor.go
haritabh17: Agree, updated to invoke `preExecutionUnavailableError`.

go/services/multigateway/poolergateway/load_balancer.go
haritabh17: Ack, updated.

go/services/multigateway/poolergateway/availability_errors.go
haritabh17: Right, the only side effect of upstream PG changing the error message is that we won't buffer them and immediately throw the error back to user.

go/services/multigateway/buffer/shard_buffer.go
haritabh17: Yes, the successive-failure scenario can still cause each request to exhaust its bounded retries. That is intentional because we decided not to re-buffer after draining. Updated to ensure retries remain limited by `DrainConcurrency` rather than becoming one large concurrent wave.

PR #1336, opened 2026-07-29
fix(pgctld): reset pgbackrest pgpass during server bootstrap

niconosenzo:
Trigger an explicit chmod to set pgbackrest pgpass permission during pgctld service startup, this fixes any permission modification made externally, e.g. operator setting volume permissions during PVC mount.

### DIFF
--- a/go/cmd/pgctld/command/server.go
+++ b/go/cmd/pgctld/command/server.go
@@ -328,6 +328,11 @@ func NewPgCtldService(
 	if err := os.WriteFile(pgpassPath, []byte(pgpassContent), 0o600); err != nil {
 		return nil, fmt.Errorf("failed to write pgbackrest pgpass file: %w", err)
 	}
+	// enforce 0600 explicitly, as the operator might have changed these permissions
+	// during volume mount.
+	if err := os.Chmod(pgpassPath, 0o600); err != nil {
+		return nil, fmt.Errorf("failed to set pgbackrest pgpass file permissions: %w", err)
+	}
 	if err := os.Setenv("PGPASSFILE", pgpassPath); err != nil {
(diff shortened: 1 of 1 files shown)


go/cmd/pgctld/command/server.go
```diff
@@ -328,6 +328,11 @@ func NewPgCtldService(
 	if err := os.WriteFile(pgpassPath, []byte(pgpassContent), 0o600); err != nil {
 		return nil, fmt.Errorf("failed to write pgbackrest pgpass file: %w", err)
 	}
+	// enforce 0600 explicitly, as the operator migth have changed these permissions
```
cuongdo: nit: "migth"

go/cmd/pgctld/command/server.go
```diff
@@ -328,6 +328,11 @@ func NewPgCtldService(
...
+	// during volume mount.
+	if err := os.Chmod(pgpassPath, 0o600); err != nil {
```
cuongdo, line 333: Is your assertion that the operator is changing the permissions between the `WriteFile` a few lines above and this line of code? That is plausible but is a very small window.

niconosenzo: Not really, the permissions change is happening during volume mount most likely. The issue is that the lines above:
```go
    if err := os.WriteFile(pgpassPath, []byte(pgpassContent), 0o600); err != nil {
        return nil, fmt.Errorf("failed to write pgbackrest pgpass file: %w", err)
    }
```
set the permissions during file creation, if the file is already created it does not apply the permissions.

cuongdo: ok, that makes sense

cuongdo (approved):

PR #1337, opened 2026-07-29
fix(consensus): avoid rewind-readiness deadlock

dweitzman:
Two issues:
- The first is that a pooler should record that it's the highest known primary _before_ trying to write to WAL. This can help later if the postgres monitor loop needs to determine whether a previous promotion attempt got stuck and failed, in which case the pooler may need to resign leadership
- The second is that multiorch recovery isn't concurrent, so once a rule change is in progress it's not able to recognize the need to call `SetPrimary` to propagate rewind readiness and unblock the rule change. I tried to address this in a more principled way: `SetPrimary` simply distributes information and is idempotent, so it seems harmless to call it early and often. I've pulled it out into a separate loop so it's never blocked by recovery actions. Other options might have been to have rule change try to propagate rewind readiness itself or make recovery actions run async in the background while the recovery loop can keep running, but separating out a SetPrimary loop seemed simpler / lower risk.

### DIFF
--- a/go/common/consensus/proposals.go
+++ b/go/common/consensus/proposals.go
@@ -66,6 +66,20 @@ type RecruitmentResult struct {
 	EligibleLeaders []*clustermetadatapb.ConsensusStatus
 }
 
+// ReplicationPrimaryFromProposal builds the ReplicationPrimary a cohort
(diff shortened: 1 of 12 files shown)


haritabh17 (approved):
LGTM 🚀

PR #1339, opened 2026-07-30
refactor(logging): make structured log output consistent

GuptaManan100:
## Summary

- Make structured log output consistent across services by enforcing a small set of conventions with `sloglint` (in CI) and shaping every record through one handler.
- Fix the biggest offender: `eventlog.Emit` used a fixed `"multigres.event"` message, so every lifecycle event collapsed to the same value in the message column. It now uses the event's canonical type.
- Centralize the logger construction, remove a duplicated legacy logger, and add a conventions doc.
- The bulk of the diff (72 files) is a mechanical sweep to satisfy the newly-enabled linters — no logic changes.

## What this fixes

Events used to serialize with an opaque message:

```
event_message: "multigres.event"   →   event_message: "node.join"
```

Now every log line — event or not — carries a meaningful message, and the error field, message casing, key casing, and context usage are uniform.

## Enforced conventions (`sloglint`, fails CI)

- **`context: scope`** — use `InfoContext`/`ErrorContext`/… whenever a `context.Context` is in scope (so trace correlation is never dropped); plain variants only where there is no context.
- **`forbidden-keys: [err]`** — error values go under the canonical `error` key, never `err`.
- **`msg-style: lowercased`** — messages start lowercased.
- **`key-naming-case: snake`** — attribute keys are snake_case.

Two documented exceptions: messages intentionally prefixed with an operation name or proper noun (`MonitorPostgres: …`, `PostgreSQL is ready`) keep their casing via inline `//nolint`, and `querylog.go` keeps its OpenTelemetry semantic-convention dotted keys (shared with the query's trace and metric attributes). See `docs/general/logging.md`.

## Reviewing this PR

**Files that actually need review (9):**

- `.golangci.yml` — the four `sloglint` rules and the `querylog.go` snake_case exclusion.
- `go/common/servenv/logging.go` — the single `buildHandler`; removal of the duplicated package-level logger.
- `go/common/eventlog/eventlog.go` — `Emit` now uses `event.EventType()` as the message.
- `go/test/endtoend/shardsetup/events.go` — the event-log parser now keys off the `event_type` attribute instead of the old `"multigres.event"` sentinel (coupled to the `Emit` change).
- `go/common/servenv/logging_test.go`, `go/common/servenv/logging_consistency_test.go`, `go/common/eventlog/eventlog_test.go` — tests pinning the guarantees.
- `docs/general/logging.md`, `docs/README.md` — the conventions doc and its index link.

**Files changed only to satisfy the new linters (72):** everything else. Purely mechanical — lowercasing a message's first letter, renaming `err` → `error`, or converting a camelCase attribute key to snake_case. No behavior change. Safe to skim.


docs/general/logging.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Guidance (not enforced)
+
+These require judgment and cannot be mechanically enforced, but reviewers should
+watch for them.
+
+### Log at the right level
+
+`ERROR` means something needs attention. Normal control flow and expected
+conditions are not errors:
```
mkindahl: I think we should probably be even more clear on what error, warning, info, etc. mean and when they are going to be used. I've seen them misused quite often, resulting un unnecessary work when supporting users.

[These are a good guidance](https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-SEVERITY-LEVELS):
- ERROR: Reports an error that caused the current action to abort or unable to proceed. The action or operation has failed and as such has not had the intended effect.
- WARNING: Provides warnings of likely problems, but action or operation proceeded and completed successfully. This could be an action that is probably correct but could require investigation to ensure that it did what was intended and expected.
- INFO: Provides information that might be helpful to users but does not affect execution. For example, truncation of identifiers, the start of execution of an intermediate step of an action, an idempotent operation that was already executed, or similar.
- DEBUG: Information that is useful when debugging the system but usually not useful for an end-user.

go/cmd/pgctld/command/crash_recovery.go
```diff
@@ -104,7 +104,7 @@ func (s *PgCtldService) runCrashRecoveryInDir(
 	logger := s.logger
 	signalPath := s.standbySignalPath()
 	if _, err := os.Stat(signalPath); err == nil {
-		logger.InfoContext(ctx, "Temporarily removing standby.signal for single-user crash recovery",
+		logger.InfoContext(ctx, "temporarily removing standby.signal for single-user crash recovery",
```
mkindahl, line 107: I prefer this, but could be good to explain why this is better.

go/cmd/pgctld/command/restart.go
```diff
@@ -99,21 +99,21 @@ func (s *PgCtldService) RestartPostgreSQLWithResult(mode string, asStandby bool)
...
-		logger.Info("Stopping PostgreSQL server")
+		logger.Info("stopping PostgreSQL server")
 		stopResult, err := s.StopPostgreSQLWithResult(mode)
 		if err != nil {
 			return nil, fmt.Errorf("failed to stop PostgreSQL during restart: %w", err)
 		}
 		result.StoppedFirst = stopResult.WasRunning
 	} else {
-		logger.Info("PostgreSQL is not running, proceeding with start")
+		logger.Info("PostgreSQL is not running, proceeding with start") //nolint:sloglint // message intentionally starts with an operation name or proper noun
```
mkindahl: We are using "Postgres" rather than "PostgreSQL" in all documentation, so it is probably a good idea to stick to that in error messages as well.

```suggestion
		logger.Info("Postgres is not running, proceeding with start") //nolint:sloglint // message intentionally starts with an operation name or proper noun
```

go/cmd/pgctld/command/restart.go
```diff
@@ -151,10 +151,10 @@ func (s *PgCtldService) RestartPostgreSQLWithResult(mode string, asStandby bool)
 	result.PID = startResult.PID
 	if asStandby {
 		result.Message = "PostgreSQL server restarted as standby successfully"
-		logger.Info("PostgreSQL server restarted as standby successfully", "pid", result.PID)
+		logger.Info("PostgreSQL server restarted as standby successfully", "pid", result.PID) //nolint:sloglint // message intentionally starts with an operation name or proper noun
```
mkindahl:
```suggestion
		logger.Info("Postgres server restarted as standby successfully", "pid", result.PID) //nolint:sloglint // message intentionally starts with an operation name or proper noun
```

go/cmd/pgctld/command/restart.go
```diff
@@ -151,10 +151,10 @@ func (s *PgCtldService) RestartPostgreSQLWithResult(mode string, asStandby bool)
...
 	} else {
 		result.Message = "PostgreSQL server restarted successfully"
-		logger.Info("PostgreSQL server restarted successfully")
+		logger.Info("PostgreSQL server restarted successfully") //nolint:sloglint // message intentionally starts with an operation name or proper noun
```
mkindahl:
```suggestion
		logger.Info("Postgres server restarted successfully") //nolint:sloglint // message intentionally starts with an operation name or proper noun
```

go/cmd/pgctld/command/start.go
```diff
@@ -136,7 +136,7 @@ func (s *PgCtldService) StartPostgreSQLWithResult() (*StartResult, error) {
 
 	// Check if PostgreSQL is already running
 	if isPostgreSQLRunning(config.PostgresDataDir) {
-		logger.Info("PostgreSQL is already running")
+		logger.Info("PostgreSQL is already running") //nolint:sloglint // message intentionally starts with an operation name or proper noun
```
mkindahl:
```suggestion
		logger.Info("Postgres is already running") //nolint:sloglint // message intentionally starts with an operation name or proper noun
```

go/cmd/pgctld/command/start.go
```diff
@@ -181,7 +181,7 @@ func (s *PgCtldService) StartPostgreSQLWithResult() (*StartResult, error) {
 	}
 
 	result.Message = "PostgreSQL server started successfully"
-	logger.Info("PostgreSQL server started successfully")
+	logger.Info("PostgreSQL server started successfully") //nolint:sloglint // message intentionally starts with an operation name or proper noun
```
mkindahl:
```suggestion
	logger.Info("Postgres server started successfully") //nolint:sloglint // message intentionally starts with an operation name or proper noun
```

go/cmd/pgctld/command/start.go
```diff
@@ -345,7 +345,7 @@ func waitForPostgreSQLWithConfig(logger *slog.Logger, config *pgctld.PostgresCtl
 		case <-timeout:
 			// On timeout, include diagnostic information
 			logTail := readLogTail(logPath, 20)
-			logger.Error("PostgreSQL startup timeout",
+			logger.Error("PostgreSQL startup timeout", //nolint:sloglint // message intentionally starts with an operation name or proper noun
```
mkindahl:
```suggestion
			logger.Error("Postgres startup timeout", //nolint:sloglint // message intentionally starts with an operation name or proper noun
```

go/cmd/pgctld/command/start.go
```diff
@@ -363,7 +363,7 @@ func waitForPostgreSQLWithConfig(logger *slog.Logger, config *pgctld.PostgresCtl
 				if err != nil {
 					// No PID file means PostgreSQL never started or crashed immediately
 					logTail := readLogTail(logPath, 20)
-					logger.Error("PostgreSQL process not running during startup",
+					logger.Error("PostgreSQL process not running during startup", //nolint:sloglint // message intentionally starts with an operation name or proper noun
```
mkindahl:
```suggestion
					logger.Error("Postgres process not running during startup", //nolint:sloglint // message intentionally starts with an operation name or proper noun
```

mkindahl (approved):
Looks good. Two suggestions for improvement:
- Clarify how the log levels should be used to avoid misunderstandings.
- Documentation uses "Postgres" rather than "PostgreSQL" so it is probably sensible for us to do the same regarding log messages.

GuptaManan100:
Done! Addressed all review comments ❤️

PR #1342, opened 2026-07-30
docs(query-serving): add setseed backend-pinning design doc

shedges:
## Summary

- Adds a design doc for the setseed() connection-pinning, mirroring advisory_locks_design.md's structure.
- Covers why setseed() needs backend pinning, the acquire-only detection, and the property that sets it apart from every other reservation reason: it is sticky and survives DISCARD ALL, only releasing on real connection teardown.


haritabh17 (approved):
LGTM 🚀

GuptaManan100 (approved):

PR #1344, opened 2026-07-31
fix(pgctld): live-include extra postgresql.conf instead of copying at initdb

GuptaManan100:
## Summary

- `--pg-initdb-extra-conf` files are now live-included into `postgresql.conf` via `include_if_exists` directives instead of having their bytes copied in at initdb, so edits to those files take effect on restart/reload.
- This fixes the multigres-operator flow where a per-shard config ConfigMap is mounted read-only into the pod and updated in place: the change is now picked up on the next restart or `pg_reload_conf()` SIGHUP.
- Adds an e2e test (plus unit-test updates) that edits the referenced file in place and confirms the new value applies via both SIGHUP reload and a full restart.

## Problem

`GeneratePostgresServerConfig` runs only from `pgctld init`, which early-returns when the data dir is already initialized. `appendExtraConfFiles` copied the *contents* of each extra-conf file into `PGDATA/postgresql.conf` on the persistent data volume. So the operator's config was baked into `PGDATA/postgresql.conf` once, at first initdb. On a later restart the data dir already exists → init is skipped → the mounted file (now holding new content) is never re-read, and postgres loads the stale copy. A `pg_reload_conf()` SIGHUP has the same problem — postgres reads `PGDATA`, not the mount. Config changes therefore applied only at first cluster creation.

## Solution

`appendExtraConfFiles` now writes an `include_if_exists '<absolute path>'` directive per file, in order, at the end of `PGDATA/postgresql.conf`, instead of copying the bytes. `postgresql.conf` then points at the live mounted file(s): every server start and every `pg_reload_conf()` SIGHUP re-reads the current content.

- Precedence is preserved — includes are emitted after the generated template, one per file in order, so extras still override the template (same last-write-wins as before).
- `include_if_exists` (not `include`) means a missing/unmounted file is silently skipped rather than failing startup. Paths are absolute and quoted per postgresql.conf syntax.
- `ReadPostgresServerConfig`/`parseConfigInto` already follow `include_if_exists`, so pgctld's in-memory struct still resolves the values correctly (covered by existing + updated tests), and the values remain effective at first initdb (postgres reads `postgresql.conf` → follows the include → mounted file).
- Flag help and env-var docs updated: the file is now live-included, not appended verbatim at init time. The `--pg-initdb-extra-conf` / `POSTGRES_INITDB_EXTRA_CONF` name is kept as-is.

## Testing

- Unit: `postgresconfig_gen_test.go` now asserts the `include_if_exists` directives are written (and contents are *not* copied), a missing file no longer errors, and a new liveness case confirms re-reading reflects an edited file.
- E2e: `TestPgInitdbExtraConfLiveReload` starts real postgres, edits the referenced file in place, and verifies the new value applies via both `pg_reload_conf()` (SIGHUP, `work_mem`) and `pgctld restart` (`max_connections`). Verified red against the old copy-the-bytes behavior and green with the fix.

### DIFF
--- a/go/services/pgctld/postgresconfig_gen.go
+++ b/go/services/pgctld/postgresconfig_gen.go
@@ -162,10 +162,10 @@ func (cnf *PostgresServerConfig) appendExtraConfFiles(paths []string) error {
 	for _, p := range paths {
 		abs, err := ExpandToAbsolutePath(p)
 		if err != nil {
-			return fmt.Errorf("resolving extra postgres config %q: %w", p, err)
+			return fmt.Errorf("failed when resolving extra postgres config %q: %w", p, err)
 		}
 		if _, err := fmt.Fprintf(f, "include_if_exists %s\n", quoteConfValue(abs)); err != nil {
-			return fmt.Errorf("appending include for %q: %w", p, err)
+			return fmt.Errorf("failed to append include for %q: %w", p, err)
 		}
 	}
 	return nil
--- a/go/services/pgctld/postgresconfig_gen.go
+++ b/go/services/pgctld/postgresconfig_gen.go
@@ -49,7 +49,8 @@ func ExpandToAbsolutePath(dir string) (string, error) {
 }
 
 // GeneratePostgresServerConfig writes postgresql.conf from the embedded template,
-// appends extraConfFiles verbatim (postgres last-write-wins), then reads it back.
+// appends include_if_exists directives for extraConfFiles (postgres
+// last-write-wins), then reads it back.
 func GeneratePostgresServerConfig(poolerDir string, pgUser string, extraConfFiles []string) (*PostgresServerConfig, error) {
 	// Create minimal config for template generation
 	if poolerDir == "" {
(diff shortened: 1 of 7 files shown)


go/services/pgctld/postgresconfig_gen.go
```diff
@@ -146,26 +156,27 @@ func (cnf *PostgresServerConfig) appendExtraConfFiles(paths []string) error {
...
+	}
 	for _, p := range paths {
-		data, err := os.ReadFile(p)
+		abs, err := ExpandToAbsolutePath(p)
 		if err != nil {
-			return fmt.Errorf("reading extra postgres config %q: %w", p, err)
-		}
-		if _, err := fmt.Fprintf(f, "\n## %s\n", p); err != nil {
-			return fmt.Errorf("appending %q: %w", p, err)
+			return fmt.Errorf("resolving extra postgres config %q: %w", p, err)
```
mkindahl:
```suggestion
			return fmt.Errorf("failed when resolving extra postgres config %q: %w", p, err)
```

go/services/pgctld/postgresconfig_gen.go
```diff
@@ -146,26 +156,27 @@ func (cnf *PostgresServerConfig) appendExtraConfFiles(paths []string) error {
...
 		}
-		if _, err := f.Write(data); err != nil {
-			return fmt.Errorf("appending %q: %w", p, err)
-		}
-		if len(data) > 0 && data[len(data)-1] != '\n' {
-			if _, err := f.WriteString("\n"); err != nil {
-				return fmt.Errorf("appending %q: %w", p, err)
-			}
+		if _, err := fmt.Fprintf(f, "include_if_exists %s\n", quoteConfValue(abs)); err != nil {
+			return fmt.Errorf("appending include for %q: %w", p, err)
```
mkindahl:
```suggestion
			return fmt.Errorf("failed to append include for %q: %w", p, err)
```

mkindahl (approved):
Minor comment about error message wording, but otherwise looks good so approving.

GuptaManan100:
Thank you, applied the fixes!!

PR #1345, opened 2026-07-31
feat(multipooler): add ReloadConfig RPC

GuptaManan100:
## Summary

- Adds a `ReloadConfig` RPC to the multipooler manager service that triggers a PostgreSQL config reload and confirms it took effect, so the operator can apply reload-safe GUC changes without recreating the pod.
- The reload is triggered via pgctld (SIGHUP, no superuser needed) and confirmed by waiting for `pg_conf_load_time()` to advance past a baseline captured before the trigger.
- Wires the RPC through the gRPC service, the `rpcclient.MultipoolerClient` interface (real client + fake), and adds unit + endtoend coverage.

## Description

The multigres-operator needs a single per-pod RPC to apply reload-safe PostgreSQL config changes (sighup/user/superuser-context GUCs) via SIGHUP instead of a rolling restart, and to verify the reload actually landed. multipooler is the right home: the operator already dials it, and it already holds both a pgctld client and a live PostgreSQL connection. pgctld's own `ReloadConfig` only delivers the signal and can't verify, so the new RPC reuses it to trigger and then reads PostgreSQL to confirm.

The response is a single `config_load_time` timestamp. Set means the reload was confirmed; unset means PostgreSQL was not running so nothing reloaded (the caller treats that as retryable). A hard gRPC error means a genuine failure (the reload fired but `pg_conf_load_time()` never advanced within the budget, or the read failed). The caller verifies its config-file change was re-read by comparing the returned time against its own write — no GUC values cross the wire, and reload-vs-restart classification stays entirely with the caller.

Confirmation uses a `time.Now()` baseline captured before the SIGHUP and accepts any `pg_conf_load_time()` at or after it. Because that is an absolute reference, every prior reload is strictly in the past, so a pooled poll that lands on a backend which has not yet processed the signal simply reports its old value and the loop keeps polling. This relies on multipooler and PostgreSQL sharing a clock, which holds since they are colocated in the pod.

This assumes the separate pgctld change that includes the operator-provided extra config file (so a SIGHUP re-reads updated values) rather than copying it into `postgresql.conf` at startup. No ALTER SYSTEM path and no classification logic are included here.

### DIFF
--- a/go/services/multipooler/internal/manager/rpc_reload_config.go
+++ b/go/services/multipooler/internal/manager/rpc_reload_config.go
@@ -21,12 +21,9 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/multigres/multigres/go/common/mterrors"
-	"github.com/multigres/multigres/go/common/timeouts"
-	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
 	multipoolermanagerdatapb "github.com/multigres/multigres/go/pb/multipoolermanagerdata"
 	pgctldpb "github.com/multigres/multigres/go/pb/pgctldservice"
-	"github.com/multigres/multigres/go/services/multipooler/internal/executor"
-	"github.com/multigres/multigres/go/tools/retry"
+	"github.com/multigres/multigres/go/services/multipooler/internal/manager/consensus"
 )
 
 // ReloadConfig triggers a PostgreSQL configuration reload on this multipooler's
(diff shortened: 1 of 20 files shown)


go/services/multipooler/internal/manager/rpc_reload_config.go
```diff
@@ -0,0 +1,127 @@
...
+	return &multipoolermanagerdatapb.ReloadConfigResponse{
+		ConfigLoadTime: timestamppb.New(loadTime),
+	}, nil
+}
+
+// waitForConfigReload polls pg_conf_load_time() until it reports a time at or
+// after baseline, returning that time. It backs off between polls and gives up
+// with a DEADLINE_EXCEEDED error if the reload is not observed within the
+// budget (e.g. the SIGHUP never reached the backend).
+func (pm *MultipoolerManager) waitForConfigReload(ctx context.Context, baseline time.Time) (time.Time, error) {
```
mkindahl: There is a function ReloadPostgresConfig in consensus package. Can you reuse it?

mkindahl (commented):
Just one question: there is a similar function consensus.ReloadPostgresConfig which does close to the same thing done. Is there anything that can be re-used?

GuptaManan100:
@mkindahl Yes! There was an overlap, I've extracted that part out ❤️

mkindahl (approved):

PR #1346, opened 2026-07-31
feat(pgctld): seed day-0 role passwords and settings at init

mkindahl:
## Summary

Service roles created by the init SQL (e.g. `authenticator`, `supabase_auth_admin`) are
created **without** passwords. Today those passwords are set only by a post-boot control-plane
write over the gateway; on a freshly provisioned cluster that write can fail (e.g. the gateway
is unreachable during the first minutes) with no retry, leaving the roles passwordless and the
dependent services unable to authenticate.

This adds an **init-time** path so a newborn cluster is usable with no post-boot network write.

## What's in this PR

- **`--pg-init-secrets-file`** (`POSTGRES_INIT_SECRETS_FILE`): a JSON file of per-project day-0
  state — role passwords and per-database settings:

  ```json
  {
    "roles": { "authenticator": "…", "supabase_auth_admin": "…" },
    "database_settings": { "postgres": { "app.settings.jwt_exp": "3600" } }
  }
  ```

- **`applyInitSecrets`** runs as the last step of `postInitdbSetup`, after the init SQL has
  created the roles. It emits `ALTER ROLE … PASSWORD` / `ALTER DATABASE … SET`, fed to `psql`
  over **stdin** so secret values never appear in argv or `/proc/<pid>/cmdline`.
- **Verify-and-fail:** after applying, it queries `pg_authid` and fails init if any login role
  is still passwordless, so the cluster never comes up "healthy but passwordless".
- **Retry-safe rollback:** `initdb` writes `PG_VERSION` before this step, so a failed setup
  would otherwise read as initialized and a retry would no-op past it. On any `postInitdbSetup`
  failure the data directory is now removed, so a retry re-runs the full init — transient
  failures (e.g. the Secret not yet mounted) self-heal, persistent ones stay loud.
- **Log-safe:** the apply session disables statement logging (`SET log_statement = 'none'`,
  `log_min_error_statement = 'panic'`, `log_min_duration_statement = -1`) so the plaintext
  password isn't written to the PostgreSQL server log — the generated config uses
  `log_statement = 'ddl'`, which would otherwise capture it.
- **Docs:** `docs/pgctld-init.md` documents the init flags, execution order, superuser password
  resolution, and init secrets.

Role values are plaintext (Postgres hashes per `password_encryption`); the code is
value-agnostic — a pre-hashed `SCRAM-SHA-256$…` verifier would apply verbatim — but plaintext
is the supported input mode.

## Testing

- **Unit:** SQL builder (deterministic sorted output, identifier/literal quoting, rejects empty
  role/password/db/key, absent/empty role and settings combinations), loader (parse + edge
  cases), and error-output redaction.
- **E2E (real PostgreSQL, verified on 18):** a seeded role authenticates over SCRAM, a wrong
  password is rejected, `pg_authid` shows the password set, and `app.settings.jwt_exp` is
  applied; an unseeded login role fails init; a failed init rolls back the data dir and a
  corrected retry succeeds; and the plaintext password does not leak into any server log.

## Context

Part of MINT-60 (provisioning leaves service roles without passwords, auth/postgrest). This is
the pgctld/multigres piece only — it does not on its own close MINT-60: the multigres-operator
still needs to mount the per-project `init-secrets` Secret into the pgctld init step, and the
control plane needs to materialize it. Those are tracked separately.

### DIFF
--- a/go/cmd/pgctld/command/init_secrets.go
+++ b/go/cmd/pgctld/command/init_secrets.go
@@ -1,0 +1,226 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package command
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/multigres/multigres/go/common/parser/ast"
+)
+
+// initSecrets is the day-0 state handed to pgctld at init via
+// --pg-init-secrets-file: per-project role login credentials and database
+// settings that cannot be baked into the shared image. The file is a
+// per-project Kubernetes Secret; see the MINT-60 plan for the delivery model.
+//
+// Role values are opaque: a plaintext password (Postgres hashes it per
+// password_encryption) or a pre-hashed "SCRAM-SHA-256$..." verifier (stored
+// verbatim). pgctld does not distinguish the two — it passes the value straight
+// into ALTER ROLE ... PASSWORD.
+type initSecrets struct {
+	// Roles maps additional login role names to its password/verifier. The
+	// superuser is NOT expected here and it is wrong to add it. It is set at
+	// initdb via --pwfile from its own Secret.
+	Roles map[string]string `json:"roles"`
+
+	// DatabaseSettings maps a database name to GUCs applied via
+	// ALTER DATABASE <db> SET <key> TO <value> (e.g. app.settings.jwt_exp).
+	DatabaseSettings map[string]map[string]string `json:"database_settings"`
+}
+
+// loadInitSecrets reads and parses the init-secrets JSON file. Unknown fields
+// are ignored so a newer platform payload does not break an older pgctld.
+func loadInitSecrets(path string) (*initSecrets, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read init secrets file %q: %w", path, err)
+	}
+	var s initSecrets
+	if err := json.Unmarshal(data, &s); err != nil {
+		return nil, fmt.Errorf("cannot parse init secrets file %q: %w", path, err)
+	}
+	return &s, nil
+}
+
+// applyInitSecrets sets role passwords and database settings from the
+// init-secrets file over the transient socket. It runs as the last step of
+// postInitdbSetup, after the init-scripts have created the roles.
+//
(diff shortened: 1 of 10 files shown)


go/cmd/pgctld/command/pg_instance.go
```diff
@@ -142,16 +143,21 @@ func (p *pgInstance) stop() {
 }
 
 // psql runs a psql command against this instance connected to database,
-// appending args after the standard -h/-p/-U/-d connection flags.
-// Returns combined stdout+stderr and any error.
-func (p *pgInstance) psql(database string, args ...string) ([]byte, error) {
+// appending args after the standard -h/-p/-U/-d connection flags. Returns
+// combined stdout+stderr and any error. You can pass in a Reader represeting
```
niconosenzo:
```suggestion
// combined stdout+stderr and any error. You can pass in a Reader representing
```

go/cmd/pgctld/command/init_secrets.go
```diff
@@ -0,0 +1,197 @@
...
+// failure rather than a silently-broken cluster.
+func applyInitSecrets(logger *slog.Logger, pg *pgInstance, cfg PgCtldServiceConfig) error {
+	if cfg.InitSecretsFile == "" {
+		return nil
+	}
+	secrets, err := loadInitSecrets(cfg.InitSecretsFile)
+	if err != nil {
+		return err
+	}
+	if len(secrets.Roles) == 0 && len(secrets.DatabaseSettings) == 0 {
```
niconosenzo, line 79: shouldn't it be `||` instead of `&&` ?

niconosenzo (commented):
Solid! left 1 likely blocker comment and 1 nit. Let's also hear others' opinions about the approach.

go/cmd/pgctld/command/init_secrets.go
mkindahl: Well, that would generate an ~error~ message for this configuration (note the missing `database_settings`):

```json
{
  "roles": {
    "authenticator": "…",
    "supabase_auth_admin": "…"
  }
}
```

However, I see that it is good to extend the tests to make it clear what is supported and what is not.

We could also generate a more precise message that tells which of them were not set.

go/cmd/pgctld/command/init_secrets.go
```diff
@@ -0,0 +1,206 @@
...
+	var sb strings.Builder
+
+	for _, role := range sortedKeys(secrets.Roles) {
+		if role == "" {
+			return "", errors.New("init secrets: empty role name")
+		}
+		if secrets.Roles[role] == "" {
+			return "", fmt.Errorf("init secrets: empty password for role %q", role)
+		}
+		fmt.Fprintf(&sb, "ALTER ROLE %s PASSWORD %s;\n",
```
cuongdo, line 153: https://github.com/multigres/multigres/blob/11675fd7651b9b34358e0bccab27c0084a61bc02/config/postgres/template.cnf#L190

turns on statement logging for DDLs. I belive ALTER ROLE is considered a DDL, so the password would make it into Postgres logs?

cuongdo: One easy fix is to set `log_statement` to `none` before executing the ALTER ROLE, and reset it to its previous value after.

mkindahl: Thanks for catching this! I looked around for best practices to see if there are other alternatives, but this seems to be what is recommended (turning of `log_statements` for the duration of the DDLs).

niconosenzo (approved):

PR #1347, opened 2026-07-31
fix(multipooler): stop noisy/misleading logging in the health and replication paths

dweitzman:
Four independent logging fixes bundled together since they're all small and low-risk:

- Heartbeat reader/writer and connpoolmanager's user-pool rebalancer were logging at INFO/ERROR every tick regardless of whether anything changed — now only log on actual state transitions.
- Sync-replication config logging fired unconditionally on every postgres-monitor tick instead of only when the config actually changed.
- SetPrimary no-op path now logs which coordinator authored the current rule, to make no-op investigations tractable.
- Several log call sites were logging enum fields as raw ints instead of their string name.


cuongdo (approved):

Verolop (approved):
lgtm!

PR #1348, opened 2026-07-31
fix(multiorch): cache successful SetPrimary target to stop redundant re-sends

dweitzman:
leader-info propagation was re-issuing SetPrimary to a follower on every recovery cycle even after it had already succeeded, because it had no memory of what it last successfully sent. Adds a comparison helper and caches the last successfully-propagated target so propagation only re-sends when something has actually changed.


Verolop (approved):
lgtm!

PR #1352, opened 2026-08-04
feat(query-serving): analyze PL/pgSQL bodies for unsafe session state

GuptaManan100:
## Summary

- **Closes the Tier 1 vector**: a `DO` or `CREATE FUNCTION`/`PROCEDURE` body could change backend session state (`SET`, `set_config`) or reach a blocklisted call invisibly to the gateway's session tracker. The planner now extracts the body, parses it (PL/pgSQL or SQL), and runs the existing expression-level filter over every embedded SQL fragment. Only `DO` and the `AS '…'`-string forms carry an opaque body; `CREATE RULE`/`TRIGGER`/`EVENT TRIGGER` and the SQL-standard `BEGIN ATOMIC` body are already reached by the top-level `FuncCall` walk.
- **Session state in a body — rejected unless self-reverting**: `SET LOCAL`, `SET TRANSACTION`, and `set_config(…, is_local := true)` are allowed (they revert at transaction end and cannot leak to the next pooled client); plain `SET`, `RESET`, `DISCARD`, `LISTEN`, and non-local `set_config` are rejected, since a body may apply them conditionally (in an `IF`/`LOOP`/handler) and the gateway can't faithfully mirror them. Cluster-managed GUCs stay rejected even as `SET LOCAL`.
- **Dynamic `EXECUTE` — injection-safe forms accepted**: a string literal is analyzed inline; an assembled string proven structurally fixed — built only from `quote_ident` / `quote_literal` / `format()` with a constant `%I`/`%L`-only template / `||` of those — is accepted (its skeleton and every interpolated value are analyzed, so a blocklisted call hidden in an argument is still caught); anything that can inject arbitrary statement text (`%s`, a bare variable, raw concatenation) is rejected.
- **Languages**: `plpgsql`/`sql` bodies are parsed and analyzed; `c`/`internal` are allowed (a symbol reference into a library/binary — nothing session-state-shaped to inspect, and creation is gated elsewhere); other opaque languages (`plperl`, `plpython`, …) fail closed.
- **Rejections keep the transaction open**: a gateway policy rejection is raised before the statement reaches a backend, so it no longer aborts the surrounding transaction — the session stays in-block and the client can continue instead of being forced to `ROLLBACK` (and the rest of a transaction no longer cascades into "current transaction is aborted"). Applies to every rejection layer.
- **`--unsafe-pooler-mode`** (default off) disables all unsafe-statement rejections for trusted, single-tenant deployments that accept the risk; routing signals (tracked `set_config`, advisory-lock pinning, `current_setting` rewrites) are still gathered, so routing is unaffected.

## PL/pgSQL parser

Body analysis needs each embedded query to parse as stand-alone SQL. The PL/pgSQL parser gained: **`INTO`-clause extraction** from execsql statements (so `SELECT … INTO x` yields valid SQL plus a separate target list, while `INSERT/MERGE/IMPORT … INTO` are left alone); **seeded parsing** (`ParseSeed`) that pre-declares the names `do_compile` would — each parameter as `$N` and by name, trigger `NEW`/`OLD`/`TG_*`, event-trigger `tg_event`/`tg_tag` — so `param := …` / `new.field := …` resolve as assignment targets instead of being misread as SQL and rejected; and support for **`OPEN` cursor arguments**, **nested/subscripted and composite-field assignment targets**, and **cursor `FOR` loops**.

## Compatibility validation

Validated end-to-end against the pgregress compatibility suite (regression + isolation + contrib + external extensions). Divergences the feature introduces are recorded as accepted-divergence patches across core regression (`guc`, `alter_table`, `plpgsql`, `join_hash`, `memoize`, `select_parallel`, …) and the external extensions **PostGIS, hypopg, pgtap, plpgsql_check, http, pg_partman**. Where an extension's test *scaffolding* (not its subject) trips the analysis, the harness accommodates it rather than weakening the feature — a `--unsafe-pooler-mode` gateway for scaffolding that must run, and primary-side / per-test helper seeding so the honest single rejection line is the only divergence recorded.

## Docs & tests

`docs/query_serving/unsafe_statement_rejection.md` documents the Tier 1 body analysis and the non-aborting transaction behavior; `docs/query_serving/plpgsql_parser.md` covers the parser work. Covered by planner unit tests (accept/reject/bypass/fail-closed, an accept/reject body corpus, injection-safe-`EXECUTE` and dynamic-`FOR` coverage) and an end-to-end suite through a real multigateway → PostgreSQL.


go/services/multigateway/planner/unsafe_plpgsql.go
```diff
@@ -0,0 +1,747 @@
...
+	}
+	s, ok := nl.Items[0].(*ast.String)
+	if !ok {
+		return ""
+	}
+	return s.SVal
+}
+
+// bareFuncName returns a FuncCall's unqualified, lower-cased name, or "" if the
+// name is schema-qualified (not a trusted pg_catalog builtin for our purposes).
```
shedges, line 570: we also need to handle user defined schema qualified scenario along with pg_catalog.

e.g. here user change schema name to `myschema` and invoking `quote_ident()`. After this, an unqualified quote_ident(...) inside a body resolves to the user function, not `pg_catalog`
```
  CREATE FUNCTION myschema.quote_ident(text) RETURNS text AS $$
  BEGIN RETURN 'x''); DROP TABLE users; --'; END;
  $$ LANGUAGE plpgsql;
  SET search_path = myschema, pg_catalog;
```

shedges (approved):
Rest LGTM 🚀

go/services/multigateway/handler/handler.go
```diff
@@ -351,7 +351,9 @@ func (h *MultigatewayHandler) HandleQuery(ctx context.Context, conn *server.Conn
 		if err != nil {
 			// If we're in an active transaction and the query failed,
 			// transition to aborted state. The client must ROLLBACK to recover.
-			if conn.TxnStatus() == protocol.TxnStatusInBlock {
+			// A gateway policy rejection (feature_not_supported) never reached the
+			// backend, so its transaction is still open — leave the session in-block.
+			if conn.TxnStatus() == protocol.TxnStatusInBlock && !mterrors.IsGatewayRejection(err) {
```
haritabh17, line 356: Should we force-kill the xact even on gateway errors? This can potentially behave oddly with some clients where the expectation is that any error returned on the wire results in xact being closed.

haritabh17 (approved):
Rest looks good 👍

go/services/multigateway/planner/unsafe_plpgsql.go
GuptaManan100: This is a valid vulnerability. Unfortunately, there isn't a great fix:
- (A) Only trust explicitly pg_catalog.-qualified calls → safe but rejects the common bare form (many false rejections).
- (B) Keep bare trusted but reject if the name could be user-defined → the gateway can't know the catalog/search_path, so it can't detect shadowing reliably.

go/services/multigateway/handler/handler.go
GuptaManan100: The reason I introduced this was because otherwise we get a lot of cascaded failure errors for pg_regress tests which create these functions as part of a long transaction. Even if we pre-create them, the creation fails, and this was the only way to not have `aborted transaction` errors everywhere.

PR #1354, opened 2026-08-05
refactor(multipooler): single source of truth for primary_conninfo

mkindahl:
## What

Collapse the two hand-maintained definitions of `primary_conninfo` — an inline builder in `setPrimaryConnInfoLocked` and a field-picking comparator (`matchesRecorded`) — into a single source of truth.

## Why

The recently-merged passfile-drift fix (`fix(multipooler): detect missing passfile in primary_conninfo drift check`) patched the immediate symptom: it folded `passfile=` into the drift comparison so a passwordless conninfo (`fe_sendauth: no password supplied`) self-heals. But it left the underlying hazard in place — the value is still *written* in one place and *compared* in another, and the two can silently diverge. `matchesRecorded` still compares only host, port, and passfile; a conninfo that drifted in `user` or `application_name` reads as "no drift" and is silently accepted — the same blind-spot pattern one field over.

This change removes the bug class rather than another symptom.

## How

- Reuse the `PrimaryConnInfo` proto as the single in-memory representation.
- `buildPrimaryConnInfo(ci)` — the ONE place the conninfo string is rendered (inverse of `parseAndRedactPrimaryConnInfo`). `setPrimaryConnInfoLocked` writes via it.
- `expectedPrimaryConnInfo(target)` / `expectedPrimaryConnInfoAt(host, port)` — assemble the expected value from the authoritative inputs (recorded leader host/port, `connPoolMgr.PgUser()`, `servicePoolerID` app name, `pgpassFilePath()`). Both the write path and the drift check assemble here, so they cannot diverge.
- `connInfoDrifted(actual, expected)` — the single comparison site, across ALL managed fields (host, port, user, application_name, passfile), keeping the "only flag drift we can fix" guard for an unknown passfile. Removes the `passfileDrift` special-case and the duplicated host/port-only checks.
- Split the two concerns `primaryConnInfoDiffersFromRecorded` mixed: `connInfoReconcileAllowed()` (may we reconcile at all? — the gates) vs `connInfoDrifted()` (did the value drift? — pure, testable).
- DRY the host/port equality via `connInfoPointsAt(actual, host, port)`, shared with `standbyStuckDiverged`.

## Behavioral delta

The drift check now also detects and self-heals a conninfo that drifted in `user` or `application_name`, which today is silently accepted. No change to what gets written on the happy path.

## Tests

- `TestConnInfoDrifted` — per-field table that forces every builder-emitted field through the comparison; if a field is added to the builder but not the comparator, the round-trip case fails.
- `TestBuildPrimaryConnInfo` / `TestPrimaryConnInfoRoundTrip` — builder ↔ parser inverse.
- Updated `TestPrimaryConnInfoDiffersFromRecorded` for the full-field comparison (added `user` and `application_name` mismatch cases, plus a runtime guard tying the literal identity strings to what the manager derives).

Verified locally: `go build ./go/...`, full `manager` package with `-race`, golangci-lint (0 issues), and the conninfo-focused multipooler e2e tests + multiorch `TestFixReplication` / `TestSpuriousFailoverRecovery` against PostgreSQL 17.

### DIFF
--- a/go/services/multipooler/internal/manager/pg_config.go
+++ b/go/services/multipooler/internal/manager/pg_config.go
@@ -123,6 +123,26 @@ func parseSynchronousStandbyNames(value string) (*SyncStandbyConfig, error) {
 	}, nil
 }
 
+// buildPrimaryConnInfo renders a PrimaryConnInfo into a PostgreSQL
+// primary_conninfo connection string. This is the ONE place the string is
+// constructed: setPrimaryConnInfoLocked writes it via ALTER SYSTEM, and
+// expectedPrimaryConnInfo assembles the value the drift check compares against,
+// so a field added here is emitted on the write path and (via connInfoDrifted)
+// forced through the comparison — the two can no longer silently diverge.
+//
+// It is the inverse of parseAndRedactPrimaryConnInfo: fields it emits are the
+// fields that function parses back. The passfile clause is omitted when empty,
+// mirroring "no passfile written yet" (pgpassPath unknown); the password itself
+// is never embedded — it lives in the pgpass file the passfile= path points at.
+func buildPrimaryConnInfo(ci *multipoolermanagerdata.PrimaryConnInfo) string {
+	connInfo := fmt.Sprintf("host=%s port=%d user=%s application_name=%s",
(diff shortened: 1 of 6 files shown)


haritabh17 (approved):
LGTM ✅

PR #1357, opened 2026-08-06
fix(pgctld): recover a crashed standby via standby-mode recovery

mkindahl:
## Summary

pgctld's crash-recovery path ran single-user (`postgres --single`) recovery for **any** not-cleanly-stopped standby. Single-user recovery runs in primary mode and does not follow timeline-history switches, so for a healthy standby that had followed a leader's timeline switch it advanced the checkpoint on the node's **old** timeline, past the point where the leader's timeline forked. The subsequent standby start then FATAL-looped `requested timeline N is not a child of this server's history`, and the pooler monitor retried it every few seconds forever while the pod kept reporting Ready.

This PR scopes the single-user step to nodes that may actually have diverged, and lets a clean follower recover through ordinary standby-mode crash recovery (which follows the timeline switch).

Part of [MUL-1009](https://linear.app/supabase/issue/MUL-1009) — linked intentionally *without* a closing keyword, because this PR does not fully resolve the issue on its own. It restores automatic recovery of a crashed standby, but closing MUL-1009 also needs the readiness gap fixed so a postmaster-dead pod is marked NotReady (see follow-ups). The issue should stay open until that lands.

## Problem (before)

**Why pgctld runs single-user recovery at all:** it is how a node reaches a *clean shutdown*, which is the state `pg_rewind` requires. It was introduced for the demote path (a SIGKILLed former primary that must be rewound onto the new leader's timeline) and later reused to rescue a standby wedged by an earlier `pg_rewind` that stamped `minRecoveryPoint` onto the wrong timeline. In both cases the node has genuinely diverged and needs a rewind, and single-user recovery is the prerequisite for it — a standby-only concern, so the step is gated on `standby.signal` (`postgres --single` refuses to run with it present, hence the remove/recreate).

**The bug** is that `Start(allow_crash_recovery)` applied that step to **every** not-cleanly-stopped standby. The crashed node here is a **clean follower**: its WAL is a clean descendant of the current leader's timeline, with nothing un-replicated. It needs nothing more than ordinary standby-mode crash recovery. Instead, pgctld force-runs single-user recovery, which strips `standby.signal`, recovers **in primary mode on the node's old timeline**, and recreates `standby.signal`. The node can then never start as a standby of the current timeline:

```mermaid
sequenceDiagram
    participant PG as postmaster
    participant Mon as pooler monitor
    participant Ctl as pgctld
    Note over PG: crash / SIGKILL of a healthy streaming standby
    Note over PG: local WAL is a clean descendant of the leader timeline (no real divergence)
    loop every few seconds, forever
        Mon->>Ctl: Start(allow_crash_recovery)
        Ctl->>Ctl: remove standby.signal
        Ctl->>PG: postgres --single (primary-mode recovery)
        PG-->>Ctl: clean shutdown on OLD timeline, checkpoint now past the fork
        Ctl->>Ctl: recreate standby.signal
        Ctl->>PG: start as standby
        PG-->>Ctl: FATAL requested timeline N is not a child
        Ctl-->>Mon: error, postgres still down
    end
    Note over Mon,Ctl: pod keeps reporting Ready, multiorch loops fix-replication
```

The pod stays `Ready` throughout (readiness does not cover the postmaster), and multiorch spins its fix-replication action about once per second because it can't reach the dead postgres to verify replication status.

### Where does the divergence come from?

It is generated by single-user recovery itself — nothing foreign is received. `postgres --single` (with `standby.signal` removed) does plain crash recovery on the node's **own** timeline and does not follow the timeline-history switch. It replays the old timeline to its end and then, on its clean exit, writes a brand-new `CHECKPOINT_SHUTDOWN` onto that old timeline — past the point where the leader's timeline forked. `pg_waldump` of a wedged node makes it concrete. Past the fork (`0/5000000`) the old timeline TL1 carries exactly two records, both written by the `--single` run:

```
0/05000028  CHECKPOINT_SHUTDOWN  tli 1      # the real end of TL1, at the fork
0/050000A0  SWITCH                          # written by --single
0/06000028  CHECKPOINT_SHUTDOWN  tli 1      # written by --single; pg_control now points here
```

while the timeline the standby should follow forks cleanly at the same LSN and continues with the real data:

```
0/05000028  END_OF_RECOVERY   tli 2; prev tli 1   # TL2 branches here
0/050000B8  CHECKPOINT_ONLINE tli 2 ...            # real WAL continues on TL2
```

So `pg_control`'s latest checkpoint ends up at `0/6000028` on TL1, beyond where TL2 forked (`0/5000000`). A later standby start asked to follow TL2 finds the node's own history already extends past the TL2 fork point, hence "requested timeline 2 is not a child of this server's history." Without the `--single` step, standby-mode recovery replays TL1 only up to `0/5000000`, reads the TL2 history, switches, and follows TL2 — the node never writes anything onto TL1, so there is no divergence.

### Why not just pg_rewind it?

Because **there is nothing to rewind** — the divergence is self-inflicted. Before the crash the standby held only WAL from the leader's timeline; standby-mode crash recovery would replay it, read the timeline-history file, follow the switch, and resume streaming in a couple of seconds. The `--single` step is what *creates* the divergence (primary-mode replay advances the checkpoint past the fork on the abandoned timeline), so "just rewind it" would only repair damage pgctld itself caused. The correct fix is to not cause it.

`pg_rewind` is the right tool for a node that genuinely diverged — a former primary with un-replicated WAL past the fork — and that path is preserved (see below); it is also strictly heavier (it needs the leader reachable and rewind-ready and adds downtime), so using it to undo self-inflicted damage on a clean follower is the wrong trade even where it would work.

## Fix (after)

`StartRequest` gains a `suspected_divergence` signal. pgctld runs single-user recovery for a standby only when the caller sets it — i.e. a node that may genuinely have diverged (a former primary being demoted, or one already flagged for rewind), for which single-user recovery is the correct way to reach the clean-shutdown state `pg_rewind` needs. A clean follower is left to the postmaster's own standby-mode crash recovery, which follows the timeline switch and resumes streaming — no rewind required. The postgres monitor passes `SuspectedDivergence` from consensus.

```mermaid
sequenceDiagram
    participant PG as postmaster
    participant Mon as pooler monitor
    participant Ctl as pgctld
    Note over PG: crash / SIGKILL of a healthy streaming standby
    Mon->>Ctl: Start(allow_crash_recovery, suspected_divergence=false)
    Note over Ctl: clean follower, skip single-user recovery
    Ctl->>PG: start as standby (standby.signal kept)
    PG->>PG: standby-mode crash recovery, follows the timeline switch
    PG-->>Ctl: reached consistency, streaming from leader (no rewind needed)
    Ctl-->>Mon: started
    Note over Mon: postgres Ready within seconds
```

A genuinely-diverged node (`suspected_divergence=true`, set by the demote/rewind paths) still takes the `--single` route to reach a clean shutdown and is then rewound against the leader — unchanged by this PR.

## Testing

- New end-to-end regression test `TestStandbyRecoversAfterPostmasterKill` (`go/test/endtoend/multiorch`): brings a standby to a provably non-diverged, caught-up, streaming state, SIGKILLs its postmaster, and asserts the monitor auto-restarts it. It fails on the pre-fix code (standby stays down for the full window) and passes after the fix (recovers in a few seconds). Restart latency is recorded in the e2e timing summary.
- Its primary analog, `TestPostgresMonitorControl`, already covers the primary case.
- `TestPgCtldServiceStart_AsPrimaryCrashRecoveryMatrix` is extended with a `suspected_divergence` dimension: a clean follower with `allow_crash_recovery` no longer runs single-user recovery, while a suspected-diverged standby still does.
- No regression: `TestDemoteStalePrimary_SIGKILL` / `_GracefulShutdown` confirm the single-user + `pg_rewind` path for a genuinely-diverged former primary still works; `TestPgRewind_AfterCrash*` and the manager/pgctld unit suites are green.

## Behavior on a genuinely-diverged (unflagged) standby

A SIGKILLed standby that is *not* a clean follower — it durably flushed WAL past the fork on the old timeline (e.g. a phantom transaction from a sync-replication ack the old primary later lost) and consensus did not flag it — is handled **fail-safe**, not silently. Standby-mode recovery replays the local WAL and then postgres's own timeline check refuses to attach to the leader's timeline (`requested timeline N is not a child`), so the node stays **down**: it never comes up serving divergent data, never promotes, never corrupts. It simply does not *auto-recover* in this PR (the monitor retries the start). This is not a regression — the old unconditional `--single` also wedged this case (it clean-shut the node past the fork, so the standby start still FATAL-looped). Auto-healing it is the follow-up below.

Note the recovery contract is unchanged for the common torn-tail case: a partial final WAL record from the crash is discarded by ordinary crash recovery (records are length- and CRC-checked), which is not divergence.

## Follow-ups (not in this PR)

- **Escalate genuine divergence to pg_rewind**: when a clean-follower standby start FATALs with a timeline `not a child` error, treat it as suspected divergence and route through `--single` + `pg_rewind` — so the rare unflagged-diverged case (above) self-heals instead of retrying. The standby-mode start is a reliable, non-destructive detector for this.
- **Readiness gap**: a postgres-dead pod still reports Ready. Wiring postmaster liveness into multipooler readiness so the pod goes NotReady is a separate change.
- **multiorch escalation**: the fix-replication action retries without escalating. With this fix the pooler self-heals in the common case, so the loop clears on its own, but bounding/escalating the retry deserves its own follow-up.

### DIFF
--- a/go/cmd/pgctld/command/server.go
+++ b/go/cmd/pgctld/command/server.go
@@ -622,31 +622,39 @@ func (s *PgCtldService) Start(ctx context.Context, req *pb.StartRequest) (*pb.St
 		}
 	}
 
-	// When the caller allows it, make sure a node that was not cleanly shut down
-	// is crash recovered before the start below. crashRecoveryRan reports whether
-	// recovery occurs (here, or via the postmaster on the normal start) so the
-	// caller can treat it as evidence the node was not cleanly shut down.
+	// When the caller allows it, force single-user crash recovery for a standby
+	// the caller flagged as possibly diverged, so it reaches the clean-shutdown
+	// state pg_rewind needs before the start below. crashRecoveryRan reports
+	// whether that single-user recovery ran (matching
+	// StartResponse.crash_recovery_ran); a clean follower, or an as_primary start,
+	// is instead crash-recovered by the postmaster on the normal start below and
+	// does not set it.
 	var crashRecoveryRan bool
 	if req.GetAllowCrashRecovery() {
 		needed, nErr := s.crashRecoveryNeeded(ctx)
 		if nErr != nil {
 			s.logger.WarnContext(ctx, "could not determine clean-shutdown state before start (continuing)", "error", nErr)
-		} else if needed {
-			// crash recovery happens either way, so report it. A node without a
-			// standby.signal is crash-recovered by the postmaster on the normal start
-			// below, so it needs no explicit step. A standby may not be: if an early
-			// pg_rewind stamped minRecoveryPoint onto the wrong timeline, standby
-			// startup FATAL-loops and never reaches a clean state. Force single-user
-			// recovery for it — runCrashRecovery removes standby.signal first, since
-			// postgres --single refuses to run with it (that incompatibility, not
-			// standby.signal blocking the postmaster's own recovery, is why the
-			// explicit step is gated on standby.signal).
+		} else if needed && s.hasStandbySignal() && req.GetSuspectedDivergence() {
+			// A not-cleanly-stopped standby that the caller suspects may have
+			// diverged (a former primary being demoted, or a node already flagged
+			// for rewind) is force-recovered in single-user mode. runCrashRecovery
+			// removes standby.signal first — postgres --single refuses to run with
+			// it — and recreates it afterwards; this reaches the clean-shutdown
+			// state pg_rewind needs (e.g. to unwedge a node whose earlier pg_rewind
+			// stamped minRecoveryPoint onto the wrong timeline).
+			//
+			// A clean follower is deliberately NOT sent here: single-user
+			// recovery runs in primary mode and does not follow timeline-history
+			// switches, so it would finalize the node on its old timeline past the
+			// leader's fork and wedge the standby start ("requested timeline N is not
+			// a child"). Its crash recovery — and that of a node without
+			// standby.signal — is handled by the postmaster on the normal start
+			// below, which in standby mode follows the timeline switch. crashRecoveryRan
+			// therefore reports specifically whether single-user recovery ran.
 			crashRecoveryRan = true
-			if s.hasStandbySignal() {
-				if rcErr := s.runCrashRecovery(ctx); rcErr != nil {
-					// Best effort: the start below may still surface a clearer error.
-					s.logger.WarnContext(ctx, "standby crash recovery before start failed (continuing)", "error", rcErr)
-				}
+			if rcErr := s.runCrashRecovery(ctx); rcErr != nil {
+				// Best effort: the start below may still surface a clearer error.
+				s.logger.WarnContext(ctx, "standby crash recovery before start failed (continuing)", "error", rcErr)
 			}
 		}
 	}
--- a/go/cmd/pgctld/command/server_test.go
+++ b/go/cmd/pgctld/command/server_test.go
@@ -187,16 +187,28 @@ func TestPgCtldServiceStart_AsPrimaryCrashRecoveryMatrix(t *testing.T) {
 		name                 string
 		asPrimary            bool
 		allowCrashRecovery   bool
+		suspectedDivergence  bool
 		wantStandbySignal    bool
 		wantCrashRecoveryRan bool
 	}{
 		{
 			name:      "standby, no crash recovery",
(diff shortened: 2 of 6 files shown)


GuptaManan100 (approved):
Love it! Took me a while to understand, but I got it now 👍

PR #1358, opened 2026-08-06
feat(multipooler): report NotReady when postgres is not accepting

mkindahl:
Part of MUL-1009.

## Problem

The multipooler registered **zero** readiness checks, so its HTTP `/ready` always returned 200. A pod whose PostgreSQL postmaster is dead — a crashed or FATAL-looping standby — still reported Ready, hiding the down node from Kubernetes and operators.

## Change

Register a readiness check in the multipooler that dials its configured postgres leg (the Unix socket, or a TCP fallback matching how `init.go` selects the postgres address) and returns 503 when nothing is accepting. This mirrors pgctld's postmaster-aware `/ready` (`go/cmd/pgctld/command/ready.go`). A successful dial distinguishes a live postmaster from a dead one that left a stale socket file behind.

Also wires a `readinessProbe` onto the multipooler container in the demo StatefulSet so the pod actually flips to NotReady.

## Why this is safe for the PRIMARY case

Readiness gates only Kubernetes endpoint membership and rollout. Query routing and failover are driven by the health stream and topology, **not** the HTTP `/ready` probe (verified: nothing internal consumes the pooler's `/ready`). So a primary briefly flapping NotReady during a legitimate postgres restart does not trigger failover or eviction.

## Testing

- Unit: `TestPostgresAccepting` — live/missing/stale Unix socket, TCP fallback, wildcard host, closed port.
- E2E: `TestMultipoolerReadinessReflectsPostmaster` — baseline `/ready` 200 → SIGKILL the primary's postgres → `/ready` returns 503 → re-enable restarts → recovers to 200.

## Scope / follow-ups

- This makes the in-repo multipooler `/ready` honest and wires the **demo** probe. Production/staging pod readiness is defined by the operator, **outside this repo**; those pods won't flip to NotReady until the operator's multipooler `readinessProbe` points at `/ready` instead of `/live` (`/live` is hardcoded-OK). That operator change is tracked as the follow-up **MUL-240** — which is why this PR is **Part of MUL-1009** (non-closing) rather than a fix. MUL-1009 closes once this PR and MUL-240 are both deployed.
- Companion recovery-half PR #1357 (a clean follower whose postmaster is SIGKILLed auto-recovers via standby-mode crash recovery) is now merged.
- Out of scope: auto-escalating a genuinely-diverged standby to `pg_rewind` so it self-heals instead of FATAL-looping (separate, larger follow-up).
- The postgres socket-dial helper is deliberately self-contained (a small mirror of pgctld's), rather than a shared dependency; consolidating both into `go/tools/netutil` is a noted follow-up.

### DIFF
--- a/demo/k8s/k8s-multipooler-statefulset.yaml
+++ b/demo/k8s/k8s-multipooler-statefulset.yaml
@@ -141,6 +141,18 @@ spec:
               port: 16000
             initialDelaySeconds: 10
             periodSeconds: 10
+          # Readiness reflects whether this pod's PostgreSQL is accepting
+          # connections. A dead or FATAL-looping postmaster makes /ready return
+          # 503 so the pod goes NotReady and a wedged/down node is visible to
+          # operators (MUL-1009). This only affects Service endpoint membership
+          # and rollout gating; query routing and failover run off the health
+          # stream + topology, not this probe.
+          readinessProbe:
+            httpGet:
+              path: /ready
+              port: 16000
+            initialDelaySeconds: 5
+            periodSeconds: 5
           volumeMounts:
             # We use the subPathExpr to ensure that each pod
             # gets its own subdirectory.
--- a/go/services/multipooler/init.go
+++ b/go/services/multipooler/init.go
@@ -191,6 +191,7 @@ func NewMultipooler(telemetry *telemetry.Telemetry) *Multipooler {
 			Links: []Link{
 				{"Config", "Server configuration details", "/config"},
 				{"Live", "URL for liveness check", "/live"},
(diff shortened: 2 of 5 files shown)


mkindahl:
Closing this since if we do not report a pooler as ready when Postgres is down, there is a risk that a misconfiguration of the pod makes the pod entirely unreachable, which might mean that we cannot even reach it to repair it.

mkindahl:
Replacement: #1363 — gates `/ready` on the gRPC control plane being up (for both multipooler and pgctld) instead of on postgres, so a pooler whose postgres is down stays Ready and reachable for control RPCs / its health stream.

PR #1360, opened 2026-08-06
fix(query-serving): isolate temporary backend state

haritabh17:
## Summary

- Destroy reserved backends after temporary-object access so backend-local temp state and pending statistics cannot leak across logical sessions.
- Pin `current_schema()` calls that can instantiate `pg_temp`, and reconnect once when `temp_buffers` validation encounters a temp-contaminated backend.
- Normalize backend-sensitive `stats` and `sysviews` output while retaining only genuine pooling divergences in `stats.patch`.

## Problem

PostgreSQL temporary namespaces and some statistics remain tied to a physical backend after logical session cleanup. Reusing those backends caused the `temp`, `stats`, and `sysviews` compatibility tests to fail under transaction pooling.

## Solution

Treat temporary-object access as terminal for a pooled backend, route namespace-instantiating calls through reserved connections, retry safe `temp_buffers` validation on a fresh backend, and canonicalize only the catalog/statistics rows whose values legitimately depend on backend selection.

## Verification

- `go build ./go/...`
- Relevant multipooler and planner unit tests
- pgregress verifier normalizer unit tests


go/common/pgprotocol/server/startup.go
```diff
@@ -506,6 +507,25 @@ func (c *Conn) handleStartupMessage(protocolVersion uint32, reader *MessageReade
 	delete(c.params, "replication")
 	c.replicationMode = replicationMode
 
+	// A search_path supplied at connect time (directly or via options=-c ...)
```
shedges: Good !!

go/services/multipooler/internal/pools/reserved/reserved_conn.go
```diff
@@ -72,6 +72,15 @@ type Conn struct {
 	// released indicates whether this connection has been released.
 	released atomic.Bool
 
+	// closeOnRelease prevents temporary-object access from leaking to another
```
shedges: Very very Minor and no need to block this PR.

This variable not limited to Temp buffer feature. Currently temp buffer is the only thing using it but we may find new use-cases in future. We should keep the comment more general.

go/common/pgsettings/restricted.go
```diff
@@ -0,0 +1,63 @@
...
+var restrictedGUCs = map[string]string{
+	"synchronous_commit": "replication durability is managed by the cluster",
+}
+
+// RestrictedGUCError returns a feature_not_supported rejection if name is a
+// cluster-managed GUC, or nil otherwise. The message names the GUC, gives the
+// reason, and points the user at RESET, which can only restore the managed
+// value. Shared by the planner's statement/expression guards and the engine's
+// execute-time bound-name resolution so all surfaces give the same message.
+func RestrictedGUCError(name string) error {
```
shedges, line 54: We need to perform same check in startup path also? 

e.g. setting pass as connection param  `options=-c synchronous_commit=off`

shedges (approved):
Rest LGTM 🚀

go/services/multipooler/internal/pools/reserved/reserved_conn.go
haritabh17: Ack, done.

go/common/pgsettings/restricted.go
haritabh17: Ack, done.

PR #1362, opened 2026-08-07
feat(parser): resolve PL/pgSQL variables (T_DATUM) for PG parity

GuptaManan100:
## Summary

- Reverses the PL/pgSQL parser's founding "no variable resolution" decision: it now builds a compile-time namespace from the `DECLARE` section and resolves statement-level identifiers to `T_DATUM`, exactly as PostgreSQL's `plpgsql_yylex`.
- Brings the parser to exact accept/reject parity with PostgreSQL on every check that needs only a namespace (no catalog), and fixes the one case where we were *stricter* than PG.
- Types the datum family for the catalog-free cases (`RECORD`/`%ROWTYPE`, `rec.field`, comma-list rows); a variable declared with a named composite type still stays a scalar (no catalog).
- Consolidates the three working docs into the single `docs/query_serving/plpgsql_parser.md`.

## Description

**Mechanism** (ports of PG): a namespace stack + flat datum list (`namespace.go`, from `plpgsql_ns_*`); mode-gated resolution (`NORMAL`/`DECLARE`/`EXPR`) so only statement-level tokens resolve and embedded-SQL capture stays byte-identical; the `AT_STMT_START` guard (resolve a statement-leading word only when it's followed by `:=`/`=`/`[`, so a variable named like a statement keyword can still be an assignment target); and `tok_is_keyword` recovery (a variable named like an unreserved keyword doesn't hijack a keyword position such as a `FETCH` direction or `RAISE` option). The grammar reintroduces `stmt_assign` and the namespace push/pop lifecycle; the AST gains `PLpgSQL_rec` / `_row` / `_recfield`.

**Now rejected with PG's message** (previously accepted): assignment / `GET DIAGNOSTICS` / loop target that is `CONSTANT` (including a field of a constant record and a constant comma-list member); `EXIT`/`CONTINUE` outside a loop, a nonexistent label, or a block label under `CONTINUE`; a cursor `FOR` over an unbound cursor; an undeclared single target of a query/dynamic `FOR` or `FOREACH`; a `GET DIAGNOSTICS` record/row target. **Now accepted** (previously wrongly rejected — the one subset divergence): a variable named like an unreserved keyword used as an assignment target (`forward := forward * 2`). Also: `rec.field` targets resolve, `ALIAS` reuses its target's datum (so an alias of a `CONSTANT` isn't assignable), and the implicit `sqlstate`/`sqlerrm` handler `CONSTANT`s are created.

**Out of scope** (needs a live catalog): unknown-type rejection, `%TYPE`/collation resolution, and record typing for *named composite types*; plus the unrecognized-exception-condition table (a small independent follow-up). `PLpgSQL_expr.Parsed` stays `nil` — sub-parsing embedded SQL is a separate planner-side step.

Verified: full PL/pgSQL + AST test suites pass, the PostgreSQL regression corpus (240 + 280 bodies) round-trips, and the grammar is `%expect 0` (zero shift/reduce, zero reduce/reduce).


go/common/parser/plpgsql/read_construct.go
```diff
@@ -187,48 +229,220 @@ func (l *lexer) readDatatype() *plpgsqlast.PLpgSQL_type {
...
+	case v.row != nil:
+		l.checkAssignable(v.row)
+	case v.scalar != nil:
+		l.checkAssignable(v.scalar)
+	case v.compound:
+		// Unresolvable compound target: could be a field of a named-composite
+		// record we cannot resolve without a catalog. Accept (superset) rather than
+		// reject a body PG accepts, matching the compound comma-list member handling.
+	default:
+		l.Error("loop variable of loop over rows must be a record variable or list of scalar variables")
```
shedges: Is this error from postgres code? Or we can make it generic as it may misleading in below scenario.

Here we are erroring out if a query for loop target didn't resolve to a declared variables.  parser can only see variables declared in the DECLARE section and it can't see the routine's own arguments. We will reject function whose loop variables is function argument. 

`FUNCTION f(myrow record) ... BEGIN FOR myrow IN SELECT ... LOOP ... END`

go/common/parser/plpgsql/namespace.go
```diff
@@ -0,0 +1,360 @@
...
+// qualified match to a *scalar* variable is disregarded; likewise a non-empty
+// name2 disregards an unqualified match to a scalar variable. ("" is the NULL
+// sentinel — a real identifier is never empty.)
+func (ns *namespace) lookup(nsCur *nsItem, localmode bool, name1, name2, name3 string) (*nsItem, int) {
+	// Outer loop iterates once per block level in the namespace chain.
+	for nsCur != nil {
+		var item *nsItem
+
+		// Check this level for an unqualified match to a variable name.
+		for item = nsCur; item.itemType != nsTypeLabel; item = item.prev {
```
shedges: here we need check `item != nil` ?

GuptaManan100: Fixex

go/common/parser/plpgsql/read_construct.go
GuptaManan100: Fixed

shedges (approved):
LGTM

PR #1363, opened 2026-08-10
feat(multipooler,pgctld): base /ready on gRPC serving, not postgres

mkindahl:
Part of MUL-1009. Supersedes #1358.

## What

Make the multipooler and pgctld `/ready` endpoints report ready based **only on whether their own gRPC control plane is accepting connections** — deliberately decoupled from postgres.

- **multipooler** (`init.go` + new `ready.go`): registered no ready checks before, so `/ready` was always 200. Adds a gRPC-accepting readiness check (and a `/ready` status-page link).
- **pgctld** (`server.go`): already gated `/ready` on the postgres socket **and** gRPC. Narrows it — drops the postgres-socket check so `/ready` reflects gRPC only. No new check is added; the postgres coupling is removed.
- **demo** (`k8s-multipooler-statefulset.yaml`): wires the multipooler container's `readinessProbe` to `/ready`.

Liveness probes and the pgctld gRPC-accepting check are unchanged.

## Why not gate readiness on postgres (the change from #1358)

#1358 made the pooler's `/ready` return 503 when postgres was down. Gating pod readiness on postgres couples **pod reachability** to **postgres availability**: a NotReady pod is dropped from Service endpoints and — unless the headless Service sets `publishNotReadyAddresses: true` — loses its DNS record. Since multigres components reach the pooler by its topology-registered `Hostname:grpcPort` (resolved via that headless DNS), a routine "stop postgres" could make the pod unreachable for the very control RPCs and health stream used to observe and recover it. That's a latent, worst-timed coupling whose correctness depends on a separate Service flag staying set.

Gating on gRPC serving removes the coupling entirely:

- A pooler whose postgres is down stays **Ready** and reachable; its monitor restarts postgres.
- postgres liveness is already signalled **out-of-band via the health stream**, which is what drives routing/failover. `multiorch` has no Kubernetes client and never reads pod readiness, so readiness was never an input to failover — its only role is external k8s visibility + rollout gating, which "is the control plane serving?" expresses correctly.
- Correctness no longer depends on `publishNotReadyAddresses` (a NotReady pod now only ever means "gRPC isn't serving", i.e. it genuinely shouldn't take traffic).

Liveness remains the tool for "restart a wedged process"; readiness here answers "is my gRPC control plane up?".

## Testing

- Unit (`go/services/multipooler/ready_test.go`): `TestGrpcAccepting` / `TestTCPAccepting` / `TestUnixSocketAccepting`. pgctld's own ready-probe tests are unchanged.
- E2E (`go/test/endtoend/multipooler/readiness_test.go`): `TestMultipoolerReadinessReflectsGRPCNotPostgres` — isolated single-node shard; `/ready` is 200 while up and **stays 200 after postgres is SIGKILLed** (auto-restart disabled), proving readiness is decoupled from postgres.

## Follow-ups

- Operator (**MUL-240**): point the multipooler/pgctld `readinessProbe`s at `/ready`. With `/ready` now postgres-independent, that wiring no longer needs `publishNotReadyAddresses: true` for correctness of this probe.
- The pooler's gRPC-accepting helper is a deliberate small mirror of pgctld's; consolidating both into `go/tools/netutil` is a noted follow-up.

### DIFF
--- a/demo/k8s/k8s-multipooler-statefulset.yaml
+++ b/demo/k8s/k8s-multipooler-statefulset.yaml
@@ -141,6 +141,18 @@ spec:
               port: 16000
             initialDelaySeconds: 10
             periodSeconds: 10
+          # Readiness reflects only whether the multipooler's own gRPC control
+          # plane is serving — NOT postgres. A pooler whose postgres is down
+          # stays Ready so it keeps its Service endpoint/DNS and remains
+          # reachable for control RPCs and its health stream while its monitor
+          # brings postgres back. postgres liveness is signalled via the health
+          # stream, not this probe.
+          readinessProbe:
+            httpGet:
+              path: /ready
+              port: 16000
+            initialDelaySeconds: 5
+            periodSeconds: 5
           volumeMounts:
             # We use the subPathExpr to ensure that each pod
             # gets its own subdirectory.
--- a/go/cmd/pgctld/command/server.go
+++ b/go/cmd/pgctld/command/server.go
@@ -163,19 +163,12 @@ func (s *PgCtldServerCmd) runServer(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Register /ready probe: postgres socket accepting + gRPC accepting.
-	// Replication health is intentionally excluded (see pgctldReadyHandler).
-	pgSocketPath := filepath.Join(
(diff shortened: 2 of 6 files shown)


niconosenzo (approved):
lgtm

PR #1364, opened 2026-08-10
feat(multiadmin): expose backup start/stop timestamps and job_id

niconosenzo:
GetBackups dropped start_timestamp, stop_timestamp, and job_id when converting pooler BackupMetadata to BackupInfo, so backup_time was always zero over the wire.

Add start_timestamp/stop_timestamp as separate fields plus job_id, and deprecate backup_time in favor of the two explicit fields.


haritabh17 (approved):
🚀

PR #1365, opened 2026-08-10
feat(multipooler): quarantine pooler when postgres is unrecoverable

mkindahl:
## What & why

A multipooler whose PostgreSQL is *unrecoverably* failing to start — the classic case being a genuinely diverged standby that keeps FATAL-looping — is invisible to Kubernetes today. postgres is a child of pgctld, so the pgctld and multipooler containers stay `Running` and the pod stays `Ready`; the monitor just logs "failed to start Postgres, will retry" every 5s forever with no terminal state and no backoff on the start path. Nothing ever flags the node for replacement.

This PR adds **Layer 1** of the fix: the pooler detects that its postgres is unrecoverable, marks itself not-usable, and exposes a durable signal that an external actor can act on. It is deliberately orthogonal to the readiness probe (#1363 intentionally keeps `/ready` on the gRPC control plane, not postgres); this is the separate "needs replacement" signal.

## What it does

The postgres monitor now classifies "postgres continuously failing to start/rewind/restore for longer than a configurable timeout, across at least a small floor of genuine attempts" as an unrecoverable FATAL loop, and quarantines the pooler:

- Writes `LIFECYCLE_QUARANTINED` to its own topology record — the durable signal an external replacer watches (a pre-existing, purpose-built lifecycle value; see the design note below).
- Flips cohort eligibility to `INELIGIBLE` so the coordinator drops it from consensus.
- Disables further restart attempts so the node stops FATAL-looping.

The verdict needs no new proto or health-stream surface. The topology `lifecycle_status` is the single signal: the operator (Layer 2) reads it, and multiorch already carries the topology `Multipooler` record — including `lifecycle_status` — in its pooler-watch cache, so it needs no new plumbing to observe it. multiorch's behavioural response (dropping the node from consensus) comes from the `INELIGIBLE` cohort-eligibility signal above, which it already acts on. (multiorch only acts on the `SHUTDOWN` lifecycle value today; it deliberately does not tear down a `QUARANTINED` pooler — the node is kept for replacement, not tombstoned.)

## Design notes

**Reuses a state the codebase was already designed for — this PR only supplies the missing producer.** `LIFECYCLE_QUARANTINED` is not new. It has existed in the proto with a detailed doc comment describing exactly this scenario — "it cannot automatically recover to a functioning state (e.g. it could not complete a pg_rewind, could not restore from backup to start postgres…) … Only the pooler sets this on itself … the operator/provisioner should treat a QUARANTINED pooler as absent … and provision a replacement" — and the operator **already** reads it (`isLifecycleQuarantined`, used to drop such poolers from the healthy set and from primary-candidate selection). Until now, though, *nothing ever set it*: the value was defined and consumed but had **no producer**. This PR is that producer. That the enum, its precise semantics, its pooler-owns-it rule, and its operator-side consumer were all put in place up front is a strong signal that a pooler-owned, lifecycle-based signal is the sanctioned shape of this fix — we are completing an intended design, not bolting on a new mechanism.

**Ownership — only the pooler writes its own topology record. Why not the orchestrator?** A fair question is whether the orchestrator should set `LIFECYCLE_QUARANTINED` instead. It can't, durably: the pooler continuously republishes its own topology record, so any lifecycle value another actor writes is overwritten on the next publish — the pooler is the sole writer by construction. This isn't specific to this PR; `fix_replication` already carries a TODO to "signal this durably via the pooler's lifecycle stage … Orch must not write the pooler's topology Type — the pooler owns its own record and would clobber an external write."

Beyond that hard constraint, the pooler is also the *right* owner on the merits: it has first-hand ground truth (the actual failed start/rewind/restore attempts) where the orchestrator can only infer liveness from a distance, and by ruleguard the orchestrator may not even read role/serving fields for identity. This matches the codebase's direction of travel — orchestrator-driven RPCs like `RewindToSource` and `DemoteStalePrimary` were removed in favor of pooler-autonomous self-heal.

The orchestrator does retain a role for the one case the pooler can't judge alone — "this node has fallen irrecoverably behind while its peers replicate fine" — but that is a future orchestrator-*initiated* request that asks the pooler to quarantine itself; the pooler stays the writer.

**Time-based gating with an attempts floor.** The gate is elapsed time since the first failure (`--postgres-unrecoverable-timeout`, e.g. `5m`) plus a minimum-attempts floor (`--postgres-unrecoverable-min-attempts`, default 3, validated to be in `[2, 10)`), rather than a raw attempt count. This is independent of the monitor interval and, importantly, robust to the fact that a fast start-FATAL (seconds) and a slow restore failure (minutes) would otherwise map to wildly different wall-clock budgets under a pure count. The timeout is the real gate; the floor just guards the degenerate cases (a single hung attempt, sparse ticks) — hence the small upper bound.

**Defaults to disabled.** Quarantining disables restarts and marks the node ineligible, so it is only safe once an actor exists to replace quarantined nodes. `--postgres-unrecoverable-timeout` defaults to `0` (disabled); the operator's manifests opt in when Layer 2 ships.

## Follow-up (Layer 2, separate PR)

The operator (multigres/multigres-operator) will watch `LIFECYCLE_QUARANTINED` and drive remediation — delete the pod, wipe its data PVC, and let the node re-bootstrap from backup — which is the only path that actually heals genuine on-disk divergence (a same-PVC restart would FATAL-loop identically). It lives in the operator repo, so it is a separate PR; it needs no proto change from this one — the operator already reads the `QUARANTINED` lifecycle value.

## Testing

- Unit: `quarantine_test.go` — 7 tests with an injectable clock covering both gates independently (timeout holds despite many attempts; attempts floor holds despite elapsed timeout; reset re-anchors the window on recovery; disabled; only genuine failed recovery actions count; idempotent latch).
- E2E (real postgres): `multiorch/pooler_quarantine_test.go` — brings up a 3-pooler shard with multiorch on an isolated fixture, corrupts a standby's `pg_control` so postgres persistently FATALs on start, and asserts the standby self-quarantines in topology.
- Full multipooler package suite green; `golangci-lint` clean. No proto or generated-file changes.

## Related PRs

- **Layer 2 (operator):** multigres/multigres-operator#568 — watches the `LIFECYCLE_QUARANTINED` signal produced here and drives remediation (delete pod + wipe data PVC + re-bootstrap from backup).

Fixes MUL-644. Part of MUL-1009.

### DIFF
--- a/go/services/multipooler/internal/manager/postgres_monitor.go
+++ b/go/services/multipooler/internal/manager/postgres_monitor.go
@@ -277,6 +277,10 @@ func (pm *MultipoolerManager) monitorPostgresIteration(ctx context.Context) (pos
 			if !pm.consensusMgr.SuspectedDivergence() {
 				pm.consensusMgr.ResetRewindBackoff()
 			}
+			// Postgres came up: the FATAL-loop streak (if any) is broken. This is
+			// the common healthy path, which returns before reaching
+			// trackRecoveryOutcome, so reset the classifier here too.
+			pm.resetUnrecoverableTracking()
 		}
 		return currentState, nil
 	}
@@ -305,7 +309,13 @@ func (pm *MultipoolerManager) monitorPostgresIteration(ctx context.Context) (pos
 	action = pm.determineRemedialAction(lockCtx, currentState)
 
 	// Take remedial action with lock held
-	pm.takeRemedialAction(lockCtx, action, currentState)
+	actionErr := pm.takeRemedialAction(lockCtx, action, currentState)
+
+	// Feed the unrecoverable-FATAL-loop classifier: a postgres that keeps
+	// failing to start (or rewind/restore) for long enough is quarantined so it
+	// gets replaced instead of spinning forever. Runs under the same action lock
+	// so the quarantine record write is serialised with the rest of the tick.
+	pm.trackRecoveryOutcome(lockCtx, action, currentState, actionErr)
 
 	return currentState, nil
 }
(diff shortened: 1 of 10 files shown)


go/services/multipooler/internal/manager/quarantine.go
```diff
@@ -0,0 +1,86 @@
...
+//
+// The verdict does not auto-clear within a process lifetime — recovery happens
+// by replacing the pod, which starts a fresh manager. The caller must hold the
+// action lock (record.Mutate asserts this); the monitor calls it from within
+// monitorPostgresIteration, which already holds the lock.
+func (pm *MultipoolerManager) markPoolerQuarantinedLocked(ctx context.Context, reason string) {
+	// Idempotent: skip the publish + side effects if already quarantined.
+	if pm.record.Snapshot().GetLifecycleStatus().GetStatus() == clustermetadatapb.PoolerLifecycleStatus_LIFECYCLE_QUARANTINED {
+		return
+	}
```
GuptaManan100: I'm not so sure about this check. What if pm.record succeeds, but the SetCohortEligibility call fails. Wouldn't you want to retry that in case of a duplicate call?

go/services/multipooler/internal/manager/postgres_monitor.go
```diff
@@ -448,6 +458,105 @@ func (pm *MultipoolerManager) matchesRecorded(target *clustermetadatapb.PoolerAd
 	return expectedPassfile == "" || ci.GetPassfile() == expectedPassfile
 }
 
+// defaultUnrecoverableMinAttempts is the floor of genuine failed recovery
+// attempts required before a pooler may quarantine itself, regardless of
+// elapsed time. It guards against quarantining on too few real attempts — e.g.
+// a single attempt that hangs for the whole timeout, or sparse monitor ticks.
+// Overridable via --postgres-unrecoverable-min-attempts; used when the manager's
+// unrecoverableMinAttempts is unset (0), e.g. in tests.
+const defaultUnrecoverableMinAttempts = 3
```
GuptaManan100: Maybe move this to the constants package. Just a cleaner place to keep all constants in case they are ever reused across packages.

GuptaManan100 (approved):
Rest LGTM

go/services/multipooler/internal/manager/postgres_monitor.go
mkindahl: Agree. Done.

go/services/multipooler/internal/manager/quarantine.go
mkindahl: Hmmm... good point.

mkindahl: Ok, I dealt with it by running each step if it is not already in the right state. I broadcast the health only if I had to change some field since there is a regular broadcast anyway and nothing new happened here.

go/services/multipooler/internal/manager/quarantine.go
```diff
@@ -0,0 +1,115 @@
+// Copyright 2026 Supabase, Inc.
```
dweitzman, line 1: Quarantining feels close enough to HA/consensus that probably this file and the related state in multipooler manager should all move under the consensus/ subpackage (which should maybe be renamed to ha/ instead of consensus/)

internal/manager and MultipoolerManager tend to grow without bounds into disarray if we don't aggressively split out feature-specific logic into feature-specific sub-managers and sub-packages

PR #1367, opened 2026-08-11
test(queryserving): add PostgREST spec suite through the multigateway

GuptaManan100:
## Summary

- Adds an end-to-end suite (`go/test/endtoend/queryserving/postgresttests`) that runs PostgREST's upstream hspec suite (1303 examples) with PostgREST pointed at the multigateway — client → multigateway → multipooler → postgres — using a real libpq client from a pinned Docker image.
- The direct-PostgreSQL baseline is an asserted invariant (every non-skipped spec passes on plain postgres), so the default run is **gateway-only** and any gateway failure is a divergence. `POSTGREST_FULL_BASELINE=1` re-runs the baseline and classifies, to re-verify the invariant after bumping the PostgREST tag or editing fixtures.
- Runs in CI behind the `Run Extended Query Serving Tests` / `Run all Query Serving Tests` labels; the divergence count lands in the job summary via `report.go`. Expected red until the tracked divergences in `DIVERGENCES.md` are closed.

## Status

49 gateway divergences, 0 environment failures (baseline fully green). ~42 of the divergences share one root cause: PostgREST's mutation SQL calls `set_config('pgrst.inserted', …, true)` inside an INSERT `WHERE` (wrapped in a CTE), and the gateway's unsafe-funccall analyzer only allows `set_config` as a top-level `SELECT` target — so it rejects the transaction-scoped (`is_local=true`) call postgres accepts anywhere. Closing that one closes the majority; repro in `DIVERGENCES.md`.

## Harness notes

- Provides the DB (a multigres cluster), loads PostgREST's fixtures directly, and normalizes environment differences that would otherwise masquerade as gateway bugs: timezone + planner GUCs via `ALTER DATABASE`, and the RangeSpec `ANALYZE` hook run in the loader instead of the container. Skips specs needing extensions we don't build (`pg-safeupdate`).
- The hspec container reaches the host gateway/postgres via `--network=host` + `127.0.0.1` (Linux CI and OrbStack). The spec image (`testdata/Dockerfile.spec`) is a non-root one-shot build/test image that passes Checkov's Docker policies.

### DIFF
--- a/.github/workflows/test-postgrest.yml
+++ b/.github/workflows/test-postgrest.yml
@@ -1,0 +1,119 @@
+# Copyright 2026 Supabase, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: PostgREST Spec Suite
+
+# Runs PostgREST's upstream hspec suite with PostgREST pointed at the
+# multigateway, classifying each failure as a gateway divergence (fails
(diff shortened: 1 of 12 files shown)


haritabh17 (approved):
LGTM 🚀

shedges (approved):
LGTM

PR #1369, opened 2026-08-12
feat(multiadmin): add JWT authentication

shedges:
## Summary

Multiadmin's HTTP, Connect, REST, proxy, and pprof surfaces currently accept requests without authenticating the caller. Envoy's edge JWT policy only protects traffic that enters through that route, so in-cluster callers can bypass it by reaching Multiadmin directly.

This PR adds JWT verification to Multiadmin, disabled by default.

## What changes

- Adds `--enable-auth` to require JWT bearer tokens on Multiadmin's HTTP, Connect, REST, proxy, `/config`, and pprof endpoints. Native gRPC is deliberately unaffected by this flag.
- Adds a reusable `jwt` servenv authenticator with issuer, JWKS URI, optional audience, and repeatable subject allow-list configuration.
- Verifies signatures using cached JWKS keys, restricts accepted signing algorithms, requires token expiration, and validates issuer plus any configured audience/subjects.
- Retries the initial JWKS fetch with bounded exponential backoff and fails startup if keys cannot be loaded.
- Uses the same generic authentication-failure response across HTTP, Connect, REST, and gRPC while logging the specific rejection reason server-side.
- Resolves authentication plugins safely during startup and fails closed while JWT initialization is pending.
- Exports `mg.servenv.auth.attempts`, tagged by plugin and outcome.
- Leaves `/live`, `/ready`, and `/version` unauthenticated. Services that do not opt into HTTP authentication retain their existing `/config` and pprof behavior.

The generic `--grpc-auth-mode=jwt` integration remains available for services that explicitly want JWT authentication on native gRPC. Multiadmin's `--enable-auth` path selects JWT only for its HTTP-based surfaces, so an independently configured gRPC auth mode cannot change HTTP behavior while `--enable-auth` is disabled.

## Configuration

Enabling Multiadmin HTTP authentication requires:

```text
--enable-auth
--grpc-auth-jwt-issuer=<issuer>
--grpc-auth-jwt-jwks-uri=<jwks-uri>
```

Optional constraints:

```text
--grpc-auth-jwt-audience=<audience>
--grpc-auth-jwt-allowed-subs=<subject>  # repeat for multiple subjects
```

Deployment wiring in `multigres-operator` and token acquisition for the CLI/web UI remain separate follow-up work. Authentication stays off unless explicitly enabled.

## Validation

- `go test -race ./go/common/servenv ./go/services/multiadmin ./go/observability/metriccatalog`
- Repository lint hook for `go/common/servenv`
- Generated metric catalog refreshed after merging current `main`


haritabh17 (approved):
Should we gate `/config` and `/version` as well? Rest all looks good.

shedges:
> Should we gate `/config` and `/version` as well? Rest all looks good.

Added auth support for /config also. /version kept open as it only provides version info and low risk. It will be easy for other services to retrieve version without token.

PR #920, opened 2026-04-28
chore(github): assign proto file ownership by service domain

mkindahl:

PR #1370, opened 2026-08-12
fix(github): restore CODEOWNERS catch-all

alcimerio:
## Summary

I created #1368 and that didn't fall inside any of the rules, so it has no reviewers. I noticed the default rule was introduced in #682 and removed in #920. This PR restores that - if that was intentional, feel free to close this.


mkindahl (approved):
It was intentional because everybody got a lot of reviews, but we should not miss reviews to approving. We can look at something less noisy later.

alcimerio:
@mkindahl Makes sense! What is the process to get it merged?

PR #1371, opened 2026-08-12
fix(multipooler): write primary_conninfo to auto.conf before standby start

haritabh17:
## What

Write `primary_conninfo` directly into `postgresql.auto.conf` (while postgres is stopped) in `restartAsStandbyLocked`, before the standby restart — instead of relying only on the post-start `ALTER SYSTEM` write.

## Why

`primary_conninfo` SET was the last config write in multigres that happened only over live SQL. A standby that cannot reach consistency can never accept that SQL, and without `primary_conninfo` it can never stream the WAL it needs to reach consistency — a circular deadlock:

1. On a suspected-divergence start, the monitor deliberately drops `primary_conninfo` from auto.conf (held start) so the node won't stream from a leader it hasn't been rewound to. That part is correct and unchanged.
2. The comment there promises the rewind path re-establishes conninfo afterwards.
3. But re-establishment ran via `ALTER SYSTEM` **after** postgres restarts. If the restarted node needs leader WAL before it can accept connections, the SQL step never runs → held blind forever.

Not theoretical: on 2026-08-12 a k3s cluster (image `sha-7edf67a`, shard `postgres/default/0-inf`) wedged exactly this way for 25+ minutes until the PVC was wiped. A gracefully demoted primary diverged ~16MB past the branch point (recruit LSN `8/B20000A0` vs final WAL `8/B3000040`), its pg_rewind self-heal did not complete, and the node sat at "the database system is starting up" / "waiting for WAL to become available at 8/B3000040" while multiorch looped `ReplicaNotReplicating` and elections degraded to 64s on action-lock timeouts (`promote_ms 64459`).

Precedent for file-based recovery config: Patroni writes recovery params including `primary_conninfo` into generated config before starting postgres; CloudNativePG's instance manager owns `primary_conninfo`/`restore_command` in operator-managed files by role; core's `pg_basebackup -R` writes `primary_conninfo` into `postgresql.auto.conf` while the server is down.

## How

- **`setAutoConfSetting(ctx, name, value)`** — SET variant on the existing `editAutoConf` primitive: replaces an existing entry in place, drops duplicates (postgres reads the last occurrence, so a stale duplicate below the fresh entry would win), appends when absent, and is an idempotent no-op when the entry already matches. Token-boundary name matching (so `primary_conninfo` doesn't touch `primary_conninfo_extra`) is shared with `dropAutoConfSettings` via an extracted `autoConfLineSets`.
- **`ast.QuoteConfValue`** — shared GUC-config-file quoting: single quotes AND backslashes doubled (the config-file lexer processes backslash escapes inside quoted values), matching ALTER SYSTEM's own `write_auto_conf_file`. Lives next to `QuoteStringLiteral` with a doc contrast, since that SQL-literal quoter is the natural wrong reach (its `E'...'` form is SQL-lexer syntax a config file would read as a literal `E`). Also adopted by pgctld's `include_if_exists` writer, which previously had its own quote-only-doubling copy with the same latent backslash bug.
- **Single source of truth**: now that #1354 is merged, the file write assembles its value via `expectedPrimaryConnInfoAt` and renders it via `buildPrimaryConnInfo` — the same assembler/renderer the SQL path and the drift comparator use — so the file path cannot drift from either, and it emits the same missing-passfile warning as `setPrimaryConnInfoLocked`.
- **`restartAsStandbyLocked`**: after `runPgRewind` and the post-rewind auto.conf surgery (`fixPgBackRestPaths` / `dropRestoreCommandFromAutoConf` — pg_rewind copies the source's auto.conf, which is a primary's and has no conninfo), and **before** the `Restart` call, write the source's conninfo into auto.conf. Unconditional across the rewind and no-rewind branches, matching the helper's documented contract ("postgres is a working standby of source on return"). Best-effort like the neighboring surgery: on failure the post-start SQL write still covers nodes that can reach consistency.
- **Admin-pause honored**: the file write is skipped when `walReceiverManuallyStopped` is set, so it cannot bypass the StopReplication pause that `setPrimaryConnInfoLocked` enforces as the single conninfo-writer gate; the SQL path then refuses loudly, preserving existing behavior.
- The SQL path (`setPrimaryConnInfoLocked`) is kept for re-pointing a running standby (PG13+ reloads `primary_conninfo` on SIGHUP) and as post-start verification; the held-start drop in the monitor is unchanged. `suspectedDivergence` clearing semantics are untouched. No cache to keep coherent: the drift check (`primaryConnInfoDiffersFromRecorded`, #1332) reads the live GUC, which reflects auto.conf once postgres starts.
- `MockPgCtldService` gains a `RestartFunc` hook (mirroring `StatusFunc`) so tests can observe auto.conf at the moment postgres would restart.

## Tests

- `TestSetAutoConfSetting`: append when absent (incl. empty file), replace in place, idempotence proven by a no-write against a read-only file, token boundaries (`primary_conninfo_extra` untouched), embedded-single-quote doubling, duplicate collapse, missing-file error.
- `TestRestartAsStandby_RestoresPrimaryConnInfo`: now seeds a rewind-style auto.conf (no conninfo) and asserts — via the `RestartFunc` snapshot — that the source's conninfo is in the file **before** the restart call. This is the regression the change exists to prevent: a rewound standby whose postgres cannot accept SQL still ends up with streaming configured.
- `TestRestartAsStandby_NoDivergence_StillSetsPrimaryConnInfo`: seeds a stale conninfo from a previous run and asserts it is replaced in place (not shadowed) before restart.
- Validated: `go build ./go/...`, `go test -short ./go/services/multipooler/... ./go/cmd/pgctld/...`, golangci-lint 0 issues.

## Related

- Merged: #1303 (self-heal diverged standbys locally — created this flow), #1187 (wait for rewind-ready), #1337 (rewind-readiness deadlock), #1190 (retry pg_rewind), #1287 (start as standby by default), #1332 (conninfo drift check).
- Open: #1365 (quarantine unrecoverable postgres — this change turns many would-be quarantines into self-heals), #1160 (SwitchPrimary), #1268 (recruit LSN drift), #835 (always start in recovery mode).
- **Rebased on #1354** (single source of truth for primary_conninfo, now merged): this PR no longer extracts its own conninfo builder — the pre-start file write goes through #1354's `expectedPrimaryConnInfoAt`/`buildPrimaryConnInfo`, extending its single-source-of-truth guarantee to the file path.


go/common/parser/ast/sql_utils.go
```diff
@@ -179,6 +179,22 @@ func QuoteStringLiteral(value string) string {
...
+// backslash would be reinterpreted; doubling matches ALTER SYSTEM's own writer
+// (write_auto_conf_file in guc.c).
+//
+// Not interchangeable with QuoteStringLiteral: that one produces SQL-lexer
+// syntax, whose escape-string form (E'...') a config file would read as a
+// literal E followed by a quoted string.
+func QuoteConfValue(value string) string {
+	escaped := strings.ReplaceAll(value, `\`, `\\`)
+	escaped = strings.ReplaceAll(escaped, `'`, `''`)
+	return `'` + escaped + `'`
```
mkindahl: You probably want to have a check mirroring the check in Postgres that the string does not contain any newlines. It would be a bug in our source, yes, but if not caught it would result in a configuration file that generates errors when read. It might be a good idea to not accept things like vertical space, line breaks, etc., even though Postgres will not raise alarms about it.

go/common/parser/ast/sql_utils_test.go
```diff
@@ -0,0 +1,51 @@
...
+	tests := []struct {
+		name  string
+		value string
+		want  string
+	}{
+		{"plain", "host=h port=5432", "'host=h port=5432'"},
+		{"empty", "", "''"},
+		{"single_quote_doubled", "pa'ss", "'pa''ss'"},
+		{"backslash_doubled", `a\tb`, `'a\\tb'`},
+		{"backslash_and_quote", `a\'b`, `'a\\''b'`},
```
mkindahl, line 39: If you add the check for newline above, this test needs to be extended with some negative tests for things that should not be accepted.

go/services/multipooler/internal/manager/rpc_manager.go
```diff
@@ -1158,6 +1199,50 @@ func (pm *MultipoolerManager) dropAutoConfSettings(ctx context.Context, names ..
...
+// setAutoConfSetting writes a "name = 'value'" entry into postgresql.auto.conf,
+// replacing an existing entry for name in place (dropping any duplicates —
+// postgres reads the last occurrence, so stray duplicates must not survive a
+// set) or appending one when absent. Idempotent: no write when the entry
+// already matches. Like dropAutoConfSettings it edits the file directly, so it
+// is safe while postgres is stopped — which is its whole reason to exist:
+// ALTER SYSTEM needs a postgres that can accept connections, and the paths
+// that need this helper run exactly when postgres can't.
+func (pm *MultipoolerManager) setAutoConfSetting(ctx context.Context, name, value string) error {
+	entry := name + " = " + ast.QuoteConfValue(value)
```
mkindahl: Might be good to check that the name is a valid name. Even though it would indicate an error in our end, it would create a corrupt configuration file.

go/services/multipooler/internal/manager/rpc_manager.go
```diff
@@ -1158,6 +1199,50 @@ func (pm *MultipoolerManager) dropAutoConfSettings(ctx context.Context, names ..
...
+			}
+			if replaced {
+				// Duplicate entry: drop it (last-occurrence-wins would otherwise
+				// override the entry written above).
+				changed = true
+				continue
+			}
+			replaced = true
+			out = append(out, entry)
+			if strings.TrimSpace(line) != entry {
```
mkindahl: Nit: not really needed (entry will not have trailing spaces) but just a good measure.
```suggestion
			if strings.TrimSpace(line) != strings.TrimSpace(entry) {
```

mkindahl (approved):
Two minor things for hardening, but it would work with the current files so approving.

go/common/parser/ast/sql_utils.go
haritabh17: Done — QuoteConfValue now returns an error for values containing \n or \r, mirroring ALTER SYSTEM

go/common/parser/ast/sql_utils_test.go
haritabh17: Added — rejects_line_breaks cases in TestQuoteConfValue

go/services/multipooler/internal/manager/rpc_manager.go
haritabh17: Done — setAutoConfSetting now validates the name against `^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)?$`

go/services/multipooler/internal/manager/rpc_manager.go
haritabh17: Ack. Applied as suggested.

PR #1372, opened 2026-08-12
refactor(query-serving): remove ReasonSetConfig capture reservation

GuptaManan100:
## Summary

- Removes the `ReasonSetConfig` per-statement capture reservation and its entire lifetime machinery (pooler-side unwind, the post-tracking release RPC, the `PersistingSetConfig` plan flag, `ReleaseSetConfigReservations`).
- An unpinned session-persisting `set_config(..., false)` now mirrors an unpinned `SET`: it is rewritten to `is_local := true` so it reverts on the pooled backend, and the value lives only in the gateway map, replayed at the next checkout. A pinned session (or a statement that reserves its own backend) persists it for real on the reserved backend.
- Behaviour-preserving: same-session still observes the value, a fresh client does not, and gateway-managed GUCs never persist on a pooled backend.

## Description

The pinned/unpinned decision is live session state, but `SELECT set_config` plans are cacheable, so it can't be baked into the plan. A new `SessionStateBranch` engine primitive carries both routes (verbatim-persist and reverted) and picks per live session state at execute time, keeping plans cacheable. The dynamic `pg_settings` shape flips `is_local` the same way inside `ResolveTrackSetConfig`. SQL `EXECUTE` is non-cacheable, so `planExecuteStmt` rewrites the prepared body's `is_local` for an unpinned execution and runs it verbatim otherwise. The reservation-reason proto enum value is deleted outright — we're in alpha with no wire-compatibility guarantee, so there's no need to reserve it.

`Route.PortalStreamExecute` previously reissued the client's original portal, ignoring the rewritten route query — so on the extended (Parse/Bind/Execute) protocol the revert never reached the backend and an unpinned `set_config` persisted on the pooled backend, leaking across clients. It now reissues `r.Query` (keeping the client's binds) when it differs from the portal's prepared statement, so the rewrite applies on the portal path too; plain routes are untouched.

## Validation

- `go test -short` across the affected multigateway / multipooler / protoutil packages.
- e2e suite (`go/test/endtoend/queryserving/set_config_connstate_repro_test.go`): deterministic cross-client leak/isolation on both the simple and extended (portal) protocols, prepared `EXECUTE` tracked-and-isolated, suspended set_config portal, dynamic gateway-managed containment, failed-commit stamping.


go/services/multigateway/engine/session_state_branch.go
```diff
@@ -0,0 +1,164 @@
...
+		Pinned:     pinned,
+		Unpinned:   unpinned,
+	}
+}
+
+// choose picks the pinned branch when the set_config will land on a
+// session-affine backend: the session is already pinned, OR this statement
+// reserves its own backend (see StatementReservesBackend), OR reservesBackend
+// is set by the caller for a path that reserves at runtime (a row-limited
+// portal — the multipooler reserves any maxRows>0 portal for possible
```
shedges, line 109: there is scenario where session state may leak if portal exit with less rows than maxRows. Let's say client set maxRows = 10 and postgres has only 3 rows then  PostgreSQL returns all 3 rows and marks the portal as complete. But the gateway already chose the is_local=false version, so the setting is now stuck on a pooled backend.

go/services/multigateway/engine/session_state_branch.go
```diff
@@ -0,0 +1,164 @@
...
+		return true
+	}
+	return state != nil && state.HasReservedConnectionFor(tableGroup, shard)
+}
+
+// StatementReservesBackend reports whether the statement itself reserves its
+// backend for the session (via its own PlanExecInfo intent), even when the
+// session was not already pinned. This matters for a set_config that shares a
+// statement with a backend-pinning call — e.g.
+// `SELECT set_config('x','y',false), pg_advisory_lock(1)`: SessionPinned is
```
shedges, line 50: there is scenario where `pg_try_advisory_lock` returns false when the lock is unavailable. In that case, no lock is acquired, the backend is never reserved and it goes straight back to the pool. The `set_config` already ran with `is_local=false`, so the GUC change is now stuck on a pooled backend.

shedges (approved):
Rest LGTM

go/services/multigateway/engine/session_state_branch.go
GuptaManan100: Fixed!

go/services/multigateway/engine/session_state_branch.go
GuptaManan100: Fixed!

PR #1374, opened 2026-08-12
refactor(multiorch): drop IsLastCheckValid in favor of health recency

dweitzman:
Analyzers were often treating a temporary blip in network connectivity as a sign that a pooler is unhealthy

Networks are unreliable, so we shouldn't read anything too strong into a brief interruption. This PR updates some analyzers to consider health data stale based on how old it is, not based on whether there's a TCP connection that appears currently active to that pooler


proto/multiorchdata.proto
```diff
@@ -36,9 +36,6 @@ message PoolerHealthState {
   // serving_status, hostname, port_map.
   clustermetadata.Multipooler multipooler = 1;
 
-  // Computed fields (cached by multiorch)
-  bool is_last_check_valid = 3;
```
dweitzman, line 40: Any code that was previously using `is_last_check_valid` can just consult `last_seen != nil` and/or `stream_connected` instead, if that really represents what the code wanted to know

go/services/multiorch/grpcserver/server.go
```diff
@@ -202,7 +202,7 @@ func (s *MultiorchServer) buildPoolerHealthList(req *multiorchpb.ShardStatusRequ
 
 		healthList = append(healthList, &multiorchpb.PoolerHealth{
 			PoolerId:      h.Multipooler.Id,
-			Reachable:     h.IsLastCheckValid,
+			Reachable:     h.StreamConnected,
```
Verolop: Should this be based on how recent LastSeen is instead? As it is now, a dropped TCP connection would  immediately flip Reachable to false, which seems like exactly what this PR is trying to avoid. Maybe a separate field?

Verolop (approved):
left a small non-blocking comment. Otherwise lgtm.

go/services/multiorch/grpcserver/server.go
dweitzman: Renamed `Reachable` to `StreamConnected` and replaced `LastCheck` with `LastSeen` so the e2e code can be more explicit about what it wants. None of this is used in production, just tests

PR #1375, opened 2026-08-13
fix(multigateway): buffer pre-execution transport failures during failover

haritabh17:
## Problem

During a failover, requests whose error is a raw gRPC transport failure (`connection refused` dialing a dead multipooler) are returned to the client immediately instead of being buffered — even while the failover buffer is armed for that shard. `classifyError` only buffers no-writable-primary / MTF01 / pre-execution-marked errors, and `mterrors.FromGRPC` turns a dial failure into a bare `UNAVAILABLE`, which is `actionFail`.

Observed in the field: bursts of `portal execution failed: Code: UNAVAILABLE ... connect: connection refused` against incrementing stale pod IPs, continuing ~200ms after the buffer armed. The auth path already handles this case (`GetAuthCredentials` marks `UNAVAILABLE` pre-execution, safe there because auth is idempotent); the query paths had no equivalent.

## Fix

**Mark stream-start failures as pre-execution** (`markStreamStartFailure` in `grpc_query_service.go`): a non-EOF `UNAVAILABLE` from a generated streaming-client constructor is provably a NewStream/pick failure — the request message never left the gateway, so no query can have executed and buffering a retry is safe. This holds structurally in gRPC-Go: `CloseSend` always returns nil, and `SendMsg` failures surface as `io.EOF`, which `FromGRPC` never maps to `UNAVAILABLE`. Mid-stream `Recv` failures and unary `ExecuteQuery` stay unmarked — there the request may have executed, and replaying could double-apply a write. `classifyError` itself is unchanged: markers assert transport facts at the site where they are known; the classifier consumes them.

## Also in this PR

- **feat(multigateway):** `multigateway.buffer.requests.failed_unbuffered` counter (labeled by mtrpc code) — fires when an error passes through as `actionFail` while the shard buffer is BUFFERING/DRAINING. During a healthy failover no infrastructure error should reach a client, so this is the alarm for the next classification gap.
- **fix(pgprotocol):** a diagnostic-less transport `UNAVAILABLE` (e.g. a reserved-connection query against a dead pooler, which bypasses buffering by design) now reaches clients as SQLSTATE `08006` `connection_failure` instead of `XX000` wrapping raw gRPC text.
- **fix(mterrors):** `%v` renders error chains on a single line, message-first. Previously the cause landed on a second line that log tooling drops, hiding the `connection refused` detail behind `Code: UNAVAILABLE`.

## Tests

- Contract tests pin both halves of the boundary: a dial failure at stream start **is** marked pre-execution; an `UNAVAILABLE` after data has streamed is **not**.
- `TestPoolerGateway_BuffersDialFailureAgainstStalePrimary` reproduces the field failure deterministically: a frozen-health stale routing leader at a refused port + continuous writes must buffer and drain against the promoted primary. Fails without the fix with the exact production error; passes with it.
- `classifyError` table pins the safety property: an unmarked transport `UNAVAILABLE` never buffers.
- Full unit suites (incl. `-race`) and the e2e buffer suite (`TestBuffer*`) pass.


go/services/multigateway/poolergateway/pooler_gateway.go
```diff
@@ -373,14 +374,27 @@ func (pg *PoolerGateway) withBuffering(
 			"pooler_id", conn.ID())
 
 		err = inner(conn)
-		if err != nil && classifyError(err, target, retryReadOnlyError) == actionBuffer {
-			continue
+		if err != nil {
+			if classifyError(err, target, retryReadOnlyError) == actionBuffer {
+				continue
+			}
+			pg.recordUnbufferedFailure(ctx, sk, err)
```
GuptaManan100, line 381: On a context-cancelled request during an active buffer, this records code=CANCELED into the alarm counter. Should we special case to not record that? A query timed out for whatever reason shouldn't surface here.

GuptaManan100 (approved):
Rest LGTM!

go/services/multigateway/poolergateway/pooler_gateway.go
haritabh17: Good catch — agree, fixed.

PR #1376, opened 2026-08-13
fix(topo): expire gateway/orch registrations via etcd leases and retry deregistration

haritabh17:
## Problem

Multigateway and multiorch register themselves in etcd as **permanent keys** (no lease), and `toporeg.Unregister()` makes a **single delete attempt with a hardcoded 1s timeout**, logging failure and moving on. Registration, by contrast, retries with backoff until it succeeds — the asymmetry is the bug. Two ways to leak:

- **Graceful shutdown**: any etcd round-trip over ~1s loses the delete. `Unregister` also runs late inside `Shutdown()` under the shared OnClose window, so it can be skipped entirely.
- **Crash path**: SIGKILL / OOM / node loss never runs shutdown at all — no shutdown-side fix can cover this class.

Either way, one missed delete = a stale registration that lives forever. Observed on a live cluster across rolling updates, scale-downs, and node restarts (a node restart re-ran gateway containers in place; each generated a fresh random service ID, orphaning the old registrations under still-live hostnames).

This is not cosmetic: each stale gateway record permanently holds one of 2047 PID prefixes used for query-cancel routing and poisons the cancel-routing prefix cache; a stale multiorch record can be picked persistently by multiadmin's rule-change path.

## Fix

Three pieces: two for registration lifecycle, one for prefix ownership.

**1. Entries expire when their owner dies.** A new `Conn.PutEphemeral` binds gateway/orch registrations to a lease shared per connection (TTL from the existing `--topo-etcd-lease-ttl`), renewed for the connection's lifetime. When the process dies, renewals stop and etcd deletes the entries within one TTL — covering the crash path. `memorytopo` implements it as a plain upsert (no liveness concept).

**2. Entries come back when the owner is alive.** A lease can also end while the process is healthy — expiry after a topology outage, a reconnection replacing the underlying connection, or a cell reconfiguration replacing it wholesale. Only the component knows its record should exist, so re-creation lives with the component: `toporeg`'s `WithReassert` rewrites the record periodically. One loop covers every way an entry can go missing, **including ones the connection layer cannot observe**, so neither `etcdtopo` nor `WrapperConn` tracks what it wrote. The loop stops before deregistration, so it can never resurrect a record the component just removed.

**multipooler is deliberately excluded** from both: its records are persistent by design (lifecycle state plus orchestrator-side tombstone cleanup in `recovery/bookkeeping.go`), and it already republishes on its own schedule — a second writer would churn every watcher of those keys.

**3. PID prefixes are claimed atomically, decoupled from registration records.** The old allocator was scan-guess-verify: infer free prefixes from gateway records, pick one, register, then check for a concurrent duplicate. Ownership inferred from record presence is inherently racy — two starting gateways can pick the same free prefix and rely on the verify step to converge — and lease-backed records make it worse: a record absent during an outage makes its prefix look free. Prefix ownership now lives in its own claim file (`gateway_prefixes/<n>`, global topology) claimed via a new `Conn.ClaimEphemeral`: atomic create when absent, refresh when held by the same gateway (rebinding to the current lease), `NodeExists` when held by another. Two racing gateways cannot both win, so the post-registration collision check is deleted — and so is the unverified reuse of a prefix from an old record. Each process claims fresh; a dead claimant's file expires with its lease.

If a re-assertion finds the claim held by another gateway (possible only after the claim expired during an outage), the loss is fatal: a live gateway cannot renumber — its prefix is baked into its listeners and cancel manager, and clients hold PIDs stamped with it — so the gateway **fails its readiness check (draining new connections) and initiates its own graceful shutdown** via a new `ServEnv.InitiateShutdown`. The supervisor (Kubernetes `restartPolicy`, systemd, docker-compose) restarts the process, which claims a fresh prefix — encoding the restart in the process itself keeps recovery independent of probe configuration (a failing readiness probe alone restarts nothing). The graceful path also unregisters the gateway's record promptly, so no stale record advertises the reassigned prefix alongside its new owner. Claims are **never explicitly released** — lease expiry is the only release path, so a gateway can never delete a claim that has passed to another gateway (the prefix-lost restart path would otherwise do exactly that, briefly freeing the new owner's prefix for a third gateway to claim). A stopped gateway's prefix frees itself within one TTL; with 2047 slots and one claim per process, reuse has no urgency. The lease is deliberately not revoked on connection close: connections also close while the process is alive (reconnections), and an eager revoke would delete a healthy gateway's claim at every swap — the ≤TTL lingering is the grace period that bridges them.

**4. `toporeg.Unregister` retries with backoff for up to 5s** instead of one 1s attempt, for prompt cleanup on the graceful path. `NoNode` counts as success (a previous attempt's delete may have been applied even though its response was lost).

## Design note: why the connection layer keeps no bookkeeping

Re-creating lost ephemeral files from inside the connection layer looks natural but cannot be complete: a connection can be replaced wholesale — `WrapperConn` swaps its underlying conn on retriable errors, and `store.ConnForCell` replaces the cached conn when a cell's addresses change — and the replacement has no memory of what the old one wrote. Only the owner knows its registration should exist, so re-creation lives solely with the owner, and `etcdtopo`/`WrapperConn` deliberately track nothing. The cost is one idempotent rewrite every 10s per gateway/orch; nothing watches those keys, so there is no watcher fan-out.

## Mixed-version notes

- Gateways from before prefix claims advertise a prefix without holding a claim file; the claim loop scans records once and avoids their prefixes. The scan narrows nothing once the fleet writes claim files.
- Registrations leaked by earlier versions carry no lease and never expire on their own; they need one-time cleanup (`etcdctl del`, or operator-side pruning — tracked separately for the multigres-operator).

## Follow-ups (out of scope)

- **`WrapperConn.handleConnectionError` never classifies etcd timeouts as retriable** (pre-existing): `convertError` maps etcd Unavailable/DeadlineExceeded to `TopoError{Timeout}`, whose `mterrors.Code` is UNKNOWN and whose message lacks the "context " prefix, so no reconnection is scheduled for that class on any operation. Re-assertion makes registrations recover regardless, but the gap deserves its own fix.
- One-time cleanup of pre-lease leaked registrations (above).

## Tests

- Conformance suite (memorytopo **and** etcd): `checkPutEphemeral` (create/overwrite/delete/re-create) and `checkClaimEphemeral` (atomic claim, competitor rejected without overwriting the holder, holder refresh, claimable again after the claim is removed).
- etcd end-to-end: `testEphemeralLeaseRenewal` (files share the connection's lease; revoking deletes them; the next write lands on a fresh lease) and `testEphemeralExpiresOnClose` (renewals stop → etcd deletes the entry within one TTL — the SIGKILL promise).
- toporeg: re-assertion rewrites periodically, is off by default, stops before deregistration deletes, survives transient failures, and restores a record deleted out-of-band through a real store (verified to fail when re-assertion is disabled).
- Gateway prefix claims: distinct claims under concurrent claiming; mixed-version avoidance (record-advertised prefix never taken, proven by exhaustion); theft-detection round-trip (holder refresh ok → competitor rejected → claim expiry (simulated) → competitor claims → old holder's refresh fails); lost claim is fatal exactly once (healthy refresh and transient topology errors trigger nothing; a stolen claim sets the readiness flag and initiates shutdown once, never twice); `ServEnv.InitiateShutdown` queues one termination signal and never blocks on repeat calls.
- `WrapperConn`: `PutEphemeral`/`ClaimEphemeral` forward; explicitly does **not** re-create after a reconnection — pinning owner-side re-assertion as the single source of recovery.
- Full `-short` unit suite, `-race` on changed packages, `TestEtcd2Topo`, multiadmin e2e (real gateways claiming prefixes through real etcd), multiorch e2e.

### DIFF
--- a/go/common/servenv/servenv.go
+++ b/go/common/servenv/servenv.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 	"os"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/multigres/multigres/go/common/mterrors"
@@ -247,6 +248,25 @@ func (se *ServEnv) RegisterReadyCheck(f func() error) {
 	se.readyMu.Lock()
 	defer se.readyMu.Unlock()
 	se.readyChecks = append(se.readyChecks, f)
+}
+
+// InitiateShutdown triggers the same graceful shutdown sequence as SIGTERM:
+// lameduck, OnTerm/OnTermSync hooks, then OnClose hooks, then process exit.
+//
+// It exists for components that detect a fatal condition only a process
+// restart can repair — for example a gateway whose PID prefix claim has
+// passed to another gateway. Exiting from inside the process makes the
+// restart independent of deployment configuration: Kubernetes restartPolicy,
+// systemd, and docker-compose all restart an exited process, whether or not
+// any probe is wired up.
+//
+// Safe to call at any time and from any goroutine; repeat calls while a
+// shutdown is already pending are no-ops.
+func (se *ServEnv) InitiateShutdown() {
+	select {
+	case se.exitChan <- syscall.SIGTERM:
+	default: // A shutdown signal is already pending.
+	}
 }
 
 // GetHTTPPort returns the HTTP port value
--- a/go/common/servenv/servenv_test.go
+++ b/go/common/servenv/servenv_test.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"syscall"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -99,5 +100,33 @@ func TestRegisterReadyCheck(t *testing.T) {
 		w := httptest.NewRecorder()
 		se.mux.ServeHTTP(w, httptest.NewRequest(http.MethodGet, "/ready", nil))
 		require.Equal(t, http.StatusOK, w.Code)
+	})
+}
+
+func TestInitiateShutdown(t *testing.T) {
+	t.Run("delivers a termination signal", func(t *testing.T) {
+		se := NewServEnv(viperutil.NewRegistry())
+		se.InitiateShutdown()
+		select {
+		case sig := <-se.exitChan:
+			require.Equal(t, syscall.SIGTERM, sig)
+		default:
+			t.Fatal("InitiateShutdown should have queued a termination signal")
+		}
+	})
+
+	t.Run("never blocks on repeat calls", func(t *testing.T) {
+		// The channel holds one pending signal; further calls must be
+		// no-ops rather than blocking the caller (a re-assert goroutine).
+		se := NewServEnv(viperutil.NewRegistry())
+		se.InitiateShutdown()
+		se.InitiateShutdown()
+		se.InitiateShutdown()
+		require.Equal(t, syscall.SIGTERM, <-se.exitChan)
+		select {
+		case sig := <-se.exitChan:
+			t.Fatalf("expected exactly one queued signal, got a second: %v", sig)
+		default:
+		}
 	})
 }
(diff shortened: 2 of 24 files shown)


shedges (approved):
🚀

PR #1377, opened 2026-08-13
feat(multipooler): report reloaded settings against expected values

GuptaManan100:
## Summary

- `ReloadConfig` now takes an optional `expected_settings` map (reload-safe GUC name → desired value) and returns a per-setting verdict alongside the existing `config_load_time`: `all_applied`, `mismatches`, and `needs_restart`.
- After the reload, the multipooler reads `pg_file_settings` (joined with `pg_settings` for `pending_restart`) and compares the file postgres just re-read against the desired values, so a caller can tell a stale config file apart from a change that requires a restart.
- Empty `expected_settings` preserves the existing behavior (just reload); the verdict fields stay at their zero values.

## Description

The reload path previously confirmed only that *a* reload happened (`pg_conf_load_time()` advanced). That does not tell an operator whether the file postgres actually read carries the change they wrote — a not-yet-synced file passes the old check silently.

`ReloadConfig` now verifies the outcome. It reads `pg_file_settings` because that view reflects the config file the SIGHUP read (a stale file still shows the old value → the mismatch is visible) and its `setting` column is the raw file token (`32MB`), which matches what the operator wrote without unit normalization. `pg_settings` is consulted only for `pending_restart`: a restart-requiring change reports the generic `"setting could not be applied"` in `pg_file_settings`, so the restart signal has to come from `pg_settings.pending_restart` (confirmed against the PostgreSQL `guc.c` apply path).

For each expected setting the response reports whether it is present in the file, the actual file value, whether it was applied, and any error; `needs_restart` is set when a correctly-written value could not take effect without a restart.

Verification uses the internal query service, which connects as the configured superuser and can read `pg_file_settings` without extra grants.

## Tests

- Unit tests for the verdict logic (all-applied, stale/wrong value, missing, needs-restart, shadowed duplicate occurrence, query failure).
- End-to-end tests that write `postgresql.conf` on disk directly (the path this feature exists for) and drive the RPC against a real PostgreSQL: a match/mismatch case on `work_mem` and a restart-required case on `max_prepared_transactions`, restoring the file on cleanup.


go/services/multipooler/internal/manager/rpc_reload_config.go
```diff
@@ -85,7 +96,153 @@ func (pm *MultipoolerManager) ReloadConfig(ctx context.Context) (*multipoolerman
...
+	// pg_file_settings has one row per occurrence of a setting across all config
+	// files. Ordering by seqno lets us pick the effective occurrence per name
+	// (see effectiveFileSetting). pg_settings.pending_restart is per-setting, so
+	// a LEFT JOIN by name is sufficient; unknown/custom GUCs have no pg_settings
+	// row and their pending_restart reads as NULL (treated as false).
+	queryCtx, cancel := context.WithTimeout(ctx, timeouts.PostgresConfigTimeout)
+	defer cancel()
+	result, err := qs.Query(queryCtx, `SELECT fs.name, fs.setting, fs.applied, fs.error, s.pending_restart
+FROM pg_file_settings fs
+LEFT JOIN pg_settings s ON s.name = fs.name
```
shedges, line 178: join `s.name = f.name` only matches name of setting and return if restart needed or not. We should also check value i.e. `fs.setting` is matches with `pg_settings.setting` or not. It might be possible that we perform reload before file flush to disk and compare time with file `pg_conf_load_time`. 

1/ If local file synchronization is guaranteed complete before the RPC starts then current changes are sufficient.
2/ Also this verification confirm changes on current backend. It might be possible that other busy backend not processed yet.

Sharing my thoughts please confirm timing
````
10:00  Caller expects work_mem=64MB
10:01  File on the machine still says work_mem=32MB
10:02  Multipooler sends reload                                        <<< RPC
10:03  PostgreSQL reads 32MB
10:04  pg_conf_load_time advances
10:05  File synchronization finally writes 64MB
10:06  Code queries pg_file_settings and sees 64MB

The checks now see:

Reload time advanced:            yes
Current file contains 64MB:      yes
File entry can be applied:       yes
````

Comment (GuptaManan100) [reply go/services/multipooler/internal/manager/rpc_reload_config.go:178]: Good catch, I reorded the check vs the reload to fix this!

File: go/services/multipooler/internal/manager/rpc_reload_config.go [hunk 2]
```diff
@@ -85,7 +125,152 @@ func (pm *MultipoolerManager) ReloadConfig(ctx context.Context) (*multipoolerman
...
+		// "reload won't help, restart" case from a stale/wrong file value.
+		requiresRestart := present && fs.setting == want && !fs.applied && fs.context == "postmaster"
+		if requiresRestart {
+			verdict.needsRestart = true
+		}
+
+		verdict.mismatches = append(verdict.mismatches, &multipoolermanagerdatapb.SettingMismatch{
+			Name:            name,
+			Expected:        want,
+			Actual:          fs.setting,
```
Verolop: Could we avoid returning `fs.setting` here, or redact sensitive GUCs? Since expected_settings is caller-controlled, a client can send a bogus expectation for `primary_conninfo` and get the raw file value back in actual, which may contain a replication password. We already redact `primary_conninfo` elsewhere. I think the other mismatch fields are enough without returning the raw value.

shedges (approved):
LGTM

proto/multipoolermanagerdata.proto
```diff
@@ -633,18 +633,85 @@ message SetPostgresRestartsEnabledResponse {
...
+message SettingMismatch {
+  // GUC name.
+  string name = 1;
+
+  // Value the caller expected (from ReloadConfigRequest.expected_settings).
+  string expected = 2;
+
+  // Raw value found in the config file (pg_file_settings.setting). Empty when the
+  // setting is absent from the file entirely (see present).
+  string actual = 3;
```
mkindahl: Do you need to pass this back? This will effectively export the value from the server, which might leak information.

proto/multipoolermanagerdata.proto
```diff
@@ -633,18 +633,85 @@ message SetPostgresRestartsEnabledResponse {
...
+
+// SettingMismatch reports one expected GUC whose state in the config file (read
+// before the reload) does not match what the caller wrote, or cannot be applied
+// by a reload.
+message SettingMismatch {
+  // GUC name.
+  string name = 1;
+
+  // Value the caller expected (from ReloadConfigRequest.expected_settings).
+  string expected = 2;
```
mkindahl: Do you need to return this? What would it be used for?

proto/multipoolermanagerdata.proto
```diff
@@ -633,18 +633,85 @@ message SetPostgresRestartsEnabledResponse {
...
+
+  // Raw value found in the config file (pg_file_settings.setting). Empty when the
+  // setting is absent from the file entirely (see present).
+  string actual = 3;
+
+  // Whether the setting appears in the config file at all. False means the file
+  // PostgreSQL would read does not mention this setting — typically a stale file
+  // that has not yet synced the caller's write.
+  bool present = 4;
```
mkindahl: If you don't return the actual value, there is little reason to return if it is present as well.

mkindahl (approved):
It seems you fixed some of the comments I had while I was reviewing (passing back values seems risky).

Please correct me if I am wrong, but the problem seems to be that there is a race condition between the update of the ConfigMap and the RefreshConfig RPC. It seems to be because these events are driven by an external entity, e.g., the operator.

When the ConfigMap is updated, it does not immediately take effect and can be delayed. If the operator sends a RefreshConfig RPC before the file is updated physically, the refresh has no effect.

An ideal solution to this problem would be to add a sidecar or otherwise make pgctld reload the configuration on the config file change (similar to how [configmap-reload](https://github.com/jimmidyson/configmap-reload) works). This would allow the operator to just update the ConfigMap and then wait for it to take effect.

However, there are some unsolved problems remaining:

- The new ConfigMap could contain option values that require a restart, and automatically triggering a restart could be disruptive, so we should not do it blindly.
- The new ConfigMap could contain error entries, which could cause unexpected problems.

So I think we should merge this PR for now, and think about the alternative options after that.

go/services/multipooler/internal/manager/rpc_reload_config.go
GuptaManan100: Ooooh wow! execellent point, it didn't even cross my mind!

GuptaManan100: Removed!

proto/multipoolermanagerdata.proto
GuptaManan100: Removed!

proto/multipoolermanagerdata.proto
GuptaManan100: Removed!

proto/multipoolermanagerdata.proto
GuptaManan100: Removed!

PR #1378, opened 2026-08-14
chore(proto): clean up field definitions before MVP freeze

mkindahl:
## What

Phase 1a of the pre-MVP protobuf cleanup (MUL-524): remove reserved field numbers and prune dead field definitions so the schema starts from a clean slate before the release freeze, when field churn is still cheap and there are no wire-compatible deployed consumers to protect.

- **query.proto** — drop `reserved 10, 11` (`ExecuteOptions`) and `reserved 6` (`ReservationOptions`), and densely renumber `ExecuteOptions` from `1,2,4,5,7,8,9,12` to `1..8`.
- **multiadminservice.proto** — delete the deprecated `BackupInfo.backup_time` field (already superseded by `start_timestamp`/`stop_timestamp`) and renumber the message to `1..15`.
- **multipoolerservice.proto** — delete the "unused in v1" placeholders: `ReplicationMode.REPLICATION_MODE_TRUE` and both `send_unixnano` fields.

Go and TypeScript bindings are regenerated. The stream-replication test now uses an out-of-range mode value instead of the removed `REPLICATION_MODE_TRUE` constant, preserving its "reject unimplemented mode" coverage.

## Scope decisions

- **Targeted renumber, not a blanket one.** A full gap scan showed most numbering gaps are intentional design — grouped fields in `multiorchdata.PoolerHealthState`, the deliberate `reserved_state = 5` symmetry across the multipooler response messages, and the powers-of-2 `ReservationReason` bitmask. Only the two messages actually edited here (which carry no cross-message constraint) were renumbered; intentional numbering is left untouched.
- **`PoolerType.type` and `DRAINED` are deliberately not removed here.** They were deprecated-in-place specifically to flush out remaining readers, so removal is deferred to a Phase 1b that first confirms no in-repo, operator, or persisted-metadata readers remain.

## Verification

- `go build` of affected packages: OK
- `go test -short ./go/services/multipooler/grpcpoolerservice/... ./go/pb/...`: pass (including the query redact-annotation lock test)
- No remaining references to the removed symbols in hand-written Go or the multiadmin web app

## Note for reviewers

This is an intentionally wire-breaking change, which is safe pre-MVP. It's exactly the kind of change the planned `buf breaking` CI gate (a later phase) is meant to block once the schema is frozen.

### DIFF
--- a/proto/multipoolerservice.proto
+++ b/proto/multipoolerservice.proto
@@ -368,7 +368,6 @@ message CopyBidiExecuteResponse {
 enum ReplicationMode {
   REPLICATION_MODE_UNSPECIFIED = 0;
   REPLICATION_MODE_DATABASE = 1;  // replication=database (logical)
-  REPLICATION_MODE_TRUE = 2;      // replication=true (physical) — reserved, not used in v1
 }
 
 message StreamReplicationInit {
@@ -392,9 +391,6 @@ message StreamReplicationRequest {
     StreamReplicationInit init = 1;  // MUST be the first message
     bytes data = 2;                  // opaque client->server bytes
   }
-  // Reserved for future end-to-end latency measurement; unused in v1.
-  // Carried in the envelope, never inside `data`, to stay protocol-blind.
-  int64 send_unixnano = 3;
 }
 
 message StreamReplicationReady {
@@ -411,7 +407,6 @@ message StreamReplicationResponse {
     bytes data = 2;                    // opaque server->client bytes
     StreamReplicationError error = 3;  // infrastructure error (NOT a PG ErrorResponse)
   }
-  int64 send_unixnano = 4;             // reserved, unused in v1
 }
 
 // ReservationReason indicates why a connection needs to be reserved.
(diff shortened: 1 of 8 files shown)


proto/multipoolerservice.proto
```diff
@@ -392,9 +391,6 @@ message StreamReplicationRequest {
     StreamReplicationInit init = 1;  // MUST be the first message
     bytes data = 2;                  // opaque client->server bytes
   }
-  // Reserved for future end-to-end latency measurement; unused in v1.
-  // Carried in the envelope, never inside `data`, to stay protocol-blind.
-  int64 send_unixnano = 3;
```
mkindahl, line 397: This was reserved for future use, but it is straightforward to add a new field once we need it and this was currently unused, so I removed it for now.

proto/multipoolerservice.proto
```diff
@@ -368,7 +368,6 @@ message CopyBidiExecuteResponse {
 enum ReplicationMode {
   REPLICATION_MODE_UNSPECIFIED = 0;
   REPLICATION_MODE_DATABASE = 1;  // replication=database (logical)
-  REPLICATION_MODE_TRUE = 2;      // replication=true (physical) — reserved, not used in v1
```
mkindahl, line 371: This was unused, except in a test, so I saw no reason to keep it.

GuptaManan100 (approved):
Excellent catch! Thank you!

PR #1379, opened 2026-08-14
feat(multipooler): add logical slot primitives and slot-name helper

mkindahl:
## Summary

First of six steps implementing logical replication slot-failover ([MUL-482](https://linear.app/supabase/issue/MUL-482)). This step adds the multipooler-manager slot primitives and the deterministic slot-name helper. It is deliberately dormant: nothing in the lifecycle (recruit / SetPrimary / promote / demote) calls the new code yet, so there is no runtime behaviour change in a running cluster. Later steps wire it in.

Closes [MUL-1316](https://linear.app/supabase/issue/MUL-1316/slot-failover-slot-primitives-and-naming).

## What's included

- `EnsureLogicalSlot(name, plugin, failover)` → `pg_create_logical_replication_slot(...)`, idempotent (checks `pg_replication_slots` first; no error if the slot already exists).
- `DropLogicalSlot(name)`.
- `GetSlotState(name)` — projects a `pg_replication_slots` row into `LogicalSlotState{Name, CatalogXmin *int64, InvalidationReason *string, FailoverReady bool}`, where `FailoverReady` is computed in SQL as `synced AND NOT temporary AND invalidation_reason IS NULL`. Returns a wrapped `ErrLogicalSlotNotFound` when the slot is absent. Nullable columns are surfaced as `nil` rather than a sentinel.
- `LogicalSlotExists(name)` — canonical `SELECT EXISTS(...)` existence check, matching the idiom already used in this package.
- `LogicalSlotName(name)` — derives a deterministic, cluster-unique slot name from the multipooler component name (`ID.Name`, not the `{cell}_{name}` application_name): lower-cased, `-`→`_`, restricted to `[a-z0-9_]`, capped at 63 characters, and prefixed `mg_` so managed slots cannot collide with client-created ones. Returns an error if the name contains `_` (not a valid `ID.Name` character). The transform is collision-free; a short deterministic hash is appended (and the body truncated) only when the input leaves the injective `[a-z0-9-]` domain or would exceed the length cap.

All slot operations run as admin SQL through the pooler's existing superuser query path (the same `pm.exec` / `pm.query` mechanism used for `pg_promote()` and `ALTER SYSTEM`), and bind slot names as query parameters — never string-interpolated.

## Design

Part of the logical slot-failover design, Part 3 §3.2.1 and §3.2.3. The primary creates one slot per follower, named after that follower via `LogicalSlotName`, and each follower independently computes the same name for its `primary_slot_name` — hence `LogicalSlotName` is a standalone helper and the slot operations take the name as a parameter rather than deriving it from the local node's own identity.

## Testing

Unit tests cover the naming helper (sanitization, lower-casing, `-`→`_`, length cap, `mg_` prefix, underscore rejection, and collision-freeness across the injective/lossy boundary) and the slot operations (idempotent create, create-when-absent, drop, state read including NULL `catalog_xmin`/`invalidation_reason` and not-found, and the existence check). Run via the `/mt-dev` skill.

### DIFF
--- a/go/services/multipooler/internal/manager/logical_slots.go
+++ b/go/services/multipooler/internal/manager/logical_slots.go
@@ -1,0 +1,271 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package manager
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"strings"
+	"time"
+
+	"github.com/multigres/multigres/go/common/mterrors"
+	"github.com/multigres/multigres/go/common/timeouts"
+	"github.com/multigres/multigres/go/services/multipooler/internal/executor"
+
+	mtrpcpb "github.com/multigres/multigres/go/pb/mtrpc"
+)
+
+// ============================================================================
+// Logical Replication Slots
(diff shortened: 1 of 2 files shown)


mkindahl:
Closed in favor of #1394 .

PR #1380, opened 2026-08-14
fix(ci): tolerate elapsed-less final test events in coverage reporting

mkindahl:
## Problem

The **Code Coverage** jobs (`Direct test coverage` and `Subprocess coverage`) flake intermittently across PRs, failing a run in which every test actually passed. The failure is always the same, and it comes from the reporting step — not the tests:

```
Error: missing elapsed on final test event
Processing test results from coverage-direct-test-results.jsonl failed
```

These jobs pipe `go test -json` to a `*-test-results.jsonl` file and hand it to `dorny/test-reporter`'s `golang-json` parser. That parser groups events by `(Package, Test)`, takes each group's **last** event, and aborts the whole job if that event has no `elapsed` field. Only terminal events (`pass`/`fail`/`skip`) carry `elapsed`; `run`/`output`/`pause`/`cont` events do not.

A complete, all-passing stream can still end a group on an elapsed-less event. In the integration suite, background goroutines and child processes write directly to stdout/stderr; `test2json` wraps that stray output into an `output` event and attributes it to whichever test was most recently active. When a line lands in the narrow window *after* a test's terminal event but *before* the next test starts, that test's group ends on an `output` event — and the reporter throws. A killed test binary (timeout `SIGQUIT`, panic, OOM) produces the same shape via a truncated stream. This is why a green run turns red for a purely cosmetic reason, and why only the coverage jobs are affected: they consume raw `go test -json`, whereas the `gotestsum`-based jobs (`test-short`, `test-race`, `test-integration`) do not.

In the run that prompted this (PR #1365, job `Direct test coverage`), all packages reported `ok` and the go-test step succeeded; the job failed solely because the reporter choked on one such group.

## Fix

Add `go/tools/gotestjson.Normalize` and a thin `normalize` CLI, and run it over each results file *before* the reporter step in both coverage jobs. `Normalize` guarantees every `(Package, Test)` group ends on a terminal event that carries `elapsed`, **without hiding real failures**:

- If the group already reached a real terminal earlier (a later stray line displaced it as the last event), it re-emits that real result and duration — a passed test stays a pass, a failed test stays a fail.
- If the group never reached a terminal (interrupted/truncated stream), it appends a synthetic `fail` — an interrupted test is surfaced as a failure, not a silent pass.
- Unparseable trailing lines from a killed process are dropped so the reporter never hits `Invalid JSON`.

The reporter step keeps gating nothing on its own; the go-test step's exit code (with its existing whole-suite retry) is still what fails the job on a genuine test failure. An interrupted-stream repair also emits a `::warning::` so it stays visible.

No new GitHub Actions are introduced (the step is a `go run` of in-repo tooling), and job permissions are unchanged.

## Testing

- `go/tools/gotestjson/gotestjson_test.go` mirrors dorny's exact grouping check to **reproduce the crash on the raw stream and confirm it is gone after normalization**, and covers: stray-output-after-pass (pass preserved), a real failure followed by stray output (failure preserved), an interrupted test (marked failed), a truncated final line (dropped), a clean stream (byte-for-byte unchanged), and package-level events (ignored).
- Verified end to end with the `normalize` CLI against a crafted results file matching the CI shape: before, the reporter's contract is violated on the passing test's group; after, it parses cleanly with the pass preserved and a sibling failing test still reported as failed.
- `golangci-lint` clean; `gofmt`/`gofumpt` clean.

## Notes

The same `golang-json` reporter is used by several other workflows; they use `gotestsum` output and have not shown this flake, so this PR intentionally scopes the change to the two coverage jobs where it recurs. The normalizer is reusable if we want to extend it later.

Tracked in Linear (project MUL).

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -91,6 +91,17 @@ jobs:
           name: coverage-direct
           path: coverage-direct.txt
           retention-days: 7
+      # The golang-json reporter below fails the whole job if any test group's
+      # final event lacks an "elapsed" field ("missing elapsed on final test
+      # event"). A complete, all-passing `go test -json` stream can still end a
+      # group on an elapsed-less event (stray async output from the integration
+      # suite attributed to a test after its pass), which turns a green run red
+      # for a purely cosmetic reason. Normalize the stream first so the reporter
+      # never chokes on it; genuine failures are preserved. See the gotestjson
+      # package doc for details.
+      - name: Normalize test results for the reporter
+        if: ${{ !cancelled() }}
+        run: go run ./go/tools/gotestjson/normalize coverage-direct-test-results.jsonl
       - name: Test Report
         uses: dorny/test-reporter@a43b3a5f7366b97d083190328d2c652e1a8b6aa2 # v3.0.0
         if: ${{ !cancelled() }}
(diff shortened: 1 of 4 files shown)


mkindahl:
Superseded by #1381 — same commit, branch renamed to `ci/normalize-gotest-json-for-reporter` to match the `ci/…` convention (the rename closed this PR instead of repointing it).

PR #1384, opened 2026-08-14
feat(query-serving): chunk large DataRows across gRPC frames

shedges:
## Problem

Query results are already streamed from multipooler to multigateway in roughly 2 MiB batches, but batching currently happens only after the PostgreSQL client has read a complete DataRow. A single row larger than --grpc-max-message-size therefore becomes one oversized QueryResult protobuf message and fails with RESOURCE_EXHAUSTED. Raising the client and server limits only moves that ceiling; PostgreSQL permits individual values far larger than the default 16 MiB gRPC limit.

The opaque row path also materializes the complete DataRow body before batching it. Raising the gRPC limit toward PostgreSQL's maximum would consequently increase both transport exposure and peak multipooler memory rather than provide robust large-row support.

## Fix

**1. Opaque DataRows are read incrementally into bounded passthrough blocks**:

- The PostgreSQL message reader can now read the type and length header separately, leaving the reader positioned at the message body.
- Client-facing opaque DataRows larger than the 2 MiB streaming batch size are copied directly from the PostgreSQL buffered reader into bounded blocks. The complete row is never allocated in the multipooler.
- The original DataRow type and length header are preserved in the first block. Concatenating blocks in stream order reconstructs the exact PostgreSQL wire bytes.
- Existing complete rows already accumulated in a batch are flushed before an oversized row. Intermediate fragments carry a row count of zero; the final fragment carries one and is flushed immediately.
- The five-byte PostgreSQL message header is represented by `protocol.MessageHeaderSize` (`1 + PacketHeaderSize`) instead of repeated literals, keeping frame-size accounting explicit and consistent.

**2. Existing small-row behavior is preserved**:

- DataRows that fit within the streaming batch size remain complete and continue to be packed together efficiently.
- All five response readers use the same incremental path: simple query, Execute, Bind+Execute, Prepare+Execute, and Bind+Describe+Execute.
- Structured result consumers retain the existing parsed-row path.

**3. The protobuf envelope is extended compatibly**:

- passthrough_block now means the next ordered chunk of raw DataRow wire bytes; it may contain complete frames or a frame fragment.
- passthrough_row_count is the number of DataRows completed in that block and may be zero for an intermediate fragment.
- Existing passthrough field numbers and RPC shapes do not change. A new boolean field records whether a block ends inside a DataRow; existing poolers continue to send complete blocks with its default value of false.

**4. Interrupted rows terminate the client connection safely**:

- `passthrough_row_in_progress` explicitly reports whether a block ends inside a DataRow frame.
- If PostgreSQL or the gRPC stream fails after an incomplete row fragment reached the gateway, the pgwire server discards buffered output and closes the client connection. It does not append an ErrorResponse or ReadyForQuery that the client would misinterpret as row bytes.
- Failures between complete rows retain the existing frame-safe behavior and can still return a normal PostgreSQL error.

## Known trade-off

The bounded path applies to opaque client-facing rows. Structured internal query consumers still materialize and decode complete DataRows because they inspect column values rather than forwarding wire bytes. Those paths are intentionally used for small control queries and are not the arbitrary client result path addressed here.

## Tests

- Response-loop matrix covers all five simple/extended readers in opaque and structured modes. It uses a DataRow larger than the batch threshold and verifies bounded block size, a zero-count intermediate fragment, exact byte-for-byte reconstruction, and final row accounting.
- Affected package suites pass for pgprotocol/client, sqltypes, multipooler/grpcpoolerservice, and multigateway/poolergateway.
- Query-serving end-to-end regression now returns one 20 MiB field through both simple and extended protocols, exceeding the default 16 MiB gRPC message limit.
- Simple- and extended-protocol regressions verify that an interrupted DataRow produces no trailing ErrorResponse/ReadyForQuery, while an error after a complete opaque row remains frame-safe.
- Writer teardown coverage verifies buffered partial-frame bytes are discarded rather than flushed.
- Full production binary build passes.

### DIFF
--- a/go/common/pgprotocol/client/extended.go
+++ b/go/common/pgprotocol/client/extended.go
@@ -451,7 +451,7 @@ func (c *Conn) processExecuteResponses(ctx context.Context, callback func(ctx co
 	}
 
 	for {
-		msgType, body, streamed, err := c.readBatchMessage(batcher, flushBatch)
+		msgType, body, dataRowStreamed, err := c.readQueryResponseMessage(batcher, flushBatch)
 		if err != nil {
 			return false, responseReadError(firstErr, err)
 		}
@@ -469,7 +469,7 @@ func (c *Conn) processExecuteResponses(ctx context.Context, callback func(ctx co
 			}
 
 		case protocol.MsgDataRow:
-			if streamed {
+			if dataRowStreamed {
 				break
 			}
 			if err := batcher.addDataRow(body); err != nil {
@@ -854,7 +854,7 @@ func (c *Conn) processBindAndExecuteResponses(ctx context.Context, callback func
 	}
 
 	for {
-		msgType, body, streamed, err := c.readBatchMessage(batcher, flushBatch)
+		msgType, body, dataRowStreamed, err := c.readQueryResponseMessage(batcher, flushBatch)
 		if err != nil {
 			return false, responseReadError(firstErr, err)
 		}
@@ -875,7 +875,7 @@ func (c *Conn) processBindAndExecuteResponses(ctx context.Context, callback func
 			}
 
 		case protocol.MsgDataRow:
-			if streamed {
+			if dataRowStreamed {
 				break
 			}
 			if err := batcher.addDataRow(body); err != nil {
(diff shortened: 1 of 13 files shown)


haritabh17 (approved):
LGTM 🚀

PR #1385, opened 2026-08-15
feat(query-serving): unblock PostgREST spec suite

GuptaManan100:
## Summary

- Allow a transaction-local `set_config(name, value, true)` to appear outside a top-level SELECT target (in a WHERE clause, subquery, or CTE) by passing it through to the backend untracked. It reverts at transaction end, so the pooler needn't mirror it. This unblocks PostgREST's mutation row-count trick, which calls `set_config('pgrst.inserted', …, true)` inside an `INSERT … WHERE`.
- Quote col_name-keyword named-argument labels when deparsing, so a call like `f("varchar" => 1)` no longer round-trips to the invalid `f(varchar => 1)`. This fixes PostgREST RPC calls whose parameter names are keywords (`varchar`, `boolean`, `integer`, `json`, …).
- With both fixes the PostgREST spec suite runs fully green through the gateway (1303 examples, 0 failures, 1 intentionally-pending), so the divergence-tracking `DIVERGENCES.md` is removed.

## Description

**Transaction-local `set_config` pass-through.** The unsafe-funccall analyzer only accepted `set_config` as a top-level SELECT target and rejected it everywhere else, because a tracked SET can't reproduce the conditional/repeated evaluation those positions imply. A transaction-scoped call is different: it reverts at transaction end and leaves nothing for the pooler to track, so `allowTransactionLocalSetConfig` now lets it through untracked. It fails closed for everything else — `is_local=false` (would leak untracked backend state), a bound/dynamic `is_local` or `name` (unresolvable at plan time), a cluster-managed GUC, or a gateway-managed variable (must never reach the backend).

**Named-argument label quoting.** A named-argument label is a `param_name` (`type_function_name`), which excludes the col_name keyword category — so a col_name keyword is invalid bare there and must be quoted, even though it is valid bare as a column or type name. `NamedArgExpr.SqlString` used `QuoteIdentifier`, which keeps col_name keywords bare (correct for column/type deparsing), so the reconstructed query dropped the quotes and the backend rejected it with `syntax error at or near "=>"`. New `QuoteArgLabel` mirrors PostgreSQL's `quote_identifier` (quote every keyword whose category is not unreserved) and is used for labels.

Both paths are covered by planner unit tests, parser round-trip cases (including case-folding parity), and end-to-end tests through a real multigateway → PostgreSQL.

## Note

The `set_config` change and PR #1352 (`--unsafe-pooler-mode` / Tier 1 body analysis) both rewrite the same branch of `analyzeFunctionCalls`, so whichever lands second needs a small reconciliation — the safe always-on carve-out here should run before that PR's operator-opt-out `reject` gate.


shedges (approved):
LGTM

PR #1389, opened 2026-08-17
fix(cmd): Change cluster start command to it's correct setting

saifalharthi:
This is a tiny fix with not effect on current functionailty. While trying to start a cluster, the command suggests running `multigres cluster up` instead of `multigres cluster start` which is the correct command. This commit fixes that.


cuongdo (approved):
Thanks @saifalharthi!

PR #1391, opened 2026-08-17
fix(topo): scope component listings to requested cell

haritabh17:
## Summary

- Scope pooler, gateway, and orch listings to the requested cell.
- Add regression coverage for cells sharing the same topology server and root.

## Problem

When cells share a topology path, each per-cell listing returns every component, causing duplicate multiadmin API results.

## Solution

Filter unmarshaled records by their own cell in the typed `Get*ByCell` methods while continuing to support shared topology layouts.

### DIFF
--- a/go/common/topoclient/multipooler.go
+++ b/go/common/topoclient/multipooler.go
@@ -216,6 +216,9 @@ func (ts *store) GetMultipoolersByCell(ctx context.Context, cellName string, opt
 		if err := proto.Unmarshal(listResults[n].Value, multipooler); err != nil {
 			return nil, err
 		}
+		if multipooler.GetId().GetCell() != cellName {
+			continue
+		}
 		if opt != nil && opt.DatabaseShard != nil && opt.DatabaseShard.Database != "" {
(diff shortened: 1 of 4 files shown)


go/common/topoclient/multipooler.go
```diff
@@ -216,6 +216,9 @@ func (ts *store) GetMultipoolersByCell(ctx context.Context, cellName string, opt
 		if err := proto.Unmarshal(listResults[n].Value, multipooler); err != nil {
 			return nil, err
 		}
+		if multipooler.GetId().GetCell() != cellName {
```
Verolop, line 222: tiny thing before approving: could we apply the same cell filter to the three Get*IDsByCell methods as well? With a shared topo, they still return ids from both cells. It’d be great to include them in the new regression test too

Verolop (approved):
Looks good; would be nice to include the extra case in the test, but otherwise lgtm!

go/common/topoclient/multipooler.go
haritabh17: Got it, updating.

PR #1393, opened 2026-08-19
fix(multipooler): finish an in-flight pg_rewind past the RPC timeout, quarantine if it can't

mkindahl:
## Situation

A crash-killed primary rejoins the cluster as a standby by running `pg_rewind` against the current leader. `pg_rewind` rewrites the target data directory in place and is **not** transactional: there is a mutating phase (overwriting data/WAL blocks, then rewriting `pg_control`/`backup_label`) that is the point of no return. If that phase is interrupted the directory is left partially rewound — unstartable and, per [PostgreSQL guidance](https://www.postgresql.org/docs/current/app-pgrewind.html), generally unrecoverable (an interrupted rewind can even truncate the control file, so a plain re-run also fails; rebuild-from-backup is the only reliable recovery).

The concrete trigger is an **RPC timeout that kills the in-flight `pg_rewind`**. multiorch drives the stale-primary demote under a bounded action budget — `FixReplicationAction` `Timeout: 45s` / `DemoteStaleLeader` `60s`, applied to the action context in `recovery_loop.go`. That deadline flows down: `SetPrimary → setPrimaryLocked →` (stale/diverged primary) `→ restartAsStandbyLocked(ctx) → runPgRewind(ctx) → pgctldClient.PgRewind(ctx) → executil.Command(ctx, "pg_rewind")`, and `executil.Command` SIGTERM/SIGKILLs its subprocess when that context is cancelled. Because rewind runtime scales with retained `pg_wal` (below), a rewind can outlive the budget — so the deadline fires mid-write and leaves the half-rewound, unstartable directory. The monitor then starts postgres on it anyway; the node comes up in recovery, waits forever for WAL it can never fetch, reports `running` (which resets the unrecoverable-recovery classifier every tick), and spins until the volume is wiped.

**This is a real field failure, not a hypothetical.** The incident logs (`multigres-pg-rewind-recreated-primary-2026-07-22`) come from a server running the earlier `RewindToSource` code: 101 `"RewindToSource RPC called"` multipooler log lines, with the node looping rewind → start-failure → re-rewind ~100 times. That `RewindToSource` RPC has since been removed, but **the same RPC-timeout hazard persists on current `main` via the `SetPrimary` path above** (and via the same 45s `FixReplication` budget).

**Why rewinds run long enough to trip the budget** (from the `pg_rewind --debug` traces): ~80% of a 58s rewind was copying 116–121 × 16 MB WAL segments (~1.8 GB) for ~120 bytes of real divergence — **runtime scales with retained `pg_wal`, not divergence** — so under load / IO-throttling a rewind of 70s+ easily exceeds a 45s budget.

### Part A (primary) — let a started rewind finish regardless of the RPC timeout

`restartAsStandbyLocked` now detaches at the **point of no return**. The pre-stop measurement still runs on the caller's context (safe to abort — postgres is up and the directory is untouched), but the moment we stop postgres the stop → `pg_rewind` → restart-as-standby sequence runs on a context **detached from the caller's cancellation** (`ctxutil.Detach`) under its own generous timeout, while **keeping the action lock** (new `actionlock.CarryLock`, since `Detach` drops context values). When multiorch's deadline fires its RPC returns `DeadlineExceeded`, but the pooler keeps driving the rewind to a valid standby (or a definitive failure); the monitor stays blocked on the lock, so it cannot start postgres on the in-flight directory; multiorch simply retries and finds the node already healed.

```mermaid
sequenceDiagram
    participant Orch as multiorch
    participant Pool as Pooler
    participant Lock as Action lock
    participant Ctl as pgctld
    participant PG as Postgres

    Note over Pool: SetPrimary RPC handler holds the lock
    Orch->>Pool: SetPrimary, RPC deadline 45s
    Pool->>Lock: Acquire
    Note over Pool: stale primary, restartAsStandbyLocked
    Note over Pool: point of no return, detach from RPC deadline and keep the lock
    Pool->>Ctl: stop postgres
    Ctl->>PG: stop
    Pool->>Ctl: PgRewind, writes sentinel first
    Ctl->>PG: run pg_rewind on the stopped data dir
    Note over PG: rewinding, runtime scales with pg_wal, about 70s
    Note over Orch,Pool: 45s deadline elapses
    Pool-->>Orch: RPC returns DeadlineExceeded, handler keeps running
    Note over Pool,Lock: sequence continues on the detached context, lock still held
    Note over Lock: monitor blocks here, cannot start postgres mid rewind
    PG-->>Ctl: pg_rewind exits 0, Done, about 70s
    Ctl-->>Pool: PgRewind response
    Pool->>Ctl: restart as standby
    Ctl->>PG: start and verify in recovery
    Pool->>Ctl: remove sentinel
    Pool->>Lock: Release
    Orch->>Pool: SetPrimary retry next cycle
    Pool-->>Orch: already a healthy standby, no op
```

### Part B (safety net) — quarantine an interrupted rewind that cannot finish

Part A removes the RPC-timeout interruption, but a rewind can still be force-killed (a pod SIGKILLed past its shutdown grace). For that, a durable, fsynced on-disk **rewind sentinel** (mirroring the bootstrap sentinel) is written just before the mutating `pg_rewind` and removed only after a healthy standby is verified back up. On a later monitor tick (or a replacement pod's first tick) its presence forces the repair path instead of a blind start: it re-arms the `suspectedDivergence` flag (in-memory only, lost across a restart) so the node is repaired via a `pg_rewind` re-run, and it keeps the unrecoverable classifier counting even while postgres reports `running` — so a genuinely unrecoverable node is **quarantined** (`LIFECYCLE_QUARANTINED` + cohort `INELIGIBLE`) for operator replacement instead of spinning forever. A directory that can still be rewound just completes the re-run, clears the sentinel, and rejoins.

```mermaid
sequenceDiagram
    participant Life as Pod lifecycle
    participant Mon as Pooler monitor
    participant Ctl as pgctld
    participant PG as Postgres

    Note over Mon,PG: rewinding, sentinel on disk
    Life-->>Ctl: SIGKILL, shutdown grace exceeded
    Note over PG: data dir half-rewound, sentinel still present
    Note over Life,PG: replacement pod starts
    Mon->>Mon: sentinel present, re-arm suspected divergence
    alt directory still rewindable
        Mon->>Ctl: pg_rewind re-run
        Ctl-->>Mon: success
        Mon->>PG: restart as standby, verify healthy
        Mon->>Ctl: remove sentinel
        Note over Mon,PG: node rejoins the cohort
    else unrecoverable, e.g. control file truncated
        loop bounded by the unrecoverable-recovery budget
            Mon->>Ctl: pg_rewind re-run
            Ctl-->>Mon: fails
        end
        Mon->>Mon: quarantine, LIFECYCLE_QUARANTINED and cohort INELIGIBLE
        Note over Mon: operator replaces the pod, a clean node rejoins
    end
```

## What changed

- `go/services/multipooler/internal/manager/actionlock/action_lock.go` — add `CarryLock` (carry the action-lock marker across a detached context).
- `go/services/multipooler/internal/manager/rpc_manager.go` — Part A detach at the point of no return in `restartAsStandbyLocked`, under `rewindOperationTimeout`; write the rewind sentinel before the mutating `pg_rewind` and remove it after a verified-healthy standby.
- `go/common/constants/postgres.go` — add `RewindSentinelFile`.
- `go/services/multipooler/internal/manager/rewind_sentinel.go` (new) — sentinel helpers (on the pooler's local volume, fsynced).
- `go/services/multipooler/internal/manager/postgres_monitor.go` — surface `rewindSentinelPresent`; `remedialActionMarkRewindInterrupted` re-arms suspected divergence; the classifier no longer resets the failure streak on `postgresRunning` while the sentinel is present.

## Orchestrator interaction

No orchestrator change is needed; the fix relies only on existing mechanisms, but a rewinding pooler and a quarantined pooler are **two distinct states** worth spelling out:

- **While rewinding, the pooler stays `ELIGIBLE`** and remains a cohort member — it is *not* marked `INELIGIBLE`. Cohort eligibility is downgraded to `INELIGIBLE` in only two places, and the rewind path is neither: graceful shutdown, and quarantine (`markPoolerQuarantinedLocked`). What keeps a mid-rewind member from harming the shard is not eligibility but (a) the **recruit-position floor** (`SetRecruitBlockedUntil`, set before the rewind), which bars it from being counted toward a new rule-change quorum or re-ADDed to the cohort until it has caught back up, and (b) **ANY-N synchronous replication**, so its missing ack does not stall commits as long as one other cohort member is caught up.
- **Only when repair is exhausted does the pooler quarantine and flip to `INELIGIBLE`.** That is the signal the orchestrator acts on (verified end to end): an `INELIGIBLE` follower is dropped from the sync cohort (`CohortMismatchAnalyzer → ReconcileCohortAction` REMOVE), a quarantined/diverged node cannot be elected, and a quarantined ex-primary triggers re-election.

## Testing

- Unit: `CarryLock`; the point-of-no-return detach (a cancelled caller context neither cancels the operation nor drops the lock); sentinel round-trip and discovery; the divergence re-arm; and the classifier change (a running-but-mid-rewind node keeps counting and quarantines; a clean sentinel-free node still resets). Full multipooler manager suite green.
- Integration (real postgres + live multiorch): `TestInterruptedRewindQuarantinesStandby` models an interrupted rewind (sentinel + truncated `pg_control`) and asserts the terminal `LIFECYCLE_QUARANTINED` verdict; `TestRewindDivergedReplica` covers the happy-path rewind through the detached sequence.

## Follow-ups (out of scope here)

- No explicit multiorch "shard write-unavailable / durability degraded" signal for the degenerate case where a quarantined member drops a minimal cohort below its durability floor (writes stall by design). Tracked separately.
- No metric/event for `pg_rewind` **execution** time (only `rewind.checkpoint_wait.duration`, the pre-rewind wait). Worth adding given rewinds can now legitimately run long.
- `pg_rewind` copies the whole retained `pg_wal` regardless of divergence — a separate upstream-mitigation write-up.

---

Resolves [MUL-1343 — Prevent interrupted pg_rewind from wedging replicas](https://linear.app/supabase/issue/MUL-1343/prevent-interrupted-pg-rewind-from-wedging-replicas).

### DIFF
--- a/go/common/constants/postgres.go
+++ b/go/common/constants/postgres.go
@@ -106,6 +106,17 @@ const (
 	// backups.
 	BootstrapSentinelFile = ".multigres-bootstrap-in-progress"
 
+	// RewindSentinelFile marks an in-progress pg_rewind. Written before the actual
+	// (mutating) pg_rewind runs and removed only after postgres is verified back up
+	// as a standby; its presence on startup means a prior rewind was interrupted
+	// (e.g. the pod was killed mid-rewind) and the data directory is partially
+	// rewound — unstartable and, per PostgreSQL guidance, generally unrecoverable.
+	// The monitor uses it to force the rewind-repair path instead of starting
+	// postgres on the half-rewound directory, and to quarantine if repair keeps
+	// failing. Lives in pooler_dir (not PGDATA) so it stays out of pgBackRest
+	// backups, and on the local volume so it survives a pod restart on the same PVC.
+	RewindSentinelFile = ".multigres-rewind-in-progress"
+
 	// StandbySignalFile is PostgreSQL's marker file (in PGDATA) whose presence
 	// puts the server into standby mode. Notably, postgres --single refuses to
 	// run with it present, so crash recovery removes and recreates it.
--- a/go/services/multipooler/internal/manager/actionlock/action_lock.go
+++ b/go/services/multipooler/internal/manager/actionlock/action_lock.go
@@ -183,3 +183,22 @@ func AssertActionLockHeld(ctx context.Context) error {
 
 	return nil
 }
+
+// CarryLock copies the action-lock ownership marker from src onto dst (if src
+// holds one) and returns the augmented context; it is a no-op returning dst when
+// src holds no lock.
+//
+// It exists so a context detached from the caller's cancellation (e.g. one built
+// with ctxutil.Detach, which drops all context values) can keep proving ownership
+// of a lock that was acquired on the request context. This lets a long-running,
+// must-complete operation — such as a pg_rewind that has to outlive the caller's
+// RPC deadline — continue to hold the action lock (so the monitor cannot act
+// underneath it) while running deadline-immune. The same *actionLockValue is
+// shared between the two contexts, so the caller's Release still releases the one
+// lock and AssertActionLockHeld observes the release on both.
+func CarryLock(dst, src context.Context) context.Context {
+	if val, ok := src.Value(actionLockKey{}).(*actionLockValue); ok {
+		return context.WithValue(dst, actionLockKey{}, val)
+	}
+	return dst
+}
--- a/go/services/multipooler/internal/manager/actionlock/action_lock_test.go
+++ b/go/services/multipooler/internal/manager/actionlock/action_lock_test.go
@@ -212,3 +212,26 @@ func TestActionLock_ReleasePanicsWithWrongContext(t *testing.T) {
 		lock.Release(ctx1)
 	})
 }
+
+func TestCarryLock(t *testing.T) {
+	lock := NewActionLock()
+
+	held, err := lock.Acquire(context.Background(), "op")
+	require.NoError(t, err)
+
+	// A bare context does not hold the lock; carrying it over from the held
+	// context makes it prove ownership without re-acquiring.
+	bare := context.Background()
+	require.Error(t, AssertActionLockHeld(bare), "precondition: bare context holds no lock")
+
+	carried := CarryLock(bare, held)
+	require.NoError(t, AssertActionLockHeld(carried), "carried context should prove ownership")
+
+	// The same lock value is shared: releasing via the original context also
+	// invalidates the carried one (they are the one lock, not two).
+	lock.Release(held)
+	require.Error(t, AssertActionLockHeld(carried), "carried context should observe the release")
+
+	// No-op when the source holds no lock: returns dst unchanged.
+	require.Error(t, AssertActionLockHeld(CarryLock(bare, context.Background())))
+}
--- a/go/services/multipooler/internal/manager/postgres_monitor.go
+++ b/go/services/multipooler/internal/manager/postgres_monitor.go
@@ -120,6 +120,13 @@ type postgresState struct {
 	connInfo                 *multipoolermanagerdatapb.PrimaryConnInfo
 	pgMode                   pgmode.Mode
 	bootstrapSentinelPresent bool
+	// rewindSentinelPresent is true when a pg_rewind sentinel is on disk, meaning a
+	// prior rewind did not verifiably complete (see rewind_sentinel.go). It is the
+	// durable, restart-surviving signal that the data directory may be
+	// half-rewound; the monitor uses it to force the rewind-repair path rather than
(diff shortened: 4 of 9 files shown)


haritabh17 (approved):
LGTM 🚀

PR #1394, opened 2026-08-19
feat: logical replication slot failover via native PostgreSQL 17 slot sync

mkindahl:
Closes MUL-482

## Summary

Today Multigres only allows **temporary** logical replication slots: the Multigateway rejects any non-temporary `CREATE_REPLICATION_SLOT` because a persistent slot cannot survive a primary failover — slot state is not in the WAL stream, so a promoted standby never has it — and a mid-stream failover simply tears the tunnel down. A logical subscriber therefore cannot continue across a promotion; it has to drop and recreate its subscription and re-`COPY` every table.

This PR makes a persistent logical replication slot survive a primary failover using PostgreSQL 17's native slot-sync, so a subscriber can resume on the new primary instead of re-seeding. Everything is behind a dynamic feature flag (`--enable-slot-based-replication`, default off), so the branch is inert until an operator opts in.

## How it works

- **Multipooler** — logical-slot primitives (`EnsureLogicalSlot` / `DropLogicalSlot` / `GetSlotState` / `LogicalSlotName`); slot-based physical replication (per-follower physical slots, `primary_slot_name`, `hot_standby_feedback`, `sync_replication_slots`, `wal_level = logical`, and a `dbname` in `primary_conninfo`) maintained across the promote / demote / add-standby / remove-standby paths; `synchronized_standby_slots` on the primary; and failover-slot readiness that is **logged, not gated** at promotion (a fresh slot's `catalog_xmin` does not advance until a consumer consumes, so a not-ready slot is expected — gating would hang, and readiness can never be reached once the source primary is gone).
- **Multigateway** — the replication preamble now admits a non-temporary LOGICAL slot registered for failover — both the walsender `CREATE_REPLICATION_SLOT ... (FAILOVER)` command and the `pg_create_logical_replication_slot(..., failover => true)` SQL form — while every other non-temporary slot stays rejected.
- **Multiorch** — leader appointment is slot-aware: failover-slot readiness (reported in the health stream) breaks ties among WAL-equal candidates. WAL position always wins, so this never trades data safety for slot readiness.
- **Failover model** — client-driven: the consumer reconnects and resumes from its own durably-tracked LSN, which the slot synced to the new primary can still serve; a failover before a freshly created slot has synced degrades to a safe client-driven re-seed. There is deliberately **no** creation-time durability barrier (an ack-hold was prototyped and removed — it cannot deliver durable-before-ack because a fresh slot's `catalog_xmin` only advances once the client consumes, which the held ack would block).

## Message sequence charts

The consumer always connects to the **Multigateway**, which relays the logical stream to the current leader's Multipooler/PostgreSQL and, on a leader change, routes the reconnect to the new primary. Physical replication and slot-sync happen PostgreSQL-to-PostgreSQL between the poolers, below the gateway.

### 1. Happy path — subscribe and stream, slots synced to the standbys

A persistent failover slot is created **once** on the primary. The consumer streams through the gateway, and PostgreSQL 17's slot-sync workers on each standby independently copy that slot (clamped to the standby's own replay LSN) so every standby can serve a resume after a future promotion.

```mermaid
sequenceDiagram
    participant C as Consumer
    participant G as Multigateway
    participant P as Primary
    participant S1 as Standby 1
    participant S2 as Standby 2

    C->>G: CREATE_REPLICATION_SLOT sub LOGICAL (FAILOVER)
    G->>P: relay to the current leader
    Note over P: create persistent logical slot sub, failover=true
    P-->>G: slot created at consistent_point
    G-->>C: slot created
    C->>G: START_REPLICATION SLOT sub LOGICAL at lsn
    G->>P: relay START_REPLICATION
    P-->>G: XLogData, decoded WAL
    G-->>C: XLogData, relayed with LSN-aware tap
    C->>G: StandbyStatusUpdate flushed=lsn
    G->>P: relay StandbyStatusUpdate
    Note over P: advance confirmed_flush_lsn and catalog_xmin
    P-->>S1: stream WAL via physical slot mg_s1
    P-->>S2: stream WAL via physical slot mg_s2
    S1->>P: slot-sync worker reads failover slots
    Note over S1: create synced copy of sub, clamped to replay LSN
    S2->>P: slot-sync worker reads failover slots
    Note over S2: create synced copy of sub, clamped to replay LSN
    Note over S1,S2: synced copies become failover_ready as catalog_xmin advances
```

### 2. Failover while the consumer is streaming — disconnect and reconnect

A mid-stream primary failure drops the gateway's upstream, and the consumer's connection with it. Multiorch promotes the standby whose synced slot is failover-ready and re-bases the other standby onto it. The consumer reconnects to the gateway, which routes it to the new primary, and resumes from its own durably-tracked LSN — the synced slot is already present and retained there, so there is no re-`COPY`.

```mermaid
sequenceDiagram
    participant C as Consumer
    participant O as Multiorch
    participant G as Multigateway
    participant P as Primary
    participant S1 as Standby 1
    participant S2 as Standby 2

    C-->>G: streaming sub
    G-->>P: relay stream, XLogData and StandbyStatusUpdate
    Note over P,S2: sub is synced and failover_ready on S1 and S2
    Note over P: old primary fails
    P-xG: upstream stream drops
    G-xC: consumer connection drops
    O->>S1: promote, WAL-highest and failover-ready
    Note over S1: new primary, sub present and retained
    O->>S2: SetPrimary S1
    S2-->>S1: re-base and stream WAL
    C->>G: reconnect START_REPLICATION SLOT sub LOGICAL at confirmed_lsn
    G->>S1: route to the new leader
    S1-->>G: resume from confirmed_lsn, no re-COPY
    G-->>C: stream resumes
```

### 3. Standby added after the primary — keeping the slot-based path from stranding a late joiner

This deadlock is one the slot-based path *introduces*, not a pre-existing bug — so this MSC shows how the PR avoids it rather than fixing something older. With slot-based replication on, a standby streams through a per-follower physical slot and cannot start until that slot exists on the primary. The natural ordering — create a follower's slot only at the promotion hook (for the already-committed cohort), or via a `COHORT_ADD` that itself requires the standby to already be streaming — would deadlock a standby that joins after promotion: it cannot stream, so it is never admitted to the cohort, so its slot is never created. To keep the feature from stranding late joiners, the discovery-driven `ReconcileFollowers` RPC has the primary pre-create the slot the moment Multiorch sees the standby, ahead of streaming. (Surfaced in kind testing; see §2.6 of the design doc.)

```mermaid
sequenceDiagram
    participant C as Consumer
    participant O as Multiorch
    participant G as Multigateway
    participant P as Primary
    participant S1 as Standby 1
    participant S2 as Standby 2

    Note over C,P: initial cohort P and S1, physical slot mg_s1 exists
    C->>G: START_REPLICATION SLOT sub LOGICAL
    G->>P: relay to the current leader
    S2->>O: starts and reports health, REPLICA and not yet streaming
    O->>P: ReconcileFollowers followers=S1,S2
    Note over O,P: not a consensus message, level-triggered notification
    Note over P: create missing physical slot mg_s2 ahead of streaming
    O->>S2: SetPrimary P
    Note over S2: set primary_slot_name to mg_s2
    S2-->>P: stream WAL via mg_s2 and catch up
    S2->>O: now streaming, admitted to the cohort
    Note over S2: slot-sync then creates a synced copy of sub
    Note over P,S2: without ReconcileFollowers mg_s2 never exists and S2 deadlocks
```

## Design doc

`docs/ha/logical_slot_failover_design.md` covers the PostgreSQL background, the Multigateway/Multipooler machinery, the correctness invariants, the scenarios where data can actually be lost (§4.4.1), the bootstrap slot-creation deadlock the slot-based path introduces and how `ReconcileFollowers` avoids it (§2.6), and why the creation-time barrier was rejected (§4.4). A fully-DIY Patroni-model alternative is recorded in Appendix A.

## Testing

- Unit tests across the touched packages (Multipooler manager, Multigateway handler, Multiorch consensus, `rpcclient`, `parser/ast`).
- End-to-end: slot-sync propagation of a failover slot to a standby (driving a real consumer so `catalog_xmin` advances the slot to `failover_ready`); a subscription-wire test verifying the FAILOVER option rides in the `CREATE_REPLICATION_SLOT` a PostgreSQL 17 subscriber sends; and a regression test (`TestSlotBasedReplicationLateStandbyJoins`) that adds a standby after promotion and asserts it joins the cohort — reproducing the deadlock the slot-based path would otherwise hit (§2.6) and proving `ReconcileFollowers` avoids it.

### DIFF
--- a/go/services/multipooler/internal/connpoolmanager/manager.go
+++ b/go/services/multipooler/internal/connpoolmanager/manager.go
@@ -94,9 +94,14 @@ const (
 //	regularConn, _ := mgr.GetRegularConn(ctx, user)
 //	reservedConn, _ := mgr.NewReservedConn(ctx, settings, user)
 type Manager struct {
-	config     *Config
-	logger     *slog.Logger      // Set by Open()
-	connConfig *ConnectionConfig // Stored for lazy pool creation
+	config *Config
+	logger *slog.Logger // Set by Open()
+
+	// connConfig holds the connection settings, stored once at Open and read
+	// lock-free elsewhere — PgDatabase (off the postgres-monitor loop) and
+	// buildClientConfig (on pool creation). An atomic pointer so those reads never
+	// race the Open write; nil until the first Open.
+	connConfig atomic.Pointer[ConnectionConfig]
 
 	ctx           context.Context          // Manager lifecycle context, used for lazy pool creation
 	adminPool     *admin.Pool              // Shared admin pool for kill operations
@@ -186,7 +191,7 @@ func (m *Manager) Open(ctx context.Context, connConfig *ConnectionConfig) {
 	defer m.createMu.Unlock()
 
 	m.ctx = ctx
-	m.connConfig = connConfig
+	m.connConfig.Store(connConfig)
 	emptyPools := make(map[string]*UserPool)
 	m.userPoolsSnapshot.Store(&emptyPools)
 
@@ -277,16 +282,20 @@ func (m *Manager) Open(ctx context.Context, connConfig *ConnectionConfig) {
 // this manager. They are honored only on TCP connections (libpq parity); the
 // client startup code skips SSLRequest when SocketFile is set.
 func (m *Manager) buildClientConfig(user, password string) *client.Config {
+	// Non-nil on every call: buildClientConfig runs only after Open has stored
+	// connConfig (the admin pool is built later in Open; user pools are built
+	// lazily thereafter).
+	cc := m.connConfig.Load()
 	return &client.Config{
-		SocketFile:     m.connConfig.SocketFile,
-		Host:           m.connConfig.Host,
-		Port:           m.connConfig.Port,
-		Database:       m.connConfig.Database,
+		SocketFile:     cc.SocketFile,
+		Host:           cc.Host,
+		Port:           cc.Port,
+		Database:       cc.Database,
 		User:           user,
 		Password:       password,
-		SSLMode:        m.connConfig.SSLMode,
-		SSLNegotiation: m.connConfig.SSLNegotiation,
-		TLSConfig:      m.connConfig.TLSConfig,
+		SSLMode:        cc.SSLMode,
+		SSLNegotiation: cc.SSLNegotiation,
+		TLSConfig:      cc.TLSConfig,
 		DialTimeout:    m.config.DialTimeout(),
 	}
 }
@@ -532,6 +541,16 @@ func (m *Manager) teardownLocked() {
 // PgUser returns the configured PostgreSQL user for system queries.
 func (m *Manager) PgUser() string {
 	return m.config.PgUser()
+}
+
+// PgDatabase returns the PostgreSQL database the pooler connects to, taken from
+// the connection config supplied at Open. Empty before Open.
+func (m *Manager) PgDatabase() string {
+	cc := m.connConfig.Load()
+	if cc == nil {
+		return ""
+	}
+	return cc.Database
 }
 
 // PgPassword returns the resolved PostgreSQL superuser password and an "ok"
--- a/go/services/multipooler/internal/manager/pg_replication.go
+++ b/go/services/multipooler/internal/manager/pg_replication.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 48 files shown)


go/services/multipooler/internal/manager/pg_replication.go
```diff
@@ -494,6 +495,122 @@ func (pm *MultipoolerManager) resetPrimaryConnInfo(ctx context.Context) error {
...
+// Without it a standby's catalog horizon can outrun a freshly-created failover
+// slot, and the standby's slot-sync worker refuses to persist the synced slot
+// ("synchronization could lead to data loss"). An empty cohort clears it to ”
+// so logical decoding is not stalled when no follower slot exists yet. Applied
+// via ALTER SYSTEM + reload; no-op unless slot-based replication is enabled.
+func (pm *MultipoolerManager) setSynchronizedStandbySlots(ctx context.Context, cohort []*clustermetadatapb.ID) error {
+	if !pm.slotBasedReplicationEnabled() {
+		return nil
+	}
+	slots, err := pm.followerPhysicalSlotNames(cohort)
```
cuongdo, line 582: I think it's necessary to further filter `slots` with something like this query:

```sql
SELECT slot_name FROM pg_replication_slots WHERE slot_name = ANY(slots) AND active
```

Otherwise, an inactive slot (which would be common during a failover) would stall logical decoding on the primary.

go/services/multipooler/internal/connpoolmanager/manager.go
```diff
@@ -534,6 +534,15 @@ func (m *Manager) PgUser() string {
 	return m.config.PgUser()
 }
 
+// PgDatabase returns the PostgreSQL database the pooler connects to, taken from
+// the connection config supplied at Open. Empty before Open.
+func (m *Manager) PgDatabase() string {
+	if m.connConfig == nil {
+		return ""
+	}
+	return m.connConfig.Database
```
cuongdo: should this read be done under lock?

cuongdo (commented):
I took a cursory look. Overall, this is nice. I commented on some things that stood out to me.

mkindahl:
Discovered a bug during testing. A late follower would not have a physical replication slot on the primary, so added a `ReconcileFollower` message that is sent when a change in the "candidate cohort" is discovered.

go/services/multipooler/internal/manager/pg_replication.go
mkindahl: Well, yes, it would stall logical decoding on the primary, but for a good reason. If all slots go inactive even for a short moment, we would send the change to the consumer, and if we have a crash and a failover at that moment, we would have sent the consumer a change that was never acknowledged. Note that if `synchronized_standby_slots` is empty, it sends the change immediately.

Here is the scenario. Servers are A[1], B[1], and C[1]. A is primary, each holds transaction 1 in the WAL, and we assume we've set `synchronized_standby_slots` to contain only active slots.

1. A receives a transaction 2, so A[1,2].
2. There is a brief network hiccup.
3. A[1,2] blocks because neither B[1] nor C[1] has responded.
4. We set `synchronized_standby_slots` to the empty set, because the WAL receivers on the standbys are down.
5. Logical decoding processes transaction 2 and sends the change to the consumer.
6. A[1,2] crashes, or was isolated from the others and is explicitly killed.
7. B[1] is elected primary and C[1] is standby.
8. Transaction 2 was lost, which is fine because it was never acknowledged. It was, however, sent to the consumer, which is wrong.

Now, if `synchronized_standby_slots` is not empty it would work, but making sure the list contains exactly the correct slots is going to be tricky, and I am wondering if there are some corner-cases here if we set the wrong slot name, so I think a safer option right now is to include the entire cohort.

Normally the primary and the standbys are chugging along fine, so there is only a minor delay before the change is sent to the consumer (we are fine with a single ack, but will wait until all ack before we send out the change).

However, if we have a permanent failure, we will reconcile the cohort and modify `synchronous_standby_names` to exclude the failed server, at which point we should also update `synchronized_standby_slots`.

go/services/multipooler/internal/connpoolmanager/manager.go
mkindahl: Fixed.

mkindahl:
## Manual end-to-end validation

Validated the feature (a persistent `failover = true` logical slot surviving a primary failover without a re-seed) end-to-end, found and fixed a bootstrap slot-creation deadlock, and stress-tested it with a repeated-failover loop. **No committed data was lost in any scenario.**

### Environments

- **Local kind** (`demo/k8s`, operator-free) on macOS/Docker Desktop. PostgreSQL 17.
- **Dev EKS**, deployed operator-free with raw manifests (the shared operator currently has no active reconciler), 3 poolers pinned to one node sharing a hostPath backup repo, real Linux nodes / EBS. PostgreSQL 17.

### What was validated

**Bootstrap — 1 primary + 2 standbys, `AT_LEAST_2`:** standbys come up with `wal_level=logical`, `hot_standby_feedback=on`, `sync_replication_slots=on`, `primary_slot_name=mg_<n>`; the primary holds a per-follower physical slot (`mg_1`, `mg_2`) and `synchronized_standby_slots` lists them.

**Slot admission + sync:** `CREATE SUBSCRIPTION … WITH (failover = true)` from an external PG17 subscriber is admitted through the multigateway (the §3.2 guard lift), and the slot reaches `failover_ready` (`synced AND NOT temporary AND invalidation_reason IS NULL`) on both standbys once consumption advances `catalog_xmin`.

**Planned switchover** (`SwitchPrimary`, kind): promotes a slot-ready standby; the subscriber resumes through the gateway with **no re-COPY** and byte-identical data — no committed-data loss.

**Unplanned failover** (`kubectl delete pod` on the primary; kind **and** EKS): multiorch promotes a `failover_ready` standby; the consumer reconnects (client-driven, §3.1.1) and resumes with **no re-COPY**; the crashed old primary is recreated by the StatefulSet and **rejoins as a streaming standby**; source and consumer end byte-identical.

### Bug found and fixed in this PR

On a staggered bootstrap the primary created a per-follower physical slot for only **one** of the two standbys. Root cause is a circular dependency: a follower's slot was created only when it joined the committed cohort, but cohort-add (`isAdditionCandidate`) requires the follower's WAL receiver to be **already streaming**, which needs the slot — so a late-joining standby stranded, looping on no-op `pg_rewind`. This is the ordering constraint now documented in design doc §2.6.

Fixed here by the discovery-driven **`ReconcileFollowers`** reconcile plus a **`FixReplication`** backstop (multipooler + multiorch commits; §2.6). After the fix: fresh bootstrap reaches 1P+2S with both standbys streaming and **no manual slot creation**, and on crash-rejoin the new primary **auto-creates the returning node's physical slot with no deadlock**.

### 10× failover loop (EKS; client and consumer as local Docker containers)

A writer container inserted random rows every 0.1 s through the gateway; a consumer container held a `failover = true` subscription. Loop: kill the primary → wait for a new primary → wait for the old primary to rejoin as a streaming standby — repeated 10 times.

- **All 10 cycles succeeded** — every killed primary was auto-recreated by the StatefulSet and rejoined as a synchronous-quorum streaming standby (10/10 crash-rejoins, no deadlock).
- After stopping the writer and draining, the consumer's table was **byte-identical** to the source: **303 rows, identical row and id-set checksums** — the consumer saw every committed row, no gaps, zero data loss across the 10 failovers.

_Harness note: the containers' gateway `kubectl port-forward` dropped mid-run and was replaced with a self-restarting one; that lowered the total row count but does not affect the exact source↔consumer equality._

cuongdo (approved):

PR #1395, opened 2026-08-19
feat(multipooler): add pg_rewind execution-duration metric

mkindahl:
## Situation

pg_rewind runtime scales with retained `pg_wal` — it copies the whole retained WAL, not just the divergence (~1.8 GB / ~120 WAL segments observed for ~120 bytes of actual divergence) — so a rewind can run for minutes under load. That runtime now matters operationally: it is what can exceed a pod's shutdown grace, and it is what the detached-rewind backstop (`rewindOperationTimeout`) bounds.

But nothing measured the actual pg_rewind runtime. The only rewind metric, `multipooler.rewind.checkpoint_wait.duration`, measures how long a rewind *waited for the leader to become rewind-ready* — not the rewind itself.

## Change

Add `multipooler.rewind.execution.duration`, a histogram of the pg_rewind subprocess runtime, labelled by `phase`:

- `dry_run` — the read-only scan (connect to source, read the timeline history, find the last common checkpoint, scan target WAL); and
- `rewind` — the mutating copy (the `-R` invocation whose runtime is dominated by the WAL it copies).

It is recorded in `runPgRewind` around each pg_rewind invocation, on success and failure, so a slow-then-failed rewind is captured too. pgctld also logs the raw subprocess duration on completion and failure. The metric catalog is regenerated (`make metrics`) so the new series is kept by the Prometheus keep-list.

Follow-up to the interrupted-pg_rewind fix ([MUL-1343](https://linear.app/supabase/issue/MUL-1343/prevent-interrupted-pg-rewind-from-wedging-replicas) / #1393).

## Testing

- Unit: the metric records one data point per phase with durations converted to seconds; nil-receiver / nil-histogram recorder guards.
- `make build` and `make metrics` (catalog regenerated and committed); multipooler manager and pgctld command suites green; golangci-lint clean.


haritabh17 (approved):
LGTM 🚀

PR #1396, opened 2026-08-19
fix(pgprotocol): harden frontend message validation

shedges:
## Problem

Multigateway's pgwire frontend parser diverges from PostgreSQL for several malformed or boundary-case messages. The affected paths can accept invalid input, allocate from an untrusted count or frame length, report an error at the wrong protocol stage, or put the client and gateway into different extended-query states.

- **Signed and unvalidated counts:** `Parse` and `Bind` read wire counts as signed 16-bit values and allocate count-sized slices before proving that the corresponding entries fit in the received body. A valid count above 32767 is misread, while a tiny malformed message can claim a much larger count and induce disproportionate allocation.
- **Trailing and invalid fields:** `Parse`, `Bind`, `Execute`, `Describe`, and `Close` do not consistently require exact body consumption. Unsupported Describe/Close target discriminators and extra bytes can therefore reach normal processing even though PostgreSQL rejects them as protocol violations.
- **Malformed empty-body messages:** `Sync` and `Flush` must have empty bodies, but non-empty variants are not consistently rejected or consumed. Leftover bytes can be interpreted as another frontend message and extended-query recovery can diverge from PostgreSQL.
- **Late Bind format validation:** parameter and result format codes other than text (`0`) or binary (`1`) can survive Bind parsing and fail later. PostgreSQL rejects the unsupported value during Bind, before portal execution.
- **Incorrect extended-query recovery:** parsing failures can escape as internal Multigres errors or make the connection ready too early. PostgreSQL sends the appropriate error, ignores extended-query messages until `Sync`, and only then sends `ReadyForQuery`.
- **Oversized frontend allocation:** the client-controlled frame length is used to obtain a body buffer without first applying PostgreSQL's large-message limit. A client can declare a near-1 GiB body without sending it and make the gateway attempt the allocation.
- **Oversized-frame response divergence:** after detecting an oversized frame, the normal gateway error path sends an internal error and `ReadyForQuery`. PostgreSQL silently closes because the stream cannot safely recover from a framing-level failure.

## Fix

**1. Parse structured frontend messages strictly.** `Parse`, `Bind`, `Execute`, `Describe`, and `Close` now validate every required field and require exact body consumption. Describe/Close accept only statement (`S`) and portal (`P`) targets, and Bind validates text/binary format codes at Bind time.

**2. Validate counts before allocation.** Protocol counts are decoded as unsigned 16-bit values and checked against the bytes remaining in the message before allocating their slices. This preserves the full wire range and prevents a small truncated body from driving a disproportionate count allocation.

**3. Match PostgreSQL extended-query recovery.** Malformed extended-query messages produce SQLSTATE `08P01` and enter the existing drain-until-Sync state. Non-empty Sync and Flush bodies are consumed and handled at the appropriate recovery boundary; invalid Bind formats produce SQLSTATE `22023` during Bind.

**4. Reject oversized frames before body allocation.** `ReadMessageLength` applies PostgreSQL's `PQ_LARGE_MESSAGE_LIMIT`-equivalent body limit before consulting the buffer pool or allocating a body. The oversized condition is distinguishable from ordinary handler errors so the server discards buffered output and closes silently, matching PostgreSQL.

## Design note: preserve the existing protocol state machine

The fix reuses the existing extended-query error-drain mechanism rather than introducing a second recovery path. Parsing handlers emit the protocol error and mark the batch failed; the connection continues consuming frontend messages until a valid Sync restores normal processing. Oversized frames are intentionally separate because their declared boundary is not trustworthy and the stream cannot be resynchronized safely.

## Tests

- Unit coverage for exact message consumption, invalid target discriminators, non-empty Sync/Flush bodies, malformed counts, Bind format validation, and oversized-frame rejection before allocation.
- A valid Parse message with 32768 parameter OIDs verifies unsigned count handling.
- PostgreSQL differential end-to-end coverage compares raw wire traces for 11 cases: oversized frame, truncated high Parse count, Parse/Bind/Execute/Close trailing bytes, invalid Describe/Close targets, non-empty Sync/Flush, and invalid Bind format.
- `go test -count=1 ./go/common/pgprotocol/... ./go/common/mterrors`
- `make build`
- `PATH=/Users/shedges/workspace/multigres/bin:$PATH go test -count=1 -run='^TestFrontendValidationMatchesPostgres$' -v ./go/test/endtoend/queryserving`


haritabh17 (approved):
LGTM 🚀

PR #1397, opened 2026-08-20
fix(pgctld): stop child reaper from stealing subprocess exit status

mkindahl:
## Summary

After the change that stopped pgctld from cancelling `pg_rewind`, a production cluster ran much longer but eventually wedged with two writers on the same timeline — an unrecoverable split, because `pg_rewind` reports "no rewind required" for same-timeline peers. The trigger was pgctld's SIGCHLD reaper.

pgctld runs a reaper (only under PID 1, container init) to reap the postmaster, which `pg_ctl start -W` forks and then leaves reparented to pgctld. That reaper reaped with `Wait4(-1)`, which races os/exec's own `Wait` for every other subprocess pgctld runs (`pg_ctl`, `initdb`, `pg_rewind`, `pg_isready`, `psql`, `postgres --single`). Both waiters call `wait4` on the same child; whichever wins consumes the exit status. When the reaper won, os/exec's `Wait` saw `ECHILD` ("waitid: no child processes") and reported a failure for a command that had actually succeeded.

That false failure is what wedged the cluster: a `pg_ctl start` that in fact succeeded was reported as failed, so a redundant `Start` RPC treated the node as needing crash recovery, removed `standby.signal`, and single-user crash-recovered a still-live postmaster into a second read-write server on the primary's timeline.

## Before

```mermaid
sequenceDiagram
    autonumber
    participant O as multiorch
    participant P as pgctld (PID 1)
    participant C as pg_ctl
    participant R as SIGCHLD reaper
    participant M as postmaster

    O->>P: Start (allow_crash_recovery)
    P->>C: exec pg_ctl start -W  (cmd.Run → Wait)
    C->>M: fork postmaster, then exit
    Note over M,P: postmaster reparented to pgctld (PID 1)
    R-->>C: Wait4(-1) reaps pg_ctl's status first  (RACE)
    Note over P: cmd.Wait() → ECHILD "waitid: no child processes"
    P-->>O: Start FAILED  (but postgres is actually up)
    O->>P: Start (retry)
    Note over P: not cleanly stopped + standby.signal present
    P->>M: remove standby.signal, run single-user recovery
    Note over M: mid-startup postmaster comes up READ-WRITE on the<br/>old timeline → two writers on one TLI (pg_rewind can't fix)
```

## After

The reaper now waits only on the specific postmaster PIDs registered after a successful `Start`/`Restart`, and never calls `Wait4(-1)` — so it can never touch an os/exec-owned child, and os/exec always observes the true exit status.

```mermaid
sequenceDiagram
    autonumber
    participant O as multiorch
    participant P as pgctld (PID 1)
    participant C as pg_ctl
    participant R as childReaper
    participant M as postmaster

    O->>P: Start (allow_crash_recovery)
    P->>C: exec pg_ctl start -W  (cmd.Run → Wait)
    C->>M: fork postmaster, then exit
    Note over M,P: postmaster reparented to pgctld (PID 1)
    Note over R: reaper waits only on tracked postmaster PIDs, never Wait4(-1)
    C-->>P: cmd.Wait() returns pg_ctl's true exit status
    P->>R: TrackPID(postmaster)
    P-->>O: Start OK
    O->>P: Start (retry / duplicate)
    Note over P: guard 2 — postmaster already running → skip crash recovery
    Note over P: guard 3 — never remove standby.signal while a postmaster is live
```

## Changes

- **Root fix** (`reaper.go`): replace the `Wait4(-1)` SIGCHLD reaper with a `childReaper` that tracks exact postmaster PIDs and reaps only those. It still runs only under PID 1, and is now started from `OnRun` alongside the other background workers.
- **Defense in depth — guard 2** (`server.go`): a `Start` with `allow_crash_recovery` skips the crash-recovery branch when a postmaster is already running. A duplicate `Start` against a live node is an idempotent no-op instead of a destructive recovery.
- **Defense in depth — guard 3** (`crash_recovery.go`): `runCrashRecovery` refuses to remove `standby.signal` while a postmaster is live, closing the residual race where a postmaster appears between guard 2's check and the removal. The transient absence of `standby.signal` is what let a mid-startup postmaster finish recovery as a primary.

Each of the three cuts a different link in the chain, so any one of them alone breaks the cascade.

## Testing

- New `reaper_test.go`, including `TestChildReaper_DoesNotStealOsExecChildren` — a regression guard that runs many concurrent os/exec children with the reaper active. It fails intermittently against the old `Wait4(-1)` reaper and passes deterministically with the scoped reaper.
- New `TestRunCrashRecoveryInDir_SkipsWhenPostgresRunning` covering guard 3.
- `go test ./go/cmd/pgctld/command` passes with `-race`; `golangci-lint` clean.

## Follow-up

A timeline/consensus-term backstop (refuse to come up read-write on an already-claimed timeline) is the deeper invariant but a larger change; tracked separately rather than folded in here.

Fixes MUL-1370

### DIFF
--- a/go/cmd/pgctld/command/crash_recovery.go
+++ b/go/cmd/pgctld/command/crash_recovery.go
@@ -102,6 +102,20 @@ func (s *PgCtldService) runCrashRecoveryInDir(
 	r *retry.Retry,
 ) error {
 	logger := s.logger
+	dataDir := s.pgConfig.PostgresDataDir
+
+	// Never disturb a live postmaster. The standby.signal removal below opens a
+	// window in which the file is absent; a postmaster that is running — or still
+	// mid-startup after a Start whose success was misreported — can observe that
+	// absence and finish recovery as a primary on a timeline it must not claim.
+	// Callers gate this path on a stopped node; this is the last-line guard for
+	// the residual race where a postmaster appears between that check and here.
+	// A running node needs no crash recovery, so skipping is also the correct no-op.
+	if isPostgreSQLRunning(dataDir) {
+		logger.InfoContext(ctx, "skipping crash recovery: postgres is running", "data_dir", dataDir)
+		return nil
+	}
+
 	signalPath := s.standbySignalPath()
 	if _, err := os.Stat(signalPath); err == nil {
 		logger.InfoContext(ctx, "temporarily removing standby.signal for single-user crash recovery",
--- a/go/cmd/pgctld/command/crash_recovery_test.go
+++ b/go/cmd/pgctld/command/crash_recovery_test.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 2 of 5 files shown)


haritabh17 (approved):
Nice!! 🚀

PR #1398, opened 2026-08-20
fix(multiorch): make cohort-reconcile's CAS match its safety check

dweitzman:
ReconcileCohortAction independently re-derived "the current rule" for both its removal-safety check and its `UpdateConsensusRule` CAS — a second read that could disagree with what the engine's pre-execution recheck already verified.

- Bundles the redetected problem and its verified rule into one `RecheckedProblem` value threaded through `RecoveryAction.Execute`, so a mismatched pair is a compile error.
- Drops the now-redundant safety re-check (the analyzer already proves it against that exact rule).
- Makes rejection of an undecided/in-flight rule proposal explicit instead of an incidental side effect of the leader-liveness check.

### DIFF
--- a/go/services/multiorch/recovery/actions/reconcile_cohort.go
+++ b/go/services/multiorch/recovery/actions/reconcile_cohort.go
@@ -75,8 +75,20 @@ func NewReconcileCohortAction(
 	}
 }
 
-// Execute applies the cohort change on the shard leader.
-func (a *ReconcileCohortAction) Execute(ctx context.Context, problem types.Problem) error {
+// Execute applies the cohort change on the shard leader. rechecked.HighestKnownRule
+// is the exact position the engine's recheck just re-verified this problem
+// against (recovery_loop.go's recheckProblem re-runs
+// CohortMismatchAnalyzer.Analyze against it). Because that analyzer only
+// emits ProblemCohortMemberIneligible when IsCohortMemberRemovalSafe already
+// holds for this same rule — a pure, deterministic function of (rule,
+// targetID) — a REMOVE problem surviving the recheck already proves removal
+// is safe against the rule; Execute does not re-verify it. CAS-ing on the
+// same rule (ExpectedOutgoingRule below) rather than a fresh, independent
+// store read means the mutation can't apply against any rule other than the
+// one just proven safe.
+func (a *ReconcileCohortAction) Execute(ctx context.Context, rechecked types.RecheckedProblem) error {
+	problem := rechecked.Problem
+	rule := rechecked.HighestKnownRule
 	a.logger.InfoContext(ctx, "executing reconcile cohort action",
 		"shard_key", problem.ShardKey.String(),
 		"pooler", problem.PoolerID.Name,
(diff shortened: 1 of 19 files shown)


go/services/multiorch/recovery/actions/reconcile_cohort.go
```diff
@@ -75,8 +75,20 @@ func NewReconcileCohortAction(
...
+// against (recovery_loop.go's recheckProblem re-runs
+// CohortMismatchAnalyzer.Analyze against it). Because that analyzer only
+// emits ProblemCohortMemberIneligible when IsCohortMemberRemovalSafe already
+// holds for this same rule — a pure, deterministic function of (rule,
+// targetID) — a REMOVE problem surviving the recheck already proves removal
+// is safe against the rule; Execute does not re-verify it. CAS-ing on the
+// same rule (ExpectedOutgoingRule below) rather than a fresh, independent
+// store read means the mutation can't apply against any rule other than the
+// one just proven safe.
+func (a *ReconcileCohortAction) Execute(ctx context.Context, rechecked types.RecheckedProblem) error {
```
dweitzman, line 89: The main thing here is making sure the action knows more information about what the highest known rule was at the time this problem was detected

Previously you could imagine a word where new health information causes the highest known rule to change between the problem detection in the analyzer and fixing in here in the action, making the compare-and-swap on the highest known decision below misleading.

We want to compare-and-swap relative to the rule at the time analysis happened, which is what the code here does now

go/services/multiorch/recovery/actions/reconcile_cohort.go
```diff
@@ -134,7 +165,7 @@ func (a *ReconcileCohortAction) Execute(ctx context.Context, problem types.Probl
 	req := &multipoolermanagerdatapb.UpdateConsensusRuleRequest{
 		Operation:            op,
 		StandbyIds:           []*clustermetadatapb.ID{targetID},
-		ExpectedOutgoingRule: members.HighestKnownPosition.GetDecision().GetRuleNumber(),
+		ExpectedOutgoingRule: rule.GetDecision().GetRuleNumber(),
```
dweitzman, line 168: The is the compare-and-swap in question

mkindahl (approved):
LGTM.

This is a failure scenario before:

```mermaid
sequenceDiagram
    participant EN as Engine cycle
    participant EXB as attemptRecovery B
    participant EXC as attemptRecovery C
    participant ST as Pooler cache
    participant LD as Leader multipooler
    Note over EN,LD: cohort {leader, A, B, C} at rule R<br/>B and C both signal INELIGIBLE
    EN->>EXB: dispatch REMOVE B in goroutine
    EN->>EXC: dispatch REMOVE C in goroutine
    Note over EXB,EXC: distinct poolers run in parallel
    EXB->>ST: recheck - analyze snapshot
    ST-->>EXB: REMOVE B re-detected at R
    EXC->>ST: recheck - analyze snapshot
    ST-->>EXC: REMOVE C re-detected at R
    EXC->>LD: UpdateConsensusRule REMOVE C - expected R
    LD-->>EXC: CAS ok - leader rule advances R to R2
    LD->>ST: health stream frame ConsensusStatus R2
    Note over ST: THIS advances the cache rule<br/>HighestPosition now R2, cohort {leader, A, B}
    EXB->>ST: Execute second read FindShardMembers
    ST-->>EXB: HighestKnownPosition R2 - not the rechecked R
    EXB->>LD: UpdateConsensusRule REMOVE B - expected R2
    LD-->>EXB: CAS ok - leader is at R2
    Note over LD: B removed against R2, never safety-checked<br/>under R2 recruited is {A}, below N=2<br/>cohort {leader, A} cannot survive next leader loss
```

This is how it looks after your fix:

```mermaid
sequenceDiagram
    participant EN as Engine cycle
    participant EXB as attemptRecovery B
    participant EXC as attemptRecovery C
    participant ST as Pooler cache
    participant LD as Leader multipooler
    Note over EN,LD: cohort {leader, A, B, C} at rule R
    EN->>EXB: dispatch REMOVE B in goroutine
    EN->>EXC: dispatch REMOVE C in goroutine
    EXB->>ST: recheck - analyze snapshot
    Note over ST: gate IsCohortMemberRemovalSafe(R, B) holds
    ST-->>EXB: RecheckedProblem B, HighestKnownRule R
    EXC->>LD: UpdateConsensusRule REMOVE C - expected R
    LD-->>EXC: CAS ok - leader rule advances R to R2
    LD->>ST: health stream frame ConsensusStatus R2
    Note over EXB: Execute anchors on rechecked R<br/>no second read of the cache
    EXB->>LD: UpdateConsensusRule REMOVE B - expected R
    LD-->>EXB: CAS fails - leader is at R2 not R
    Note over EXB,LD: rejected - retry next cycle<br/>recheck re-runs the gate against R2<br/>removing B now unsafe so no REMOVE is emitted
```

PR #1400, opened 2026-08-21
fix(multipooler): re-resign on failed promote; avoid sync-commit stall on rule-CAS read

dweitzman:
- Re-resign as leader if `Promote` fails after the promotion hook already cleared it — previously left the node stuck, neither leader nor signaling for replacement.
- `readCurrentRule`'s `FOR UPDATE NOWAIT` precheck ran under implicit auto-commit, so its row lock could stall on `synchronous_commit`. Fixed via `BeginAdmin` + explicit rollback (only `COMMIT` waits, never `ROLLBACK`).
- Added `BeginAdmin` (admin-pool transactions), sharing implementation with the existing `Begin` via a new `genericTx`/`txConn` abstraction.
- Split `admin.Conn`'s retry-capable stateless methods from its no-retry transactional path (`TxConn`), so the two can't be mixed up on one type.

### DIFF
--- a/go/services/multipooler/internal/executor/admin_query.go
+++ b/go/services/multipooler/internal/executor/admin_query.go
@@ -16,8 +16,10 @@ package executor
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/multigres/multigres/go/common/sqltypes"
+	"github.com/multigres/multigres/go/services/multipooler/internal/pools/admin"
 )
 
 // The QueryAdmin* methods are the InternalQueryService access path for the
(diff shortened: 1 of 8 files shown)


simonbtomlinson (approved):
Looks good to me @dweitzman !

I verified the behavior you describe around offline sync standbys and `SELECT FOR UPDATE NOWAIT`. Wrapping it in a transaction and rolling back fixes it as you describe.

The other changes make sense to me as well.

GuptaManan100 (approved):

PR #1403, opened 2026-08-24
feat(multiadmin): add only_reachable filter to GetGateways

haritabh17:
## Problem

After gateway pod churn, gateway topo records left behind by dead pods have no liveness signal, so `GET /api/v1/gateways` returns stale registrations alongside live gateways (multiple records per cell sharing a replicaset hash with different pod suffixes). The UI cannot filter them, and its gateway count reflects every registered entry.

Fixes MUL-1368.

## Change

Adds an `only_reachable` flag to `GetGatewaysRequest` (`GET /api/v1/gateways?onlyReachable=true`). When set, multiadmin probes each gateway's `hostname:port_map["grpc"]` with a concurrent 1s TCP dial and drops non-responders. Records without a grpc port cannot be probed and are treated as stale.

- Response shape is unchanged — existing clients are unaffected.
- Probes run concurrently, so the request waits at most ~one probe timeout regardless of gateway count.
- Regenerated Go pb and web TS client, so the UI can pass the flag directly.

## Scope note

This is an interim, UI-facing filter. The root cause (no lease on gateway records + best-effort unregister) is addressed by lease-backed registration in MUL-1311 (#1376); the probing here is marked for deletion once records self-expire. Filtering the response also does nothing for the other stale-record side effects (held PID prefixes, cancel routing) — those are only fixed by MUL-1311.

## Testing

- New unit test `TestMultiadminServerGetGatewaysOnlyReachable`: live listener kept, refused port dropped, missing grpc port dropped, unfiltered behavior unchanged.
- `go build ./go/...` and full multiadmin package tests pass.

### DIFF
--- a/go/pb/multiadmin/multiadminservice.pb.go
+++ b/go/pb/multiadmin/multiadminservice.pb.go
@@ -552,7 +552,10 @@ func (x *GetDatabaseNamesResponse) GetNames() []string {
 type GetGatewaysRequest struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
 	// cells is a comma-separated list of cell names to filter by (optional)
-	Cells         []string `protobuf:"bytes,1,rep,name=cells,proto3" json:"cells,omitempty"`
+	Cells []string `protobuf:"bytes,1,rep,name=cells,proto3" json:"cells,omitempty"`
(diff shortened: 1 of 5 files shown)


rafael (approved):
LGTM - If I recall correctly we also had a similar flag in vitess

PR #1405, opened 2026-08-24
test(pgregress): accept stats visibility-lag divergence under pooling

haritabh17:
## Summary

The `stats` regression test has been a persistent residual failure in the nightly compatibility runs (two hunks beyond the existing `stats.patch`). This records them as accepted divergences, with the mechanism documented in the patch header.

## Root cause

Both hunks are cumulative-stats **visibility lag**, not wrong accounting:

- Since PG 15, each backend accumulates pending stats and flushes them to shared memory at idle, throttled to once per second; `pg_stat_force_next_flush()` only arms the flush flag **in the backend that executes it**.
- The test's explicit `BEGIN … TRUNCATE … COMMIT/ROLLBACK` transactions run on reserved backends, and the `enable_bitmapscan=off` indexscan runs on a different settings-bucket backend. When the test calls `pg_stat_force_next_flush()` and immediately reads `pg_stat_user_tables`, those backends are idle in the pool with their counters still unflushed — they arrive ~1s after the assertions have already run.
- The per-transaction truncate/subxact stats semantics are computed **correctly** (each transaction is pinned to a single backend); only the read-after-write timing diverges. PgBouncer behaves identically. This matches the class of divergence `stats.patch` already accepts ("assumes one physical backend per psql session").

## Why a patch (and not a product change)

The diff is deterministic — byte-identical across the last two scheduled CI runs (32442932057, 32686056557) — so the patch holds. A future product fix that would make the upstream expectations pass for real is fanning `pg_stat_force_next_flush()` out to idle pooled backends; that is noted in the patch header.

## Verification

The two new hunks are net-zero line deltas, merged ahead of the existing hunks (whose offsets are unchanged). Verified mechanically with the harness's own pipeline: reconstructed the CI run's actual output (stock REL_17_6 `expected/stats.out` → old patch → CI residual diff) and ran `suiteutil.VerifyPatch` against the merged patch — applies cleanly and matches byte-for-byte.

Not touched here: the remaining residual failures have their own tracks — `brin*`/`explain`/`partition_prune` are addressed by #1352 and #1302.

### DIFF
--- a/go/test/endtoend/pgregresstest/testdata/pg17/patches/stats.patch
+++ b/go/test/endtoend/pgregresstest/testdata/pg17/patches/stats.patch
@@ -4,8 +4,44 @@
 # statistics differ. Backend-selection-sensitive counters are normalized by the
 # pg_regress verifier; SQL still runs through multigateway rather than directly
 # against PostgreSQL.
+#
+# The trunc_stats_test*/tenk2 hunks record cumulative-stats VISIBILITY lag, not
+# wrong accounting: since PG 15 each backend holds pending stats and flushes
+# them at idle, throttled to once per second, and pg_stat_force_next_flush()
+# only arms the one backend that executes it. Under pooling the explicit
+# transactions run on reserved backends and the enable_bitmapscan=off indexscan
+# runs on a different settings-bucket backend, so their pending counters are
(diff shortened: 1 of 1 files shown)


shedges (approved):
LGTM 🚀

PR #1407, opened 2026-08-25
test(pgregress): record temporary function divergence in explain

haritabh17:
## Summary

- record the intentional rejection of schema-qualified temporary functions under connection pooling
- accept the resulting `pg_temp.mysin` lookup failure in the `explain` regression fixture
- restore the orphaned failover-slot cleanup call after manager control-plane queries moved to the admin pool
- isolate the fixture update from the larger query-serving changes in #1302 and #1352

## Testing

- `go test -short ./go/services/multipooler/internal/manager -count=1`
- the extracted `explain` hunk was exercised by the PostgreSQL compatibility run on #1302
- `explain` on `main` still has the separate prepared-statement residual addressed by #1302


shedges (approved):
LGTM 🚀

PR #1408, opened 2026-08-25
docs(sharding): add database-owned shards design

GuptaManan100:
## Summary

- Adds the sharding design doc: databases split into shards, shards grouped into table groups, each table group owned by one database, each shard served by one pooler cohort.
- The sharding function lives on each `Table` (not the `TableGroup`), so one group can mix functions over the same shards — e.g. co-locating a secondary index with the table it routes for.
- Reuses existing types: resolution produces `clustermetadata.ShardKey`, `query.Target` already carries it, and `Shard.key_range` reuses `clustermetadata.KeyRange`, so no new wire types.

## Description

Doc-only. Defines the model (`Database` / `TableGroup` / `Table` / `Shard` / `ShardingFunction`), the two-step `value → keyrange-id → shard` resolution, the illustrative proto shape, and worked examples (unsharded, hash, co-location, reference, range, resharding, co-located secondary index).

Deliberately out of scope: DDL syntax, resharding execution, cross-shard 2PC, and where the catalog is stored (topology vs postgres) — so no proto or tables are committed yet.

### DIFF
--- a/docs/sharding/design.md
+++ b/docs/sharding/design.md
@@ -1,10 +1,10 @@
 # Sharding: database-owned shards
 
-A database's data is split into **shards**. Shards are grouped into **table
-groups**; each table group belongs to one **database**, and each shard belongs
-to one table group. A shard is served by one **pooler cohort** (postgres + its
-multipoolers). Ownership runs top to bottom — a table group, its shards, and
-their poolers all belong to a single database.
+A database's data is split into **shards**. Each shard belongs to one **table
+group**, and each table group belongs to one **database**. A shard is served by
+one **pooler cohort** (postgres + its multipoolers). Ownership runs top to
+bottom — a table group, its shards, and their poolers all belong to a single
+database.
 
 ```text
 database ─┬─ table group ─┬─ shard ── pooler cohort (postgres + multipoolers)
@@ -15,7 +15,7 @@ database ─┬─ table group ─┬─ shard ── pooler cohort (postgres + 
 Placement is structural: a table's physical location is determined by its
 position in the `database → table group → shard` tree, and a pooler holds
 exactly one database's shard. Each pooler carries its identity as
-`{database, table_group, shard, key_range}`.
+`{database, table_group, shard}`.
 
 ## Terms
 
@@ -25,12 +25,14 @@ groups, shards, and the poolers serving them belong to it alone.
 **TableGroup** — a co-location unit within one database: a set of member tables
 that share one set of shards (key ranges tiling the keyrange-id space). Tables in
 the same group co-locate — they can be joined and transacted locally. The group
-defines the shards; each member table brings its own sharding function.
+defines the shards; each member table brings its own sharding function — keeping
+it per-table lets one group host tables keyed differently on the same shards
+(e.g. a table and its differently-keyed lookup index).
 
 **Table** — a member of a group, with a sharding function and the shard-key
 columns that function is evaluated on. Because the function is per-table, one
 group can mix functions over the same shards: a table sharded by
-`hash(customer_id)` can sit beside the `hash(invoice_id)` secondary index that
+`hash(customer_id)` can sit beside the `hash(invoice_id)` lookup index that
 routes lookups against it, both on the same shards and pooler cohorts.
 
 **Shard** — one key-range slice of a table group, served by a pooler cohort
@@ -44,11 +46,12 @@ key_range)`; its name is derived from the range as `hex(start)-hex(end)` — so
   scatters.
 - **range** — order-preserving; a range scan touches only the overlapping
   shards.
-- **lookup** — resolves the keyrange-id through a secondary-index table, which
+- **lookup** — resolves the keyrange-id through a lookup-index table, which
   can itself be a member of the same table group.
 - **reference** — owns no shards; the table is replicated onto every shard in the
-  database, so it joins locally from anywhere. (Unset function ⇒ unsharded: the
-  group has one full-range shard.)
+  database, so it joins locally from anywhere.
+
+Unset function ⇒ unsharded: the group has one full-range shard.
 
 ## The two functions
 
@@ -115,7 +118,7 @@ message RangeFunction {}
 message ReferenceFunction {}
 message LookupFunction {
   string schema = 1;
-  string table  = 2;   // secondary-index table in this database (often a group member)
+  string table  = 2;   // lookup-index table in this database (often a group member)
   string from_column = 3;
   string to_column   = 4;
 }
@@ -132,7 +135,7 @@ new wire types.
 Keyranges are `hex(start)-hex(end)`: `-80` is `[0x00, 0x80)`, `80-` is
 `[0x80, +inf)`.
 
-### Unsharded database
+### Unsharded table
 
 One table group, one full-range shard; every query routes there.
 
@@ -212,7 +215,7 @@ the ranges.
 ]}
 ```
 
-### Secondary index co-located
+### Lookup index co-located
 
 A `sales` group holds three tables that route on different columns but share one
 set of shards:
@@ -228,7 +231,7 @@ TableGroup{ name: "sales", shards: [-80, 80-], tables: [
 ```
 
 `invoice` is hashed on `customer_id`; `invoice_line` routes through a lookup;
-`invoice_customer_idx` — the secondary index backing that lookup — is hashed on
+`invoice_customer_idx` — the lookup index backing that lookup — is hashed on
 `invoice_id`. All three are members of `sales`, so they share its shards and
 pooler cohorts.
 
--- a/docs/sharding/design.md
+++ b/docs/sharding/design.md
@@ -1,0 +1,264 @@
+# Sharding: database-owned shards
+
+A database's data is split into **shards**. Shards are grouped into **table
+groups**; each table group belongs to one **database**, and each shard belongs
+to one table group. A shard is served by one **pooler cohort** (postgres + its
+multipoolers). Ownership runs top to bottom — a table group, its shards, and
+their poolers all belong to a single database.
+
+```text
+database ─┬─ table group ─┬─ shard ── pooler cohort (postgres + multipoolers)
+          │               └─ shard ── pooler cohort
+          └─ table group ─── shard ── pooler cohort
+```
+
+Placement is structural: a table's physical location is determined by its
+position in the `database → table group → shard` tree, and a pooler holds
+exactly one database's shard. Each pooler carries its identity as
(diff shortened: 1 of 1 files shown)


docs/sharding/design.md
```diff
@@ -0,0 +1,264 @@
+# Sharding: database-owned shards
+
+A database's data is split into **shards**. Shards are grouped into **table
```
haritabh17: > Shards are grouped into **table groups**

Very minor nit / totally fine this was too: Will removing this bit improve clarity? Then we are drawing inference only in one direction, currently its saying the same thing from 2 directions 😃

docs/sharding/design.md
```diff
@@ -0,0 +1,264 @@
...
+```text
+database ─┬─ table group ─┬─ shard ── pooler cohort (postgres + multipoolers)
+          │               └─ shard ── pooler cohort
+          └─ table group ─── shard ── pooler cohort
+```
+
+Placement is structural: a table's physical location is determined by its
+position in the `database → table group → shard` tree, and a pooler holds
+exactly one database's shard. Each pooler carries its identity as
+`{database, table_group, shard, key_range}`.
```
haritabh17: > shard, key_range}

`key_range` introduced here for the first time.

docs/sharding/design.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Terms
+
+**Database** — the top of the hierarchy and the ownership boundary. Its table
+groups, shards, and the poolers serving them belong to it alone.
+
+**TableGroup** — a co-location unit within one database: a set of member tables
+that share one set of shards (key ranges tiling the keyrange-id space). Tables in
+the same group co-locate — they can be joined and transacted locally. The group
+defines the shards; each member table brings its own sharding function.
```
haritabh17: Do we want to specify here why sharding functions are associated with tables and not table groups?

docs/sharding/design.md
```diff
@@ -0,0 +1,264 @@
...
+
+- **hash** — even distribution; a point lookup on the key hits one shard, a scan
+  scatters.
+- **range** — order-preserving; a range scan touches only the overlapping
+  shards.
+- **lookup** — resolves the keyrange-id through a secondary-index table, which
+  can itself be a member of the same table group.
+- **reference** — owns no shards; the table is replicated onto every shard in the
+  database, so it joins locally from anywhere. (Unset function ⇒ unsharded: the
+  group has one full-range shard.)
```
haritabh17: > (Unset function ⇒ unsharded: the group has one full-range shard.)

Note as a separate point??

docs/sharding/design.md
```diff
@@ -0,0 +1,264 @@
...
+identifier every `Multipooler` already advertises (`Multipooler.shard_key`) and
+`query.Target` already carries — so a resolved target routes to poolers with no
+new wire types.
+
+## Examples
+
+Keyranges are `hex(start)-hex(end)`: `-80` is `[0x00, 0x80)`, `80-` is
+`[0x80, +inf)`.
+
+### Unsharded database
```
haritabh17: Unsharded _table_?

haritabh17 (approved):
Rest looks good 🚀

docs/sharding/design.md
```diff
@@ -0,0 +1,264 @@
...
+```diff
+ TableGroup{ name: "by_user", shards: [
+-  {key_range: -80},
++  {key_range: -40},
++  {key_range: 40-80},
+   {key_range: 80-},
+]}
+```
+
+### Secondary index co-located
```
haritabh17: Lets call it `Lookup Seconday index co-located` ?

haritabh17: Lets call it `Lookup index co-located` ?

docs/sharding/design.md
GuptaManan100: It is an usharded database, with unsharded tables. Could be more than 1.

docs/sharding/design.md
```diff
@@ -0,0 +1,267 @@
+# Sharding: database-owned shards
+
+A database's data is split into **shards**. Each shard belongs to one **table
+group**, and each table group belongs to one **database**. A shard is served by
+one **pooler cohort** (postgres + its multipoolers). Ownership runs top to
+bottom — a table group, its shards, and their poolers all belong to a single
+database.
```
mkindahl, line 7: It is not clear what the difference is between a shard and a pooler cohort. They seem to be identical.

docs/sharding/design.md
```diff
@@ -0,0 +1,267 @@
...
+Unset function ⇒ unsharded: the group has one full-range shard.
+
+## The two functions
+
+Two composable steps map a query to a shard:
+
+| Step                  | Owned by                                           | Answers                                                   |
+| --------------------- | -------------------------------------------------- | --------------------------------------------------------- |
+| `value → keyrange-id` | the **table's** function + the **table's** columns | how a row's shard-key value becomes a point in `[00, FF]` |
+| `keyrange-id → shard` | the **table group's** shard key ranges             | which shard owns that point                               |
```
mkindahl: Maybe I am misunderstanding, but the mapping from value to keyrange-id seems to suggest that there is a maximum of 256 shards (0x00-0xFF). This also hint at hash-based sharding, but that is not explicitly said.

Would it not make sense to support general range sharding? The comment above seems to suggest that, but the mapping here does not. Maybe I am misunderstanding what a key range-id is.

Hmm... reading further below it seems this is an integer range that you map all sharding keys to (which is fine, we only ever expect to represent keys that have a total order). Isn't this going to break down quite quickly if a user wants to, for example, shard a table on day of the year? There are 365 days in a year, and it seems like a reasonable straightforward mapping that a user might request.

docs/sharding/design.md
```diff
@@ -0,0 +1,267 @@
...
+**ShardingFunction** — how a table's rows are placed:
+
+- **hash** — even distribution; a point lookup on the key hits one shard, a scan
+  scatters.
+- **range** — order-preserving; a range scan touches only the overlapping
+  shards.
+- **lookup** — resolves the keyrange-id through a lookup-index table, which
+  can itself be a member of the same table group.
+- **reference** — owns no shards; the table is replicated onto every shard in the
+  database, so it joins locally from anywhere.
```
mkindahl: I would call it something like "global" or "unsharded". Reference tables we usually are familiar with, but this will be read as a "reference to some other table", i.e., an alias.

docs/sharding/design.md
```diff
@@ -0,0 +1,267 @@
...
+`[0x80, +inf)`.
+
+### Unsharded table
+
+One table group, one full-range shard; every query routes there.
+
+```text
+Database{ name: "db2", table_groups: [
+  TableGroup{ name: "default",
+              tables: [{schema:"sc2", name:"t3"}],   // no function => unsharded
```
mkindahl, line 189: It is not clear what the difference between "unsharded" and "reference" is. Both are "global" in the sense that they exist on every shard, right?

docs/sharding/design.md
```diff
@@ -0,0 +1,267 @@
...
+`t1` and `t2` in the same group on the same key. For any `user_id` both rows
+share a keyrange-id, so `t1 JOIN t2 ON user_id` runs within one shard.
+Co-location is group membership.
+
+```text
+tables: [ {schema:"sc1", name:"t1", function:{hash:{}}, columns:["user_id"]},
+          {schema:"sc1", name:"t2", function:{hash:{}}, columns:["user_id"]} ]
+```
+
+### Reference table
```
mkindahl, line 240: Although I know this is a common name, it sounds like an "alias table" to me: a table that references another table, which is not what it is.

This is a minor comment, no need to change since it is an established term. Just want to point out the risk of confusion.

docs/sharding/design.md
GuptaManan100: They map one-to-one, but logically mean different things, but yes they are the same. Would it be better if we only used one term. the shard is the logical slice of a database. While the pooler
cohort is the physical processes (postgres + its multipoolers) that serve it. I'll add a clarification.

docs/sharding/design.md
GuptaManan100: I don't mind renaming to a different term, but Vitess used `Reference` table nomenclature. and `unsharded` already means something else.

docs/sharding/design.md
GuptaManan100: They're similar but not quite the same, so I've spelled it out in the **reference** bullet. An unsharded table has **one physical copy**, on its group's single full-range shard (the group has exactly one shard). A reference table is **copied onto every shard of every group** in the database, so joins against it are local anywhere. So unsharded is "one place"; reference is "everywhere", they're not the same kind of global.

The reason this distinction is useful is because for an unsharded table, we don't have to do much, but for reference table, we need to make sure DMLs to that table are reflected in all the shards!

docs/sharding/design.md
GuptaManan100: Yes, I get the confusion, nomenclature is hard 😆. Its the one unsolved problem of CS, lol.

docs/sharding/design.md
GuptaManan100: I explained it in more detail so that its clearer to understand, let me know if it helps

docs/sharding/design.md
mkindahl: > They're similar but not quite the same, so I've spelled it out in the **reference** bullet. An unsharded table has **one physical copy**, on its group's single full-range shard (the group has exactly one shard). A reference table is **copied onto every shard of every group** in the database, so joins against it are local anywhere. So unsharded is "one place"; reference is "everywhere", they're not the same kind of global.

Gotcha.

> The reason this distinction is useful is because for an unsharded table, we don't have to do much, but for reference table, we need to make sure DMLs to that table are reflected in all the shards!

Well, there is one distinction in that case: an unsharded table only exists on a single shard, so we'd better route any query that references that table to the correct shard.

mkindahl (approved):

docs/sharding/design.md
mkindahl: LGTM.

docs/sharding/design.md
```diff
@@ -0,0 +1,293 @@
...
+to both shards, gather.
+
+### Co-located tables
+
+`t1` and `t2` in the same group on the same key. For any `user_id` both rows
+share a keyrange-id, so `t1 JOIN t2 ON user_id` runs within one shard.
+Co-location is group membership.
+
+```text
+tables: [ {schema:"sc1", name:"t1", function:{hash:{}}, columns:["user_id"]},
```
shedges, line 236: add multi-column examples

PR #1409, opened 2026-08-25
feat(multigateway): gate keep-transaction-on-gateway-rejection behind a flag

GuptaManan100:
## Summary

- A gateway policy rejection (feature_not_supported) never reaches the backend, so the transaction stays open. The gateway used to always leave the session in-block, which surprises clients that expect any wire error to close the transaction (PostgreSQL's contract, raised in #1352 review).
- Make that opt-in via `--keep-transaction-on-gateway-rejection` (default off, dynamic): off, rejections abort the transaction like any other error; on, they leave it in-block.
- pg_regress and the associated compatibility suites opt in, so a single refused statement in a long transaction doesn't cascade into "current transaction is aborted".


haritabh17 (approved):
Perfect! 🚀

PR #1414, opened 2026-08-26
fix(query-serving): PL/pgSQL parser hardening and unsafe-pooler-mode follow-ups

GuptaManan100:
## Summary

- Follow-ups to #1352: fix two ways malformed client-supplied `DO` / `CREATE FUNCTION` bodies could break the analyzer, and one place `--unsafe-pooler-mode` didn't actually relax rejections.
- Also gates the keep-transaction-on-gateway-rejection behavior behind a flag (default off).

## Fixes

- **INTO target EOF panic** — a body ending right after an `INTO` target (`SELECT a INTO x`) sliced with the EOF token's zero offset and panicked, dropping the client connection. Now reports a normal unterminated-statement error.
- **Cursor OPEN infinite loop** — an unterminated bound-cursor argument list (`OPEN c(1`) looped forever appending args and growing memory, since `readSQLConstruct` reports EOF non-fatally. Now breaks on any non-`,` terminator and requires `)`.
- **unsafe-pooler-mode set_config** — `validateDynamicSetConfigShape` and `validateAcceptedSetConfig`'s argument-shape checks rejected regardless of the opt-out, so statements like `SELECT set_config('work_mem', v, false) FROM gucs` still failed with `feature_not_supported`. They now pass through to PostgreSQL untracked, matching the documented contract.
- **quoting-shadow limitation** — documented (KNOWN LIMITATION) that the dynamic-EXECUTE safe-skeleton trusts `quote_ident` / `quote_literal` / `format` by unqualified name, which a tenant can shadow. A real fix needs catalog/search_path resolution the plan-time analyzer lacks.

## Also included

- `--keep-transaction-on-gateway-rejection` (default off): off, a gateway policy rejection aborts the open transaction like any other wire error (PostgreSQL's contract); on, it leaves the session in-block. pg_regress and the compatibility suites opt into the flag.

### DIFF
--- a/go/common/parser/plpgsql/read_construct.go
+++ b/go/common/parser/plpgsql/read_construct.go
@@ -760,12 +760,32 @@ func (l *lexer) readExecSQLIntoTarget() (strict bool, target string, endPos int)
 		strict = true
 		tok = l.scanNext()
 	}
+	if tok.tok == 0 {
+		// End of input (or a lexical error) right after INTO [STRICT], with no
+		// target name. Push the terminator back so the caller's scan loop reports
+		// the unterminated statement, rather than slicing with the EOF token's
+		// zero position below. -1 endPos keeps the caller off the INTO slice path.
+		l.pushBack(tok)
+		return strict, "", -1
+	}
 	targetStart := tok.pos
 	for {
 		next := l.scanNext()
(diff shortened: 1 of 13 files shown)


haritabh17 (approved):
Nice 🚀

PR #1415, opened 2026-08-26
feat(multipooler): scrub pooled connections for session-state divergence

haritabh17:
Builds the connection state consolidator (MUL-1158): a background scrubber that verifies pooled backends' real session state against the state the pool tracks for them, alarms on divergence, and replaces affected connections.

## Why

Under the gateway-authoritative session model (#1306), the pool trusts each connection's settings label absolutely — pointer-equal reuse and clean-stack checkout run zero SQL. Any backend mutation that escapes tracking (a `set_config` hidden in a routine body, a tracking bug, out-of-band DDL) therefore silently leaks to the next borrower. The scrubber is the detection net for that failure class — the "verification sampler" deferred out of scope during the #1306 design. The tracking and rejection gates (#1352) remain the correctness boundary; the scrubber bounds the leak window to roughly one sweep and raises an alarm that tracking was bypassed.

## Design

- **Divergence probe** (`regular.Conn.VerifySessionState`): one round trip comparing the tracked label against real backend state, classifying findings as untracked / phantom / value-mismatch. Verified on PostgreSQL 17 that `role`, `session_authorization`, and custom placeholder GUCs are `GUC_NO_SHOW_ALL` (never visible in `pg_settings`), so identity is probed via `current_setting('role')` + `session_user` (compared against the connection's login user via a new `client.Conn.User()` accessor) and tracked custom GUCs via `current_setting(name, missing_ok)`. Value spellings are normalized through a statement-local `set_config(..., is_local := true)` probe so `'65536'` vs `'64MB'` never counts as divergence. Findings carry GUC names only, never values.
- **Scrubber worker** (`connpool`): pops one idle connection per tick, rotating across the clean and settings stacks; clean connections return with their idle clock intact (scrubbing never defeats idle-timeout shrinking); divergent connections are closed and eagerly replaced into the same slot (a freed slot cannot wake a waitlisted client — same rationale as `closeIdleResources`). Replace, never reconcile: divergence means tracking was bypassed, and what a checker observes is only part of what the untracked code did.
- **Pluggable checkers**: the scrubber runs registered `connpool.ConnChecker`s (`Name` + `Check`); the session-state probe is the first, registered by every regular pool at construction (covering the reserved pool's underlying pool). Future checkers — prepared statements, advisory locks, temp state — plug in the same way and populate the same metrics via their name label.
- **Bootstrap invariant**: the untracked rule requires that connection bootstrap creates no session-source GUC state outside the settings label — bootstrap settings must arrive via startup-packet parameters (`source='client'`) or be reflected in the label, or the scrubber would replace every backend each sweep. Stated at the connector and in the docs, and pinned by an e2e canary (below).
- **Operations**: `--connpool-session-scrub-interval` (default 10s, `0` disables). OTel counters `mg.pooler.session_scrub.{checked,divergence,errors}` with `pool_type`, `checker`, and divergence-`kind` attributes, registered in the generated Prometheus metric catalog; any nonzero divergence warrants investigation. Divergence log lines carry names only.

Known blind spot (documented): an untracked custom placeholder GUC is unenumerable from SQL; the creation-time rejection gates are the defense for that class.

## Testing

- Unit: probe classification (identity, custom GUCs, normalization, fail-closed probe errors) against fakepgserver; scrubber slot accounting, idle-clock preservation, multi-checker merge, worker end-to-end with mock conns. All race-clean.
- E2E (`TestSessionScrubberReplacesHiddenState`): first a canary asserts normal traffic produces zero divergence (pinning the bootstrap invariant); then a `set_config` hidden in a SQL function body installed directly on the primary contaminates a pooled backend; the scrubber detects it, logs the GUC by name (asserted never the value), terminates the backend, and no later client observes the leak.
- Full queryserving suite run with scrubbing default-on: all session/SET/prepared/transaction tests pass; the only divergence event across the run was the e2e test's intentional one (no false positives).
- GUC semantics (`pg_settings` visibility, RESET source restoration, `is_local` autocommit revert, normalization forms) verified against a live PostgreSQL 17.

Docs: scrubber section added to `docs/query_serving/connection_pooling.md`; the hidden-routine-body limitation in `docs/query_serving/session_settings.md` now references it.

### DIFF
--- a/go/services/multipooler/internal/pools/connpool/scrub.go
+++ b/go/services/multipooler/internal/pools/connpool/scrub.go
@@ -1,0 +1,155 @@
+// Copyright 2026 Supabase, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connpool
+
+import (
+	"context"
+	"time"
+)
+
+// The session-state scrubber is the safety net for the gateway-authoritative
+// session model: the pool trusts each connection's tracked settings label
+// absolutely (pointer-equal reuse and clean-stack reuse run zero SQL), so any
+// session mutation that escaped tracking — a set_config hidden in a routine
+// body, a tracking bug, an out-of-band DDL — silently leaks to the next
+// borrower. The scrubber periodically probes idle connections through
+// SessionStateVerifier, emits metrics when the real backend state diverges
+// from the label, and closes and replaces divergent connections.
+//
+// It only touches idle connections, so it adds no latency to the query path.
+// It is a sampler: it narrows the leak window and raises an alarm, but the
+// gateway's tracking gates remain the correctness boundary.
+
+// scrubProbeTimeout bounds one verification probe. Kept well under
+// PoolCloseTimeout so an in-flight probe cannot stall pool shutdown.
+const scrubProbeTimeout = 5 * time.Second
+
+// scrubStackCount is the number of idle stacks the scrubber rotates over:
+// the clean stack plus the settings stacks.
+const scrubStackCount = stackMask + 2
+
+// scrubStack maps a rotation index to an idle stack: 0 is the clean stack,
+// 1..stackMask+1 are the settings stacks.
+func (pool *Pool[C]) scrubStack(i int) *connStack[C] {
+	if i == 0 {
+		return &pool.clean
+	}
+	return &pool.states[i-1]
+}
+
+// scrubPop pops an idle connection like pool.pop, but also returns the
+// connection's pre-borrow timeUsed value so the scrubber can restore it: a
+// scrub must not refresh the idle clock, or a small pool would have every
+// connection kept forever-fresh by scrubbing and the idle-timeout worker
+// could never shrink it. The stamp read before borrow is reliable: the only
+// concurrent writer for an in-stack connection is the expirer, and if it won
+// the race the borrow fails and the connection (already closed by the
+// expirer) is skipped.
+func (pool *Pool[C]) scrubPop(stack *connStack[C]) (*Pooled[C], time.Duration) {
+	for conn, ok := stack.Pop(); ok; conn, ok = stack.Pop() {
+		stamp := conn.timeUsed.get()
+		if conn.timeUsed.borrow() {
+			return conn, stamp
+		}
+	}
+	return nil, 0
+}
+
+// scrubOne probes one idle connection for session-state divergence, rotating
+// the starting stack across calls so every bucket gets coverage. A clean
+// connection returns to the pool with its idle clock intact; a divergent one
+// is closed and replaced. Returns false to stop the worker when the pool's
+// connections do not implement SessionStateVerifier (a wiring error).
+func (pool *Pool[C]) scrubOne(cursor *int) bool {
+	if pool.Capacity() == 0 {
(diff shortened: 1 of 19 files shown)


go/services/multipooler/internal/pools/connpool/scrub.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	return nil, 0
+}
+
+// scrubOne runs every registered checker against one idle connection,
+// rotating the starting stack across calls so every bucket gets coverage. A
+// clean connection returns to the pool with its idle clock intact; a
+// divergent one is closed and replaced. The bool return keeps the runWorker
+// signature; it is always true.
+func (pool *Pool[C]) scrubOne(cursor *int) bool {
```
GuptaManan100, line 76: When we grab a connection from the stack, we don't increment the borrowed, so that can lead to the `Available` value being wrong, and it can cause `Get` to block on a connection that isn't free at all.

go/services/multipooler/internal/pools/connpool/scrub.go
```diff
@@ -0,0 +1,172 @@
...
+	type finding struct {
+		checker string
+		div     Divergence
+	}
+	var findings []finding
+	var checkErr error
+	var errChecker string
+	ctx, cancel := context.WithTimeout(pool.ctx, scrubProbeTimeout)
+	for _, checker := range pool.checkers {
+		div, err := checker.Check(ctx, conn.Conn)
```
GuptaManan100, line 110: This could potentially delay the close functionality. Since Check can take upto 5 seconds, a `Close()` call will wait an additional that much time. Should we be checking for closeChan as well while this call runs?

go/services/multipooler/internal/pools/regular/session_verify.go
```diff
@@ -0,0 +1,320 @@
...
+//     current_setting(name, missing_ok := true), which returns NULL when the
+//     session has never seen the GUC.
+//
+// Known blind spot: a custom GUC set behind tracking's back on a connection
+// whose label does not contain it is undetectable — placeholder GUCs cannot
+// be enumerated from SQL. The creation-time rejection gates remain the
+// defense for that class.
+const sessionSourceProbe = "SELECT name, current_setting(name), 'session' FROM pg_settings WHERE source = 'session'" +
+	" UNION ALL SELECT 'role', pg_catalog.current_setting('role'), 'identity'" +
+	" UNION ALL SELECT 'session_authorization', session_user::text, 'identity'"
```
GuptaManan100: nit: we should keep constants in the `constants` package.

go/services/multipooler/internal/pools/regular/session_verify.go
```diff
@@ -0,0 +1,320 @@
...
+
+// sessionStateQuery returns the probe SQL, extending the constant part with
+// one 'custom' row per tracked placeholder GUC. customNames must be sorted
+// for deterministic SQL. Single quotes are escaped by doubling, as in
+// Settings.ApplyQuery.
+func sessionStateQuery(customNames []string) string {
+	var b strings.Builder
+	b.WriteString(sessionSourceProbe)
+	for _, name := range customNames {
+		escaped := strings.ReplaceAll(name, "'", "''")
```
GuptaManan100: nit: should we use the ast package for quoting identifier?

GuptaManan100 (approved):
Rest LGTM!

go/services/multipooler/internal/pools/connpool/scrub.go
haritabh17: Ack updated.

go/services/multipooler/internal/pools/connpool/scrub.go
haritabh17: Added ctx.

go/services/multipooler/internal/pools/regular/session_verify.go
haritabh17: Ack, moved

go/services/multipooler/internal/pools/regular/session_verify.go
haritabh17: Yes, moved.

PR #1418, opened 2026-08-27
feat(query-serving): accept provably-safe runtime-built EXECUTE in PL/pgSQL bodies

GuptaManan100:
## Summary

Widens the Tier 1 dynamic-`EXECUTE` policy to accept _provably-safe_ runtime-built statements in PL/pgSQL bodies, without weakening tenant isolation. Purely additive vs `main`: it only accepts runtime-built `EXECUTE`s that were rejected before, and never accepts anything unsafe.

## What's now accepted

- **`EXECUTE v` (intra-body dataflow)** — a scalar's `:=` assignments and its DECLARE initializer are traced, so `EXECUTE v` / `RETURN QUERY EXECUTE v` is allowed when every value `v` can hold reduces to a safe skeleton (`format('…%I…', …)`, safe `||` concat with bind params, transitive `w := v`).
- **`format()` with constrained `%s`** — a top-level `format()` may use `%s` fed by a bounded constant set (a literal, an all-literal `CASE`, or a variable proven to hold only those); the finite set of concrete statements is enumerated and each re-analyzed as a static statement. Positional specifiers (`%n$I` / `%n$L` / `%n$s`) are supported.

## Soundness (fail-closed)

Collection is flow-insensitive, so acceptance only holds when _every_ value the variable can hold is provably safe. Rejected as unprovable: function parameters and `ALIAS`es, `CALL`-mutated arguments, `INTO`/`FETCH`/loop/`GET DIAGNOSTICS` targets, names out of lexical scope at the `EXECUTE`, and the `NULL`→empty-`%s` case. Identifiers are compared as PostgreSQL folds them (quoted names keep their case). The quoting builtins (`quote_ident`/`quote_literal`/`format`) are trusted by unqualified name only — a documented limitation that needs catalog/search_path resolution to close.

## Tests

Accept **and** adversarial soundness-reject cases for each rule (param aliasing, `CALL` taint, nested-scope shadow, `NULL`-empty exposure, quoted-name aliasing, injected `set_config`). Full `multigateway/...` suite green; the external-extension regression patches (`plpgsql_check`, `postgis_topology`) are regenerated to reflect the wider acceptance.


go/services/multigateway/planner/unsafe_plpgsql.go
```diff
@@ -563,6 +755,227 @@ func expandFormatSkeleton(f string, sb *strings.Builder) bool {
...
+// to a top-level format() (not one nested in a `||` chain) to keep the fixed
+// skeleton single-valued outside the enumerated holes.
+//
+// Returns (err, true) when it owns the decision (the payload is such a format()
+// with all %s constrained); (nil, false) to let the caller fall through to the
+// normal rejection. res may be nil, in which case only literal/CASE-of-literal
+// %s args qualify (no variable resolution).
+func analyzeConstrainedFormatExecute(exprText string, res *varResolver) (error, bool) {
+	fc, ok := executeArgExpr(exprText).(*ast.FuncCall)
+	if !ok || bareFuncName(fc.Funcname) != "format" {
```
haritabh17, line 979: We also need to ensure that the `search_path` is resolving to `pg_catalog`, otherwise it can resolve to a user defined function with same name.

go/services/multigateway/planner/unsafe_plpgsql.go
```diff
@@ -183,11 +183,181 @@ func analyzePLpgSQLFunction(fn *plpgsqlast.PLpgSQL_function) error {
...
+	r := &varResolver{
+		assigns:   map[string][]*plpgsqlast.PLpgSQL_expr{},
+		tainted:   map[string]bool{},
+		resolving: map[string]bool{},
+	}
+	plpgsqlast.Rewrite(fn, func(cursor *plpgsqlast.Cursor) bool {
+		switch n := cursor.Node().(type) {
+		case *plpgsqlast.PLpgSQL_stmt_assign:
+			if name, ok := simpleIdent(n.Target); ok {
+				r.assigns[name] = append(r.assigns[name], n.Expr)
```
haritabh17, line 275: Consider example where , inside a function with a parameter q followed only by `IF false THEN q := 'SELECT 1'; END IF; EXECUTE q` is currently accepted, although EXECUTE receives the caller's arbitrary `q`.

haritabh17 (approved):
Rest looks good 🚀

go/services/multigateway/planner/unsafe_plpgsql.go
```diff
@@ -183,11 +183,181 @@ func analyzePLpgSQLFunction(fn *plpgsqlast.PLpgSQL_function) error {
...
+// non-simple assignment targets). See varResolver for the soundness argument.
+func collectVarAssignments(fn *plpgsqlast.PLpgSQL_function) *varResolver {
+	r := &varResolver{
+		assigns:   map[string][]*plpgsqlast.PLpgSQL_expr{},
+		tainted:   map[string]bool{},
+		resolving: map[string]bool{},
+	}
+	plpgsqlast.Rewrite(fn, func(cursor *plpgsqlast.Cursor) bool {
+		switch n := cursor.Node().(type) {
+		case *plpgsqlast.PLpgSQL_stmt_assign:
```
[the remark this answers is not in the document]
GuptaManan100: Fixed

go/services/multigateway/planner/unsafe_plpgsql.go
GuptaManan100: Good catch. Fixed

go/services/multigateway/planner/unsafe_plpgsql.go
GuptaManan100: You're right, and this is the same known limitation already documented, because we can't fix it without a catalog lookup...

PR #1420, opened 2026-08-27
feat(multipooler): add metrics for logical replication failovers

kabochya:
## Summary

Adds observability metrics for logical-replication failovers (MUL-1453), following the pattern established in #1355 (reserved connection metrics). Instruments the promotion path in `go/services/multipooler/internal/manager` — specifically the logical-replication-slot-management span of `promoteLocked`'s `promotionHook` (`ensureFollowerPhysicalSlots`, `setSynchronizedStandbySlots`, the advisory unready-slot check), plus the slot-cleanup paths (`dropOrphanedFailoverSlots`, `ReconcileFollowers`) — with OTel metrics, all colocated in the existing `managerMetrics` struct (the same file/pattern used for the pg_rewind metrics) rather than a new package, since the instrumented code lives directly in `manager`, not in a standalone subsystem like `replicationstats`/`replication`.

New instruments (all under `mg.pooler.logical_failover.*`):

| Metric | Type | What it measures | Category |
|---|---|---|---|
| `duration` | Histogram, labelled `status` | Duration of just the LR-slot-management span inside promotion (excludes `ClearResignedLeaderAtTerm` and `pg_promote`, which run on every failover regardless of logical replication) | Troubleshooting — attributes a slow/failed promotion to the slot-management step specifically |
| `count` | Counter, labelled `status` | Count of that same span by outcome | Troubleshooting — a rising failure count flags slot-sync instability |
| `slots_dropped` | Counter, labelled `reason` (`orphaned`/`departed_follower`) | Slots dropped during cleanup in `dropOrphanedFailoverSlots` / `ReconcileFollowers` | Troubleshooting — a silent stop here would leak slots (retained WAL) with no other symptom |
| `slots` | ObservableGauge, labelled `status` (`ready`/`unready`) | Steady-state failover-slot readiness, sampled continuously via the existing health-heartbeat loop from `failoverSlotReadiness` | Health — visible before a failover is ever needed, not just reactively |

A generic whole-failover duration/count metric (covering `ClearResignedLeaderAtTerm` and `pg_promote` too) is left for a follow-up PR.

The metric catalog (`go/observability/metriccatalog/catalog_generated.go`) was regenerated via `make metrics` to include the 4 new series, scoped to the `multipooler` binary.

## Test plan

- [x] `go test -short -race ./go/services/multipooler/internal/manager/...` — all pass, including new/updated unit tests for every new instrument (value correctness, attribute labels, nil-receiver safety)
- [x] `go test -short ./go/observability/metriccatalog/...` — catalog consistency tests (`TestKeepListMatchesEverySeries`, `TestKeepListByBinaryConsistent`, etc.) pass against the regenerated catalog
- [x] `gofmt -l` / `go vet` clean on touched packages
- [ ] Manual verification against a running local cluster (scrape `/metrics` during a real promotion) not yet done


go/services/multipooler/internal/manager/rpc_consensus.go
```diff
@@ -569,26 +569,18 @@ func (pm *MultipoolerManager) promoteLocked(ctx context.Context, req *consensusd
...
+		status := logicalFailoverStatusSuccess
+		if slotsErr != nil {
+			status = logicalFailoverStatusFailure
 		}
-		// Hold the failover logical slots back to the followers' physical slots so
-		// a standby cannot outrun a slot's catalog_xmin and block slot-sync. Set
-		// before pg_promote so it is in effect the moment this node is primary.
-		if err := pm.setSynchronizedStandbySlots(hookCtx, cohortMembers); err != nil {
-			return err
+		pm.metrics.recordLogicalFailover(hookCtx, status, time.Since(slotsStart))
```
mkindahl: This means that `recordLogicalFailover` runs on every promotion even when slot-based replication is disabled. In that case `manageLogicalFailoverSlots` is a near-instant no-op (each sub-step returns early on the flag check), but this still records `status=success` with a ~0s duration.

I think you need to check if logical replication slot failover is enabled using  `pm.slotBasedReplicationEnabled()` or recording inside `manageLogicalFailoverSlots` after an enabled check.

go/services/multipooler/internal/manager/metrics.go
```diff
@@ -154,10 +192,71 @@ func newManagerMetrics() (*managerMetrics, error) {
...
+	)
+	errs = append(errs, slotsErr)
+	if slotsErr == nil {
+		m.logicalFailoverSlots = slots
+		if _, err := meter.RegisterCallback(
+			func(_ context.Context, o metric.Observer) error {
+				ready := m.logicalFailoverSlotsReady.Load()
+				total := m.logicalFailoverSlotsTotal.Load()
+				o.ObserveInt64(m.logicalFailoverSlots, ready, metric.WithAttributes(attribute.String("status", "ready")))
+				o.ObserveInt64(m.logicalFailoverSlots, total-ready, metric.WithAttributes(attribute.String("status", "unready")))
```
mkindahl: This computes `total - ready` from two separate atomic `Load()`s, while `setFailoverSlotReadiness` stores `ready` and `total` using independent atomic stores. For that reason, you can read `total` and `ready` that are not in sync.

Example: node goes from (ready=0,total=0) to (ready=3,total=5). The heartbeat does `Store(ready=3)`, then this callback runs before `Store(total=5)`: it reads ready=3, total=0 → `unready = 0 - 3 = -3`.

You might want to pack both counts into a single atomic value and avoid this.

go/services/multipooler/internal/manager/health_provider.go
```diff
@@ -326,6 +326,16 @@ func (pm *MultipoolerManager) runHealthHeartbeat(ctx context.Context, interval t
 					pm.healthStreamer.SetReplicationLag(lag.Nanoseconds())
 				}
 			}
+			// Steady-state failover-slot readiness, so it's visible before a
+			// failover is ever needed, not just via the advisory check at
+			// promotion time. Best-effort, like replication lag above. This
+			// backs mg.pooler.logical_failover.slots, unrelated to the health
+			// stream broadcast above.
+			if pm.slotBasedReplicationEnabled() {
```
mkindahl: This samples readiness on every node via the heartbeat, but `failoverSlotReadiness` counts a slot ready only when `synced=true`. On the current _primary_ the failover slots have `synced=false`, so the primary continuously reports `ready=0` and `unready=total`.

Consider if you want to emit it on the primary, since it is not a useful metric for the primary.

go/services/multipooler/internal/manager/health_provider.go
```diff
@@ -326,6 +326,16 @@ func (pm *MultipoolerManager) runHealthHeartbeat(ctx context.Context, interval t
...
+				if ready, total, err := pm.failoverSlotReadiness(ctx); err == nil {
```
mkindahl, line 348: Nit: this adds a second database round-trip per heartbeat (on top of `ReplicationLag`) whenever slot-based replication is enabled. You might want to include the slot readiness in the `ReplicationLag` query and avoid an extra database round-trip.

go/services/multipooler/internal/manager/metrics.go
```diff
@@ -110,8 +111,45 @@ func (m *healthMetrics) recordTransition(ctx context.Context, from, to clusterme
...
+	logicalFailoverSlotsDropped metric.Int64Counter
+
+	// logicalFailoverSlots is a steady-state gauge of this node's logical
+	// failover-slot readiness, labelled by status (ready/unready). Its value
+	// comes from these atomics rather than a live query, since an
+	// ObservableGauge callback runs synchronously during metric collection and
+	// must not block a scrape on a DB round-trip; setFailoverSlotReadiness is
+	// called periodically by the manager's health heartbeat to keep them fresh.
+	logicalFailoverSlotsReady atomic.Int64
+	logicalFailoverSlotsTotal atomic.Int64
```
mkindahl: These are coupled: you probably want to store them as a structure and use a pointer. See the other comment below where they are used.

mkindahl (commented):
A few comments inline below, otherwise looks good.

go/services/multipooler/internal/manager/health_provider.go
kabochya: I'll skip this for now, right now the replication lag query seems to come from a separate async loop. I think we might want to follow-up on these admin/o11y queries later on to provide a generic way of exposing database metrics (e.g. some sort of heartbeat loop hook) for easier extension in the future.

go/services/multipooler/internal/manager/health_provider.go
kabochya: Added a gating helper method to skip this for primaries

go/services/multipooler/internal/manager/health_provider.go
```diff
@@ -326,6 +339,16 @@ func (pm *MultipoolerManager) runHealthHeartbeat(ctx context.Context, interval t
...
+			// Steady-state failover-slot readiness, so it's visible before a
+			// failover is ever needed, not just via the advisory check at
+			// promotion time. Best-effort, like replication lag above. This
+			// backs mg.pooler.logical_failover.slots, unrelated to the health
+			// stream broadcast above.
+			if pm.shouldPollFailoverSlotReadiness() {
+				if ready, total, err := pm.failoverSlotReadiness(ctx); err == nil {
+					pm.metrics.setFailoverSlotReadiness(ready, total)
+				}
+			}
```
mkindahl, line 353: This updating the snapshot on the primary, but nothing ever clears it. This means that a standby that is promoted keeps emitting that frozen value, which looks like it is valid, but it is not useful for the primary. You should probably clear it. For example, something like this:

```suggestion
			if pm.shouldPollFailoverSlotReadiness() {
				if ready, total, err := pm.failoverSlotReadiness(ctx); err == nil {
					pm.metrics.setFailoverSlotReadiness(ready, total)
				}
			} else if pm.metrics != nil {
				pm.metrics.failoverSlotReadinessSnapshot.Store(nil)
			}
```

Idea is that it should be clear if it is the primary or if the option is disabled. If you want, you can introduce a clear function to be symmetric with `setFailoverSlotReadiness` and use that instead.

mkindahl (approved):
One minor thing to fix, but otherwise looks good. Approving since probes can filter on this if necessary (e.g., "don't check the metric on primaries") so leaving to you to decide  about the merge.

go/services/multipooler/internal/manager/health_provider.go
kabochya: Thanks for catching this!

cuongdo:
Congrats on your first Multigres PR!! 🎉🎉🎉

PR #1421, opened 2026-08-28
fix(multigateway): handle parameter-bound set_config for gateway-managed variables

GuptaManan100:
## Summary

- PostgREST applies role-level settings per request via `set_config($1, $2, true)` (dynamic name). When `$1` resolves to a gateway-managed variable like `statement_timeout` (set via `ALTER ROLE`), the gateway rejected the statement, so every request for such a role failed.
- The gateway now accepts the `is_local=true` form: it tracks the value in gateway state and lets the paired Route run the `set_config` transaction-locally on the backend, which reverts and never persists on a pooled connection.

## Problem

A bound `set_config` name is invisible to the planner's gateway-managed rewrite, so `engine.resolveSetConfig` fail-closed on any bound name resolving to a gateway-managed variable — rejecting PostgREST's per-request role-setting query. The upstream PostgREST hspec suite we run never triggers this: its fixtures configure no role-level `statement_timeout`, so the dynamic-name form is never emitted; that coverage lives only in PostgREST's separate Python `test/io` suite, which we don't run.

## Solution

Gate the rejection on `is_local`: allow `is_local=true` (the value runs transaction-locally and reverts, and the gateway owns it for `SHOW`/enforcement), keep rejecting `is_local=false` (a pinned session would persist it on the reserved backend — a documented limitation PostgREST never hits). Adds a dual-target reproduction (`TestStatementTimeoutBoundName`) plus engine unit coverage for the applied and rejected cases.

### DIFF
--- a/go/services/multigateway/engine/apply_session_state.go
+++ b/go/services/multigateway/engine/apply_session_state.go
@@ -297,20 +297,33 @@ func (s *ApplySessionState) resolveSetConfig(resolver setConfigParamResolver) (r
 				"set_config name argument cannot be NULL")
 		}
 		name = v
-		// A gateway-managed variable must never reach a backend, but a
-		// parameter-bound name is invisible to the planner's rewrite (it only
-		// strips literal gateway-managed names from the routed query), so the
-		// real set_config would execute there. This resolution runs during the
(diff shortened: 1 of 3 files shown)


haritabh17 (approved):
Nice 🚀

PR #1422, opened 2026-08-28
fix(consensus): restrict MultiCellPolicy to 2 cells

simonbtomlinson:
# Summary

`MultiCellPolicy{N: 3+}` did not correctly require replicas in `sync_standby_names` to be in different cells. For now, simply restrict to `N=2` and make `MultiCellPolicy` require `1+` standby in a different cell to the primary. In the future we could expand support to an arbitrary number of cells.

# Description

Removes `MultiCellPolicy.N` and hard-codes it to 2. Does not make changes at the protobuf layer, instead simply errors if the protobuf tries to construct a `MultiCellPolicy` with `N != 2`.

Makes `multiCellPolicy` private, it was not used outside its package before, and this helps to protect the public API in case we support N cells again.

### DIFF
--- a/go/common/consensus/durability.go
+++ b/go/common/consensus/durability.go
@@ -57,7 +57,7 @@ type DurabilityPolicy interface {
 	//
 	// cohort is the full set of poolers participating in the term, including
 	// the primary. The method derives the eligible standby set per policy
-	// (e.g., MultiCellPolicy excludes the primary's cell). Passing the full
+	// (e.g., multiCellPolicy excludes the primary's cell). Passing the full
 	// cohort keeps the caller-side contract simple.
 	//
 	// On success the config is always non-nil — every promotion explicitly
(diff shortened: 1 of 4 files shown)


go/common/consensus/durability.go
```diff
@@ -107,11 +107,12 @@ func NewPolicyFromProto(policy *clustermetadatapb.DurabilityPolicy) (DurabilityP
 		}
 		return AtLeastNPolicy{N: int(policy.RequiredCount)}, nil
 	case clustermetadatapb.QuorumType_QUORUM_TYPE_MULTI_CELL_AT_LEAST_N:
-		// N=0 would make revocation (|uncovered cells| < N) unsatisfiable for any recruitment.
-		if policy.RequiredCount < 1 {
```
simonbtomlinson, line 111: I removed the support for `RequiredCount == 1` here, which we could have kept. But it's kind of useless - it's equivalent to no replication at all. And removing it simplified the code and let me remove `MultiCellPolicy.N`. I'm curious if reviewers agree.

dweitzman: Seems reasonable to me, we could support policies from the at least N policy without needing to involve multi-cell

dweitzman (approved):

PR #1424, opened 2026-08-31
feat(query-serving): make unsafe-pooler-mode a per-connection direct-connection setting

GuptaManan100:
## Summary

- Replace the global `--unsafe-pooler-mode` operator flag with a per-connection setting, `multigres.direct_connection`. A client opts itself into unsafe-statement bypass instead of the whole gateway being unsafe.
- Requestable two ways: at connect time (`options=-c multigres.direct_connection=on`) or mid-session via `SET multigres.direct_connection = on`. It is a one-way latch — once on it cannot be turned off (RESET/off errors), because a connection that has run unsafe statements may hold untracked backend state.
- A direct connection gets its own dedicated backend, pinned and quarantined for the connection's lifetime (new `RESERVATION_REASON_DIRECT_CONNECTION`) and discarded at teardown, so its untracked session state can never leak into the shared pool.

## Description

The old design was all-or-nothing: `--unsafe-pooler-mode` disabled the unsafe-statement rejections (Tier 1 PL/pgSQL body analysis, the Tier 2 statement blocklist, the restricted-GUC guard, and the expression-level function blocklist) for the entire gateway. This makes the same opt-out a property of an individual connection.

`multigres.direct_connection` is a gateway-only control, never forwarded to a backend:

- **Connect time** (`server/startup.go`): the param is extracted and stripped before restricted-GUC vetting and before anything reaches a backend; a malformed Boolean is a FATAL startup error. A truthy value latches the connection.
- **Mid-session** (`SET multigres.direct_connection = on`): planned as a new gateway-local primitive (`EnableDirectConnection`) that latches the flag and replies with a bare `SET` — it touches no backend.

Once latched (`Conn.DirectConnection()`), every statement on the connection:

- skips the unsafe-statement rejections in the planner (planning signals are still gathered, so routing stays correct);
- is kept off the shared plan cache, since its plan depends on the per-connection opt-out;
- reserves and quarantines its backend via `ReasonDirectConnection`, OR-ing the reason onto any existing reservation so the pin survives after other reasons (e.g. a transaction) end.

`ReasonDirectConnection` is *sticky* like `ReasonSetSeed`: it survives `DISCARD ALL` and releases only at real teardown. Unlike setseed, the tainted backend is closed on release rather than returned to the pool (`AddReservationReason` sets `closeOnRelease` unconditionally, at add time, since even a failed statement may have mutated untracked state). A shared `StickyReasons` mask and `HasStickyReason` helper capture this class.

The pgTAP/pg_regress test harness previously spun up a second, dedicated unsafe gateway; it now exercises the same behavior through the per-connection option on a single gateway, so the separate unsafe-gateway setup is removed.


haritabh17 (approved):
LGTM 🚀

PR #1427, opened 2026-09-01
refactor(query-serving): rename direct-connection to unsafe-connection

GuptaManan100:
## Summary

- Rename the per-connection direct-connection mode to unsafe-connection across the codebase, aligning it with the existing "unsafe" family (unsafe-pooler-mode, unsafe_funccall).
- The user-facing option/GUC is now `multigres.unsafe_connection`; the former `multigres.direct_connection` is kept as a deprecated alias, recognized at connect time and via `SET`, until all clients migrate off it.
- Internal identifiers, files, tests, and the reservation-reason proto enum (`RESERVATION_REASON_UNSAFE_CONNECTION`, numeric value unchanged at 256, so wire-compatible) are all renamed.

## Description

`SET` error messages now echo the name the client actually supplied, so a client using the deprecated alias sees the name it typed rather than the canonical one. Both recognition points (connect-time in `startup.go` and `SET` in `variable_set_stmt.go`) accept either name.

Unit tests for both recognition points are parametrized over the primary name and the deprecated alias, and an end-to-end subtest proves the alias still enables the mode at connect time and via `SET`. Unrelated "direct connection" usages (direct TLS, the raw-write path, pgregress `isolation.go`, pgbench docs) were left untouched.

### DIFF
--- a/go/common/pgprotocol/server/unsafe_connection_test.go
+++ b/go/common/pgprotocol/server/unsafe_connection_test.go
@@ -23,11 +23,12 @@ import (
 	"github.com/multigres/multigres/go/common/constants"
 )
 
-// TestExtractDirectConnectionParam covers the connect-time parse: the parameter
+// TestExtractUnsafeConnectionParam covers the connect-time parse: the parameter
 // is stripped from the params map (so it never reaches a backend), a truthy value
-// latches direct connection on the connection, a falsy value does not, and a
-// malformed value is a FATAL.
-func TestExtractDirectConnectionParam(t *testing.T) {
+// latches unsafe connection on the connection, a falsy value does not, and a
+// malformed value is a FATAL. It runs against both the current parameter name and
+// the deprecated multigres.direct_connection alias, which must behave identically.
+func TestExtractUnsafeConnectionParam(t *testing.T) {
 	tests := []struct {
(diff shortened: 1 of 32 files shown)


go/common/pgprotocol/server/unsafe_connection_test.go
```diff
@@ -43,24 +44,30 @@ func TestExtractDirectConnectionParam(t *testing.T) {
...
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			c := &Conn{params: map[string]string{}}
-			if tt.present {
-				c.params[constants.DirectConnectionParam] = tt.value
-			}
+	params := map[string]string{
+		"unsafe_connection":        constants.UnsafeConnectionParam,
+		"direct_connection(alias)": constants.DirectConnectionParam,
+	}
```
shedges, line 50: This is for backward compatibility?

shedges (approved):
LGTM

go/common/pgprotocol/server/unsafe_connection_test.go
GuptaManan100: Just for the moment, once existing clients move to unsafe_connection, I'll drop it

PR #1429, opened 2026-09-01
fix(multipooler): enforce 0600 permissions on pgbackrest pgpass file

niconosenzo:
Replace pgpass through a 0600 temporary file so stale permissions
cannot expose refreshed credentials before chmod.


cuongdo (approved):
Not necessary for this PR, but it would be good to reduce the duplication between this code and the code in MultipoolerManager.

niconosenzo:
> Not necessary for this PR, but it would be good to reduce the duplication between this code and the code in MultipoolerManager.

Agreed. Since both paths now perform the same atomic pgpass creation, I’ll extract that logic into a shared helper.

PR #1432, opened 2026-09-02
test(query-serving): run PostgREST test/io suite through the multigateway

GuptaManan100:
## Summary

- Adds `TestPostgRESTIO`, a sibling of the hspec `TestPostgREST` suite that runs the proxy-relevant subset of PostgREST's upstream pytest `test/io` suite with PostgREST pointed at the multigateway.
- Covers paths the in-process hspec suite cannot reach: per-request role settings hoisted via `set_config($1, $2, true)`, `statement_timeout`, hoisted transaction settings, prepared statements, and LISTEN/NOTIFY config/schema reloads.
- Opt-in only, gated on the same PR labels / `RUN_EXTENDED_QUERY_SERVING_TESTS` as the hspec suite.

## Description

The io suite spawns real `postgrest` binaries and drives them over HTTP, so it exercises the extended-protocol / prepared-statement and role-setting paths the WAI-driven hspec suite never hits. The image uses PostgREST's official prebuilt release binary for the pinned tag (checksum-pinned, no Haskell build) on a slim Python base; a `postgrest` shim repoints each spawned instance at the gateway over TCP+password without patching upstream test files. Direct PostgreSQL remains the asserted baseline, so a gateway-only failure is a divergence. `io_selection.go` is the canonical include/skip list and `io_tests.md` documents why each test is run or skipped.

## Known open divergence

`test_notify_reloading_catalog_cache` is intentionally left failing (the job stays red until fixed, like the hspec suite). After a `DROP`+`CREATE` of a table and a `NOTIFY 'reload schema'`, PostgREST's next request fails with `22P02` because the multipooler's backend prepared statement (`ppstmtN`, deduped by query text, never re-parsed or closed after a DDL) keeps the pre-DDL parameter type. Root cause and candidate fix surfaces are documented in `io_tests.md`.


haritabh17 (approved):
🚀

PR #1433, opened 2026-09-02
fix(multipooler): heal stale cached plans on Bind/Execute and Describe

shedges:
## Summary

Two related bugs in cached-plan handling around SQLSTATE 0A000 ("cached plan must not change result type"). Both stem from the same root cause: multipooler's canonical prepared statements are shared by `(query text, param types)` across every client whose pooled backend connection happens to have them prepared (see `PoolerConsolidator`), so a DDL that changes a table's shape can leave a stale plan behind for a completely unrelated later caller to inherit.

- **Bug 1 — `portalExecuteWithReserved` gets stuck on a stale cached plan.** The reserved-connection path had no recovery for 0A000, unlike the regular-connection path. A reserved connection that hits a stale plan stays stuck on the error indefinitely, and since the backend is recycled back into the pool on release, any later caller reserving that same backend inherits the same stuck error.

- **Bug 2 — `Describe` has no recovery either, on either connection kind.** PostgreSQL revalidates a prepared statement's result shape on a bare `Describe('S', name)` for any row-returning statement, not just on Bind/Execute — `exec_describe_statement_message` calls `CachedPlanGetTargetList`, which calls `RevalidateCachedQuery` — so a stale plan raises the identical 0A000 there too. A client that never touched the original statement gets hit with a raw, unrecoverable error for something that isn't its fault: unlike a normal PostgreSQL client, it has no way to know a shared backend-level statement is even involved, so it has no DEALLOCATE-and-re-PREPARE recourse.

## Fix

Both get the same treatment: on 0A000, close the stale backend statement and drop the per-connection cache entry so the statement is correctly rebuilt on its next use, then retry once — but only when the connection is actually able to run that retry. Extracted into a single generic helper, `cachedPlanRetry`, used by all three call sites (`portalExecuteWithReserved`, `portalExecuteWithRegular`, and both of `Describe`'s branches).

The retry is gated on `conn.TxnStatus()`. A connection reserved for an active explicit transaction that hits a stale plan mid-transaction can't be healed in place: PostgreSQL rejects every subsequent command in a transaction once any statement in it has errored, so retrying there just trades the diagnosable 0A000 for PostgreSQL's opaque "transaction is aborted". In that case (`TxnStatusFailed`) the fix still closes the stale statement and drops the local cache entry — PostgreSQL's Close message has no aborted-transaction guard, since it's handled directly rather than routed through the SQL executor — so the statement is correctly rebuilt the next time the connection is used, but the original 0A000 is preserved and returned as-is. The caller rolls back and retries its transaction, exactly as it would for any other transactional error.

## Test plan

- [x] Unit tests for `cachedPlanRetry`: heals on a synthetic 0A000, retries exactly once, does not retry unrelated errors (`cached_plan_retry_test.go`)
- [x] End-to-end regression, regular pool (`describe_stale_across_clients_test.go`): two client connections forced onto the same settled 1-backend pool; client A prepares and describes a statement, a DDL changes the table's shape, client B (never touched the original statement) gets the correct post-DDL shape on Describe. Verified failing (raw 0A000) without the fix and passing with it.
- [x] End-to-end regression, reserved connection / explicit transaction (`reserved_stale_across_clients_test.go`): same setup, but client B's reservation is an active transaction. Verifies client B's first attempt gets the clean 0A000 (not the masked "transaction is aborted"), and that a fresh transaction on the same connection afterward succeeds with the correct post-DDL shape.
- [x] `go test -short` passes for `go/services/multipooler/internal/executor`

### DIFF
--- a/go/services/multipooler/internal/executor/executor.go
+++ b/go/services/multipooler/internal/executor/executor.go
@@ -1131,17 +1131,22 @@ func (e *Executor) portalExecuteWithReserved(
 	// Bind and execute using the portal's own name and the canonical statement name.
 	// When the protocol layer folded Describe('P') into Execute, fuse the
 	// backend round trip too so the portal description rides on the Execute
-	// response.
+	// response. Heals a stale cached plan transparently the same way the
+	// regular-connection path does: this backend may have been reserved by an
+	// entirely different caller before a DDL invalidated the canonical
+	// statement it left prepared, so a 0A000 here is not necessarily this
+	// caller's fault, and it cannot recover on its own either.
 	params := sqltypes.ParamsFromProto(portal.ParamLengths, portal.ParamValues)
-	var completed bool
 	// Opaque row passthrough for this statement; reset after so the reserved
 	// connection does not carry the mode into later transaction statements.
 	reservedConn.Conn().SetPassthroughRow(options.GetPassthroughRow())
-	if includeDescribe {
-		completed, err = reservedConn.BindDescribeAndExecute(ctx, portal.Name, canonicalName, params, paramFormats, resultFormats, maxRows, callback)
-	} else {
-		completed, err = reservedConn.BindAndExecute(ctx, portal.Name, canonicalName, params, paramFormats, resultFormats, maxRows, callback)
+	bindExecute := func(canonicalName string) (bool, error) {
+		if includeDescribe {
+			return reservedConn.BindDescribeAndExecute(ctx, portal.Name, canonicalName, params, paramFormats, resultFormats, maxRows, callback)
+		}
+		return reservedConn.BindAndExecute(ctx, portal.Name, canonicalName, params, paramFormats, resultFormats, maxRows, callback)
 	}
+	completed, _, err := e.bindExecuteWithCachedPlanRetry(ctx, reservedConn.Conn(), preparedStatement, canonicalName, bindExecute)
 	reservedConn.Conn().SetPassthroughRow(false)
 	if err != nil {
 		return e.portalReservedError(reservedConn, portal.Name, options, newlyReserved, addedStatementLocal, err)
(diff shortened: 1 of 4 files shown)


go/services/multipooler/internal/executor/executor.go
```diff
@@ -1131,17 +1131,22 @@ func (e *Executor) portalExecuteWithReserved(
...
-		completed, err = reservedConn.BindDescribeAndExecute(ctx, portal.Name, canonicalName, params, paramFormats, resultFormats, maxRows, callback)
-	} else {
-		completed, err = reservedConn.BindAndExecute(ctx, portal.Name, canonicalName, params, paramFormats, resultFormats, maxRows, callback)
+	bindExecute := func(canonicalName string) (bool, error) {
+		if includeDescribe {
+			return reservedConn.BindDescribeAndExecute(ctx, portal.Name, canonicalName, params, paramFormats, resultFormats, maxRows, callback)
+		}
+		return reservedConn.BindAndExecute(ctx, portal.Name, canonicalName, params, paramFormats, resultFormats, maxRows, callback)
 	}
+	completed, err := cachedPlanRetry(ctx, e, reservedConn.Conn(), preparedStatement, canonicalName, bindExecute)
```
haritabh17, line 1150: Preserve the original cached-plan diagnostic for this case, while invalidating enough local state so that the statement can be rebuilt after the caller rolls back. Keep transparent close/reparse/retry only when the backend returned to idle and add reserved-connection tests covering both portal execution and Describe inside an explicit transaction.

go/test/endtoend/queryserving/describe_stale_across_clients_test.go
```diff
@@ -0,0 +1,133 @@
...
+// TestDescribeStaleAcrossClients is the regression for: multipooler's pooled
+// backend connections share an already-PREPARE'd statement (keyed by query
+// text + param types, see preparedstatement.PoolerConsolidator) across
+// unrelated client sessions. PostgreSQL revalidates a prepared statement's
+// result shape on Bind/Execute, but NOT on a bare Describe('S', name)
+// (postgres.c exec_describe_statement_message just replays whatever
+// resultDesc was recorded at the last successful Parse). So a client that
+// never touched the original statement, but happens to be handed a backend
+// that has it cached from before a DDL changed the table's shape, can be
+// told the pre-DDL shape on Describe with no error at all.
```
haritabh17: nit: the comment is stale now.

haritabh17 (approved):
Rest looks good 🚀

go/services/multipooler/internal/executor/executor.go
shedges: Done

go/test/endtoend/queryserving/describe_stale_across_clients_test.go
shedges: Done

PR #1434, opened 2026-09-02
fix(pgctld): detect a stale postmaster.pid whose PID was reused

kabochya:
## Problem

After a pod reschedule onto a different node (node drain, RWO volume reattach), `postmaster.pid` survives on the reattached data volume holding a PID from the *old* container, while the new container starts with its own fresh PID namespace.

The running check only asked whether *some* process held that PID (`Signal(0)`), which proves a process exists but not that it is our postmaster. When an unrelated process in the new namespace happened to be assigned the same PID number, the check reported a false positive and the node wedged:

1. `StartPostgreSQLWithResult` returned `AlreadyRunning=true` and never called `pg_ctl`, so postgres was never started.
2. Every subsequent `MonitorPostgres` tick reached the same conclusion, so there was no self-correction.
3. The readiness gate reports on gRPC rather than postgres (deliberately), so the pod stayed `Ready` the whole time and the shard controller saw nothing wrong.

Recovery required manually deleting `postmaster.pid`. Credit to @mkindahl, whose review on #1410 walked through this sequence precisely — and who also suggested the pg_isready-based approach this PR uses.

## Solution

Rather than trying to corroborate the recorded PID's OS-level identity (start time, process name), `checkPostgreSQLRunning` now probes the actual connection endpoint with `pg_isready` and trusts its exit code:

- **0 (accepting) or 1 (rejecting, e.g. still starting up or in crash recovery)** — postgres is alive; keep the lock file.
- **2 (no response)** — nothing is listening as postgres, whatever the recorded PID currently belongs to; remove the stale lock file so `pg_ctl`/postgres don't inherit the same false positive.
- **3 (no attempt made, e.g. invalid connection parameters), or pg_isready itself couldn't be run at all** — something is broken in the control plane, not in postgres. This is flagged as a hard error and bubbled up rather than guessed either way — every caller (`Start`, `Stop`, `Restart`, `ReloadConfig`, crash recovery, and status reporting) surfaces it instead of silently deciding the node is up or down.

`pg_isready`'s `-t` timeout is derived from the request's context deadline rather than relying on libpq's default, so the subprocess's own connection timeout fires before the context cancels it mid-wait.

This sidesteps OS process-identity entirely — no interface for injecting a fake process reader, no dependency on a real process happening to be named `postgres`, no start-time tolerance window to reason about.

### Status must be read-only

`checkPostgreSQLRunning` is split into a side-effect-free `probePostgreSQLRunningAndReady` (used only by `Status`) and `checkPostgreSQLRunning` itself, which layers the stale-lock cleanup on top for callers that act on the result (`Start`, `Stop`, `Restart`, `ReloadConfig`, crash recovery). `Status` is polled continuously and concurrently in the background, so it must not carry the side effect of deleting a lock file on a single bad probe — doing so raced a postgres instance pgctld was legitimately still managing and caused several integration tests to fail deterministically in CI.

`Restart` also now calls `StopPostgreSQLWithResult`/`StartPostgreSQLWithResult` directly rather than through separate "skip the recheck" variants, trading one extra `pg_isready` probe per restart for a simpler call graph.

## Relationship to #1410

#1410 retries `pg_ctl start` through the orphan-cleanup lock window. That is a different failure (same node, unclean kill, transient) from the reused-PID scenario this PR fixes, and it does not touch the running-check false positive that actually wedges a rescheduled pod. This change is independent of it and does not conflict.

## Commits

1. **test(pgctld): stop mock pg_ctl's background sleep from holding test stdio** — a pre-existing, unrelated bug in the mock `pg_ctl`: its backgrounded `sleep 3600 &` inherited the shell's stdout/stderr, which is the pipe `CombinedOutput` reads from, so the pipe kept a live writer after `pg_ctl` exited and never reached EOF. Same fix as in #1410, kept as its own commit so it's independently reviewable and green.
2. **fix(pgctld): detect a stale postmaster.pid whose PID was reused** — everything else described above.

## Testing

- `go test -short ./go/cmd/pgctld/...` — green.
- `make build && go test ./go/test/endtoend/pgctld/...` (full package, real compiled binary) — green.
- `golangci-lint run ./...` — 0 issues, full repo.

### DIFF
--- a/go/cmd/pgctld/command/start.go
+++ b/go/cmd/pgctld/command/start.go
@@ -15,7 +15,7 @@
 package command
 
 import (
-	"errors"
+	"context"
 	"fmt"
 	"log/slog"
 	"os"
@@ -28,6 +28,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/multigres/multigres/go/common/constants"
 	"github.com/multigres/multigres/go/common/servenv"
 	"github.com/multigres/multigres/go/services/pgctld"
 )
@@ -113,7 +114,7 @@ func (s *PgCtlStartCmd) runStart(cmd *cobra.Command, args []string) error {
 	config.Password = password
 
 	svc := &PgCtldService{logger: s.pgCtlCmd.lg.GetLogger(), pgConfig: config}
-	result, err := svc.StartPostgreSQLWithResult()
+	result, err := svc.StartPostgreSQLWithResult(cmd.Context())
 	if err != nil {
 		return err
 	}
@@ -129,16 +130,18 @@ func (s *PgCtlStartCmd) runStart(cmd *cobra.Command, args []string) error {
 }
 
 // StartPostgreSQLWithResult starts PostgreSQL with the given configuration and returns detailed result information
-func (s *PgCtldService) StartPostgreSQLWithResult() (*StartResult, error) {
-	result := &StartResult{}
+func (s *PgCtldService) StartPostgreSQLWithResult(ctx context.Context) (*StartResult, error) {
 	logger := s.logger
 	config := s.pgConfig
 
 	// Check if PostgreSQL is already running
-	if isPostgreSQLRunning(config.PostgresDataDir) {
+	running, err := checkPostgreSQLRunning(ctx, logger, config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if PostgreSQL is running: %w", err)
+	}
+	if running {
 		logger.Info("Postgres is already running") //nolint:sloglint // message intentionally starts with an operation name or proper noun
-		result.AlreadyRunning = true
-		result.Message = "PostgreSQL is already running"
+		result := &StartResult{AlreadyRunning: true, Message: "PostgreSQL is already running"}
 
 		// Get PID of running instance
 		if pid, err := readPostmasterPID(config.PostgresDataDir); err == nil {
@@ -148,6 +151,8 @@ func (s *PgCtldService) StartPostgreSQLWithResult() (*StartResult, error) {
 		return result, nil
 	}
 
+	result := &StartResult{}
+
 	// Ensure Unix socket directory exists before starting PostgreSQL
 	// This is necessary for restarts after restores, where pgBackRest only restores pg_data
 	// but not external directories like pg_sockets
@@ -155,7 +160,7 @@ func (s *PgCtldService) StartPostgreSQLWithResult() (*StartResult, error) {
 		if err := os.MkdirAll(config.UnixSocketDirectories, 0o755); err != nil {
 			return nil, fmt.Errorf("failed to create Unix socket directory %s: %w", config.UnixSocketDirectories, err)
 		}
-		logger.Info("ensured Unix socket directory exists", "socket_dir", config.UnixSocketDirectories)
+		logger.InfoContext(ctx, "ensured Unix socket directory exists", "socket_dir", config.UnixSocketDirectories)
 	}
 
 	// Enforce PGDATA permission invariant before pg_ctl start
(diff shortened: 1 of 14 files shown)


mkindahl (commented):
My understanding of the key issue is that this scenario can leave a PID file around:

1. Pod is running and healthy. PID file is on the volume.
2. Pod crashes, which leaves the PID file on the volume. A delete or graceful termination does not cause that.
3. Pod restarts and attaches to the PVC.

We are now in a situation where:

1. Pid file exists. If the PID file is _not_ there, no Postgres can be running, so then there is nothing we need to do.
2. Postmaster can be running or not running.
3. We intend to start Postgres using `pg_ctl start`.
4. We need to decide if we should remove the PID file or not.

Checking the binary name and the start time seems more complicated than necessary. I think we can just use `pg_isready` to check if Postgres is running in this way:
- If `pg_isready` exits with 0 or 1. Postgres is running and there is nothing to do.
- If `pg_isready` exits with 2: Postgres is dead, so delete PID file.
- If `pg_isready` exits with 3: Something is seriously wrong with the control plane and we should flag this as broken.

go/cmd/pgctld/command/start.go
```diff
@@ -241,20 +245,73 @@ func ensurePGDATAPermissions(logger *slog.Logger, dataDir string) error {
...
+	if err != nil || running {
+		return running, err
 	}
 
-	// Read PID from file and check if process is actually running
-	pid, err := readPostmasterPID(dataDir)
-	if err != nil {
-		return false
+	dataDir := config.PostgresDataDir
+	pidFile := filepath.Join(dataDir, "postmaster.pid")
```
mkindahl: We usually define constants for hard-coded values and store them in, for example, `go/common/constants/postgres.go`.

go/cmd/pgctld/command/start.go
```diff
@@ -241,20 +245,73 @@ func ensurePGDATAPermissions(logger *slog.Logger, dataDir string) error {
...
+	pid, _ := readPostmasterPID(dataDir) // best effort, for the log line below only
+	removeStalePostmasterPID(logger, pidFile, pid)
```
mkindahl, line 277: This code is executed when `err == nil && !running` (negation from the early exit above), so it seems to run also when there is no `postmaster.pid` because the server is not running, which seems redundant and also generate a warning message for a normal start.

go/cmd/pgctld/command/stop.go
```diff
@@ -135,16 +146,7 @@ func (s *PgCtldService) StopPostgreSQLWithResult(mode string) (*StopResult, erro
...
-		result.WasRunning = false
-		result.Message = "PostgreSQL is not running"
-		return result, nil
-	}
-
-	result.WasRunning = true
-	logger.Info("stopping Postgres server", "data_dir", config.PostgresDataDir, "mode", mode)
+	logger.InfoContext(ctx, "stopping Postgres server", "data_dir", config.PostgresDataDir, "mode", mode)
 
 	if err := s.stopPostgreSQLWithConfig(mode); err != nil {
```
mkindahl, line 150: Since Stop's goal is just that postgres ends up down, you might want to run `pg_ctl stop` here and if it fails, re-run `probePostgreSQLRunningAndReady` and treat "not running" as success.

go/cmd/pgctld/command/stop.go
```diff
@@ -125,8 +126,18 @@ func (s *PgCtlStopCmd) runStop(cmd *cobra.Command, args []string) error {
...
 
 // StopPostgreSQLWithResult stops PostgreSQL with the given configuration and returns detailed result information
-func (s *PgCtldService) StopPostgreSQLWithResult(mode string) (*StopResult, error) {
-	result := &StopResult{}
+func (s *PgCtldService) StopPostgreSQLWithResult(ctx context.Context, mode string) (*StopResult, error) {
+	// Check if PostgreSQL is running
+	running, err := checkPostgreSQLRunning(ctx, s.logger, s.pgConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if PostgreSQL is running: %w", err)
+	}
```
mkindahl: This could potentially remove the pid file from under the feet of the running Postgres server. It makes sense to check if Postgres is running when you have just attached to the PVC and want to start the server, but not sure it makes sense here. It might be a better idea to attempt the stop below and then if that fails, check if the server is running after. See comment below.

mkindahl (approved):
Looks good. I have a few minor suggestions for improvement, but since it does not affect correctness, I'm approving the PR.

PR #1436, opened 2026-09-03
feat(multipooler): per-reason active reserved-connection metric

GuptaManan100:
## Summary

- Add `mg.pooler.reserved.active_by_reason`, an observable gauge that breaks active reserved connections down by reservation reason (`transaction`, `portal`, `unsafe_connection`, etc.), so operators can see how many unsafe connections — or connections held for any other reason — are currently live.
- Document the `multigres.unsafe_connection` GUC for users, and fix stale references to the removed `--unsafe-pooler-mode` operator flag across the query-serving docs.
- Add unit tests (reserved pool + connpoolmanager) and an end-to-end test that scrapes the metric from a running cluster.

## Description

We recently added the `multigres.unsafe_connection` GUC but shipped no metrics or user docs for it. Rather than a bespoke unsafe-only metric, this adds a single `reason`-labeled gauge that covers unsafe connections and every other reservation reason at once.

**Metric.** `mg.pooler.reserved.active_by_reason{reason, routing_role}` is emitted from the connpoolmanager callback, aggregated across all per-user reserved pools, with every known reason reported each collection (0 when none are active). It lives there — not as a per-pool gauge — because reserved pools are per-user, so a per-pool observable gauge would register duplicate callbacks that don't sum. The breakdown is **overlapping**: a connection holding several reasons at once is counted under each, so the per-reason values sum to more than `reserved.active_connections`; it is not a partition of the total.

`reserved.PoolStats` gains `ActiveByReason`, and `protoutil` exposes `ReasonLabels` as the single source of truth for reason-to-label mapping (`ReasonsString` now uses it too). The generated metric catalog is regenerated to include the new series.

**Docs.** Documents `multigres.unsafe_connection` for users — what it is, when to use it, how to enable it (connect-time via `options=-c ...` and mid-session via `SET`), and its one-way-latch / backend-pinning semantics. Also corrects the query-serving docs, which still described the operator-wide `--unsafe-pooler-mode` flag that was removed in #1424 in favor of the per-connection GUC (now the only opt-out from unsafe-statement rejection).

## Testing

- Unit: reserved `Stats().ActiveByReason` (overlap + drop-on-release), and the connpoolmanager gauge (sum across pools, overlap, zero-emission for inactive reasons).
- End-to-end (`go/test/endtoend/metrics`): drives one gateway connection into unsafe mode inside a transaction, scrapes the primary pooler's Prometheus endpoint, and verifies both `unsafe_connection` and `transaction` are counted while held, the transaction sub-count drops after `COMMIT` while `unsafe_connection` stays pinned (sticky), and `unsafe_connection` returns to baseline on close. Passes under `-count=3`; full metrics suite green.

### DIFF
--- a/docs/query_serving/connection_pooling.md
+++ b/docs/query_serving/connection_pooling.md
@@ -152,6 +152,28 @@ failures for the non-reserved query path.
 - Explicit transactions (`BEGIN`/`COMMIT`/`ROLLBACK`)
 - Cursor operations requiring persistent portal state
 - Any operation requiring connection affinity
+- Unsafe connections (`multigres.unsafe_connection`), pinned to one backend for
+  the session's life via the sticky `ReasonUnsafeConnection` reason
+
+**Active-by-reason metrics:**
+
+Every pin is a reservation _reason_ (`transaction`, `portal`, `temp_table`,
+`copy`, `listen`, `logical_replication`, `session_advisory_lock`, `set_seed`,
+`unsafe_connection`). Two gauges track active reserved connections:
+
+| Metric                                  | Type  | Meaning                                         |
+| --------------------------------------- | ----- | ----------------------------------------------- |
+| `mg.pooler.reserved.active_connections` | Gauge | Total active reserved connections               |
+| `mg.pooler.reserved.active_by_reason`   | Gauge | Active reserved connections holding each reason |
+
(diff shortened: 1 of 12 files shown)


haritabh17 (approved):
Nice 🚀

shedges (approved):
LGTM

PR #1439, opened 2026-09-03
feat(multipooler): derive connpool global capacity from postgres max_connections

haritabh17:
## Description

Part of MUL-497 (infer query-serving flags from PostgreSQL instead of duplicating them).

The pooler's total demand on postgres (`--connpool-global-capacity`, default 100) and pgctld's `max_connections` (110) are hand-balanced twins maintained in files that never reference each other. When they drift, the failure is `sorry, too many clients already` at runtime — under load, the worst possible time.

This PR derives the capacity from the server itself. At admin-pool open, when `--connpool-global-capacity` was not explicitly configured (flag, env var, or config file), the manager runs one query on the admin pool it opened one line earlier:

```
capacity = max_connections − superuser_reserved_connections − reserved_connections − admin_capacity
```

- `reserved_connections` is PG 16+; `missing_ok` + COALESCE yields 0 on older servers. Functions are `pg_catalog`-qualified.
- An explicitly configured value remains authoritative. Explicitness is detected per source: `pflag.Flag.Changed` for the flag, `os.LookupEnv` for `CONNPOOL_GLOBAL_CAPACITY` (the docker image sets this today, so that path is unchanged), and key presence in the loaded config file via a new `viperutil Registry.InStaticConfig` — so a config file pinning the capacity to exactly the default still counts as explicit (viper's `IsSet` can't do this: it is unconditionally true for registered values because `Configure` installs defaults via `SetDefault`).
- Any derivation failure falls back to the configured value with a warning — `Open` is infallible by design, and the postgres monitor reopens the manager (re-running the derivation) after every postgres restart.
- A tiny budget (e.g. a derived capacity clamped to 1) previously truncated the regular pool's share to 0, letting the rebalancer disable ordinary queries entirely; the split now guarantees each pool class at least one connection, overshooting an already-unusable budget by one connection as the lesser evil.
- The `multigres_connpool_config_max_server_connections` metric now reports the value in effect (derived or configured) instead of always the configured flag.

On a default deployment (pgctld's `max_connections = 110`, PG default `superuser_reserved_connections = 3`, admin capacity 5) the derived budget is 102 — and it now follows `max_connections` automatically when an operator raises it.

## Testing

- New unit tests: derivation math incl. clamp-to-1; Open derives against fakepgserver (also asserted as the sum of both allocator capacities); derivation failure falls back to the configured value; explicit env value skips the query entirely; config-file-pinned-to-default counts as explicit (driven through the production `ViperConfig.LoadConfig` path); tiny-budget open keeps the regular pool usable across a rebalance (verified to fail pre-fix).
- `go test -short ./go/services/multipooler/...` and `./go/tools/viperutil/...` green; race-run of the new tests green.
- Integration: `go/test/endtoend/multipooler` and `.../backupfaults` pass against real postgres.

### DIFF
--- a/go/services/multipooler/internal/connpoolmanager/manager.go
+++ b/go/services/multipooler/internal/connpoolmanager/manager.go
@@ -247,8 +247,12 @@ func (m *Manager) Open(ctx context.Context, connConfig *ConnectionConfig) {
 	m.globalCapacity.Store(globalCapacity)
 	reservedRatio := m.config.ReservedRatio()
 	minPerUser := m.config.MinCapacityPerUser()
-	regularCapacity := int64(float64(globalCapacity) * (1 - reservedRatio))
-	reservedCapacity := globalCapacity - regularCapacity
+	// Guarantee each pool class at least one connection: a tiny budget (e.g. a
+	// derived capacity clamped to 1) would otherwise truncate the regular share
+	// to 0 and disable ordinary queries entirely. For such an already-unusable
+	// budget, overshooting it by one connection is the lesser evil.
+	regularCapacity := max(int64(float64(globalCapacity)*(1-reservedRatio)), 1)
+	reservedCapacity := max(globalCapacity-regularCapacity, 1)
 	regularMinPerUser := max(int64(float64(minPerUser)*(1-reservedRatio)), 1)
 	reservedMinPerUser := max(minPerUser-regularMinPerUser, 1)
 
--- a/go/services/multipooler/internal/connpoolmanager/manager.go
+++ b/go/services/multipooler/internal/connpoolmanager/manager.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log/slog"
 	"maps"
+	"strconv"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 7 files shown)


go/services/multipooler/internal/connpoolmanager/manager.go
```diff
@@ -141,6 +142,12 @@ type Manager struct {
 	regularAllocator  *FairShareAllocator
 	reservedAllocator *FairShareAllocator
 
+	// globalCapacity is the total connection budget in effect, resolved in Open:
+	// derived from the server's max_connections unless explicitly configured
+	// (see resolveGlobalCapacity). Atomic because the metrics callback reads it
+	// concurrently. Zero until the first Open.
+	globalCapacity atomic.Int64
```
shedges: it will same as max_connections?

go/services/multipooler/internal/connpoolmanager/manager.go
```diff
@@ -236,11 +243,16 @@ func (m *Manager) Open(ctx context.Context, connConfig *ConnectionConfig) {
...
+	m.globalCapacity.Store(globalCapacity)
 	reservedRatio := m.config.ReservedRatio()
 	minPerUser := m.config.MinCapacityPerUser()
-	regularCapacity := int64(float64(globalCapacity) * (1 - reservedRatio))
-	reservedCapacity := globalCapacity - regularCapacity
+	// Guarantee each pool class at least one connection: a tiny budget (e.g. a
+	// derived capacity clamped to 1) would otherwise truncate the regular share
+	// to 0 and disable ordinary queries entirely. For such an already-unusable
+	// budget, overshooting it by one connection is the lesser evil.
+	regularCapacity := max(int64(float64(globalCapacity)*(1-reservedRatio)), 1)
```
shedges: can we have something configurable like minCapacity instead of hardcoded 1

shedges (approved):

PR #1442, opened 2026-09-03
feat(multipooler): add prepared-statement, advisory-lock, and temp-object scrubber checkers

haritabh17:
Adds the `connpool.ConnChecker`s the #1415 design left open — prepared statements, advisory locks, temp-schema objects, and (from review) holdable cursors — so the pool scrubber covers every class of backend state the pool trusts without verification (MUL-1495).

## Why

#1415 shipped the scrubber with a pluggable checker interface and a single `session_state` checker. The pool also trusts four other things about an idle backend that nothing verified: that its prepared statements match what tracking recorded, that it holds no session-level advisory lock, that no cursor outlived its transaction, and that its temporary schema is empty. Each can be violated by the same failure class the scrubber exists for — state created behind tracking's back, typically inside a routine body.

## Checkers

- **`prepared_statements`** — the only two-sided diff. Idle connections legitimately keep prepared statements across borrowers (tracked in `ConnectionState`, keyed by consolidator name), so the checker compares tracked statements against `pg_catalog.pg_prepared_statements`, **names and bodies**: a backend statement tracking does not know is *untracked*, a tracked statement the backend lost is *phantom*, and a tracked statement whose backend body differs from the tracked query is *mismatched*. The body comparison matters: a hidden `DEALLOCATE` plus re-`PREPARE` keeps a tracked `ppstmt<N>` name with a new body, and `ensurePrepared` would otherwise hand the redefined statement to the next borrower's `EXECUTE` (verified on PostgreSQL 17 from a PL/pgSQL body). The unnamed statement is skipped since the view never lists it. Tracked names are consolidator identifiers and reported verbatim; **every untracked name is redacted** to the bounded label `foreign_name`, one entry per statement so counts survive — a name that reached a backend by another route may embed client data, and the consolidator's `ppstmt<N>` shape is reachable from any session's `PREPARE`, so it is no safety signal.
- **`holdable_cursors`** — an idle pooled backend must have none. Outside a transaction `pg_catalog.pg_cursors` lists only `WITH HOLD` cursors, and portals pin a reserved connection, so an open cursor here was declared behind tracking (verified: `EXECUTE 'DECLARE ... WITH HOLD'` inside a PL/pgSQL body leaves the cursor open after the function returns). Reports one `holdable_cursor` finding per cursor; names are never reported.
- **`advisory_locks`** — an idle pooled backend must hold none: acquiring functions route to a reserved connection whose release runs `pg_advisory_unlock_all`. The checker runs the existing `PgLocksAdvisoryProbeSQL` and reports a single `session_advisory_lock` finding. Lock keys are never reported.
- **`temp_objects`** — an idle pooled backend must own none: temp statements pin a reserved connection that is closed at release, and pg_temp-qualified CREATE is rejected at the gateway. The checker scans every namespace-scoped catalog the gateway's pg_temp CREATE rejection covers — `pg_class`, `pg_proc` (aggregates included), standalone `pg_type` entries (domain, enum, range, multirange; composites via `pg_class`, auto-created array types skipped), `pg_operator`, `pg_collation`, `pg_statistic_ext`, `pg_opclass`, `pg_opfamily`, `pg_conversion`, and the four text-search catalogs — in one round trip, and reports one finding per kind (`table`, `index`, `sequence`, `function`, `domain`, `enum`, `operator`, `collation`, `statistics`, ...). Object names are never reported.

  Severity, measured on a live PostgreSQL 17 rather than reasoned: relations and types are the dangerous classes, since pg_temp is searched before `pg_catalog` for unqualified relation and type names (a `pg_temp` domain named `text` captures an unqualified `::text`, verified by OID; keyword type names like `int` cannot be shadowed). Operators, collations, and the other classes are never resolved through pg_temp, even with `pg_temp` listed first in `search_path`; they are reported as stale state for completeness with the guard, not as a shadowing hole.

All of them mirror `VerifySessionState`: refuse a non-idle connection, use plain `Query` (a retry would reconnect and reset the state under inspection), and treat malformed rows as probe errors, which the scrubber handles fail-closed.

## Also

- `PgLocksAdvisoryProbeSQL` is now schema-qualified (`pg_catalog.pg_locks`, `pg_catalog.pg_backend_pid`), the same shadowing fix applied to the session probe in #1415.
- The `ConnChecker` doc still described the pre-#1415-review "connection is not punished" error path; it now states the fail-closed behavior.
- `--connpool-session-scrub-interval` help text names all five state classes.
- Each checker now gets its own `scrubProbeTimeout` budget. `scrubOne` used to create one five-second context before the checker loop, so a slow early probe could starve the later ones into a timeout that the fail-closed error path turns into a needless replacement. Pool close still cancels the parent scrub context before draining, so shutdown is unaffected.
- Docs: one paragraph per checker in the scrubber section of `connection_pooling.md`.

## Testing

- Unit: each checker against fakepgserver (in sync, untracked/phantom, redefined body, redaction of every untracked name including a `ppstmt<digits>` one, probe error, malformed boolean, cursor count-not-names, temp kinds-not-names), plus the new `ConnectionState.PreparedStatementNames` accessor. Race-clean.
- E2E: new `TestSessionScrubberReplacesHiddenAdvisoryLock` — a `pg_advisory_lock` hidden in a SQL function body installed directly on the primary leaves a lock on an unpinned pooled backend; the scrubber detects it under the `advisory_locks` checker, replaces the backend, and the log never carries the lock key. New `TestSessionScrubberReplacesHiddenTempObjects` hides a domain, an operator, a collation, and a statistics object created in pg_temp inside one SQL function body: all four kinds are detected under `temp_objects` on one sweep, the backend is replaced, and the object names are absent from the log. One test rather than one per class keeps suite time flat, since each scrubber test can wait a full stack rotation. New `TestSessionScrubberReplacesHiddenHoldableCursor` hides a `WITH HOLD` cursor in a PL/pgSQL body and asserts replacement under `holdable_cursors` with the cursor name absent from the log.
- Full queryserving suite green with all five checkers on, re-run after each review round. The last run also confirms the body comparison produces no false mismatches against the suite's PREPARE-heavy traffic.
- Every `pg_temp`-qualified CREATE the gateway guard rejects was exercised against a scratch PostgreSQL 17 to confirm each object class is creatable there, lands in its expected catalog, and is returned by the probe; the same run confirmed the shadowing semantics above. The existing scrubber canary (zero divergence from normal traffic, read across the whole suite's log) confirms the new checkers produce no false positives against the suite's PREPARE-heavy traffic.

A first full-suite run exposed that the scrubber's stack rotation (one connection per 10s tick across the clean stack plus 16 settings stacks) can take up to 170s to reach a given stack once earlier tests have populated many buckets; the advisory divergence fired at 89s against a 90s wait. Both scrubber e2e tests now wait `scrubSweepWait` (4 min) with the reasoning in a comment.

## Follow-ups (not in this PR)

The `session_state` checker still reports untracked custom GUC names verbatim, and a hidden `set_config('my.' || value, ...)` puts a value into a name — the same exposure this PR redacts for prepared statements. Worth deciding once and applying there too.

Within a stack the scrubber only ever probes the top connection: it pops the top, checks it, and pushes it back on top. Colder connections deeper in a stack are never probed unless client traffic reorders them. That coverage gap predates this PR and deserves its own task.

### DIFF
--- a/go/services/multipooler/internal/pools/regular/pool.go
+++ b/go/services/multipooler/internal/pools/regular/pool.go
@@ -63,6 +63,7 @@ func NewPool(ctx context.Context, config *PoolConfig) *Pool {
 	pool.RegisterChecker(SessionStateChecker{})
 	pool.RegisterChecker(PreparedStatementChecker{})
 	pool.RegisterChecker(AdvisoryLockChecker{})
+	pool.RegisterChecker(HoldableCursorChecker{})
 	pool.RegisterChecker(TempObjectChecker{})
 
 	return &Pool{
--- a/go/services/multipooler/internal/pools/regular/pool.go
+++ b/go/services/multipooler/internal/pools/regular/pool.go
@@ -61,6 +61,9 @@ func NewPool(ctx context.Context, config *PoolConfig) *Pool {
 
 	pool := connpool.NewPool[*Conn](ctx, config.ConnPoolConfig)
 	pool.RegisterChecker(SessionStateChecker{})
+	pool.RegisterChecker(PreparedStatementChecker{})
+	pool.RegisterChecker(AdvisoryLockChecker{})
+	pool.RegisterChecker(TempObjectChecker{})
 
 	return &Pool{
 		pool:   pool,
--- a/docs/query_serving/connection_pooling.md
+++ b/docs/query_serving/connection_pooling.md
@@ -622,18 +622,26 @@ Three more checkers cover the rest of the backend state the pool trusts:
   against `pg_prepared_statements`. Idle connections legitimately keep
   prepared statements across borrowers, so this is a two-sided comparison:
   a backend statement tracking does not know is untracked, a tracked
-  statement the backend lost is phantom. Names are consolidator identifiers,
-  never client SQL.
+  statement the backend lost is phantom. Tracked names are consolidator
+  identifiers and are reported verbatim; an untracked name is reported
+  verbatim only when it has the consolidator's shape (a tracking bug),
+  otherwise it is redacted to `foreign_name`, since a name that arrived on a
+  backend by another route may embed client data.
 - `advisory_locks` reports a single `session_advisory_lock` finding when the
   idle backend holds any advisory lock. Acquiring functions route to a
   reserved connection whose release runs `pg_advisory_unlock_all`, so a lock
   on an idle pooled backend means the acquisition escaped tracking. Lock keys
   are never reported.
 - `temp_objects` reports one finding per kind of object (`table`, `index`,
-  `sequence`, `function`, ...) in the backend's temporary schema. Temp
-  statements pin a reserved connection that is closed at release and
-  pg_temp-qualified CREATE is rejected, so any object on an idle backend
-  escaped tracking. Object names are never reported.
+  `sequence`, `function`, `domain`, `enum`, `range`, ...) in the backend's
+  temporary schema, covering `pg_class`, `pg_proc`, and standalone `pg_type`
+  entries — the same object classes the gateway's pg_temp CREATE rejection
+  recognizes. Temp statements pin a reserved connection that is closed at
+  release and pg_temp-qualified CREATE is rejected, so any object on an idle
+  backend escaped tracking. A leftover temp type is not inert: pg_temp is
+  searched before `pg_catalog` for unqualified type names as well as
+  relation names, so it would shadow a catalog type for the next borrower.
+  Object names are never reported.
 
 All four checkers run against the same idle connection each tick; the
 divergence log line and the `checker` metric attribute name which one fired.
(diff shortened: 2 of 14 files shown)


go/services/multipooler/internal/pools/regular/pool.go
```diff
@@ -61,6 +61,9 @@ func NewPool(ctx context.Context, config *PoolConfig) *Pool {
 
 	pool := connpool.NewPool[*Conn](ctx, config.ConnPoolConfig)
 	pool.RegisterChecker(SessionStateChecker{})
```
shedges, line 63: Do we want to add check for WITH HOLD cursor also? Currently we added unsafe connection mode where user can perform `EXECUTE 'DECLARE ... WITH HOLD ...'`. It can leak cursor.

shedges (approved):
Rest LGTM

go/services/multipooler/internal/pools/regular/pool.go
haritabh17: Good point, added now.

PR #1444, opened 2026-09-04
fix(test): fail fast waiting for dead processes to start, add testpoll helpers

dweitzman:
## Summary
- Add `Cmd.IsRunning()`, used by `waitForMultigatewayReady`/`WaitForPortReady`/`waitForPort` to fail immediately when a process crashed on startup, instead of polling a dead target for the full timeout.
- Add `testpoll.WaitFor`: a synchronous `assert.Eventually` alternative that fails the test immediately on timeout and lets `condition` call `require.*`/`t.Fatalf` directly to bail out even earlier. Used by `WaitForManagerReady`.
- Add `testpoll.Never` and ban `assert.Never`/`require.Never` via ruleguard: testify's `Never` spawns a goroutine per tick with no join, and its success path is exactly the case where that goroutine is still running when it returns — raced with `reassert_topo_test.go`'s deferred store teardown under `-race`. Technically `assert.Eventually()` can also leave goroutines still running after it returns, but it's more common for `assert.Never()` to be the last line in a test and that can leave the test still doing things after it already returned.

Both surfaced investigating the same CI run on an unrelated PR:
- [Race Detection Tests](https://github.com/multigres/multigres/actions/runs/33786787464/job/100753385899?pr=1289) — the `assert.Never` goroutine-leak race
- [Multigres Integration Tests](https://github.com/multigres/multigres/actions/runs/33786787434/job/100753385905?pr=1289) — a 30-min timeout that motivated failing fast on a dead process


go/tools/executil/command.go
```diff
@@ -426,6 +426,20 @@ func (c *Cmd) Resume() error {
...
+func (c *Cmd) IsRunning() bool {
+	if c == nil || c.Process == nil {
+		return false
+	}
+	// ProcessState is set once Wait() returns, meaning the process has exited.
+	if c.ProcessState != nil {
+		return false
+	}
+	// Signal 0 checks whether the process exists without actually sending one.
+	return c.Process.Signal(syscall.Signal(0)) == nil
```
Verolop, line 451: I think this may still report a process as running after it exits. Until Wait() reaps it, ProcessState stays nil and signal 0 succeeds. Maybe start Wait() in the background and use waitDone here?

go/tools/testpoll/poll.go
```diff
@@ -0,0 +1,105 @@
...
+	t.Helper()
+
+	// Passed to condition every tick (not re-created per attempt) so it can
+	// bound its own work, e.g. an RPC call that should give up once the
+	// budget is spent rather than block past it.
+	ctx, cancel := context.WithTimeout(context.Background(), waitFor) //nolint:gocritic // deliberate root context; WaitFor owns its own timeout independent of any caller context
+	defer cancel()
+
+	for {
+		if condition(ctx) {
```
Verolop: potential edge case (so non-blocking): if the tick takes us past the deadline, we still run the condition once more. That lets WaitFor pass late, and makes Never fail late.

Verolop (commented):
Looks good overall; I think it's worth tweaking the `IsRunning` before merging.

go/tools/executil/command.go
dweitzman: Good catch, I made some significant changes here, renaming `IsRunning` to `IsRunningOrZombie` and updating some of the callers to do an explicit `Wait()` if they want to detect zombies

go/tools/testpoll/poll.go
dweitzman: Fixed

Verolop (approved):
lgtm!

PR #1445, opened 2026-09-04
feat(multipooler): derive --socket-file from pooler-dir and pg-port

haritabh17:
## Description

Part of MUL-497 (infer query-serving flags instead of duplicating them); closes MUL-1518.

The postgres Unix socket path is a pure function of two other flags — `<pooler-dir>/pg_sockets/.s.PGSQL.<pg-port>` — but multipooler required `--socket-file` to be passed explicitly, and the formula was hand-copied in five places: pgctld's `PostgresSocketDir` (the value actually passed to postgres as `unix_socket_directories`), the pooler's pgBackRest path, two provisioner sites, and the shardsetup test helper.

**The formula now lives in one canonical helper** (`go/common/constants`: `PostgresSocketDir` / `PostgresSocketFilePath` — `common` because services cannot import each other) and every copy routes through it; pgctld keeps a thin alias so its ~9 call sites are untouched.

**Derivation semantics** (`resolveSocketFilePath` in multipooler init):
- An explicitly set `--socket-file` wins — *including explicitly empty* (`--socket-file=''`), which forces a TCP dial.
- Unset flag + `--pooler-dir` configured → the path is derived, so the pooler and pgctld can never disagree about where the socket lives.
- Neither → empty, TCP as before.

**Behavior change worth noting for external manifests:** previously an omitted `--socket-file` always meant TCP. After this change, a deployment that sets `--pooler-dir` but omits `--socket-file` flips from a TCP dial to the colocated socket that pgctld itself configures. All in-repo callers are updated (the local provisioner and the k8s demo manifest now omit the flag deliberately to exercise derivation; shardsetup passes explicit-empty for its PG-TLS TCP tests). An external manifest that wants to keep TCP sets `--socket-file=''`.

Also fixes the stale `--pooler-dir` help text that claimed a socket-file fallback which never existed (cleanup item from the MUL-497 audit).

## Testing

- Unit: table-driven test for `resolveSocketFilePath` (derive / explicit-wins / explicit-empty-forces-TCP / no-pooler-dir); full `go test -short ./go/...` green.
- Integration: full `go/test/endtoend/...` suite green locally (two packages that failed under machine-wide shared-memory exhaustion pass cleanly in isolation). `localprovisioner` now boots every pooler with **no** `--socket-file`, exercising derivation end-to-end; `TestPGClientTLS_*` exercise the explicit-empty TCP opt-out.
- Manual: 3-pooler local cluster — all poolers logged the derived path, `.s.PGSQL.<port>` present at the derived location, direct-socket psql and full gateway→pooler→postgres write/read verified.

### DIFF
--- a/go/services/multipooler/init.go
+++ b/go/services/multipooler/init.go
@@ -267,8 +267,8 @@ func (mp *Multipooler) RegisterFlags(flags *pflag.FlagSet) {
 	flags.String("table-group", mp.tableGroup.Default(), "table group this multipooler serves (required)")
 	flags.String("shard", mp.shard.Default(), "shard this multipooler serves (required)")
 	flags.String("service-id", mp.serviceID.Default(), "optional service ID (if empty, a random ID will be generated)")
-	flags.String("socket-file", mp.socketFilePath.Default(), "PostgreSQL Unix socket file path (if empty, TCP connection will be used)")
-	flags.String("pooler-dir", mp.poolerDir.Default(), "pooler directory path (if empty, socket-file path will be used as-is)")
+	flags.String("socket-file", mp.socketFilePath.Default(), "PostgreSQL Unix socket file path. If unset, derived as <pooler-dir>/pg_sockets/.s.PGSQL.<pg-port> when --pooler-dir is set; set explicitly to empty (--socket-file='') to force a TCP connection")
+	flags.String("pooler-dir", mp.poolerDir.Default(), "pooler directory path")
 	flags.Int("pg-port", mp.pgPort.Default(), "PostgreSQL port number")
 	flags.Int("heartbeat-interval-milliseconds", mp.heartbeatIntervalMs.Default(), "interval in milliseconds between heartbeat writes")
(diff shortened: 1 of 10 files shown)


GuptaManan100 (approved):
LGTM!

go/services/multipooler/init.go
```diff
@@ -426,11 +426,19 @@ func (mp *Multipooler) Init(startCtx context.Context) error {
...
+	}
+
 	// Validate libpq-style sslmode + sslrootcert before any pool opens. A typo
 	// or missing CA bundle should fail startup rather than silently downgrading
 	// the multipooler → postgres dials to plaintext.
 	pgHost := ""
-	if mp.socketFilePath.Get() == "" {
+	if socketFilePath == "" {
 		pgHost = mp.senv.GetHostname()
 	}
```
GuptaManan100, line 449: Will this ever trigger? Won't we just derive the value?

haritabh17: Currently the TCP opt-out e2e tests use this and they need `pgHost` set so `ValidatePGSSL` can check verify-full against the hostname and a setup with no `--pooler-dir`, where there's nothing to derive from.

shedges (approved):
🚀

PR #1446, opened 2026-09-04
fix(multipooler): probe every idle connection during pool scrubbing

haritabh17:
Closes a coverage gap in the pool scrubber from #1415 (MUL-1521): it only ever probed the top connection of each idle stack.

## Why

The scrubber popped the top of the next non-empty stack, ran the checkers, and pushed the connection back on top. Clients also take and return connections at the top, so the same hot connection was probed every tick and any connection beneath it was never checked — the sampler had a blind spot exactly where cold, long-idle backends sit. Observed in #1442's suite runs: a contaminated backend that was not at the top of its stack could wait far longer than a stack rotation.

## Design

Selection now works in **passes**, without touching stack order:

- `Pool.scrubPass` numbers the passes; `Pooled.scrubPass` records the pass in which a connection was last probed. Both are touched only by the scrub worker, so they need no atomics — with one exception: `connReopen` resets the mark to zero, since the mark belonged to the old backend and the new one has not been probed. That write happens while the connection is borrowed (the scrubber cannot hold it) and the push that follows goes through the stack mutex, which orders it before any scrubber read.
- `connStack.PopFirst(match)` unlinks the topmost connection for which `match` holds, leaving the rest in place, and reports the **depth** it was found at. Middle-node removal is safe because every stack operation, including `ForEach`, holds the mutex; the old caveat about only removing the head dates from the lock-free design. `connStack.InsertAt(conn, depth)` puts a connection back below the first `depth` entries (clamped to the bottom; depth 0 is `Push`).
- `scrubPop` takes the topmost connection not yet probed this pass (and whose idle-clock borrow succeeds) and marks it. `scrubNext` walks the stacks from the rotation cursor; if every idle connection is already marked it starts the next pass and looks once more, so a tick only comes up empty when the pool has no idle connection at all.
- The probed connection returns to its stack **at the depth it was taken from**, with its idle clock restored, via `returnConnAt(conn, depth)` — `tryReturnConn` is now `returnConnAt(conn, 0)`. Over-capacity close, direct waiter handoff, idle-limit close, and settings-bucket routing are unchanged; only the final placement takes a depth. Clients may push or pop between the probe's pop and its reinsert, so the depth is best effort for that window, exact in the case that matters (hot top, cold tail), and never lifts a connection above one that was above it.

Coverage bound per connection is now the total number of idle connections times the tick, regardless of position.

### Why the depth matters (two rejected placements)

Any placement that moves a probed connection relative to its neighbours changes which connection clients reuse next, and with traffic between ticks that refreshes a cold connection's idle clock. Returning to the *bottom* rotates the stack: every tick demotes the hot connection and promotes a colder one. Returning a middle connection to the *top* (the first cut of this PR, caught in review) promotes it once per pass. Either way, over a pass every idle connection gets refreshed by real use, none idle-expires, and the pool sits at its maximum idle size instead of shrinking to its working set — reintroducing through traffic exactly what the scrubber's idle-clock restore protects against. Coverage has to come from remembering what was probed, and placement has to leave the stack order alone.

## Testing

- Unit: `TestStackPopFirst` (head, middle, tail, no-match, empty; depth reporting; `InsertAt` at each depth, past the end, and on an empty stack; metrics callbacks fire once per operation). `TestScrubWalksEveryConnInStack` (three idle connections, each probed exactly once over three ticks, then the pass wraps). `TestScrubPassSpansStacks` (a pass covers the clean stack and a settings stack before repeating, and probed connections stay in their own stacks). `TestScrubReopenedConnIsProbedAgainInSamePass` (two connections; probe one, reopen it in place, and the next tick probes its new backend rather than moving on — fails without the reset). `TestScrubKeepsHotConnClientFacingAndColdConnsExpire` (cold idle stamps on the two lower connections, one client checkout after each of three ticks asserting the same hot connection is handed out, then `closeIdleResources` asserting both cold ones expired and the hot one survived — this fails on the top-return cut). Whole connpool package race-clean across repeated runs; scrub and stack tests across five.
- E2E: all four scrubber tests pass standalone on this branch. The `scrubSweepWait` comment is updated to describe the pass-based bound.
- Docs: scrubber section of `connection_pooling.md` describes pass-based selection and why the probed connection returns to the top.

Full queryserving suite: two runs on this branch failed only in cluster bootstrap ("Manager should become ready within 40s", `initdb` exiting non-zero), with a separate local cluster running on the same machine since 15:30 and load average around 10. All six affected tests pass standalone, their pooler logs show zero scrubber activity, and every full run before that cluster started was green. Will re-run clean before undrafting.

### DIFF
--- a/go/services/multipooler/internal/pools/connpool/scrub.go
+++ b/go/services/multipooler/internal/pools/connpool/scrub.go
@@ -31,6 +31,15 @@ import (
 // It only touches idle connections, so it adds no latency to the query path.
 // It is a sampler: it narrows the leak window and raises an alarm, but the
 // tracking gates remain the correctness boundary.
+//
+// Coverage: the scrubber works in passes. Each tick probes, from the next
+// non-empty stack, the topmost idle connection not yet probed in the current
+// pass, and marks it with the pass number; once every idle connection in
+// every stack carries the current pass number, the next pass begins. Probed
+// connections return to the top of their stack exactly as a recycled one
+// would, so LIFO reuse and idle-timeout shrinking are untouched — simply
+// re-probing the top each tick would never reach the connections beneath
+// it, since clients also take and return connections at the top.
 
 // scrubProbeTimeout bounds one checker's probe; each checker gets its own
 // budget, so a slow early probe cannot starve later checkers into a timeout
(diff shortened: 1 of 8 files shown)


go/services/multipooler/internal/pools/connpool/scrub.go
```diff
@@ -52,46 +66,77 @@ func (pool *Pool[C]) scrubStack(i int) *connStack[C] {
...
+// the idle-timeout worker could never shrink it. The stamp read before
+// borrow is reliable: the only concurrent writer for an in-stack connection
+// is the expirer, and if it won the race the borrow fails and the connection
+// (already closed by the expirer) is left for pool.pop to discard. Returns
+// false when every idle connection in the stack has already been probed
+// this pass.
+func (pool *Pool[C]) scrubPop(stack *connStack[C]) (scrubbed[C], bool) {
+	var s scrubbed[C]
+	conn, depth, ok := stack.PopFirst(func(c *Pooled[C]) bool {
+		if c.scrubPass == pool.scrubPass {
```
shedges, line 91: Minor : we need to reset `scrubpass` after `connReopen`. this will handled in next round.

shedges (approved):
Rest LGTM

go/services/multipooler/internal/pools/connpool/scrub.go
haritabh17: Ack, updated.

PR #571, opened 2026-02-02
chore(pgregress): make pg_regress tests opt-in

haritabh17:
This makes the pg_regress tests opt in:

```
haritabhgupta@Haritabhs-MBP query-serving-copy-stdin % go test -v ./go/test/endtoend/pgregresstest/...
=== RUN   TestPostgreSQLRegression
    pgregress_test.go:40: skipping pg_regress tests (set RUN_PGREGRESS=1 to run)
--- SKIP: TestPostgreSQLRegression (0.00s)
PASS
ok      github.com/multigres/multigres/go/test/endtoend/pgregresstest   0.698s
```

Runs with the env `RUN_PGREGRESS=1`:
```
haritabhgupta@Haritabhs-MBP query-serving-copy-stdin % RUN_PGREGRESS=1 go test -v ./go/test/endtoend/pgregresstest/...
=== RUN   TestPostgreSQLRegression
    main_test.go:27: Starting etcd for topology...
...
--- PASS: TestPostgreSQLRegression (76.60s)
    --- PASS: TestPostgreSQLRegression/setup_postgres_source (0.02s)
    --- PASS: TestPostgreSQLRegression/build_postgres (64.42s)
    --- PASS: TestPostgreSQLRegression/run_regression_tests (1.11s)
PASS
ok      github.com/multigres/multigres/go/test/endtoend/pgregresstest   77.662s
haritabhgupta@Haritabhs-MBP query-serving-copy-stdin % 
```

PR #1447, opened 2026-09-04
feat(multiadmin): add client-cert authentication for the HTTP listener

kabochya:
## Summary

multiadmin's HTTP listener (port 18000) serves destructive operations — `ExpireBackups`, `SwitchPrimary` — with no authentication of its own. This lets it require and verify client certificates against a subject allow-list.

Entirely opt-in: with none of these flags set, the code path is identical to today.

| Flag | Purpose |
| --- | --- |
| `--http-cert` / `--http-key` | TLS on the HTTP listener (mirrors `--grpc-cert`/`--grpc-key`) |
| `--http-ca` | CA used to verify client certificates |
| `--http-auth-mtls-allowed-subjects` | Authorized subject, repeat per subject (`CN=ns-a,O=acme`) |
| `--enable-http-mtls-auth` (multiadmin) | Enforce client identity |

## Design notes

**Subjects are compared attribute-by-attribute for exact equality.** A substring test can't express "exactly this identity" — pkix escapes commas but not `=`, so CN `evilCN=ns-team-a` renders as `CN=evilCN=ns-team-a,O=acme` and contains the entry `CN=ns-team-a,O=acme` outright. Anchoring doesn't help. Entries and the certificate's rendered subject are both tokenized by `ldap.ParseDN`, so escaping matches pkix and OID-rendered attributes are allow-listable like any other.

**The allow-list is its own flag, not gRPC's,** and repeatable rather than delimited. One shared list would mean enabling `--grpc-auth-mode=mtls` later silently grants every HTTP-side identity access to the gRPC admin RPCs; entries contain commas by construction, which is also why `--grpc-auth-jwt-allowed-subs` is repeatable. The gRPC substring path is unchanged and documents that it is unsuitable as a trust boundary.

**Probes share this port,** so the handshake uses `tls.VerifyClientCertIfGiven`. Identity is enforced per-route by middleware instead (a Connect interceptor can't see TLS state), with `/live`, `/ready` and `/version` exempt.

**The `/proxy/admin/...` self-proxy follows the listener's scheme,** trusting this listener's certificate rather than the system pool and presenting `--http-cert`/`--http-key` when enforcement is on — so that subject must also be allow-listed. Hops to cell services stay plaintext; their TLS settings aren't advertised in the topology.

**`ServEnv.validateHTTPTLS` fails startup** on unloadable TLS material, client-cert auth without the cert/key/CA that make it enforceable, or a missing/malformed allow-list.

## Known limitation

Matching is a **subset** test: unnamed attributes and values are unconstrained, so a subject carrying everything named plus more still matches. Sound where the CA issues a unique identity per caller; not sufficient where it issues non-unique CNs *and* lets a requester append attributes. Exact-set matching per attribute is a deliberate follow-up. Rejections log the presented subject, so a mismatch can be read against the configured entry.

## Dependency

`go-ldap/ldap/v3` does the RFC 4514 tokenizing. It's heavier than a DN parser suggests — 206 linked packages including `Azure/go-ntlmssp` and `go-asn1-ber`, plus `golang.org/x/crypto`, `net`, `sys`, `term` and `text` each moving forward a minor version. Hand-rolling the unescaping would avoid that; a maintained parser was preferred. `govulncheck` is green.

## Before enabling in a real deployment

- `--http-cert`/`--http-key` flips the **whole** listener to TLS, including `/live` and `/ready`. Probes with `httpGet` and no `scheme` default to HTTP and will fail — `demo/k8s/k8s-multiadmin.yaml` is in exactly that shape and needs `scheme: HTTPS`.
- The web UI is a browser Connect-web client with no certificate, so enforcement makes it unusable without a terminating proxy.
- The hop from whatever proxy fronts multiadmin is plain HTTP today and needs TLS wiring first.

Operator wiring is follow-up work, extending `MultiadminAuthConfig` from multigres-operator #571. The flag is a viperutil value, so it can live in the config file as a YAML list.

## Test plan

- [x] `go build ./...`, `gofmt`, `golangci-lint` on touched packages — clean
- [x] `go test -short ./go/common/... ./go/services/... ./go/cmd/...` — pass
- [x] `go test -race ./go/common/servenv/` — no data races
- [x] End to end on localhost with self-signed certs: allow-listed client 200; different tenant from the same CA 401; no certificate 401; `/live` and `/ready` 200; self-proxy over TLS 200; plaintext against the TLS listener 400
- [x] Rebased onto current `upstream/main`; re-verified against the new base


go/common/servenv/client_cert.go
```diff
@@ -0,0 +1,59 @@
...
+// allow-list is a trust boundary between holders of certs from one CA, anchor
+// entries on the next RDN's delimiter ("CN=ns-team-a,O=..."); pkix escapes
+// commas inside values, so a caller cannot forge that boundary from within its
+// own CN. A bare-CN subject has no trailing delimiter and cannot be anchored at
+// all. Comparing parsed identities is the real fix, and would change the
+// meaning of every existing allow-list.
+func certSubjectMatches(certs []*x509.Certificate, substrings []string) bool {
+	for _, substring := range substrings {
+		for _, cert := range certs {
+			if strings.Contains(cert.Subject.String(), substring) {
```
Verolop, line 46: can you please match the parsed identity exactly here? Adding O=supabase is not enough, e.g. CN=ns-team-a,O=supabase still accepts a cert whose CN is evilCN=ns-team-a and org is supabase. Tested this and I got a 200.
Since HTTP has its own allow-list, this can be done without changing gRPC’s behaviour. Can you please also add a test that rejects another tenant’s valid cert from the same CA?

go/common/servenv/http.go
```diff
@@ -58,15 +62,36 @@ func corsMiddleware(next http.Handler) http.Handler {
...
+	tlsConfig, err := grpccommon.BuildServerTLSConfigWithClientAuth(
+		sv.httpCert.Get(), sv.httpKey.Get(), sv.httpCA.Get(), "", tls.VerifyClientCertIfGiven,
+	)
+	if err != nil {
+		return fmt.Errorf("http tls config: %w", err)
+	}
+
+	if tlsConfig != nil {
+		server.TLSConfig = tlsConfig
+		err = server.ServeTLS(l, "", "")
```
Verolop, line 91: Multiadmin’s self-proxy still builds an http:// url when forwarding back to itself. With `--http-cert` and `--http-key enabled`, direct `/config` returned 200, but `/proxy/admin/global/config` returned 400 with “client sent an HTTP request to an HTTPS server.”
Can you please make that hop work with HTTPS too, including client-certificate authentication when enforcement is enabled?

Verolop (commented):
Neat! I like the separate HTTP and gRPC allow-lists and the verified leaf check. Left a couple of comments that'd be great to address before merging, especially the identity-matching gap.

go/common/servenv/client_cert.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (cs certSubject) matches(have map[string][]string) bool {
+	for attr, wants := range cs.attrs {
+		for _, want := range wants {
+			if !slices.Contains(have[attr], want) {
+				return false
+			}
+		}
+	}
+	return len(cs.attrs) > 0
```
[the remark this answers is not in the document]
kabochya: Thanks — the observation is accurate: `matches` is a subset test and does not reject additional RDNs. That is deliberate, and it is documented on `certSubject`.

On the example, though: `CN=ns-team-a,O=acme,OU=other` bypassing an entry of `CN=ns-team-a,O=acme` presupposes the neighbouring tenant already holds a certificate carrying the victim's CN *and* O. A CA willing to issue that has already granted the identity — such a caller would equally match a clean `CN=ns-team-a,O=acme` subject with no extra attribute at all, so the extra `OU` is not what admits them.

go/common/servenv/client_cert.go
kabochya: Added exact match for the HTTP listener using the `certSubjectEquals` method below.

go/common/servenv/servenv.go
```diff
@@ -444,14 +541,18 @@ func (se *ServEnv) registerFlags(fs *pflag.FlagSet, includeLoggerAndConfig bool)
 	fs.Bool("pprof-http", se.httpPprof.Default(), "enable pprof http endpoints")
 	fs.StringSlice("pprof", se.pprofFlag.Default(), "enable profiling")
 	fs.StringSlice("service-map", se.serviceMapFlag.Default(), "comma separated list of services to enable (or disable if prefixed with '-') Example: grpc-queryservice")
+	fs.String("http-cert", se.httpCert.Default(), "server certificate to use for the HTTP listener, requires http-key, enables TLS")
```
rafael, line 544: Comment about ergonomics: 

I worry a about flag explosion. It's not obvious to me that we would ever want to configure TLS differently for gRPC vs. HTTP.

Instead of adding independent flags, I'm wondering if we should consolidate them under something like:
--tls-cert
--tls-key
--tls-ca

We need to make sure we do it in a safe way to allow to deprecate the grpc-cert equivalent flags. 

If we agree on the premise, this can be a fast follow before we deploy changes.

rafael (approved):
I left one comment about ergonomics. Discussed with @kabochya  a few things:
1. Per suggestion from John, using SAN, instead of the Subject will be a better way of handling this. @kabochya  is thinking that can be a follow up.
2. It will be good to add an e2e tests for this, probably in [here](https://github.com/rafael/multigres/blob/fc686602665fb88e3e4780ec4788ec005e65ba85/go/test/endtoend/localprovisioner/cluster_test.go#L15). That will also require a way to hook this into the local provisioner. 
3. The other area we discussed is that right now gRPC is using `certSubjectMatches`, but now http uses `certSubjectEquals`. That seems like a foot gun for users. As part of the ergonomics change we discussed, the idea is to consolidate all this into the same configuration such that both http and grpc use the same TLS configuration / expectations.

PR #1450, opened 2026-09-07
feat(multipooler): elastic regular/reserved quotas over one shared budget

haritabh17:
## Problem

The global connection budget is split into two fixed ceilings by `connpool-reserved-ratio`, e.g. 32 regular / 8 reserved out of 40. A workload of explicit transactions saturates the reserved side while the regular side sits idle. A 32-client pgbench TPC-B run had 24 clients waiting on 8 reserved slots, with mean reserved acquisition around 104 ms and the first UPDATE inflated accordingly.

## Change

Let the existing rebalancer re-split the budget between the two classes each cycle from their summed demand, then run the existing per-user fair share inside each class:

1. Each class gets `min(demand, target)`.
2. Unused capacity is lent to whichever class demands more than its target.
3. Undemanded capacity is split by the ratio as burst headroom.
4. Each class keeps a floor of `min(userPools, target)` so no user pool is ever allocated zero capacity (the pool treats zero as closed). Borrowing therefore never leaves a class worse off than the fixed split.

Reservation IDs, backend affinity, per-user fair share, non-blocking shrink and drain semantics are unchanged. Transient overshoot during a shrink is the same as the existing per-user shrink: excess connections close on recycle.

Peak demand now resets to the current requested count rather than zero, so connections held across a whole interval (long transactions) keep counting as demand.

`--connpool-elastic-quotas=false` (`CONNPOOL_ELASTIC_QUOTAS`) restores the fixed split.

For 40 slots at 80:20:

| regular demand | reserved demand | regular | reserved |
| --- | --- | --- | --- |
| 40 | 40 | 32 | 8 |
| 8 | 32 | 8 | 32 |
| 0 (1 user) | 40 | 1 | 39 |
| 0 | 0 | 32 | 8 |

## Not included

Deliberately left out of this PR after review of the prototype design:

- A hard permit layer at the socket level. The existing per-user rebalancer already tolerates transient overshoot during shrink; a hard cap against `max_connections` is a separate concern.
- Zero-quota pool semantics and start-at-zero user pools, replaced by the per-class floor above.
- Demand-driven rebalance wakeups. The observed borrowing lag comes from the demand window and rebalance interval, both existing knobs.
- New per-class gauges. Per-pool capacity and demand stats already exist.

## Testing

- Table and property tests for the split arithmetic.
- Manager tests: 30 reserved connections on a 40-slot 80:20 budget all acquire and the split returns to 32/8 after release; with the flag off only 8 acquire.
- `connpoolmanager` and `pools/...` pass with `-race`; the elastic and dynamic-allocation tests pass 5x in a row.
- One existing test updated: the user-arrival test asserted the fixed 12 regular split; with reserved idle, regular now borrows down to one reserved slot per user (13, split 7/6).

## Rollout note

Borrowing lags demand by up to one rebalance interval plus the demand window (defaults 10 s and 30 s), in both directions. Under sustained regular saturation the reserved class shrinks to its floor, and a burst of transactions then waits up to that lag for capacity to move back.

### DIFF
--- a/docs/query_serving/connection_pooling.md
+++ b/docs/query_serving/connection_pooling.md
@@ -333,6 +333,23 @@ Derived:
   globalReservedCapacity = 500 * 0.2 = 100
 ```
 
+With **elastic quotas** (`--connpool-elastic-quotas`, on by default) these are
+targets under contention, not ceilings. Each rebalance cycle re-splits the
+global capacity from the summed demand of both classes:
+
+1. Each class gets `min(demand, target)`.
+2. Unused capacity is lent to whichever class demands more than its target.
+3. Capacity nobody demands is split by the ratio as burst headroom.
+4. Each class keeps a floor of `min(userPools, target)` slots so every user
+   pool stays usable in both classes.
+
+So a single user running only transactions can hold 499 of 500 slots on the
+reserved side, and when regular demand returns the split drifts back toward
+80/20 as reserved connections are released (nothing is preempted; the shrink
+is applied as connections are recycled, exactly like a per-user shrink).
+Borrowing lags demand by up to one rebalance interval plus the demand window.
+Set `--connpool-elastic-quotas=false` to pin the ratio as a hard split.
+
 A `FairShareAllocator` instance manages each resource type independently. This
 mirrors the `DemandTracker` design (also resource-agnostic, one per pool type).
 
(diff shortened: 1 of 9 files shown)


GuptaManan100 (approved):
LGTM!

PR #1455, opened 2026-09-08
feat: detect LeaderStuck via a quorum-commit watermark

dweitzman:
## Summary
- Heartbeat writes now carry a quorum-commit watermark (LSN + ts), deferred by one tick so a value is never exposed before its own quorum ack is proven
- Watermark exposed on both `StandbyReplicationStatus` and `PrimaryStatus` (leader's own first-hand view)
- New `LeaderStuck` cause in multiorch's leader-needs-replacement analyzer: leader looks healthy but quorum commits have stalled, even though replicas can still show raw LSN progress
- Excuses staleness during a still-undecided rule if a durability-sufficient set of the cohort shows recent WAL-receive advance (post-recruitment catch-up), but not once the rule is decided (deciding is itself proof quorum was already reached once)


go/services/multipooler/internal/manager/pg_multischema.go
```diff
@@ -160,7 +160,9 @@ func (pm *MultipoolerManager) createHeartbeatTable(ctx context.Context) error {
 	if err := pm.adminExec(execCtx, `CREATE TABLE multigres.heartbeat (
 		shard_id BYTEA PRIMARY KEY,
 		leader_id TEXT NOT NULL,
-		ts BIGINT NOT NULL
+		ts BIGINT NOT NULL,
+		quorum_commit_lsn pg_lsn,
```
dweitzman, line 164: These are the changes to the heartbeat table's schema

go/services/multipooler/internal/heartbeat/writer.go
```diff
@@ -110,23 +126,65 @@ func (w *Writer) writeHeartbeat(ctx context.Context) {
...
 		return mterrors.Wrap(err, "failed to write heartbeat")
 	}
 
+	// Candidate for the NEXT write's quorum_commit_lsn/quorum_commit_ts, paired
+	// with tsNano (also captured pre-commit, above). Best-effort: on failure,
+	// keep the previous candidate and retry next tick.
+	if result != nil && len(result.StructuredRows()) > 0 {
+		if raw, rawErr := executor.GetString(result.StructuredRows()[0], 0); rawErr == nil && raw != "" {
+			if lsn, lsnErr := pgutil.ParseLSN(raw); lsnErr == nil {
+				w.lastProven.Store(&provenWatermark{lsn: lsn, tsNano: tsNano})
```
dweitzman, line 166: After a successful heartbeat write, the heartbeat writer remembers the LSN so the next heartbeat write can include the LSN of the previous known successful commit

This seems relatively safe, but has a little delay before the quorum-ack'd LSN increases.

go/services/multiorch/recovery/analysis/leader_needs_replacement_analyzer.go
```diff
@@ -500,12 +496,106 @@ func (a *LeaderNeedsReplacementAnalyzer) classifyFollowerToLeader(sa *ShardAnaly
...
+	if sa.Leader != nil {
+		if ts := sa.Leader.Health().GetStatus().GetPrimaryStatus().GetQuorumCommitTs(); ts != nil {
+			freshest, have = ts.AsTime(), true
+		}
+	}
+	for _, pa := range sa.Analyses {
+		if pa == nil {
+			continue
+		}
+		ts := pa.Health().GetStatus().GetReplicationStatus().GetQuorumCommitTs()
```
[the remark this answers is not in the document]
dweitzman: A non-leader's replicas have been revoked during an emergency failover, so it can't successfully write and reach quorum and the quorum-commit timestamp wouldn't be updated/written.

There's technically an edge case in postgres where a commit can succeed without the sync standby policy being satisfied if processes holding locks crash or get killed, but I don't think we should worry about that for now. Long term, implementing two phase sync for replication streams in postgres will be the best way to track the minimum watermark of which transactions are known to be durable

go/services/multipooler/internal/manager/pg_multischema.go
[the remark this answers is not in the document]
dweitzman: We haven't released yet, so I don't think we're worried about that level of upgrade compatibility

go/services/multiorch/recovery/analysis/leader_needs_replacement_analyzer.go
```diff
@@ -500,12 +496,106 @@ func (a *LeaderNeedsReplacementAnalyzer) classifyFollowerToLeader(sa *ShardAnaly
...
+}
+
+// quorumCommitStuckCause checks the LeaderStuck backstop: the leader looks
+// healthy but quorum commits have stalled even though replicas can still
+// show raw LSN progress (they replay WAL ahead of the primary's own
+// synchronous-quorum ack). Absence of evidence must not convict, so this is
+// healthy (cause=="") when no pooler has reported a quorum_commit_ts yet.
+func (a *LeaderNeedsReplacementAnalyzer) quorumCommitStuckCause(sa *ShardAnalysis, cohort []*clustermetadatapb.ID, leaderID *clustermetadatapb.ID, policy commonconsensus.DurabilityPolicy) (types.ProblemCode, string, bool) {
+	freshest, have := freshestQuorumCommitTs(sa)
+	if !have || sa.Now.Sub(freshest) <= sa.Policy.QuorumCommitStaleAfter {
```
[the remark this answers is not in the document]
dweitzman: This is true, as written this is vulnerable to clock skew. In the short term, we're assuming significant clock skew is relatively uncommon (which is true in AWS). In the long term, we could revisit this and make times of recent events relative to the pooler that reported them

go/services/multipooler/internal/heartbeat/writer.go
simonbtomlinson: This seems ok to me

go/services/multipooler/internal/manager/pg_multischema.go
```diff
@@ -160,7 +160,9 @@ func (pm *MultipoolerManager) createHeartbeatTable(ctx context.Context) error {
...
+		quorum_commit_ts BIGINT
```
simonbtomlinson, line 165: Do we have a convention of bigints over `timestamp`s? Not a big deal, but I think it's a little easier to read directly out of the table if we store timestamps. I see `ts` is already a `BIGINT`, so perhaps we do.

go/services/multipooler/internal/manager/pg_replication.go
```diff
@@ -350,6 +350,12 @@ func (pm *MultipoolerManager) queryReplicationStatus(ctx context.Context) (*mult
 			if advanceTime, ok := reader.LastReceiveLSNAdvance(); ok {
 				status.LastReceiveLsnAdvanceTime = timestamppb.New(advanceTime)
 			}
+			if commitTs, ok := reader.QuorumCommitTs(); ok {
```
simonbtomlinson, line 353: We acquire the mutex twice here, and could read a later `QuorumCommitLSN()` than we do `QuorumCommitTs()`. I think this is OK - both should be relatively up-to-date and we really only care about the timestamp being "recent enough".

simonbtomlinson (approved):
LGTM. I had one question about the type of the column, but that's non-blocking.

go/services/multipooler/internal/manager/pg_multischema.go
dweitzman: I think maybe the general tradeoffs here are that `timestamp` is easier for humans but loses precision vs `BIGINT` retains precision while being harder for humans to interpret? I don't have strong feelings on this, but in this particular case I think there's an argument that we don't need the precision of BIGINT

dweitzman: Changed to timestampz

go/services/multipooler/internal/manager/rpc_manager.go
```diff
@@ -652,6 +653,18 @@ func (pm *MultipoolerManager) getPrimaryStatusInternal(ctx context.Context) (*mu
 	status.SyncReplicationConfig = syncConfig
 	status.MaxWalSenders = maxWalSenders
 
+	// The leader's own first-hand view of its heartbeat writer's last
+	// proven-quorum-acked write, available even when no follower is reachable
+	// to relay the replicated row.
+	if pm.replTracker != nil {
+		if writer := pm.replTracker.HeartbeatWriter(); writer != nil {
+			if lsn, ts, ok := writer.LastProven(); ok {
```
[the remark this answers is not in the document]
dweitzman: An old heartbeat write wouldn't suppress failover if the leader says it's not able to fulfill the obligations of the current term, and if new writes aren't happening then eventually we'll still learn that a failover is needed

go/services/multipooler/internal/heartbeat/writer.go
```diff
@@ -110,23 +126,65 @@ func (w *Writer) writeHeartbeat(ctx context.Context) {
...
+		lsnArg = lastProven.lsn.String()
+		provenTsArg = lastProven.ts
+	}
 
-	_, err := w.queryService.QueryAdminArgs(ctx, `
-		INSERT INTO multigres.heartbeat (shard_id, leader_id, ts)
-		VALUES ($1, $2, $3)
+	result, err := w.queryService.QueryAdminArgs(ctx, `
+		INSERT INTO multigres.heartbeat (shard_id, leader_id, ts, quorum_commit_lsn, quorum_commit_ts)
+		VALUES ($1, $2, $3, $4::pg_lsn, $5::timestamptz)
```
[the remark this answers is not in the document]
dweitzman: A premise of our HA implementation is that synchronous_commit is on with non-empty synchronous_standby_names

go/services/multipooler/internal/heartbeat/reader.go
```diff
@@ -163,6 +173,17 @@ func (r *Reader) readHeartbeat(ctx context.Context) {
...
 		r.haveReceiveLSN = true
 	}
+	// quorum_commit_lsn/quorum_commit_ts are just the latest observed values,
+	// not advance-tracked like receive_lsn above -- quorum_commit_ts is
+	// already a leader-stamped instant, not something we'd time ourselves.
+	if haveQuorumCommitLSN {
+		r.lastQuorumCommitLSN = quorumCommitLSN
+		r.haveQuorumCommitLSN = true
+	}
+	if haveQuorumCommitTs {
```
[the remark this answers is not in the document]
dweitzman: There's no lasting harm in being aware that a prior leader did a write. We always have the option to modify the code to consider if the write was by the current leader, if we want.

PR #1458, opened 2026-09-11
perf(multipooler): send reservation state with the last result message

shedges:
## Problem

For a statement on a reserved connection, the streaming handlers reply to multigateway with the result and then a second message that carries only the reservation state. A typical single-chunk statement inside a transaction (every statement in TPC-C) therefore produces two DATA frames plus trailers on the gateway to pooler hop where one would do. Each extra message is a loopy-writer wakeup, a buffer flush, a TLS record and a socket write on the pooler, and a matching read and decode on the gateway. In the nightly TPC-C profiles the primary pooler spends about 75% of its CPU on this hop and about 10% on Postgres work, and the largest parts of the hop cost scale with the number of frames per statement.

The protocol does not require the split: the response message already has both `result` and `reserved_state`, and the gateway's receive loop already handles a message that carries both.

Refs MUL-1579.

## Change

Pooler only. In `StreamExecute` and `PortalStreamExecute`, the terminal chunk of a statement is held until the executor returns and the reservation state is attached to it.

The pooler knows which chunk is terminal from the Postgres protocol: the client emits it on `CommandComplete`, and it is the only chunk that carries a `CommandTag`. Chunks flushed mid-result when the 2 MB batch threshold is passed have no tag.

- Intermediate chunks (no `CommandTag`) are sent as soon as they are produced, so streaming of results larger than one batch is unchanged.
- The terminal chunk waits for the executor's return (microseconds of bookkeeping) and carries the reservation state.
- Suspended portals (`PortalSuspended`, rows flushed without a tag) and empty queries end without a terminal chunk and keep sending the reservation state on its own.
- Multi-statement simple queries produce one tagged chunk per statement; each is released when the next statement produces output, and the last carries the reservation state.
- Notices keep their position ahead of the data they belong to.
- Error path: the pending chunk is still delivered with the authoritative reservation state the gateway needs for an aborted backend.
- No reserved connection: no reservation message at all.
- Backpressure is unchanged: `Send` is still called synchronously on the executor's callback path, so gRPC flow control still stalls the pooler's read from Postgres when the gateway is slow. The only added buffering is the single held terminal chunk.
- The handlers are split into `streamExecuteTo` and `portalStreamExecuteTo` (executor passed in) so the wire sequence is unit-tested without a live pooler, following `copyBidiExecuteToStdout`.

Compatibility: no proto change, no gateway change. New pooler with old gateway and old pooler with new gateway both work, since the gateway has always accepted result and reservation state in one message.

## Measurements

Kind replica of the staging configuration (operator f1a1612e, same images, internalTLS, topoTLS, metrics-only OTEL, same postgres config, 2 CPU limits), TPC-C 56 warehouses, only `spec.images.multipooler` swapped between runs. Per statement on the primary pooler:

| | baseline | this change |
|---|---|---|
| `serverStream.SendMsg` | 16.2 to 23.6 µs | 7.5 to 8.8 µs (-47 to -63%) |
| `Codec.Marshal` | 12.5 to 16.6 µs | 4.5 to 5.9 µs (-53 to -70%) |
| `bufWriter.Flush` | 35 to 74 µs | 31 to 66 µs (-8 to -15%) |
| syscalls / TLS write | | -6 to -8% / -11 to -15% |
| total pooler CPU per statement, paired single-connection run at identical 925 stmt/s | 362 µs | 320 µs (-12%) |
| total pooler CPU per statement, 8-thread runs | 200 µs | 202 µs (within noise, primary on different nodes) |
| gateway exec latency, single connection | 0.895 ms avg | 0.902 ms avg |

Throughput was unchanged within run-to-run noise (four 8-thread samples fall on one time-drift line regardless of image). This is a bounded saving, roughly 0 to 10% of pooler CPU, one of about five frames per statement. It is not the main cost of the hop.

## Testing

- 9 unit tests in `service_stream_coalesce_test.go`: single chunk on a reserved connection, intermediate chunks observed on the stream before the next chunk is produced, suspended portal, pooled connection (no reservation state), notice ordering, error after a chunk, error without data, simple-query single result, simple-query multi-statement ordering.
- `golangci-lint run` on the package: 0 issues.
- End to end on the replica: four rolling image swaps (planned failovers), TPC-C at 1, 4 and 8 threads, no errors or restarts.


go/services/multipooler/grpcpoolerservice/service.go
```diff
@@ -405,44 +483,51 @@ func (s *poolerService) PortalStreamExecute(req *multipoolerpb.PortalStreamExecu
...
-		// (typically in an aborted transaction, awaiting ROLLBACK). Send the
-		// authoritative state before returning the gRPC error so the gateway doesn't
-		// drift from the multipooler and accidentally route follow-up cleanup to a
-		// different backend.
-		if reservedState.GetReservedConnectionId() > 0 {
-			if sendErr := stream.Send(&multipoolerpb.PortalStreamExecuteResponse{ReservedState: reservedState}); sendErr != nil {
-				return mterrors.ToGRPC(sendErr)
-			}
+		if sendErr := flushPending(); sendErr != nil {
+			return mterrors.ToGRPC(sendErr)
```
GuptaManan100, line 525: nit: Minor asymmetry between the handlers: `streamExecuteTo` discards the flush error on the error path (`_ = flushPending()`) and wraps the success flush in `ToGRPC(...)`, while `portalStreamExecuteTo` returns the send error over the exec error and returns the success flush raw. Both are defensible (a broken stream means the gateway won't get anything anyway), but making them identical would read easier. Optional.

GuptaManan100 (approved):
rest LGTM!

go/services/multipooler/grpcpoolerservice/service.go
shedges: Done

PR #1459, opened 2026-09-11
perf(multipooler): cache informational reads and reduce telemetry overhead

haritabh17:
## Summary

Reduce Multigres CPU overhead from repeated informational work and avoidable telemetry allocation, while keeping health checks live, tracing configuration authoritative, and backup correctness paths on fresh data.

Resolves MUL-1540.

## Problem

Process sampling on qs-aws showed the pooler and pgctld repeatedly forking `pgbackrest info` and `psql` for results that rarely change, and profiling showed per-update metric attribute construction plus span recording with no exporter attached. Idle pooler-container CPU was ~0.22 cores in the experiment; a bundled prototype brought it to ~0.04.

## Changes

**1. Backup list cache (`GetBackups` only)**
- `Engine.CachedList` serves a per-engine 5-second cache of `pgbackrest info`. Concurrent misses share one run, bounded by `InfoTimeout` and independent of any single caller's cancellation. Failures are not cached.
- Invalidated on `SetConfigPath` / `SetBackupConfig` / `SetPgpassPath` and after `Backup`, `Expire`, `Restore`, `StanzaCreate`. Invalidation also detaches a run that started before the mutation, so a pre-mutation result can never repopulate the slot.
- Results are `proto.Clone`d so callers cannot mutate shared state.
- Recovery (`hasCompleteBackups`, restore), `Expire`, `FindByJobID` and the backup-health poller are unchanged and keep using fresh `ListBackups` / `runInfoCmd`.

**2. Postgres version memo (pgctld `Status`)**
- `SELECT version()` is memoized per postmaster identity (PID + start timestamp from `postmaster.pid`). A new identity replaces the entry; failed lookups and lookups spanning a postmaster replacement are discarded. Incomplete identity bypasses the memo.
- The live `probePostgreSQLRunningAndReady` still runs first on every `Status`; a cached version never implies health.

**3. Non-recording spans when export is explicitly disabled**
- With `OTEL_TRACES_EXPORTER=none` set explicitly, no `OTEL_TRACES_SAMPLER`, and no file-based sampler, the provider uses `ParentBased(NeverSample())`.
- Trace/span IDs and propagation are preserved. A sampled upstream parent is still honoured, so a process with export disabled never strips the sampled flag for exporting services downstream of it.
- An **unset** exporter keeps the OTel default (`parentbased_always_on`). This is deliberate: an unconfigured gateway must still root sampled traces for poolers that do export. Deployments opt in by setting the env var explicitly.
- The unset case no longer writes `OTEL_TRACES_EXPORTER=none` into the process environment; autoexport gets a no-op fallback exporter instead. Otherwise the CLI's own default would leak into every service it spawns with `os.Environ()` and be treated there as explicit.

**4. Precomputed metric attribute sets**
- `ConnectionCount` binds the pool's idle/used attribute sets once in `NewPool`; reserved transaction outcomes use a fixed attribute-set map with the dynamic path as fallback. Names, units, labels and cadence are unchanged.

## Not in this PR
- The reserved-pool `Stats()` vs `AddReservationReason` data race found while preparing this. Real but off-ticket; will follow as its own `fix(multipooler)` PR.
- Idle/loaded CPU and TPS measurement on qs-aws with a direct-Postgres control. The prototype numbers were bundled and not placement-isolated; the acceptance measurement is still to be done against this branch.

## Notes for reviewers
- No in-repo caller polls `GetBackups` periodically; the cache targets external API polling with concurrent callers, which is what the ticket describes.
- Cross-pooler staleness of up to 5 s on the informational list is accepted (multiadmin routes list/expire to followers and backup to the leader).
- No shared cache package: the two memoizers are inlined next to their only users.

## Testing
- `go test -short -race` on `go/cmd/pgctld/command`, `go/services/multipooler/...`, `go/tools/telemetry`: pass.
- Integration: `endtoend/pgctld`, `endtoend/multipooler`, `endtoend/multipooler/backupfaults`, `endtoend/multiorch`: pass.
- New tests: backup-cache coalescing, TTL expiry, cancellation independence, invalidate-during-flight, failure not cached, clone isolation, fresh-API bypass; version memo across postmaster identity change, failed query, restart race, incomplete identity; tracing explicit-none vs unset vs explicit sampler with propagation and remote-parent checks; unset exporter leaves the environment untouched across repeated initializations in one process; bound metric labels equivalence.
- `BenchmarkConnectionCountAttributes`: 296 B / 4 allocs → 16 B / 1 alloc per update.
- golangci-lint, gofumpt, goimports clean.


go/cmd/pgctld/command/status_cache.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/multigres/multigres/go/common/constants"
+	"github.com/multigres/multigres/go/services/pgctld"
+)
+
+// serverVersion memoizes `SELECT version()` for the running postmaster. The
+// version cannot change while the same postmaster runs, so there is no TTL;
+// a new postmaster identity replaces the entry. pgctld serves one data
+// directory per process, so a single slot suffices.
+//
+// ponytail: no coalescing. Status is polled by one monitor loop; a duplicate
```
cuongdo, line 36: nit: what does "ponytail" mean?

go/services/multipooler/internal/manager/backup/list_cache.go
```diff
@@ -0,0 +1,125 @@
...
+
+// listCacheTTL bounds how stale the informational backup list may be. A
+// backup or expire performed by another pooler on the shared repository can
+// take this long to show up through GetBackups.
+const listCacheTTL = 5 * time.Second
+
+// listCache memoizes the most recent `pgbackrest info` result for
+// informational callers and coalesces concurrent misses into one run.
+//
+// ponytail: single slot, one key. Re-extract into a generic helper only if a
```
cuongdo, line 36: nit: same question about "ponytail"

cuongdo (approved):

go/services/multipooler/internal/manager/backup/list_cache.go
haritabh17: 😬 thats a skill that I use with claude. it keeps TODO and note items with that tag. I can clean those up.

PR #1461, opened 2026-09-14
feat(pgctld): expose pooler_dir, host, and max_connections on Status; adopt them in multipooler

haritabh17:
## Description

Part of MUL-497 (infer query-serving flags instead of duplicating them); closes MUL-1592. Follows the merged capacity derivation (#1439) and socket-file derivation (#1445).

pgctld's `StatusResponse` already carried `data_dir`, `port`, and `version`, but every consumer discarded them — multipooler called Status in four places, read only `.Status`/`.Ready`, and re-declared the same values as its own flags. This PR makes the Status RPC the configuration channel it was shaped to be.

### pgctld side

- `StatusResponse` gains `pooler_dir`, `max_connections`, `pgbackrest_port`, `pgbackrest_cert_dir`; the structurally dead `host` field (declared since the proto's introduction, never written) is now populated. All fields are reported in the NOT_INITIALIZED branch too, so adoption works before initdb.
- `max_connections` is resolved via `postgres -C` — follows include directives and `postgresql.auto.conf`, works with postgres stopped or running. Because Status sits on the pooler's 5s monitoring path, the probe has its own 5s deadline (a hung volume cannot wedge the Status RPC), caches failures as well as successes (no fork-per-poll on persistent errors), and is invalidated **after** every config-changing operation (`InitDataDir`, `Start`, `Restart`, `ReloadConfig`, `PgRewind`) via `defer`; a generation counter additionally orders probe publication against invalidation, so a probe overlapping a Restart cannot re-publish the pre-restart value after the invalidation ran. `0` is the explicit unknown. The pgBackRest port/cert-dir pair is reported all-or-nothing, matching pgctld's own enable condition, so a half-configured pgctld cannot advertise a port it is not serving (with a startup warning naming the half-set pair).

### multipooler side

- At startup, `pg-port`, `pooler-dir`, `pgbackrest-port`, and the pgbackrest cert trio (derived from pgctld's cert-dir via the now-single-sourced `constants.PgBackRestCertFiles` contract) are adopted from pgctld's Status, polling up to **30s** — deliberately below the default kubelet liveness kill (~40s), so an unreachable pgctld produces a clear error in the logs rather than a probe kill.
- Explicitly configured flags always win, command line or config file (same `Flag.Changed` + `Registry.InStaticConfig` pattern as #1439/#1445). When all six are explicit — the entire shardsetup e2e matrix — no dial happens at all.
- **Required vs optional**: `pg-port`/`pooler-dir` are identity with no usable defaults — if they need adoption and pgctld never answers, Init fails with a clear error. A pgBackRest-only adoption instead falls back to flag values with a warning and starts degraded, preserving the manager's explicitly supported pgctld-unavailable mode (multiorch keeps visibility).
- Version skew: an older pgctld that cannot report a needed `pooler_dir` **fails startup with a clear error** (per review): the pooler directory anchors durable per-instance state — consensus promise files, backup config, recovery sentinels — and an empty value would resolve those paths into the shared working directory, letting co-located poolers overwrite each other's durable consensus terms. Rolling upgrades are unaffected: pre-upgrade manifests necessarily pass `--pooler-dir` explicitly, so adoption never fires for them.
- The `/status` page now shows the resolved socket path instead of the raw (possibly empty) flag.

### Capacity seed (absorbs the "pgctld exposes the connection budget" ticket, related MUL-379)

`max_connections` from Status seeds the connpool capacity **fallback only** — the live SQL derivation from #1439 stays authoritative, since a conf-derived value cannot see pending-restart divergence. The seed refreshes from the monitor's existing 5s Status poll (a one-shot read at Init would stay unknown on fresh clusters, where adoption precedes initdb), and the fallback is clamped to never exceed the configured value: the seed shrinks the budget on small servers, never grows it speculatively.

### Callers

The local provisioner and the k8s demo manifest stop passing all six derivable flags, so every localprovisioner e2e run and the docker smoke test exercise adoption end-to-end. The provisioner's hand-rolled pgctld-config lookup for `pgbackrest_port` is deleted, the orphaned `multipooler.pg-port`/`pooler-dir` yaml knobs are removed, and the pooler's `PGDATA` is single-sourced from pgctld's pooler directory.

## Testing

- Unit: `resolveAdoptedValues` (adopt-all / per-value explicit override / empty-status keeps flags); `TestFlagExplicitlySet` variants; seed-fallback derivation (60 → 52) and clamp (500 → 100); full `go test -short ./go/...` green.
- Integration: `endtoend/pgctld` (Status lifecycle asserts the new fields incl. unknown-before-initdb and the mocked `postgres -C`), `endtoend/localprovisioner` (adoption end-to-end against real postgres), `endtoend/multipooler` + `endtoend/metrics` (explicit-flag fast path; derived capacity still 102) — all green.
- Manual: 3-pooler local cluster — all poolers logged adopted per-cell values (`pg_port` 25432-34, pooler dirs, `pgbackrest_port` 18432-34), socket derivation composed on top, live SQL capacity derivation authoritative, gateway e2e query served.
- A high-effort multi-agent code review produced 10 findings (7 confirmed); all are addressed, including a depguard violation verified by running golangci-lint (now 0 issues). Three follow-up review comments are also addressed (pooler-dir version skew fails fast; probe/invalidation generation ordering; pgBackRest all-or-nothing reporting); the fourth (topology client open on Init error paths) is declined — the process exits immediately on any Init error, and every pre-existing validation error in Init has the same shape.


go/cmd/pgctld/command/server.go
```diff
@@ -557,6 +582,7 @@ func (s *PgCtldService) StartPgBackRestManagement() {
 }
 
 func (s *PgCtldService) Start(ctx context.Context, req *pb.StartRequest) (*pb.StartResponse, error) {
+	defer s.invalidateMaxConnections()
```
GuptaManan100: Instead of just maxConnections we should make this generic for configs we are tracking. WDYT?

go/cmd/pgctld/command/server.go
```diff
@@ -722,21 +750,102 @@ func (s *PgCtldService) ReloadConfig(ctx context.Context, req *pb.ReloadConfigRe
...
+// server is running. Returns 0 (unknown) before the data directory is
+// initialized or when the value cannot be determined. Cached because the
+// multipooler polls Status every few seconds; invalidated by every operation
+// that can change the effective config.
+func (s *PgCtldService) effectiveMaxConnections(ctx context.Context) int32 {
+	// Positive = cached value; negative = cached failure (do not re-fork until
+	// the next invalidation). Status sits on the multipooler's monitoring path,
+	// so a persistent failure must not fork postgres on every 5s poll, and a
+	// hung probe (e.g. a stuck volume) must not wedge the Status RPC: the probe
+	// gets its own short deadline independent of the caller's context.
```
GuptaManan100: well its not completely independent. The probeCtx is derived from the ctx. So it has the ctx, and an additional deadline, whichever happens first. that's fine, but the comment needs to be fixed.

GuptaManan100 (approved):
Looks good to me

go/cmd/pgctld/command/server.go
haritabh17: Agree — extracting a small per-GUC probe cache now (same gen/invalidation semantics, max_connections as the only entry) and then the concrete next consumers are the reserved-connections GUCs, which would let the capacity fallback drop its hardcoded superuser_reserved=3.

go/cmd/pgctld/command/server.go
haritabh17: Ack, this is removed now after adding the `gucProbeCache`

PR #1466, opened 2026-09-16
feat(servenv): attach service identity to log records

rafael:
## Summary

- `SetupLogging` now accepts identity attributes (`service_id`, `multigres_cell`, `multigres_shard`, `multigres_tablegroup`) derived from `ServiceIdentity` and applies them to the handler via `WithAttrs`, so every log record carries them consistently, including startup logs.
- Fields with empty values are omitted rather than logged blank.
- `UpdateTelemetryWrapper` re-applies the identity attributes whenever the telemetry wrapper is swapped, so they survive telemetry re-initialization.
- Attributes show up in both local output and OTLP log records without changing the existing OpenTelemetry resource attributes.

## Description

This makes it possible to correlate log lines to a specific service instance (cell, shard, table group) directly from log output, which is useful when tailing logs from multiple pooler/gateway instances at once. Documented the new fields in `docs/general/logging.md` and added `logging_identity_test.go`, which exercises JSON and text formats plus OTLP log records across a few representative service identities (pooler, gateway, admin with no topology fields).


go/common/servenv/servenv_unix.go
```diff
@@ -45,7 +45,7 @@ func (sv *ServEnv) Init(id ServiceIdentity) error {
 	sv.mu.Lock()
 	sv.initStartTime = time.Now()
 	sv.mu.Unlock()
-	sv.lg.SetupLogging()
+	sv.lg.SetupLogging(id.logAttributes()...)
```
dweitzman, line 48: Could potentially reuse the list of attributes attached to every span here instead of rebuilding it with different attribute names? The span attributes already have semantic convention names for these values

dweitzman (approved):
If there's no particular need to diverge from the semantic conventions for the attributes here, my preference would be to reuse the keys and values that we attach to every span for log entries

PR #1468, opened 2026-09-17
feat(pgcatalog): generate builtin catalog from vendored PG .dat files

GuptaManan100:
## Summary

- Vendors PostgreSQL REL_17_6 catalog bootstrap data (`pg_type`/`pg_proc`/`pg_operator`/`pg_cast`/`pg_collation` .dat files, byte-identical copies; tag + commit recorded in `THIRD_PARTY_NOTICES.md`).
- Adds `go/tools/pgcataloggen`, which processes the data the way PostgreSQL's own build does (`Catalog.pm`/`genbki.pl`): brace-balanced Perl-hash parsing, `BKI_DEFAULT` column defaults hand-transcribed with header-line citations, symbolic-reference resolution (including the `MULTIPLE` ambiguity convention for overloaded names), and array-type synthesis.
- Emits compiled-in Go tables in `go/common/pgcatalog` — 193 types, 3,314 procs, 799 operators, 229 casts — with OID/name lookup APIs. This is the analogue of PG's `fmgr_builtins` fast path plus the builtin slices of the four catalogs.
- Wires `make pgcatalog` into `build-all`/`generate`, so the existing `validate-generated-files` CI job enforces freshness.

## Description

First step of the gateway-side expression evaluation engine for sharded query serving: parse analysis, planning, and expression evaluation will resolve builtin types/functions/operators/casts through these tables with no runtime catalog access. User-defined objects are out of scope and arrive later via schema tracking, through a resolver seam added in a follow-up PR alongside the fmgr registry.

Struct fields cite their upstream `pg_*.h` lines at the pinned tag, and ported/derived files carry the PostgreSQL license header (same convention as `go/common/parser/ast`). Tests include known-row assertions and a cross-reference integrity suite modeled on PG's own `opr_sanity` regression test (mutual commutators/negators, array back-links, cast function signatures with binary-coercibility).


go/common/pgcatalog/pgcatalog.go
```diff
@@ -0,0 +1,238 @@
...
+
+// Proc is a builtin function: the subset of pg_proc columns multigres
+// carries - postgres/src/include/catalog/pg_proc.h:30. Src is the upstream C
+// symbol name (prosrc) that the fmgr registry binds Go implementations to;
+// for aggregates it is the placeholder "aggregate_dummy", and for
+// SQL-language builtins the sentinel "see system_functions.sql".
+type Proc struct {
+	Oid      Oid    // oid - pg_proc.h:32
+	Name     string // proname - pg_proc.h:35
+	ArgTypes []Oid  // proargtypes; pronargs == len(ArgTypes) - pg_proc.h:95
```
haritabh17, line 128: Add a variadic element oid to `Proc` and `ProcRow`? Function matching must expand the final variadic parameter while preserving normal array arguments passed with `VARIADIC`.

go/common/pgcatalog/pgcatalog.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	castsBySrcTgt = make(map[[2]Oid]*Cast, len(Casts))
+	for i := range Casts {
+		c := &Casts[i]
+		castsBySrcTgt[[2]Oid{c.Source, c.Target}] = c
+	}
+}
+
+// TypeByOid returns the builtin type with the given OID, or nil.
+func TypeByOid(oid Oid) *Type {
```
haritabh17, line 212: All the lookup APIs expose shared mutable state, is that okay? Because lookup results alias exported catalog tables and internal overload slices and hence caller mutation can alter later global resolutions.

go/common/pgcatalog/pgcatalog.go
```diff
@@ -0,0 +1,238 @@
...
+	typesByName     map[string]*Type
+	procsByOid      map[Oid]*Proc
+	procsByName     map[string][]*Proc
+	operatorsByOid  map[Oid]*Operator
+	operatorsByName map[string][]*Operator
+	castsBySrcTgt   map[[2]Oid]*Cast
+)
+
+func init() {
+	typesByOid = make(map[Oid]*Type, len(Types))
```
haritabh17, line 176: Loading everything in memory of each gateway? If we eventually put the complete catalog, an empty database in PG has ~200mb worth of catalog, so this might get too much? Esp for multi-db setups?

go/tools/pgcataloggen/main/main.go
```diff
@@ -0,0 +1,45 @@
...
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/multigres/multigres/go/tools/pgcataloggen"
+)
+
+func main() {
+	dataDir := flag.String("data", "go/common/pgcatalog/data", "directory containing the vendored .dat files")
+	outDir := flag.String("out", "go/common/pgcatalog", "directory to write generated Go files to")
```
haritabh17, line 31: Should add PG versioning?

haritabh17 (approved):
Otherwise looks fine and okay to address comments in follow up

go/tools/pgcataloggen/pgcataloggen.go
```diff
@@ -0,0 +1,810 @@
...
+			return nil, err
+		}
+		if row.Right, err = lookup(r.typeOids, right, "pg_type", false); err != nil {
+			return nil, fail("oprright", err)
+		}
+		result, err := e.Require("pg_operator.dat", "oprresult")
+		if err != nil {
+			return nil, err
+		}
+		if row.Result, err = lookup(r.typeOids, result, "pg_type", true); err != nil {
```
shedges: Is it expected to have operator result type as optional?

go/tools/pgcataloggen/pgcataloggen.go
```diff
@@ -0,0 +1,810 @@
...
+			return nil, fail("oprcom", err)
+		}
+		if row.Negator, err = lookup(r.operKeys, e.Get("oprnegate", "0"), "pg_operator", true); err != nil {
+			return nil, fail("oprnegate", err)
+		}
+		code, err := e.Require("pg_operator.dat", "oprcode")
+		if err != nil {
+			return nil, err
+		}
+		if row.Code, err = lookup(r.procKeys, code, "pg_proc", true); err != nil {
```
shedges: Sam as above, is it expected to have optional?

shedges (approved):

go/common/pgcatalog/pgcatalog.go
GuptaManan100: Yes, we'd probably need it. But I think its best to defer it to the function-matching PR that actually consumes it

go/common/pgcatalog/pgcatalog.go
GuptaManan100: That's a great catch! But copying on every lookup would add allocation on the latency-sensitive path. I've made the contract explicit in the docs, and we'll just have to treat them as read-only in practice. unfortunately there aren't great ways of enforcing them to be immutable...

go/common/pgcatalog/pgcatalog.go
GuptaManan100: For this data it doesn't scale with DB count: it's builtin catalog only. And Postgres keeps it in shared memory too!

go/tools/pgcataloggen/main/main.go
GuptaManan100: The data being imported is largely stable across postgres versions, and only functions OIDs change that too mostly additive across version upgrades, for now it seems we don't need it versioned.

PR #1469, opened 2026-09-17
refactor(parser): generate type OID constants into new pgoid package

GuptaManan100:
## Summary

- New leaf package `go/common/parser/pgoid` owns the `Oid` type (the `postgres_ext.h` analogue) and all 189 builtin type OID constants, generated from the vendored `pg_type.dat` per genbki.pl's `form_pg_type_symbol` rule and verified 1:1 against upstream's generated `pg_type_d.h`, including the `CASHOID`/`LSNOID` compatibility aliases.
- `ast` aliases the type and re-exports the constants through a generated file, so no call sites change; the hand-maintained constant list and `Oid.String()` switch are replaced by generated data, removing the drift surface (and `String()` now covers all builtin types instead of ~79).
- `pgcatalog` now depends on the tiny leaf instead of the full ast package. `pgoid` lives under the parser tree so the parser-isolation depguard rule (parser as a standalone library) holds by construction.
- Fixes asthelpergen for Go 1.22+ materialized type aliases (`*types.Alias`): an alias keeps its declared name and belongs to its declaring package, preserving deep-clone semantics for `[]Oid` fields. Regenerated ast helpers are byte-identical.

## Note

Stacked on #1468 — the first three commits belong to that PR; this change is only the last two commits (`fix(asthelpergen)` and `refactor(parser)`). Stays draft until #1468 merges, then rebases on main.


haritabh17 (approved):
This looks fine on top of the previous change.

PR #1477, opened 2026-09-21
perf(servenv): reuse gRPC handler workers across RPCs

haritabh17:
## Summary

Enable gRPC's `NumStreamWorkers` so incoming RPCs run on a fixed set of long-lived handler goroutines instead of a fresh goroutine per stream. Profiling the simple-query path showed goroutine creation and stack growth as a measurable share of pooler CPU; reusing grown stacks removes that cost without changing what any handler does.

- Default `--grpc-stream-workers=64`; `0` restores gRPC's default goroutine-per-stream behaviour (rollback switch, restart required).
- Busy workers fall back to a new goroutine, so this is **not** a concurrency limit and adds no admission queue.
- Applies to every server built through `servenv` (multipooler, multigateway, multiorch, multiadmin, pgctld), so each holds 64 resident goroutines; visible in goroutine gauges, otherwise inert.
- gRPC marks the option experimental in the pinned version (v1.81.1); it has been present since 2020 and its semantics are unchanged, but the signature could move on a future upgrade.

This is the first, self-contained half of MUL-1584. Reusable bidirectional query streams (the other half) were implemented and benchmarked separately and are deferred to a follow-up: the worker change carries most of the measured gain in about 15 lines, while stream reuse adds roughly 10 more points for ~600 lines.

## Benchmark: qs-aws EKS rig

Cluster `haritabh-perf-20260915-01`: 3 poolers (1 primary + 2 synchronous replicas on dedicated m7i.xlarge nodes, multipooler limited to 1.5 CPU), 5 gateways (t3.medium), pgbench 17 from a driver pod with 48 clients spread across all five gateways, `-M simple`, scale 10, 30 s warmup + 90 s measurement, transaction logs for latency, per-container CPU from cgroup `cpu.stat`. Two images layered on the same base with only the multipooler and multigateway binaries replaced: **off** = ee5a76ee (this PR's parent), **on** = ba0caeea (this PR). Order: off ×3 → on ×3 → off ×1 control. Zero failed transactions in all 14 runs; repetitions within 2% of each other.

| workload | metric | workers off | workers on | delta |
|---|---|---|---|---|
| simple-update | TPS | 2661 | 2890 | **+8.6%** |
| simple-update | mean latency | 18.0 ms | 16.6 ms | −7.9% |
| simple-update | p95 | 30.4 ms | 24.4 ms | −19.7% |
| simple-update | p99 | 46.9 ms | 40.5 ms | −13.6% |
| simple-update | pooler CPU / txn | 0.543 ms | 0.493 ms | **−9.2%** |
| simple-update | gateways CPU / txn (sum of 5) | 1.599 ms | 1.586 ms | −0.8% |
| simple-update | postgres CPU / txn | 0.460 ms | 0.449 ms | −2.3% |
| tpcb-like | TPS | 1091 | 1082 | −0.8% |
| tpcb-like | p99 | 169 ms | 169 ms | −0.1% |
| tpcb-like | pooler CPU / txn | 1.062 ms | 0.950 ms | **−10.5%** |
| tpcb-like | gateways CPU / txn (sum of 5) | 2.934 ms | 2.954 ms | +0.7% |
| tpcb-like | postgres CPU / txn | 0.885 ms | 0.849 ms | −4.1% |

Reading: on simple-update the pooler is the bottleneck (CFS-throttled ~90% of periods at its limit, ~70% with workers on), so the ~9% pooler CPU saving converts into throughput and a better tail. tpcb-like is bound by cross-zone synchronous commit, so the same ~10% pooler CPU saving shows as headroom rather than TPS. Gateways are unchanged, as expected for a gRPC client on the query path.

Caveat: the primary pod re-elected onto a different m7i.xlarge node between arms. Postgres CPU/txn on that same node moved only −2 to −4% while the pooler moved −9 to −10%, so node speed does not explain the gap, and the control repetition reproduced the first baseline within 1–4%.

A local laptop run (3 reps, 32 clients) had shown a larger CPU/txn drop but a noisy p99; the rig numbers above supersede it.

## Test plan

- [x] `go test ./go/common/servenv/...` (new `TestGrpcStreamWorkerConfiguration`)
- [x] endtoend `queryserving` and `multipooler` suites pass with workers enabled
- [x] `--grpc-stream-workers=0` verified effective via the `/config` page
- [x] qs-aws A/B: off ×4 vs on ×3, both workloads, zero failures (table above)


shedges (approved):
LGTM

PR #1478, opened 2026-09-21
perf(telemetry): reduce OpenTelemetry gRPC overhead per SQL statement

shedges:
## Problem

Every gRPC RPC between multigateway and multipooler, which is one per SQL statement, is recorded by OpenTelemetry's gRPC stats handler (otelgrpc) on both sides. Per RPC it extracts the trace and baggage headers, builds and hashes two attribute sets, starts and ends a span, and records an `rpc.*` duration histogram. In the nightly TPC-C profiles (staging, 12 client threads, primary pooler at 90% of its 2 cores) this is about 13% of multipooler CPU and 4% of multigateway CPU, while no trace exporter is configured on staging so every span is discarded. The instrumentation was installed unconditionally with no way to disable it.

## Changes

Three commits, independent of each other.

1. **`--grpc-otel-instrumentation` (default `true`, env `GRPC_OTEL_INSTRUMENTATION`)** on the servenv gRPC server. When `false`, the server is created without the otelgrpc stats handler and gRPC clients created by `grpccommon.NewClient` in the same process omit it too, so the gateway to pooler dial is covered. `grpccommon` cannot import servenv, so `NewGrpcServer` hands it the decision as a function evaluated at dial time; binaries without a servenv (the `multigres` CLI, provisioner) keep the default. Application spans (`gateway.query`, `shard.execute`, `consensus/*`) and `mg_*` metrics are unaffected; only the per-RPC gRPC spans and `rpc.*` metrics go.

2. **Caller identity is no longer mirrored into W3C baggage.** Since #1244 the gateway sent the client's user and application name twice: in the typed `CallerId` request field and as two baggage members in a request header. Nothing read the baggage: the pooler attributes its spans from the typed field (`annotateCaller`, in every handler that receives a `CallerId`), no baggage span processor or log enrichment exists, and no dashboard references the keys. Meanwhile every statement paid to build and encode the header on the gateway and parse it on the pooler; that parse alone was 1.4 CPU-seconds per minute on staging, about 2.5% of pooler CPU, the largest single item in the instrumentation cost. `callerid.NewContext` now stores only the typed `CallerId`; baggage set by callers for other purposes is left untouched and the Baggage propagator stays installed.

3. **`mg.pooler.query.*` metrics now cover portal execution.** `mg.pooler.query.duration`, `rows`, `errors` and `pool_acquire.duration` were recorded only on the simple-query paths. `PortalStreamExecute`, the extended-protocol Bind/Execute path that drivers use, recorded nothing on either branch, so for most production traffic the pooler's only per-statement signal was otelgrpc's `rpc.server.duration`. That gap is why the `rpc.*` metrics cannot be treated as duplicates today: the gRPC dashboards in the platform repo use them for QPS, error rate by status code and latency by method. With this commit the `mg_*` metrics cover the query path, so the dashboards can migrate later; the `rpc.*` metrics themselves are left as they are.

## What is already covered upstream

- #1459 made the tracer provider use `ParentBased(NeverSample)` when `OTEL_TRACES_EXPORTER` is explicitly `none` and no sampler is configured. An earlier version of this branch had a similar change; it was dropped in favour of #1459. Note for operators: staging currently leaves the variable **unset**, which #1459 deliberately treats as "keep the OTel default sampler", so staging keeps recording spans until `OTEL_TRACES_EXPORTER=none` is set on the operator (it already forwards that variable to every pod).
- #1458 removed the second response message per statement on the same hop.

## Testing

- servenv: flag registered and parsed through the viper binding, server option present/absent, client side follows the setting lazily (3 tests).
- grpccommon: bufconn end to end, a client RPC produces a span with instrumentation on and none with it off; lazy evaluation of the setting (3 tests).
- callerid: round trip, arbitrary `application_name`, baggage untouched, unrelated caller baggage preserved, nil no-op (6 tests).
- executor: portal execution records duration, rows, errors and pool acquire on the existing-reservation, new-reservation and regular-connection branches (4 tests).


GuptaManan100 (approved):
LGTM!

PR #1480, opened 2026-09-21
perf(rpc): reuse query streams between multigateway and multipooler

haritabh17:
## Summary

Second half of MUL-1584 (the first, gRPC handler-worker reuse, merged in #1477). Adds `ExecuteStream`, a bidirectional RPC that carries many sequential simple-query operations over one gRPC stream, plus a gateway-side pool of exclusively leased streams per pooler connection. It amortises per-RPC setup (HTTP/2 headers, metadata, otelgrpc hook, sampler, stack growth) across ~1000 statements while keeping every per-statement semantic exactly as today.

**What stays per statement.** Each message carries the complete `StreamExecuteRequest` (SQL, target, caller ID, execute/reservation options) and is fed into the unchanged `StreamExecute` handler: caller annotation, admission/fencing, reservation lookup and ownership, session settings, execution, cancellation, diagnostics. Trace context, baggage and remaining deadline travel in the message and are rebuilt per operation; nothing leaks between operations.

**Correctness rules.** Explicit completion frame per operation with the original status code, message and details. Transport EOF before completion is an error, never success. Cancellation, timeout, malformed frame or abandoned response discards the lease. A send failure after SQL was submitted is never replayed. A stream is reused only after a valid completion was consumed.

**Compatibility.** The gateway handshakes before sending SQL; an old pooler returns `UNIMPLEMENTED` and the connection falls back to `StreamExecute` for its lifetime. Calls with custom outgoing metadata, or peers requiring per-call JWT authorization, take the legacy path (the server refuses reuse before the ready frame). mTLS identity is connection-scoped and unaffected. COPY, portals, replication and commit/rollback RPCs are untouched.

**Retention.** ≤32 idle streams per pooler connection, 30 s idle expiry, retirement after 1024 operations (only between operations). Idle streams are dropped immediately when the gRPC connection leaves READY, so a pooler restart hits the retryable pre-execution path instead of a post-submission error. `--query-stream-reuse=false` (`MT_QUERY_STREAM_REUSE=false`) restores per-call RPCs.

**Observability.** `rpc.*` metrics now describe transports, not statements. New low-cardinality instruments: `mg.gateway.query_stream.operations{transport=new|reused|legacy_*}`, `.discards{reason}`, `.active`, `.idle`. Design doc: `docs/query_serving/reusable_query_rpc.md` (includes the interceptor audit).

## Benchmark: qs-aws EKS rig

Cluster `haritabh-perf-20260915-01`: 3 poolers (1 primary + 2 sync replicas, m7i.xlarge, multipooler limit 1.5 CPU, 40-connection pool), 5 gateways (t3.medium), pgbench 17 with 48 clients spread over the five gateways, `-M simple`, scale 10, 30 s warmup + 90 s measure, cgroup CPU per container. Baseline = upstream main b17ef9e9 (worker reuse merged); treatment = this PR. Order: base ×3 → streams ×3 → base ×1 control. Zero failed transactions in all 14 runs. Prometheus during the run: ~7,070 reused ops/s of 7,450 pooler statements/s, <8 new streams/s, 0 legacy fallbacks, discards 100 % `retired`.

| workload | metric | main (workers) | + streams | delta |
|---|---|---|---|---|
| simple-update | TPS | 2821 | 3110 | **+10.2%** |
| simple-update | mean latency | 17.0 ms | 15.4 ms | −9.3% |
| simple-update | p95 / p99 | 28.4 / 46.7 ms | 24.8 / 39.4 ms | −12.6% / −15.7% |
| simple-update | pooler CPU / txn | 0.505 ms | 0.461 ms | **−8.7%** |
| simple-update | gateways CPU / txn (5) | 1.577 ms | 1.411 ms | **−10.5%** |
| tpcb-like | pooler CPU / txn | 1.008 ms | 0.892 ms | **−11.5%** |
| tpcb-like | gateways CPU / txn (5) | 2.938 ms | 2.614 ms | **−11.0%** |
| tpcb-like | TPS | 1204 | 1159 | −3.8% * |

\* The primary re-elected onto a different node between arms and tpcb-like throughput is sensitive to that placement (cross-zone synchronous commit). Same-node comparison against the earlier workers-only arm on that node: tpcb-like TPS +7%, p99 −3%, pooler CPU/txn −6.5%, gateway −11%, postgres CPU/txn unchanged. The control repetition reproduced the baseline within 1%.

Gateway CPU drops too because the per-call client-side stream setup and its otelgrpc hook are amortised as well.

## Test plan

- [x] `go/common/queryrpc` transport tests: exclusive concurrent leases, response ordering and empty results, SQL status + details preserved, EOF-before-completion rejected, malformed frames, cancel/abandon/close, retirement and late-cancellation race, idle expiry, metadata/propagation fallback, old-peer fallback before SQL, no replay on transport loss, handshake honours caller deadline, dead transport drops idle streams, metrics (race, ×50)
- [x] gateway test through a real gRPC server: reservation state and retry classification (in-op UNAVAILABLE is not pre-execution)
- [x] unit suites for multigateway, multipooler grpc service, servenv, metric catalog (race)
- [x] endtoend: queryserving (incl. pgbouncer, pgparity), multipooler, backupfaults
- [x] qs-aws A/B above; rollback flag verified via `/config`


go/services/multipooler/grpcpoolerservice/execute_stream.go
```diff
@@ -0,0 +1,13 @@
+// Copyright 2026 Supabase, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
GuptaManan100, line 2: We need the full license here 😄

go/services/multigateway/query_stream_flag_test.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2026 Supabase, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
GuptaManan100, line 2: Same license thing

go/common/queryrpc/pool.go
```diff
@@ -0,0 +1,329 @@
...
+const (
+	maxIdleStreams    = 32
+	idleStreamTimeout = 30 * time.Second
+	maxStreamRequests = 1024
+)
+
+// Pool retains only idle transport streams, never PostgreSQL connections.
+// Each lease is exclusive; reservations and callers remain request fields.
+type Pool struct {
+	client      pb.MultipoolerServiceClient
```
GuptaManan100, line 33: Isn't this package better lived in multipooler package? I mean it clearly has multipooler written all over and used only there!

go/common/queryrpc/stream.go
```diff
@@ -0,0 +1,114 @@
+// Copyright 2026 Supabase, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
GuptaManan100, line 2: Same license issue, I presume its in other files too.

go/common/queryrpc/pool.go
GuptaManan100: I see its used in the gateway to create the pool. Maybe we should make it part of the QueryService interface?

go/services/multipooler/grpcpoolerservice/execute_stream.go
haritabh17: Ahh, fixed for all to the full header.

go/services/multigateway/query_stream_flag_test.go
haritabh17: Done

go/common/queryrpc/stream.go
haritabh17: Done.

go/services/multigateway/poolergateway/execute_stream_pool.go
```diff
@@ -0,0 +1,329 @@
...
+const (
+	maxIdleStreams    = 32
+	idleStreamTimeout = 30 * time.Second
+	maxStreamRequests = 1024
+)
+
+// Pool retains only idle transport streams, never PostgreSQL connections.
+// Each lease is exclusive; reservations and callers remain request fields.
+type Pool struct {
+	client      pb.MultipoolerServiceClient
```
haritabh17: Fair point. Moved the client pool into `poolergateway`.

PR #1483, opened 2026-09-22
feat(pgeval): port the Datum value model

GuptaManan100:
## Summary

- Adds `go/common/pgeval/datum`, the first piece of the gateway's expression evaluation engine: PostgreSQL's `Datum` value model ported from `postgres.h` (REL_17_6).
- `Datum` is a two-word `struct { val uint64; ptr unsafe.Pointer }` (16 bytes). By-value scalars live in `val` (allocation-free); by-reference payloads put a GC-visible pointer in `ptr` and fold their length into `val` - bare payload bytes, no varlena header.
- Ports the full `postgres.h` accessor family and adds `NullableDatum`, with correctness, GC-visibility, and allocation benchmarks.

## Description

PostgreSQL's `Datum` is a `uintptr` that is *either* an inline scalar or a hidden pointer, distinguished only by the caller's static type knowledge. That pointer-hidden-in-an-integer trick is illegal under Go's garbage collector, which must be able to find every live pointer - so this port makes `Datum` a two-word struct instead of one machine word:

- **By-value types** (bool, the sized integers, `float4`/`float8`, OIDs, xid/cid/mxid) live in `val` with `ptr` nil. Constructing and reading them touches no heap. On 64-bit PostgreSQL `USE_FLOAT8_BYVAL` makes `int64`/`float8` pass-by-value; since `val` is 64 bits this port is unconditionally byval for every scalar and needs none of upstream's `#ifdef` branches. Floats use bit-preserving `math.FloatNbits` puns (the Go analogue of upstream's union), so NaN payloads and signed zero survive byte-exact.
- **By-reference types** put their payload pointer in `ptr` (GC-visible, keeps the backing array alive) and fold the payload length into `val`. There is no TOAST and no varlena header: `ptr` addresses bare payload bytes, and the length a varlena header would carry lives in `val` instead. This is the primitive the later text/bytea/numeric/array working forms build on.
- **NULL** is never encoded in a `Datum`; it rides alongside in `NullableDatum`, matching upstream. A zero `Datum` is a legitimate value (false, 0, `InvalidOid`), not a NULL sentinel.

The `"char"` type maps to a named `Char = int8` alias - signed, because `char.c` pins `char`↔`int4` conversions to signed char, while comparisons impose unsigned ordering at each site.

The accessor functions mirror `postgres.h` one-for-one (same names, same semantics), each citing its upstream `file:line`, so ported fmgr bodies in later PRs read line-for-line against the C.

### Representation

The two-word `{val, unsafe.Pointer}` layout was chosen over two alternatives, benchmarked head-to-head on identical workloads:

- **`any`** - idiomatic, but boxes every non-cached scalar onto the heap once Datums live in a slice (i.e. always). Measured ~20× slower and ~0.5 MB / 65k allocations to build a 64Ki-element column. Rejected.
- **`{val uint64; ptr *byte; len int32}` (24 bytes)** - fully typed, zero-alloc, but ~40% slower to construct and 50% larger, and still needs `unsafe.Slice` to reconstruct byte payloads. Kept as the documented fallback if the `unsafe.Pointer` discipline proves hard to maintain.
- **`{val uint64; ptr unsafe.Pointer}` (16 bytes)** - chosen: zero-alloc, fastest to construct, smallest footprint.


haritabh17 (approved):
🚀

PR #1484, opened 2026-09-22
fix(topoclient): clarify ConnForCell error when global cell lookup fails

dweitzman:
- `ConnForCell` writes cell-local records but first resolves the target cell's address via a **global** topo read (`GetCell`); that read's raw error was indistinguishable from a cell-local write failure in logs (e.g. multiorch's "failed to re-assert topology registration").
- Wraps the `GetCell` error with context naming the global lookup explicitly.
- Error still unwraps via `errors.Is`/`errors.As` to the underlying `TopoError` — verified no call site anywhere in the repo type-checks `ConnForCell`'s error for a specific `TopoError` code.


rafael (approved):
LGTM

